PR #2, opened 2018-08-24
Feature/spanner source

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-14053

Adding spanner source plugin. 

Spanner service batch client is used to generate partitions for the spanner input query. the partitions and transaction_id are serialized and sent to InputFormat through configuration.

InputFormat creates a PartitionInputSplit for each of the spanner read partition and the InputSplit is responsible to executing the query using the deserialized transaction_id. 

SpannerRecordReader iterates through the query ResultSet for computing nextKeyValue. 
Since closing the transaction delete's the spanner session, we do not close the transaction, when the MapReduce job finishes, we close the spanner client in SpannerSource, which will release all the resources and close all the sessions.

Currently we only support the primitive types, support for Date, Timestamp, Array and Struct will be added as an improvement later.

Since the spanner client versions conflict with version of pub-sub client, have removed the pub-sub classes and created a JIRA to update pub sub and add them again - https://issues.cask.co/browse/CDAP-14185

TODO : Add documentation file for spanner source.


pom.xml
```diff
@@ -408,15 +414,16 @@
...
           <extensions>true</extensions>
           <configuration>
             <instructions>
               <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
               <Embed-Transitive>true</Embed-Transitive>
               <Embed-Directory>lib</Embed-Directory>
               <!--Only @Plugin classes in the export packages will be included as plugin-->
-              <_exportcontents>co.cask.bigquery.*,com.google.cloud.hadoop.*,co.cask.gcs.*,co.cask.speech.*,co.cask.common.*</_exportcontents>
+              <_exportcontents>co.cask.bigquery.*,com.google.cloud.hadoop.*,co.cask.gcs.*,
+                co.cask.spanner.*,co.cask.speech.*,co.cask.common.*</_exportcontents>
```
albertshau, line 426: I don't think co.cask.common should be here

pom.xml
```diff
@@ -493,7 +500,7 @@
         <version>1.0.0-SNAPSHOT</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[4.2.0,5.0.0)</parent>
+            <parent>system:cdap-data-pipeline[4.2.0,9.0.0-SNAPSHOT)</parent>
```
albertshau: this is bad practice, we cannot guarantee forward compatibility like this. Please change the upper bound for both parents to 6.0.0

widgets/Spanner-batchsource.json
```diff
@@ -0,0 +1,107 @@
...
+            "placeholder": "Spanner database name."
+          }
+        }
+      ]
+    },
+    {
+      "label" : "Query and properties",
+      "properties" : [
+        {
+          "widget-type": "textbox",
```
albertshau: a textbox is really small for a query. Should use whatever db source is using.

widgets/Spanner-batchsource.json
```diff
@@ -0,0 +1,107 @@
...
+          "label": "Query",
+          "name": "query",
+          "widget-attributes" : {
+            "placeholder": "Spanner Query."
```
albertshau: 'Spanner' is redundant here. I would follow whatever we have in db source

widgets/Spanner-batchsource.json
```diff
@@ -0,0 +1,107 @@
...
+          "widget-type": "textbox",
+          "label": "Max Partitions",
+          "name": "maxPartitions",
+          "widget-attributes" : {
+            "placeholder": "Maximum number of partitions."
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "size of paritition in bytes",
```
albertshau: typo: paritition -> partition. 

Use consistent style everwhere. Most labels capitalize. 

'Partition Size (megabytes)'

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Spanner source config
+ */
+public class SpannerSourceConfig extends SpannerConfig {
+
+  @Name("maxPartitions")
+  @Description("Spanner partition option to specify max number of partitions, " +
```
albertshau: The maximum number of partitions to use. This is only a hint. The actual number may vary.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    "this is only a hint, actual number may vary")
+  @Macro
+  @Nullable
+  public Long maxPartitions;
+
+  @Name("partitionSizeBytes")
+  @Description("Spanner partition option to specify partition size bytes, " +
```
albertshau: Partition size in megabytes. This is only a hint. The actual partition size may vary.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
albertshau: Also, is this really a hint? Maximums usually are strictly enforced.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    "this is only a hint, actual size of partition may vary")
+  @Macro
+  @Nullable
+  public Long partitionSizeBytes;
```
albertshau: bytes is a very user unfriendly unit to enter in a text box. Lets change this to megabytes.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import co.cask.spanner.common.SpannerConfig;
+
+import javax.annotation.Nullable;
+
+/**
+ * Spanner source config
+ */
+public class SpannerSourceConfig extends SpannerConfig {
+
+  @Name("maxPartitions")
```
albertshau: you don't need name annotations anywhere, they default to the variable name.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,58 @@
...
+  @Macro
+  @Nullable
+  public Long partitionSizeBytes;
+
+  @Name("query")
+  @Description("Spanner select query to execute")
+  @Macro
+  public String query;
+
+  public SpannerSourceConfig(String referenceName, String project,
```
albertshau: don't need this constructor

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.ws.rs.Path;
+
+/**
+ * Spanner batch source
+ */
+@Plugin(type = BatchSource.PLUGIN_TYPE)
+@Name(SpannerSource.NAME)
+@Description("Spanner batch source.")
```
albertshau: Descriptions should say more than just the name. Can you find a sentence or two from the Spanner site that describes what it is.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+  private Schema schema;
+  private Spanner spanner;
+
+  public SpannerSource(SpannerSourceConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
```
albertshau: you should be validating config properties in this step. The convention I've been following is for the config classes to have a validate() method that plugins call at configure time and also at prepare time (in case there are macros).

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  @Override
+  public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
+    Job job = JobUtils.createInstance();
+    final Configuration configuration = job.getConfiguration();
+    configuration.clear();
```
albertshau: if you're going to clear, there's no point in getting it from the job. I think JobUtils is used to make sure tokens and stuff are in the config. you should just be able to create a Configuration object directly.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+    Map<String, String> configurationMap = new HashMap<>();
+    configurationMap.put(SpannerConstants.PROJECT_ID, config.project);
+    configurationMap.put(SpannerConstants.SERVICE_ACCOUNT_FILE_PATH, config.serviceAccountFilePath);
+    configurationMap.put(SpannerConstants.INSTANCE_ID, config.instance);
+    configurationMap.put(SpannerConstants.DATABASE, config.database);
+    configurationMap.put(SpannerConstants.QUERY, config.query);
+
+    configurationMap.entrySet().stream()
+      .filter(e -> e.getValue() != null)
+      .forEach(e -> configuration.set(e.getKey(), e.getValue()));
```
albertshau: it's simpler and more readable to just to set these on the Configuration object directly instead of creating a map and doing this. Does it matter if they're null?

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  /**
+   * Serialize the object and store the serialized string in configuration identified by the propertyName
+   * @throws IOException
+   */
+  private void serializeObject(Object object, Configuration configuration, String propertyName) throws IOException {
```
albertshau: it's cleaner to make this method return a string and make the caller set the value in the Configuration object. Try to keep it so that methods do one thing.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+                                         Schema.nullableOf(getSchemaForCode(field.getType().getCode()))));
+      }
+    }
+    return Schema.recordOf("outputSchema", schemaFields);
+  }
+
+  /**
+   * get {@link Schema} for the {@link Type.Code},
+   */
+  private Schema getSchemaForCode(Type.Code code) throws SQLException {
```
albertshau: more accurate for this to return Schema.Type

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+      case ARRAY:
+      case DATE:
+      case TIMESTAMP:
+      case STRUCT:
+      default:
+        throw new SQLException(new UnsupportedTypeException("Unsupported SQL Type: " + code.name()));
+    }
+  }
+
+  class GetSchemaRequest {
```
albertshau: you can just use the source config as the request object. If you have this separate object, there is a hidden requirement that it has the exact same field names.

src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -0,0 +1,75 @@
...
+  @Macro
+  @Nullable
+  public String serviceAccountFilePath;
+
+  @Name("schema")
+  @Description("Schema of the Spanner table")
+  @Macro
+  public String schema;
+
+  public SpannerConfig(String referenceName, String project,
```
albertshau: is this used/needed?

src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * Spanner configuration
+ */
+public class SpannerConfig extends ReferencePluginConfig {
+  @Name("project")
+  @Description("Project ID")
```
albertshau: descriptions should be more than a name. We need to go through all these plugins and give them better descriptions. In the dataproc provisioner, we use:

Google Cloud Project ID, which uniquely identifies your project. You can find it on the Dashboard in the Cloud Platform Console.

src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -0,0 +1,75 @@
...
+  @Macro
+  public String instance;
+
+  @Name("database")
+  @Description("Database to be used")
+  @Macro
+  public String database;
+
+  @Name("serviceFilePath")
+  @Description("Service Account File Path")
```
albertshau: Need more description here.

Path on the local file system to the service account key used for access to Spanner. Does not need to be specified when running on a Dataproc cluster. When running on other clusters, the file must be present on every node in the cluster.

src/main/java/co/cask/spanner/common/SpannerUtil.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Construct and return the {@link Spanner} service for the provided credentials and projectId
+   */
+  public static Spanner getSpannerService(String serviceAccountFilePath, String projectId) throws Exception {
+    GoogleCredentials credentials = null;
+    if (serviceAccountFilePath != null) {
+      credentials = BigQuerySource.loadServiceAccountCredentials(serviceAccountFilePath);
+    }
+    SpannerOptions.Builder optionsBuilder = SpannerOptions.newBuilder();
+    if (projectId != null) {
```
albertshau: project id can be null?

src/main/java/co/cask/spanner/common/SpannerUtil.java
```diff
@@ -0,0 +1,47 @@
...
+      optionsBuilder.setProjectId(URLEncoder.encode(projectId, "UTF-8"));
```
albertshau: why do you need to encode it?

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * Spanner partition input split,
```
albertshau: input split. Performs

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+ * performs serialization and deserialization of the {@link Configuration}, {@link Partition} and
```
albertshau: it does not serialize the Configuration.

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+ * and makes the result set available to {@link SpannerRecordReader}
+ */
+public class PartitionInputSplit extends InputSplit implements Writable {
+  private static final Logger LOG = LoggerFactory.getLogger(PartitionInputSplit.class);
+  private Partition partition;
+  private Configuration configuration;
+  private BatchTransactionId batchTransactionId;
+
+  public PartitionInputSplit() {
+
```
albertshau: add a comment that this is needed for Hadoop deserialization

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+import java.io.ObjectOutputStream;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * Spanner partition input split,
+ * performs serialization and deserialization of the {@link Configuration}, {@link Partition} and
+ * {@link BatchTransactionId} received from {@link SpannerInputFormat}.
+ *
+ * Executes the spanner query for the {@link Partition}
```
albertshau: splits should not perform any reading. Their purpose is to represent a section of data. The PartitionInputSplit should only contain the Spanner Partition and should only serialize/deserialize the partition.

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+  public PartitionInputSplit(Configuration configuration, Partition partition,
+                             BatchTransactionId batchTransactionId) throws Exception {
+    this.configuration = configuration;
+    this.partition = partition;
+    this.batchTransactionId = batchTransactionId;
+  }
+
+  @Override
+  public long getLength() throws IOException, InterruptedException {
+    // TODO estimate length
```
albertshau: if this is an actual todo, it should have a jira. If we no plans on doing this, remove the comment.

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+    return 0;
+  }
+
+  @Override
+  public String[] getLocations() throws IOException, InterruptedException {
+    return new String[0];
+  }
+
+  /**
+   * Serialize configuration, partition and batchTransactionId to dataOutput
```
albertshau: don't need a javadoc for this. The contract is defined by the parent class

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+      objectOutputStream.writeObject(object);
+      objectOutputStream.flush();
+      byte[] objectBytes = byteArrayOutputStream.toByteArray();
+      // we write the byte array length, to help initialize byte array during deserialization to read from DataInput
+      dataOutput.writeInt(objectBytes.length);
+      dataOutput.write(objectBytes);
+    }
+  }
+
+  private MapWritable confToWritable(Iterator<Map.Entry<String, String>> configurationIterator) {
```
albertshau: should only serialize the Partition

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public void readFields(DataInput dataInput) throws IOException {
+    MapWritable mapWritable = new MapWritable();
+    mapWritable.readFields(dataInput);
+    partition = (Partition) readObjectFromDataInput(dataInput);
+    batchTransactionId = (BatchTransactionId) readObjectFromDataInput(dataInput);
+    configuration = writableToConf(mapWritable);
+  }
+
+  private Object readObjectFromDataInput(DataInput dataInput) throws IOException {
```
albertshau: you should only be deserializing the Partition

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,158 @@
...
+    dataInput.readFully(readArray);
+    try (ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(readArray);
+         ObjectInputStream objectInputStream = new ObjectInputStream(byteArrayInputStream)) {
+      return objectInputStream.readObject();
+    } catch (ClassNotFoundException cfe) {
+      throw new IOException("Exception while trying to deserialize object ", cfe);
+    }
+  }
+
+  public ResultSet getResultSet() throws Exception {
```
albertshau: this should be in the RecordReader, not in the split.

src/main/java/co/cask/spanner/source/SpannerInputFormat.java
```diff
@@ -0,0 +1,90 @@
...
+   * and create {@link PartitionInputSplit} using them.
+   */
+  @Override
+  public List<InputSplit> getSplits(JobContext jobContext) throws IOException, InterruptedException {
+    Configuration configuration = jobContext.getConfiguration();
+    List<InputSplit> partitionSplits = new ArrayList<>();
+    try {
+      BatchTransactionId batchTransactionId =
+        (BatchTransactionId) deserializeObject(configuration, SpannerConstants.SPANNER_BATCH_TRANSACTION_ID);
+      ArrayList<Partition> partitions =
```
albertshau: ArrayList -> List

src/main/java/co/cask/spanner/source/SpannerInputFormat.java
```diff
@@ -0,0 +1,90 @@
...
+        partitionSplits.add(split);
+      }
+    } catch (Exception e) {
+      throw new IOException("Exception while trying to initialize spanner and create partition splits", e);
+    }
+    LOG.debug("Initialized and configured {} splits", partitionSplits.size());
+    return partitionSplits;
+  }
+
+  private Object deserializeObject(Configuration configuration, String property) throws IOException {
```
albertshau: `private <T> T deserialize()`

src/main/java/co/cask/spanner/source/SpannerInputFormat.java
```diff
@@ -0,0 +1,90 @@
...
+      return objectInputStream.readObject();
+    } catch (ClassNotFoundException cfe) {
+      throw new IOException("Exception while trying to deserialize object ", cfe);
+    }
+  }
+
+  @Override
+  public RecordReader<NullWritable, ResultSet> createRecordReader(
+    InputSplit inputSplit, TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
+    return new SpannerRecordReader();
```
albertshau: Here you should be getting the transaction id from the configuration and passing it to the record reader.

src/main/java/co/cask/spanner/source/SpannerRecordReader.java
```diff
@@ -0,0 +1,79 @@
...
+  PartitionInputSplit partitionInputSplit;
+  ResultSet resultSet;
+  boolean isFinished;
+
+  @Override
+  public void initialize(InputSplit inputSplit,
+                         TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
+    partitionInputSplit = (PartitionInputSplit) inputSplit;
+    try {
+      resultSet = partitionInputSplit.getResultSet();
```
albertshau: you should only be getting the Partition from the input split. Reading logic should be moved from the split to the reader.

src/main/java/co/cask/spanner/source/SpannerRecordReader.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public ResultSet getCurrentValue() throws IOException, InterruptedException {
+    return resultSet;
+  }
+
+  @Override
+  public float getProgress() throws IOException, InterruptedException {
+    return isFinished ? 1.0f : 0.0f;
```
albertshau: I assume this means ResultSet doesn't tell you how many results there are?

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,283 @@
...
+    List<Partition> partitions = new ArrayList<>(batchReadOnlyTransaction.partitionQuery(getPartitionOptions(),
+                                                                                         Statement.of(config.query)));
+
+    // serialize batch transaction-id and partitions
+    serializeObject(batchTransactionId, configuration, SpannerConstants.SPANNER_BATCH_TRANSACTION_ID);
+    serializeObject(partitions, configuration, SpannerConstants.PARTITIONS_LIST);
+
+    // set input format and pass configuration
+    batchSourceContext.setInput(Input.of(config.referenceName,
+                                         new InputFormatProvider() {
```
albertshau: there is a SourceInputFormatProvider that you can use.

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
gsps1: https://cloud.google.com/spanner/docs/reference/rest/v1/PartitionOptions yes, its only a hint, In my pipeline the splits were always 6, even when i had set maxPartitions to 1. Also i am not sure about the note in the link where it says its currently ignored for partitionQuery request. have to check with more data.

src/main/java/co/cask/spanner/source/SpannerSource.java
gsps1: yeah, there is a precondition check to check keys and values are not null. performing a check to set key and value if not null for each config if done through a method, will result in multiple method calls. if done without method it felt like a lot of duplication, i was thinking this avoids that and went with this approach.  but if the preference is just setting it without a map, i can change it.

src/main/java/co/cask/spanner/common/SpannerConfig.java
gsps1: are you referring to the constructor or the class ? class is useful to encapsulate common properties from both source and sink.

src/main/java/co/cask/spanner/source/SpannerRecordReader.java
gsps1: thanks, that really simplifies the logic.

src/main/java/co/cask/spanner/common/SpannerConfig.java
albertshau: the constructor. These config objects are created by the platform, so unless there are unit tests that 'new' a config object, these constructors are not required

src/main/java/co/cask/spanner/source/SpannerSource.java
albertshau: yeah I think it's easier to understand if you can have a setIfNotNull() helper method and just call it a bunch of times. (also slightly less overhead from not creating a map)

pom.xml
gsps1: this is the package in this repo, not the cask-data/common. we should ideally rename the package name to avoid this confusion.

albertshau: it doesn't matter where it comes from, this will expose that package and everything under it to every plugin used in the program. It looks like somebody just threw this in here without thinking about it.

The only things that need to be exported are the packages that contain the plugins, and anything that the program needs to run the inputformat/outputformat. This usually means the input format class, output format class, and whatever key and value classes they work with.

src/main/java/co/cask/spanner/source/SpannerSource.java
gsps1: JobUtils#createInstance adds credentials information to job if security is enabled in user group info. However we seem to be following different pattern in different sources. In sources such as hbase, mongodb, we just create configuration directly and then clear it, in hive batch source we get from it job.

pom.xml
```diff
@@ -62,19 +62,20 @@
...
-    <cdap.version>4.3.1</cdap.version>
+    <cdap.version>4.3.3</cdap.version>
     <guava.version>20.0</guava.version>
     <slf4j.version>1.7.5</slf4j.version>
     <junit.version>4.12</junit.version>
     <hadoop.version>2.8.0</hadoop.version>
     <hydrator.version>1.8.0</hydrator.version>
-    <bigquery.connector.hadoop2.version>0.10.2-hadoop2</bigquery.connector.hadoop2.version>
-    <google.cloud.bigquery.version>0.25.0-beta</google.cloud.bigquery.version>
+    <bigquery.connector.hadoop2.version>0.13.1-hadoop2</bigquery.connector.hadoop2.version>
```
CuriousVini, line 71: why do we need this change? Also the latest version is 0.13.4-hadoop2 why `0.13.1-hadoop2`?

docs/GoogleSubscriber-streamingsource.md
```diff
@@ -1,13 +0,0 @@
-### Google Subscriber Streaming source
```
CuriousVini, line 1: why are we removing this?

docs/GooglePublisher-batchsink.md
```diff
@@ -1,12 +0,0 @@
-### Google Publisher Sink
```
CuriousVini, line 1: why are we removing this?

gsps1: @CuriousVini  as i mentioned in the summary, its due to version conflicts. We are depending on the recent spanner version which conflict with the older pubsub version used in this project. pubsub has to be fixed to use the latest version and need to be added again. i have also opened a JIRA for this, its in summary.

CuriousVini: I see.

src/main/java/co/cask/spanner/source/SpannerRecordReader.java
gsps1: yes.

src/main/java/co/cask/spanner/common/SpannerUtil.java
gsps1: i was getting an exception due the presence of `:` in projectId and i had to encode to avoid that issue.

pom.xml
gsps1: that package currently has few input format classes, they have to be removed. will do that as a separate PR.

gsps1:
@albertshau  thanks for the review, have updated the PR addressing the comments, please take a look again, thanks!

src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -0,0 +1,89 @@
...
+                                           "please provide the Cloud Spanner instance in your GCP project");
+    }
+    if (!containsMacro("database") && database == null) {
+      throw new IllegalArgumentException("Cloud Spanner database cannot be empty, " +
+                                           "please provide the Cloud Spanner database in the Cloud Spanner instance");
+    }
+    if (!containsMacro("schema") && schema == null) {
+      throw new IllegalArgumentException("Schema cannot be empty, " +
+                                           "please provide the output schema for Cloud Spanner");
+    }
```
gsps1: @albertshau i added this validation check, this will be useful if a property has macro enabled but at runtime value is not provided, however its not a common scenario and we don't have these kind of checks in other plugins. So i am not sure if we need to keep this validation logic or remove it.

src/main/java/co/cask/common/GCPUtils.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+/**
+ * GCP utility class to get service account credentials
+ */
+public class GCPUtils {
+  public static ServiceAccountCredentials loadServiceAccountCredentials(String path) throws Exception {
```
albertshau: whenever possible, throw a more specific exception. This seems like it can throw an IOException

src/main/java/co/cask/common/GCPUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    File credentialsPath = new File(path);
+    if (!credentialsPath.exists()) {
```
albertshau: it doesn't look like this is needed, as FileNotFoundException is thrown below

src/main/java/co/cask/common/GCPUtils.java
```diff
@@ -0,0 +1,49 @@
...
+      throw new FileNotFoundException("Service account file " + credentialsPath.getName() + " does not exist.");
+    }
+    try (FileInputStream serviceAccountStream = new FileInputStream(credentialsPath)) {
+      return ServiceAccountCredentials.fromStream(serviceAccountStream);
+    } catch (FileNotFoundException e) {
+      throw new Exception(
```
albertshau: if we just want a better error message, throw it as the same type of exception, not as Exception.

src/main/java/co/cask/spanner/common/SpannerConfig.java
albertshau: these are not needed. If the property is not annotated as nullable, CDAP will make sure it's not null

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -84,14 +82,16 @@ public SpannerSource(SpannerSourceConfig config) {
...
+    }
   }
 
   @Override
   public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
-    Job job = JobUtils.createInstance();
-    final Configuration configuration = job.getConfiguration();
-    configuration.clear();
+    config.validate();
+    final Configuration configuration = new Configuration();
```
albertshau: doesn't need to be final

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -251,18 +235,18 @@ public Schema getSchema(GetSchemaRequest request) throws Exception {
   /**
    * get {@link Schema} for the {@link Type.Code},
    */
-  private Schema getSchemaForCode(Type.Code code) throws SQLException {
+  private Schema.Type getSchemaForCode(Type.Code code) throws SQLException {
```
albertshau: nit: more accurate to call it getTypeForCode

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -28,31 +27,29 @@
...
-    this.maxPartitions = maxPartitions;
-    this.partitionSizeBytes = partitionSizeBytes;
-    this.query = query;
+  public SpannerSourceConfig(String referenceName) {
+    super(referenceName);
+  }
+
+  public void validate() {
+    super.validate();
+    if (!containsMacro("query") && query == null) {
```
albertshau: this is not needed, as it will be handled by CDAP. However, you should check partition size and max partitions are valid. For example, they should be positive numbers.

albertshau (commented):
looks much cleaner, a few comments

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,259 @@
...
+        case INT64:
+          builder.set(fieldName, resultSet.getLong(fieldName));
+          break;
+        case FLOAT64:
+          builder.set(fieldName, resultSet.getDouble(fieldName));
+          break;
+        case STRING:
+          builder.set(fieldName, resultSet.getString(fieldName));
+          break;
+        case BYTES:
```
nitinmotgi, line 163: We should handle date, datetime and timestamp

src/main/java/co/cask/common/GCPUtils.java
gsps1: i moved the method from BigQuerySource, should have checked the content, have fixed it now.

src/main/java/co/cask/spanner/source/SpannerSource.java
gsps1: we will add that in a separate PR.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,259 @@
...
+   *
+   * @param request {@link SpannerSourceConfig} containing information required for query to execute.
+   * @return schema of fields
+   */
+  @Path("getSchema")
+  public Schema getSchema(SpannerSourceConfig request) throws Exception {
+    Spanner spanner = SpannerUtil.getSpannerService(request.serviceAccountFilePath, request.project);
+    DatabaseClient databaseClient =
+      spanner.getDatabaseClient(DatabaseId.of(request.project, request.instance, request.database));
+    ResultSet resultSet = databaseClient.singleUse().executeQuery(Statement.of(request.query));
```
albertshau: is the user charged for the full query here even though we don't go through the whole result set?

albertshau: also, do you need to close anything like ResultSet, DatabaseClient, or Spanner?

gsps1: my bad. fixed to close the resources now. about the pricing, https://cloud.google.com/spanner/pricing - documentation says its based on network bandwidth utilization for queries. so i reckon it will be based on result set utilization, but i can confirm by performing some checks.

albertshau: I see. Also, did you figure out how to check if the query can be partitioned?

albertshau: can you also make sure errors are handled here? I remember there's a specific type of exception that is supposed to be throw, otherwise the message won't make it to the response (and to the UI)

gsps1: If the plugin endpoint throws an exception, `Method#invoke` will throw an `InvocationTargetException` and in ArtifactHttpHandler we handle `InvocationTargetException`. So i don't think that issue exists currently. I was able to test with an exception case.

gsps1: however i will update to throw more meaningful exception if the result set is empty.

src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * Spanner configuration
+ */
+public class SpannerConfig extends ReferencePluginConfig {
+  @Description("Google Cloud Project ID, which uniquely identifies your project." +
+    "You can find it on the Dashboard in the Cloud Platform Console")
```
albertshau: For consistency, end all descriptions with a period

src/main/java/co/cask/spanner/source/PartitionInputSplit.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+
+/**
+ * Spanner partition input split.Performs serialization and
```
albertshau: nit: space after period.

src/main/java/co/cask/spanner/source/SpannerInputFormat.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+
+/**
+ * Spanner input format, deserialize's and gets the partitions list from configuration to create input splits.
```
albertshau: deserialize's -> deserializes.

src/main/java/co/cask/spanner/source/SpannerInputFormat.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class SpannerInputFormat extends InputFormat<NullWritable, ResultSet> {
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerInputFormat.class);
+
+  /**
+   * Deserialize partitions list from configuration and create {@link PartitionInputSplit}'s using them.
```
albertshau: `{@link PartitionInputSplit PartitionInputSplits}`

src/main/java/co/cask/spanner/source/SpannerRecordReader.java
albertshau: Do you know if this ever gets called after close(). Seems like it would be a weird contract. If not, you can make this always return 0.0f and remove the isFinished flag.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,266 @@
...
+        return Schema.Type.BOOLEAN;
+      case INT64:
+        return Schema.Type.LONG;
+      case FLOAT64:
+        return Schema.Type.DOUBLE;
+      case STRING:
+        return Schema.Type.STRING;
+      case BYTES:
+        return Schema.Type.BYTES;
+      // todo support the following types
```
albertshau: can you open a jira and put the jira number here

widgets/Spanner-batchsource.json
```diff
@@ -0,0 +1,107 @@
...
+      "name": "schema",
+      "widget-type": "schema",
+      "widget-attributes": {
+        "schema-types": [
+          "boolean",
+          "int",
+          "long",
+          "float",
+          "double",
+          "string"
```
albertshau: this doesn't match the types that the plugin code supports. Bytes is missing, and int and float are not supported.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,266 @@
...
+   *
+   * @param request {@link SpannerSourceConfig} containing information required for query to execute.
+   * @return schema of fields
+   */
+  @Path("getSchema")
+  public Schema getSchema(SpannerSourceConfig request) throws Exception {
+    Spanner spanner = SpannerUtil.getSpannerService(request.serviceAccountFilePath, request.project);
+    DatabaseClient databaseClient =
+      spanner.getDatabaseClient(DatabaseId.of(request.project, request.instance, request.database));
+    String query = String.format("Select * from %s LIMIT 1;", request.table);
```
albertshau: Does spanner have any classes that can be used to build a query? Creating queries like this is generally not a good idea due to sql injection.

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -0,0 +1,266 @@
...
+    try (ResultSet resultSet = databaseClient.singleUse().executeQuery(Statement.of(query))) {
+      List<Schema.Field> schemaFields = new ArrayList<>();
+      if (resultSet.next()) {
+        List<Type.StructField> structFields = resultSet.getType().getStructFields();
+        for (Type.StructField field : structFields) {
+          // todo find out if column type is nullable from information schema.
```
albertshau: todo always requires a jira number

src/main/java/co/cask/spanner/source/SpannerSourceConfig.java
```diff
@@ -0,0 +1,56 @@
...
+  @Description("Cloud Spanner table id. Uniquely identifies your table within the Cloud Spanner database")
+  @Macro
+  public String table;
+
+  public SpannerSourceConfig(String referenceName) {
+    super(referenceName);
+  }
+
+  public void validate() {
+    if (!containsMacro("maxPartitions") && maxPartitions != null && maxPartitions < 1) {
```
albertshau: schema should also be validated. The type of each field must be validated.

albertshau (commented):
mostly comments about javadocs or descriptions. One comment about schema validation.

src/main/java/co/cask/spanner/source/SpannerSource.java
gsps1: `Statement` class is used to build SQL statement to execute. http://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/spanner/Statement.html, the only improvement we can do is to use parameter binding to bind the table name, but even that doesn't provide any additional protection.

albertshau: these types of classes generally perform checks to prevent injection

src/main/java/co/cask/spanner/source/SpannerSource.java
```diff
@@ -71,6 +72,10 @@
   "SQL (ANSI 2011 with extensions), and automatic, synchronous replication for high availability.")
 public class SpannerSource extends BatchSource<NullWritable, ResultSet, StructuredRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(SpannerSource.class);
+  private static final String TABLE_NAME = "TableName";
+  private static final Statement.Builder SCHEMA_STATEMENT_BUILDER =  Statement.newBuilder(
+    String.format("SELECT  t.column_name,t.spanner_type, t.is_nullable FROM information_schema.columns AS t WHERE " +
+                    "  t.table_catalog = ''  AND  t.table_schema = '' AND t.table_name = @%s", TABLE_NAME));
```
gsps1: this query is documented at https://cloud.google.com/spanner/docs/information-schema

widgets/Spanner-batchsource.json
gsps1: updated it, but UI doesn't seem to be using this to filter out unsupported types. However i have tested using unsupported type and that will fail validation during deploy.

src/main/java/co/cask/spanner/source/SpannerSource.java
gsps1: done

src/main/java/co/cask/spanner/source/SpannerSource.java
albertshau: can you add that as a comment in the code.

albertshau (approved):
one minor comment, otherwise lgtm

gsps1:
added comment, rebased my commits and merging.

widgets/Spanner-batchsource.json
albertshau: can you open a jira for this?

gsps1: done https://issues.cask.co/browse/CDAP-14238

PR #3, opened 2018-08-24
CDAP-14021 Adding support for logical types in source and sink

CuriousVini:
Changes:

- Maven compiler from java 7 to java 8
- Added support for Logical types in BigQuerySource
- Added support for Logical types in BigQuerySink
- Using `AvroBigQueryInputFormat` instead of `GsonBigQueryInputFormat`. This change was made because `GsonBigQueryInputFormat` was conveting integer to string. Also Avro format is more compact so its better to use it.
- Avro format converts data types in following way:
```
date -> string - format yyyy-mm-dd - example 2016-06-29
time -> string - format hh:mm:ss - example 01:57:00
timestamp -> logicaltype of timestamp-micro - example 2018-08-24 06:32:50.861782 UTC
int -> long 	
float -> double
boolean -> boolean
```
- Removing `RecordConverter` and replacing it with `AvroToStructuredTransformer`


src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -243,14 +262,38 @@ public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullW
...
+          json.addProperty(name, input.getDate(name).toString());
+          break;
+        case TIME_MILLIS:
+        case TIME_MICROS:
+          json.addProperty(name, input.getTime(name).toString());
+          break;
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          //timestamp for json input should be in this format yyyy-MM-dd HH:mm:ss.SSSSSS
+          json.addProperty(name, DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS")
```
albertshau: should create the formatter once in the initialize() method instead of for each value

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -243,14 +262,38 @@ public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullW
...
+  private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord input, Schema schema)
+    throws RecordConverterException {
+    Object object = input.get(name);
 
     if (object == null) {
       json.add(name, JsonNull.INSTANCE);
       return;
     }
 
+    Schema.LogicalType logicalType = schema.getLogicalType();
```
albertshau, line 253: this doesn't look like it handles nullable types.

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -264,12 +307,12 @@ private void decodeSimpleTypes(JsonObject json, String name, Object object, Sche
...
             json.addProperty(name, Integer.parseInt(value));
           } catch (NumberFormatException e) {
-            throw new RecordConvertorException(
+            throw new RecordConverterException(
               String.format("Unable to convert '%s' to integer for field name '%s'", value, name)
             );
           }
         } else {
-          throw new RecordConvertorException(
+          throw new RecordConverterException(
```
albertshau, line 293: almost all of this method is not needed. It should not be checking the object type after it knows the schema. If the schema says it's an int, the value should be an Integer.

This makes me wonder what sort of input was attached to this sink that the code is expecting invalid records...

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -283,25 +326,19 @@ private void decodeSimpleTypes(JsonObject json, String name, Object object, Sche
           json.addProperty(name, ((Integer) object).longValue());
         } else if (object instanceof Date) {
```
albertshau, line 305: none of these are needed, if the schema is a Long, the value is a Long.

src/main/java/co/cask/bigquery/BigQuerySink.java
albertshau: ok I see it's because it's done by the caller... this class needs a lot of cleanup (not in this PR).

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -126,7 +126,7 @@ public void prepareRun(BatchSinkContext context) throws Exception {
 
       String typeName = "STRING";
       if (type.isSimpleType()) {
-        typeName = simpleType(field.getSchema());
+        typeName = getTableDataType(field.getSchema());
       } else if (type == Schema.Type.UNION) {
```
albertshau: this is completely the wrong way to handle unions... it is assuming everything is the first thing in the union.... if NULL comes first it will treat it as null...

src/main/java/co/cask/bigquery/BigQuerySink.java
albertshau: opened CDAP-14187 for cleaning up this class, don't need to do it in this PR

src/main/java/co/cask/bigquery/BigQuerySource.java
```diff
@@ -237,6 +237,7 @@ public Schema getSchema(Request request) throws Exception {
       for (Field field : bgSchema.getFields()) {
         Field.Type type = field.getType();
         Schema.Type cType = Schema.Type.STRING;
```
albertshau: this existing logic is incorrect, we can't just assume a string. There should be an else clause at the end of the else-ifs that throws an exception about an unsupported type.

src/main/java/co/cask/bigquery/BigQuerySource.java
```diff
@@ -237,6 +237,7 @@ public Schema getSchema(Request request) throws Exception {
...
+        Schema.LogicalType lType = null;
```
albertshau: cleaner to just have a Schema here instead of cType and lType. In each clause, set a Schema.

src/main/java/co/cask/bigquery/BigQuerySource.java
```diff
@@ -249,20 +250,27 @@ public Schema getSchema(Request request) throws Exception {
...
           throw new Exception("Nested records not supported yet.");
         }
 
         if (field.getMode() == null || field.getMode() == Field.Mode.NULLABLE) {
           Schema fieldType = Schema.nullableOf(Schema.of(cType));
+          if (lType != null) {
+            fieldType = Schema.nullableOf(Schema.of(lType));
+          }
           fields.add(Schema.Field.of(field.getName(), fieldType));
         } else if (field.getMode() == Field.Mode.REQUIRED) {
```
albertshau: this won't be needed if we use a Schema variable instead of a Schema.Type above.

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ * Create StructuredRecords from GenericRecords
+ */
+public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
+  private final Map<Integer, Schema> schemaCache = Maps.newHashMap();
+
+  public StructuredRecord transform(GenericRecord genericRecord) throws IOException {
```
albertshau: doesn't look like this is used

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+        default:
+          throw new UnexpectedFormatException("field type " + fieldType + " is not supported.");
+      }
+    }
+
+    // This casting is added because bigquery up casts number fields. For example:
+    // int becomes long, float becomes double
+    switch (fieldType) {
+      case INT:
+        if (field instanceof Integer) {
```
albertshau: is this needed then? Won't it always be a long? If we're unsure about the actual type, it's better to cast to a Number and call intValue(), longValue(), etc where needed

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+          return (Integer) field;
+        } else {
+          return ((Long) field).intValue();
+        }
+      case LONG:
+        if (field instanceof Integer) {
```
albertshau: this shouldn't be needed

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          return getZonedDateTime((long) field);
+        default:
+          throw new UnexpectedFormatException("field type " + fieldType + " is not supported.");
+      }
+    }
+
+    // This casting is added because bigquery up casts number fields. For example:
+    // int becomes long, float becomes double
```
albertshau: this seems really weird... are we sure it always happens?

albertshau (commented):
Can you add unit tests for conversion of BigQuery avro records to StructuredRecord

src/main/java/co/cask/common/AvroToStructuredTransformer.java
CuriousVini: Its documented here: https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro. 
Also I verified it by having a big query table with INTEGER and FLOAT types in schema.

src/main/java/co/cask/bigquery/BigQuerySink.java
CuriousVini: yea both `BigQuerySource` and `BigQuerySink` requires clean up.

src/main/java/co/cask/common/AvroToStructuredTransformer.java
albertshau: That doc is for loading data into big query, not reading it.

CuriousVini: the conversion is both ways, for export and import: https://cloud.google.com/bigquery/docs/exporting-data#avro_export_details 

image

For example for a table with below schema, avro schema looks as below. Please note that `unique_key` is interpreted as `long` in avro schema and `latitude` is interpreted as `double`. `timestamp` is in `timestamp-micros` avro logical type and `date` and `time` are interpreted as `string`

image



```json
{
  "type": "record",
  "name": "Root",
  "fields": [
    {
      "name": "unique_key",
      "type": [
        "null",
        "long"
      ]
    },
    {
      "name": "descript",
      "type": [
        "null",
        "string"
      ]
    },
    {
      "name": "date",
      "type": [
        "null",
        "string"
      ]
    },
    {
      "name": "time",
      "type": [
        "null",
        "string"
      ]
    },
    {
      "name": "address",
      "type": [
        "null",
        "string"
      ]
    },
    {
      "name": "longitude",
      "type": [
        "null",
        "double"
      ]
    },
    {
      "name": "latitude",
      "type": [
        "null",
        "double"
      ]
    },
    {
      "name": "location",
      "type": [
        "null",
        "string"
      ]
    },
    {
      "name": "timestamp",
      "type": [
        "null",
        {
          "type": "long",
          "logicalType": "timestamp-micros"
        }
      ]
    }
  ]
}
```

src/main/java/co/cask/common/AvroToStructuredTransformer.java
CuriousVini: Its an abstract method in `RecordConverter` https://github.com/caskdata/hydrator-plugins/blob/develop/hydrator-common/src/main/java/co/cask/hydrator/common/RecordConverter.java#L39

src/main/java/co/cask/common/AvroToStructuredTransformer.java
albertshau: Looked at the docs some more and this is because there is no 'int' and 'long' in bigquery, and no 'float' and 'double'. A big query 'integer' is 8 bytes, which is a java 'long', and a big query 'float' is 8 bytes, which is a java double. So this actually means the avro int and float types should never be returned by the getSchema call

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -183,17 +184,36 @@ String unionType(Schema schema) {
...
+          throw new UnexpectedFormatException("Unsupported logical type " + logicalType );
+      }
+    }
+
     Schema.Type type = schema.getType();
     switch(type) {
       case INT:
         return "INTEGER";
+      case LONG:
+        return "LONG";
```
albertshau: this is not a big query data type

src/main/java/co/cask/bigquery/BigQuerySink.java
CuriousVini: yea I agree, I will clean it up as part of https://issues.cask.co/browse/CDAP-14187

pom.xml
```diff
@@ -493,8 +493,8 @@
         <version>1.0.0-SNAPSHOT</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[4.2.0,5.0.0)</parent>
-            <parent>system:cdap-data-streams[4.0.0,9.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline[4.2.0,6.0.0)</parent>
```
albertshau: use 6.0.0-SNAPSHOT as the upper bound. That way these plugins won't appear when we're developing CDAP 6.0.0 so we won't forget to update it if needed.

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -120,19 +122,15 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
-        continue;
-      }
-
-      String typeName = "STRING";
+      String typeName;
       if (type.isSimpleType()) {
-        typeName = simpleType(field.getSchema());
+        typeName = getTableDataType(field.getSchema());
       } else if (type == Schema.Type.UNION) {
         typeName = unionType(field.getSchema());
```
albertshau: this is still wrong. Unions are not supported unless they represent a nullable type. It should look something like:

```
if (fieldSchema.isNullable()) {
  fieldSchema = fieldSchema.getNonNullable();
}
Schema.Type type = fieldSchema.getType();
if (!type.isSimpleType()) {
  throw new IllegalArgumentException(String.format("Field '%s' is of unsupported type '%s'", name, type));
}
String typeName = getTableDataType(fieldSchema);
```

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -179,20 +178,40 @@ public void initialize(BatchRuntimeContext context) throws Exception {
     }
   }
 
-  String unionType(Schema schema) {
+  private String unionType(Schema schema) {
```
albertshau: this method is wrong and should not exist. If a field is union(string, int), it's incorrect to assume every value is of type string.

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -218,39 +238,64 @@ public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullW
...
         case LONG:
         case STRING:
         case FLOAT:
         case DOUBLE:
         case BOOLEAN:
-          decodeSimpleTypes(object, name, value, field.getSchema());
+          decodeSimpleTypes(object, name, input, field.getSchema());
           break;
 
         case UNION:
```
albertshau: remove this case. Instead, before the switch, do:

```
if (fieldSchema.isNullable()) {
  fieldSchema.getNonNullable();
}
Schema.Type type = fieldSchema.getType();
```

src/main/java/co/cask/bigquery/BigQuerySource.java
```diff
@@ -236,33 +237,38 @@ public Schema getSchema(Request request) throws Exception {
...
 
         if (field.getMode() == null || field.getMode() == Field.Mode.NULLABLE) {
-          Schema fieldType = Schema.nullableOf(Schema.of(cType));
+          Schema fieldType = Schema.nullableOf(schema);
           fields.add(Schema.Field.of(field.getName(), fieldType));
         } else if (field.getMode() == Field.Mode.REQUIRED) {
-          fields.add(Schema.Field.of(field.getName(), Schema.of(cType)));
+          fields.add(Schema.Field.of(field.getName(), schema));
+        } else {
+          throw new UnsupportedTypeException("Repeated type is not supported");
```
albertshau: better to make this an else if. Otherwise if another mode is added in a future library, it will result in a bad error message.

src/main/java/co/cask/bigquery/BigQuerySource.java
```diff
@@ -236,33 +237,38 @@ public Schema getSchema(Request request) throws Exception {
...
+        } else if (value == LegacySQLTypeName.TIMESTAMP || value == LegacySQLTypeName.DATETIME) {
+          schema = Schema.of(Schema.LogicalType.TIMESTAMP_MICROS);
+        } else {
+          // this should never happen
+          throw new UnsupportedTypeException(String.format("Big Query Sql type %s is not supported", value));
         }
 
         if (field.getMode() == null || field.getMode() == Field.Mode.NULLABLE) {
-          Schema fieldType = Schema.nullableOf(Schema.of(cType));
+          Schema fieldType = Schema.nullableOf(schema);
```
albertshau: type is a misleading name. You can just inline this in the next line.

src/main/java/co/cask/common/AvroToStructuredTransformer.java
albertshau: it should be annotated with override then.

albertshau: the one with the schema is abstract, not this one.

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+      case ARRAY:
+        // this should not happen
+        throw new UnexpectedFormatException("Field type " + fieldType + " is not supported.");
+    }
+
+    return super.convertField(field, fieldSchema);
+  }
+
+  @VisibleForTesting
+  Object getZonedDateTime(long microTs) {
```
albertshau: seems like this should return ZonedDateTime

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  @Nullable
+  protected Object convertField(Object field, Schema fieldSchema) throws IOException {
+    if (field == null) {
+      return null;
+    }
+
+    Schema.Type fieldType = fieldSchema.getType();
```
albertshau: this won't work for nullable fields.

```
Schema schema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
Schema.Type type = schema.getType();
Schema.LogicalType logicalType = schema.getLogicalType();
```

Please make sure this is covered in unit tests. Nullable types, with one record that has a null value and one that does not.

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+    Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DATE:
+          // date will be in yyyy-mm-dd format
+          return LocalDate.parse(field.toString());
```
albertshau: this should return the integer that StructuredRecord expects

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+        case TIME_MILLIS:
+        case TIME_MICROS:
+          // time will be in hh:mm:ss format
+          return LocalTime.parse(field.toString());
```
albertshau: this should return the long that StructuredRecord expects

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          return getZonedDateTime((long) field);
```
albertshau: this should return the long that StructuredRecord expects

src/main/java/co/cask/common/AvroToStructuredTransformer.java
```diff
@@ -0,0 +1,127 @@
...
+    return transform(genericRecord, convertSchema(genericRecordSchema));
+  }
+
+  @Override
+  public StructuredRecord transform(GenericRecord genericRecord, Schema structuredSchema) throws IOException {
+    StructuredRecord.Builder builder = StructuredRecord.builder(structuredSchema);
+    for (Schema.Field field : structuredSchema.getFields()) {
+      String fieldName = field.getName();
+      Object value = convertField(genericRecord.get(fieldName), field.getSchema());
+      if (value instanceof LocalDate) {
```
albertshau: instanceof is usually a sign that logic is happening in the wrong place. The convertField() method should take care of returning the correct physical type and you should just set whatever value you get back.

src/main/java/co/cask/bigquery/BigQuerySink.java
CuriousVini: removed this method

src/main/java/co/cask/common/AvroToStructuredTransformer.java
CuriousVini: made corresponding changes.

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -120,26 +122,27 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
       if (type.isSimpleType()) {
-        typeName = simpleType(field.getSchema());
+        typeName = getTableDataType(fieldSchema);
       } else if (type == Schema.Type.UNION) {
-        typeName = unionType(field.getSchema());
-      } else if (type == Schema.Type.RECORD) {
-        throw new IllegalArgumentException(
-          String.format("Current implementation of BigQuery sink does not yet support complex records.")
-        );
+        fieldSchema =  fieldSchema.isNullable()? fieldSchema.getNonNullable() : fieldSchema;
```
albertshau: this should replace the existing `if (type.isSimpleType())`. There should be no union type check.

src/main/java/co/cask/bigquery/BigQuerySink.java
```diff
@@ -218,39 +231,50 @@ public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullW
...
-              decodeSimpleTypes(object, name, value, schema);
-              break;
-            }
-          }
-          break;
+      if (!type.isSimpleType()) {
+        throw new RecordConverterException(String.format("Field '%s' is of unsupported type '%s'.", name, type));
       }
+
+      decodeSimpleTypes(object, name, input, field.getSchema());
```
albertshau: pass in fieldSchema, which will not be nullable.

albertshau (approved):
lgtm

CuriousVini:
Added avro 1.7.7 compile time dependency as a separate commit. This is needed because as part of [spanner source PR](https://github.com/data-integrations/google-cloud/pull/2/files#diff-600376dffeb79835ede4a0b285078036R71) , we modified bigquery connector version from `0.10.2-hadoop2` to `0.13.1-hadoop2`. 

However, Bigquery connector changed scope of avro dependency from compile to provided in `0.13.1-hadoop2`. [1][2]. That is why we are adding avro 1.7.7 as compile time dependency.

[1] https://mvnrepository.com/artifact/com.google.cloud.bigdataoss/bigquery-connector/0.10.2-hadoop2
[2] https://mvnrepository.com/artifact/com.google.cloud.bigdataoss/bigquery-connector/0.13.1-hadoop2

albertshau (approved):
lgtm

CuriousVini:
squashing and merging.

PR #5, opened 2018-08-30
making credentials and project id optional

gsps1:
Making serviceFilePath and projectId fields optional in the configs for in GCS, BigQuery source and sinks and actions GCS path create and delete.

We will try to get the project id from environment if its not provided and if we cannot get the project id we will throw an exception during prepareRun.

Tested the above plugins, however noticed a different issue while testing GCS Avro sink and opened a JIRA for that https://issues.cask.co/browse/CDAP-14229.

### DIFF
--- a/src/main/java/co/cask/gcs/GCPUtil.java
+++ b/src/main/java/co/cask/gcs/GCPUtil.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


src/main/java/co/cask/gcs/GCPUtil.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class GCPUtil {
+
+  /**
+   * If project id is not provided and cannot be detected from environment, throw an exception indicating that.
+   */
+  public static void validateProjectId(@Nullable String project) {
+    if (project == null && ServiceOptions.getDefaultProjectId() == null) {
+      throw new IllegalArgumentException(
+        "Google Cloud ProjectId is not provided and unable to auto detect the project id from the runtime environment");
```
albertshau: Could not detect Google Cloud project id from the environment. Please specify a project id.

albertshau: If you make this return the non-null project id you can avoid extra calls to ServiceOptions.getDefaultProjectId()

albertshau (commented):
minor comments

albertshau (approved):
lgtm

gsps1:
rebased, merging.

PR #6, opened 2018-08-30
Improve the docs for BigQuery and GCS Action plugins

albertshau:
Added information to the reference docs for BigQuery source and
sink and the Google Cloud Storage create and delete actions.
Also standardized the placeholds and labels in the widgets,
and updated the description annotations.

### DIFF
--- a/docs/BigQueryTable-batchsource.md
+++ b/docs/BigQueryTable-batchsource.md
@@ -1,9 +1,41 @@
-### Google BigQuery Table Plugin
+# Google BigQuery Table Source
 
-This plugins exports a bigquery table as source to be ingested into the processing pipeline.
-Plugin requires a service account to access the bigquery table. In order to configure
-the service account visit https://cloud.google.com. Make sure you provide right permissions
-to service account for accessing BigQuery API.
+Description
+-----------
+This source reads the entire contents of a BigQuery table.
+BigQuery is Google's serverless, highly scalable, enterprise data warehouse.
+Data from the BigQuery table is first exported to a temporary location on Google Cloud Storage,
+then read into the pipeline from there.
 
-The service account key file needs to be installed on all individual machines on the cluster
-and have to be readable for all the users running the job.
+Authorization
+-------------
+If the plugin is run on a Google Cloud Dataproc cluster, the service account key does not need to be provided.
+Credentials will be automatically read from the cluster environment.
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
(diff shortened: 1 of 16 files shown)


docs/BigQueryTable-batchsource.md
```diff
@@ -1,9 +1,41 @@
...
 
-This plugins exports a bigquery table as source to be ingested into the processing pipeline.
-Plugin requires a service account to access the bigquery table. In order to configure
-the service account visit https://cloud.google.com. Make sure you provide right permissions
-to service account for accessing BigQuery API.
+Description
+-----------
+This source reads the entire contents of a BigQuery table.
+BigQuery is Google's serverless, highly scalable, enterprise data warehouse.
+Data is first written to a temporary location on Google Cloud Storage,
```
gsps1: `Data from the bigquery table is written to`

docs/GCSBucketCreate-action.md
```diff
@@ -1,3 +1,40 @@
-## Google Cloud Storage Create
+# Google Cloud Storage Create
 
-This plugin is used for creating directories on Google Cloud Storage (GCS).
+Description
```
gsps1, line 3: unrelated to this PR : ideally these plugins should be named GCS Object Create/Delete, GCS Bucket create/delete is misleading as we are not creating/deleting buckets.

docs/GCSBucketCreate-action.md
```diff
@@ -1,3 +1,40 @@
...
+**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**Service Account File Path**: Path on the local file system of the service account key used for
+authorization. Does not need to be specified when running on a Dataproc cluster.
+When running on other clusters, the file must be present on every node in the cluster.
+
+**Bucket Name**: The bucket to create objects in.
+
+**Objects to Create**: The objects to create in the bucket.
```
gsps1, line 38: will be good to give an example for this as the backend expects it in comma separated format.

gsps1 (approved):
minor comments, lgtm otherwise

docs/GCSBucketCreate-action.md
albertshau: yeah... I also thought they just create/delete buckets at first. but renaming break compatibility. I think it's ok as long as the label is correct.

docs/GCSBucketCreate-action.md
albertshau: For our reference docs, we're writing all new ones from the perspective of a UI user, who doesn't know anything about the underlying format.

PR #7, opened 2018-09-04
Adding date time support to spanner source

CuriousVini:
**Summary**
Spanner only supports DATE and TIMESTAMP data types. DATE is a local date without time zone. TIMESTAMP has nano second level precision. However, cdap only supports micro second level precision. So this PR reduces precision from nanoseconds to microseconds.


albertshau (approved):
lgtm

CuriousVini:
Fixed typo in comment, merging PR.

PR #8, opened 2018-09-04
Skip using annotated name for service account variable 

gsps1:
Due to https://issues.cask.co/browse/CDAP-14258, the service account file path variable was null in getSchema endpoint. 

Skipping using the annotated name for serviceAccountFilePath and relevant update to the Widget JSON .

### DIFF
--- a/src/main/java/co/cask/spanner/common/SpannerConfig.java
+++ b/src/main/java/co/cask/spanner/common/SpannerConfig.java
@@ -54,15 +54,13 @@ public class SpannerConfig extends ReferencePluginConfig {
   @Macro
   public String database;
 
-  @Name("serviceFilePath")
   @Description("Path on the local file system to the service account key used for access to Spanner. " +
     "Does not need to be specified when running on a Dataproc cluster. " +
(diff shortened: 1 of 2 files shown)


src/main/java/co/cask/spanner/common/SpannerConfig.java
```diff
@@ -51,15 +51,13 @@
   @Macro
   public String database;
 
-  @Name("serviceFilePath")
   @Description("Path on the local file system to the service account key used for access to Spanner. " +
     "Does not need to be specified when running on a Dataproc cluster. " +
     "When running on other clusters, the file must be present on every node in the cluster.")
   @Macro
   @Nullable
   public String serviceAccountFilePath;
```
albertshau: should rename the variable to serviceFilePath, otherwise it will be inconsistent with other plugins.

gsps1: @albertshau updated it per your suggestion, please take a look again. thanks

albertshau (approved):
lgtm

PR #9, opened 2018-09-05
CDAP-14059 add spanner sink plugin

gsps1:
Spanner sink plugin to write to Google Cloud Spanner. 

Batch size of writes is configurable through the plugin config. The records are converted to mutations and buffered in record writer until the batch size and then written to Spanner table after it exceeds the batch size.

### DIFF
--- a/src/main/java/co/cask/spanner/sink/SpannerOutputFormat.java
+++ b/src/main/java/co/cask/spanner/sink/SpannerOutputFormat.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.spanner.sink;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.gcs.GCPUtil;
+import co.cask.spanner.SpannerConstants;
+import co.cask.spanner.common.SpannerUtil;
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.Spanner;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.OutputCommitter;
+import org.apache.hadoop.mapreduce.OutputFormat;
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Spanner output format
+ */
+public class SpannerOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerOutputFormat.class);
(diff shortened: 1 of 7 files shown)


src/main/java/co/cask/spanner/sink/SpannerOutputFormat.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Spanner output format
+ */
+public class SpannerOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(SpannerOutputFormat.class);
+
+  @Override
+  public RecordWriter<NullWritable, StructuredRecord>
+  getRecordWriter(TaskAttemptContext context) throws IOException, InterruptedException {
```
albertshau: I think it's more readable if getRecordWriter(TaskAttemptContext context) is on the previous line and throws is on this line (unless it doesn't fit?)

src/main/java/co/cask/spanner/sink/SpannerSink.java
```diff
@@ -0,0 +1,104 @@
...
+   */
+  public SpannerSink(SpannerSinkConfig config) {
+    super(config);
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    config.validate();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
```
albertshau: this is not required for a sink

src/main/java/co/cask/spanner/sink/SpannerSink.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public void prepareRun(BatchSinkContext context) throws Exception {
+    config.validate();
+    Configuration configuration = new Configuration();
+    String projectId = GCPUtil.getProjectId(config.project);
+    configuration.set(SpannerConstants.PROJECT_ID, projectId);
+    if (config.serviceFilePath != null) {
+      configuration.set(SpannerConstants.SERVICE_ACCOUNT_FILE_PATH, config.serviceFilePath);
+    }
+    configuration.set(SpannerConstants.INSTANCE_ID, config.instance);
```
albertshau: The convention in OutputFormats (InputFormats too) is to have a static method in the OutputFormat that takes a Configuration and other set of arguments. The static method is responsible for setting the values in the Configuration. That way, logic is encapsulated in that single class. 

I think it makes sense to follow that patten here and add a `SpannerOutputFormat.configure(Configuration conf, SpannerSinkConfig spannerConf)` method

src/main/java/co/cask/spanner/sink/SpannerSink.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    super.initialize(context);
+    config.validate();
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable,
```
albertshau: nit: better to keep the variable all on one line, let's move Emitter down to the next line

src/main/java/co/cask/spanner/sink/SpannerSinkConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Name("table")
+  @Description("Cloud Spanner table id. Uniquely identifies your table within the Cloud Spanner database")
+  @Macro
+  public String table;
+
+  @Name("batchSize")
+  @Description("Size of the batched writes to the Spanner table. " +
+    "When the number of buffered mutations is greater than this batchSize, " +
+    "the mutations are written to Spanner table, Default value is 100")
```
albertshau, line 40: how was 100 chosen? Is there a number Spanner recommends?

albertshau (commented):
minor comments, good job it looks very clean.

src/main/java/co/cask/spanner/sink/SpannerSinkConfig.java
gsps1: Each write to Cloud Spanner contains some overhead. To maximize bulk write throughput, maximize the amount of data stored per write. A good technique is for each commit to mutate hundreds of rows. Commits with the number of mutations in the range of 1 MiB - 5 MiB rows usually provide the best performance.

Commits with more than 5 MiB mutations don't provide extra benefit, and they risk exceeding the Cloud Spanner limits on commit size and mutations per commit.

https://cloud.google.com/spanner/docs/bulk-loading

I chose hundred as a lower estimate, since they mentioned its a good technique to mutate 100s of rows.

albertshau: I see. It would be good to include that information in the doc.

gsps1:
@albertshau thanks for the review, addressed comments, please take a look again.

albertshau (approved):
lgtm

PR #15, opened 2018-09-14
performing a bunch of cleanup

albertshau:
Fixed the cdap-maven-plugin to use a released version instead of
a snapshot version.

Moved everything to the co.cask.gcp package. This is to avoid
conflicts with other projects, especially from the co.cask.common
package.

Removed a bunch of unused and duplicate code in the file source.


CuriousVini (approved):
LGTM 👍

PR #16, opened 2018-09-19
CDAP-13274 Add FLL in BigQuery source and sink

rohitsinha54:
Emit FLL from BiqQuery Source and Sink. 

Source
image

Operations to Sink: 
image

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
+++ b/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
@@ -25,10 +25,13 @@ import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.format.UnexpectedFormatException;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.lineage.field.EndPoint;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldWriteOperation;
 import co.cask.gcp.common.GCPUtils;
 import com.google.cloud.bigquery.Field;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
@@ -53,11 +56,13 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 /**
  * This class <code>BigQuerySink</code> is a plugin that would allow users
@@ -149,9 +154,8 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
       temporaryGcsPath,
       BigQueryFileFormat.NEWLINE_DELIMITED_JSON,
       TextOutputFormat.class);
-
     
-    context.addOutput(Output.of(config.table.replace("-", "_").replace(".", "_"), new OutputFormatProvider() {
+    context.addOutput(Output.of(config.referenceName, new OutputFormatProvider() {
       @Override
       public String getOutputFormatClassName() {
         return IndirectBigQueryOutputFormat.class.getName();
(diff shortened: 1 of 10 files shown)


src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -164,6 +169,22 @@ public String getInputFormatClassName() {
...
       }
     }));
+
+    if (config.schema != null) {
+      try {
+        Schema schema = Schema.parseJson(config.schema);
+        if (schema.getFields() != null) {
+          FieldOperation operation =
+            new FieldReadOperation("Read", "Read from BigQuery table",
+                                   EndPoint.of(context.getNamespace(), config.table),
```
albertshau: these plugins should be using a reference name. Is that coming in a separate PR?

albertshau (commented):
one question about reference name. If that's coming later, lgtm.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -150,8 +155,8 @@ public void prepareRun(BatchSinkContext context) throws Exception {
       BigQueryFileFormat.NEWLINE_DELIMITED_JSON,
       TextOutputFormat.class);
 
-    
-    context.addOutput(Output.of(config.table.replace("-", "_").replace(".", "_"), new OutputFormatProvider() {
+    String outputName = config.table.replace("-", "_").replace(".", "_");
```
chtyim: What is this for?

chtyim: Nevermind, this is used as entity name.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -164,6 +169,22 @@ public String getInputFormatClassName() {
         return config;
       }
     }));
+
```
rohitsinha54, line 172: @albertshau In the lines above this line 157 where we setInput should we use referenceName now?

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -166,6 +171,15 @@ public String getOutputFormatClassName() {
         return config;
       }
     }));
+
+    if (!fields.isEmpty()) {
+      // Record the field level WriteOperation
+      FieldOperation operation = new FieldWriteOperation("Write", "Wrote to BigQuery table.",
+                                                         EndPoint.of(context.getNamespace(), outputName),
```
rohitsinha54: @albertshau Here also use the referenceName rather than the table name?

widgets/BigQueryTable-batchsink.json
```diff
@@ -7,6 +7,11 @@
     {
       "label" : "Service Account and Temporary Google Storage",
       "properties" : [
+        {
+          "widget-type": "textbox",
+          "label": "Reference Name",
+          "name": "referenceName"
```
albertshau, line 13: there should be a placeholder in the widget attributes that is consistent with what the other plugins use.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
albertshau: yes

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
albertshau: yes

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
albertshau: this is still using the table name and not the reference name.

widgets/BigQueryTable-batchsink.json
```diff
@@ -7,6 +7,14 @@
     {
       "label" : "Service Account and Temporary Google Storage",
       "properties" : [
+        {
+          "widget-type": "textbox",
+          "label": "Reference Name",
+          "name": "referenceName",
+          "widget-attributes" : {
+            "placeholder": "This will be used to uniquely identify this source/sink for lineage, annotating metadata, etc."
```
albertshau: hm for some reason I thought the other plugins had placeholder text here but that doesn't appear to be the case. Sorry, can we remove it so it's consistent.

src/main/java/co/cask/gcp/bigquery/BigQuerySourceConfig.java
```diff
@@ -19,12 +19,12 @@
...
 import co.cask.cdap.api.annotation.Macro;
 import co.cask.cdap.api.annotation.Name;
-import co.cask.gcp.common.GCPConfig;
+import co.cask.gcp.publisher.GCPReferenceConfig;
 
 /**
  * Class description here.
  */
-public final class BigQuerySourceConfig extends GCPConfig {
+public final class BigQuerySourceConfig extends GCPReferenceConfig {
```
albertshau: GCPReferenceConfig has 'source/sink' in its description, which is bad. Can you create a source config and a sink config where the source config says 'source' and the sink config says 'sink'.

albertshau (approved):
lgtm

PR #17, opened 2018-09-20
CDAP-13274 Emit FLL from Spanner source and sinks

rohitsinha54:
- Emit FLL from Spanner source and sink

Source:
image

Sink (Operations screenshot)
image

PR #21, opened 2018-09-21
CDAP-13274 Emit FLL from Spanner source and sinks

rohitsinha54:
PR was merged against develop here: https://github.com/data-integrations/google-cloud/pull/17

### DIFF
--- a/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
+++ b/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
@@ -24,15 +24,18 @@ import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
 import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.lineage.field.EndPoint;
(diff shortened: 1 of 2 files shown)


src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
```diff
@@ -132,6 +137,16 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
     // set input format and pass configuration
     batchSourceContext.setInput(Input.of(config.referenceName,
                                          new SourceInputFormatProvider(SpannerInputFormat.class, configuration)));
+    if (config.getSchema() != null) {
+      if (config.getSchema().getFields() != null) {
+        FieldOperation operation =
+          new FieldReadOperation("Read", "Read from Spanner table",
```
anwar6953, line 143: Sometimes the description has a trailing dot. Other times, it doesn't.
Let's make them consistent.

anwar6953 (approved):

src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
rohitsinha54: Lets do it in an independent PR against 0.11 so that can be pulled to develop directly.

PR #24, opened 2018-09-21
CDAP-14224 combine gcs sinks into a single sink

albertshau:
Combining the sinks so that it matches the File sink in
core-plugins. Also refactoring the logic so that it lives
in a file FileFormat class instead of in the BatchSink itself.
This is to prepare for moving the format logic into a common
module that all the file based sources and sinks can depend on.

Further down the line, when formats become actual plugins, the
new FileFormat classes can be re-used.

Also fixing the parent versions to have 5.0.0 as a lower bound.
This is because the plugins use the field level lineage APIs that
were only introduced in CDAP 5.0.0.


src/main/java/co/cask/gcp/format/output/DelimitedTextOutputProvider.java
```diff
@@ -0,0 +1,45 @@
...
+  private final String constantDelimiter;
+
+  public DelimitedTextOutputProvider(@Nullable String delimiter) {
+    this.constantDelimiter = delimiter;
+  }
+
+  @Override
+  public FileOutputFormatter<NullWritable, Text> create(Map<String, String> properties, @Nullable Schema schema) {
+    String delimiter = constantDelimiter == null ? properties.get("delimiter") : constantDelimiter;
+    if (delimiter == null) {
```
rohitsinha54: by using getOrDefault() rather than get() this if can be avoided.

rohitsinha54 (approved):
One minor comment other that LGTM.

PR #25, opened 2018-09-25
CDAP-14338 consolidate terminology and style

albertshau:
Grouped properties into Basic, Advanced, and Credentials sections
with consistent styling for descriptions and placeholders.

Changed any select dropdown for true/false into radio buttons.

Introduced an 'auto-detect' setting for project and credentials
instead of the previous behavior, which expected users to leave
them blank. This is because we found that most users expect to
have to fill out non-advanced properties. In addition, defaulted
the values for those fields to 'auto-detect', as the best user
experience is to run these plugins on dataproc clusters instead
of distributed account keys to every node on a cluster.

Refactored classes to use the same plugin config class hierarchy.

Removed settings from GCS file source for ignoring non existent
input, as it was trying to create some empty directory in a
'tmp' bucket that the user most certainly does not have access to.
Also removing the regex as it didn't seem like it was doing
anything.

Removed useless AbstractFileBatchSource and FileSourceConfig that
were just copied from another project.

Made sure properties show up in the same order in the reference
doc as they do in the plugin widget.

Fixed a bunch of warnings.


docs/BigQueryTable-batchsink.md
```diff
@@ -20,22 +21,24 @@ must be readable by all users running the job.
 
 Properties
 ----------
-**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
-It can be found on the Dashboard in the Google Cloud Platform Console.
+**Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
-**Service Account File Path**: Path on the local file system of the service account key used for
```
rohitsinha54, line 26: Is this not needed at all anymore?

docs/BigQueryTable-batchsink.md
```diff
@@ -20,22 +21,24 @@ must be readable by all users running the job.
...
 
-**Table**: The table to write to. A table contains individual records organized in rows.
+**Table**: Table to write to. A table contains individual records organized in rows.
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 
-**Bucket Name**: The Google Cloud Storage bucket to store temporary data in.
+**Temporary Bucket Name**: Google Cloud Storage bucket to store temporary data in.
 It will be automatically created if it does not exist, but will not be automatically deleted.
 Temporary data will be deleted after it is loaded into BigQuery.
```
rohitsinha54, line 38: It said earlier that it will not be deleted automatically. Is this line correct? Does it mean data will be deleted but not the bucket?

src/main/java/co/cask/gcp/gcs/source/GCSSource.java
```diff
@@ -46,77 +54,143 @@
...
+    @Description("If specified, each output record will include a field with this name that contains the file URI " +
+      "that the record was read from. Requires a customized version of CombineFileInputFormat, so it cannot be used " +
+      "if an inputFormatClass is given.")
+    private String pathField;
 
-    @Description("Name of the bucket.")
-    @Macro
-    public String bucket;
+    @Nullable
+    @Description("Output schema. If a Path Field is set, it must be present in the schema as a string.")
```
rohitsinha54, line 153: This schema seems to be used nowhere

rohitsinha54 (approved):
Few questions beside that LGTM

docs/BigQueryTable-batchsink.md
albertshau: It just got moved lower, to match the order in the widget

docs/BigQueryTable-batchsink.md
albertshau: I will clarify. The bucket will not be deleted, but the data will be.

src/main/java/co/cask/gcp/gcs/source/GCSSource.java
albertshau: yeah I'm not sure why it was there, but it will be used once we move formats here.

PR #27, opened 2018-09-26
CDAP-14273 Remove bucket config from GCS source/sinks

rohitsinha54:
- Remove the bucket config from GCS Source and Sinks.
- Remove the need of gs:// in path

Issue: https://issues.cask.co/browse/CDAP-14273

### DIFF
--- a/src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
+++ b/src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.gcp.gcs;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * A helper class for GCS configs
+ */
+public final class GCSConfigHelper {
+
+  private static final String FILE_SYSTEM_PATH_PREFIX = "gs://";
+
+  /**
+   * Returns the given path as it is if it starts with {@link GCSConfigHelper#FILE_SYSTEM_PATH_PREFIX} else appends it
+   * in the beginning.
+   *
+   * @param path the path
+   * @return {@link URI} for the given path
+   * @throws IllegalArgumentException if the given path is not valid
+   */
+  public static URI getPath(String path) {
+    try {
+      URI uri = new URI(path);
+      if (uri.getScheme() != null && (!uri.getScheme().equalsIgnoreCase("gs") || uri.getAuthority() == null)) {
+        throw new IllegalArgumentException(String.format("Invalid path '%s'. The path must be of form " +
+                                                           "'gs://<bucket-name>/path'.", path));
+      }
+      if (uri.getScheme() == null) {
+        return new URI(FILE_SYSTEM_PATH_PREFIX + uri.getSchemeSpecificPart());
+      }
+      return uri;
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException(String.format("Invalid path '%s'. The path must be of form " +
+                                                         "'gs://<bucket-name>/path'.", path), e);
+    }
+  }
+
+  /**
+   * Returns the bucket name from the given path
+   *
+   * @param path the path from which bucket name needs to be extracted
+   * @return the bucket name
+   */
+  public static String getBucket(String path) {
+    return getPath(path).getAuthority();
+  }
+}
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -35,8 +35,6 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 For example, the format 'yyyy-MM-dd-HH-mm' will result in a directory of the form '2015-01-01-20-42'.
 If not specified, nothing will be appended to the path."
 
-**Bucket Name**: Bucket to write to.
-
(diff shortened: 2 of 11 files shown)


src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Returns the given path as it is if it starts with {@link GCSConfigHelper#FILE_SYSTEM_PATH_PREFIX} else appends it
+   * in the beginning.
+   *
+   * @param path the path
+   * @return the path with which starts with {@link GCSConfigHelper#FILE_SYSTEM_PATH_PREFIX}
+   */
+  public static String getPath(String path) {
+    return path.startsWith(FILE_SYSTEM_PATH_PREFIX) ? path : FILE_SYSTEM_PATH_PREFIX + path;
```
albertshau: If somebody puts in 'gs:/abc/', this will turn it into 'gs://gs:/abc/'. If somebody puts in '/abc', this will turn into 'gs:///abc'.

I think you can return a URI instead of a String. If it can't be parsed as a URI, an exception will be thrown. If the scheme is not null and not 'gs', throw an exception. If the authority is null (ex: gs:/abc), throw an exception. If the scheme is null, create a new URI with "gs" as the scheme and the schemeSpecificPart from the current URI.

The authority for the URI will then be the bucket and you don't need that pattern matching logic.

albertshau (commented):
please remove bucket from the reference docs as well

src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
```diff
@@ -25,17 +25,31 @@
...
+   * @return {@link URI} for the given path
+   * @throws IllegalArgumentException if the given path is not valid
    */
-  public static String getPath(String path) {
-    return path.startsWith(FILE_SYSTEM_PATH_PREFIX) ? path : FILE_SYSTEM_PATH_PREFIX + path;
+  public static URI getPath(String path) {
+    try {
+      URI uri = new URI(path);
+      if (uri.getScheme() != null && !uri.getScheme().equalsIgnoreCase("gs")) {
+        throw new URISyntaxException(path, String.format("Invalid scheme. Scheme must be '%s'.",
```
albertshau: I think for any of these exceptions, the message can be:
`Invalid path. The path must be of the form 'gs://<bucket>/path'.`

src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
```diff
@@ -25,17 +25,31 @@
...
+                                                         FILE_SYSTEM_PATH_PREFIX));
+      }
+      if (uri.getScheme() != null && uri.getAuthority() == null) {
+        throw new URISyntaxException(path, "Invalid bucket name in path.");
+      }
+      if (uri.getScheme() == null) {
+        return new URI(FILE_SYSTEM_PATH_PREFIX + uri.getSchemeSpecificPart());
```
albertshau, line 44: I think you can do `new URI("gs", uri.getSchemeSpecificPart())`

albertshau (commented):
minor comments

src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
rohitsinha54: This constructor is not there. The other one needs a fragment which can be null although since it treats the second part as schemeSpecificPart and doesn't deal with authority it does not append //

src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
```diff
@@ -36,12 +36,9 @@
...
       URI uri = new URI(path);
-      if (uri.getScheme() != null && !uri.getScheme().equalsIgnoreCase("gs")) {
-        throw new URISyntaxException(path, String.format("Invalid scheme. Scheme must be '%s'.",
-                                                         FILE_SYSTEM_PATH_PREFIX));
-      }
-      if (uri.getScheme() != null && uri.getAuthority() == null) {
-        throw new URISyntaxException(path, "Invalid bucket name in path.");
+      if (uri.getScheme() != null && (!uri.getScheme().equalsIgnoreCase("gs") || uri.getAuthority() == null)) {
+        throw new URISyntaxException(path, String.format("Invalid path '%s'. The path must be of form " +
+                                                           "'gs://<bucket-name>/path'.", FILE_SYSTEM_PATH_PREFIX));
```
albertshau: wrong substitution

albertshau: also if you throw a URISyntaxException, it will be caught by the outer catch and the message will be lost. Should just throw the IllegalArgumentException directly.

Should also use this message if the path can't be parsed at all.

rohitsinha54: That is what I was trying to achieve. Since I include the exception in the illegal argument exception it will preserve the message.

albertshau: It's included as the cause. Which means the message will be different. The message is what we care most about, it is what shows up in the UI when we validate the field and pipeline deployment fails.

albertshau: speaking of which, can you make sure this is called during validation

src/main/java/co/cask/gcp/gcs/GCSConfigHelper.java
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param path the path
+   * @return {@link URI} for the given path
+   * @throws IllegalArgumentException if the given path is not valid
+   */
+  public static URI getPath(String path) {
+    try {
+      URI uri = new URI(path);
+      if (uri.getScheme() != null && (!uri.getScheme().equalsIgnoreCase("gs") || uri.getAuthority() == null)) {
+        throw new URISyntaxException(path, "Scheme must be 'gs' and a bucket name must be specified.");
```
albertshau: this should use the same error message. As a user, I won't know what to do with this error message. What's a scheme? Where does the bucket go?

albertshau: this isn't actually a syntax exception. Just throw the IllegalArgumentException with the same message, as that's what this is -- an illegal user input.

albertshau (approved):
lgtm

PR #28, opened 2018-09-27
CDAP-14400 fix pubsub sink for auto-detect credentials

albertshau:
Also fixing the widget label while we're at it.


rohitsinha54:
Code changes LGTM. Although, I don't see which change fixes the widget label?

albertshau:
that was a bad commit message, i'll remove it.

rohitsinha54:
LGTM

PR #29, opened 2018-09-27
CDAP-14401 auto create pubsub topic if it doesnt exist

albertshau:
Enhanced the pubsub sink to create the topic if it doesn't exist.
This will prevent users from running lengthy pipelines just to
realize at the end that the topic needs to be created.

### DIFF
--- a/src/main/java/co/cask/gcp/publisher/GooglePublisher.java
+++ b/src/main/java/co/cask/gcp/publisher/GooglePublisher.java
@@ -33,11 +33,20 @@ import co.cask.cdap.etl.api.lineage.field.FieldOperation;
 import co.cask.cdap.etl.api.lineage.field.FieldWriteOperation;
 import co.cask.cdap.format.StructuredRecordStringConverter;
 import co.cask.gcp.common.GCPReferenceSinkConfig;
+import co.cask.gcp.common.GCPUtils;
 import co.cask.hydrator.common.batch.sink.SinkOutputFormatProvider;
+import com.google.api.gax.rpc.AlreadyExistsException;
+import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.cloud.pubsub.v1.TopicAdminClient;
+import com.google.cloud.pubsub.v1.TopicAdminSettings;
+import com.google.pubsub.v1.ProjectTopicName;
+import com.google.pubsub.v1.Topic;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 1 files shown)


src/main/java/co/cask/gcp/publisher/GooglePublisher.java
```diff
@@ -66,8 +74,33 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-  public void prepareRun(BatchSinkContext context) {
+  public void prepareRun(BatchSinkContext context) throws IOException {
     config.validate();
+
+    TopicAdminSettings.Builder topicAdminSettings = TopicAdminSettings.newBuilder();
+    String serviceAccountPath = config.getServiceAccountFilePath();
+    if (serviceAccountPath != null) {
+      topicAdminSettings.setCredentialsProvider(() -> GCPUtils.loadServiceAccountCredentials(serviceAccountPath));
+    }
+    TopicAdminClient topicAdminClient = TopicAdminClient.create(topicAdminSettings.build());
```
rohitsinha54: We should call close() on the topicAdminClient or use with try-with-resource. The class documentation says:

```
 * <p>Note: close() needs to be called on the topicAdminClient object to clean up resources such as
 * threads. In the example above, try-with-resources is used, which automatically calls close().
```

src/main/java/co/cask/gcp/publisher/GooglePublisher.java
```diff
@@ -66,8 +74,33 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+
+    String projectId = config.getProject();
+    ProjectTopicName projectTopicName = ProjectTopicName.of(projectId, config.topic);
+
+    try {
+      topicAdminClient.getTopic(projectTopicName);
+    } catch (NotFoundException e) {
```
rohitsinha54: Is NotFoundException guaranteed to be  be thrown if it does not exists?

src/main/java/co/cask/gcp/publisher/GooglePublisher.java
albertshau: good call, fixed

src/main/java/co/cask/gcp/publisher/GooglePublisher.java
albertshau: from my experiments, this is how it behaves. Doesn't seem to return null. The client looks like an auto-generated one, so it maps pretty directly to the rest endpoint.

rohitsinha54:
LGTM.

PR #30, opened 2018-09-27
CDAP-14013 Create External Dataset with Schema

rohitsinha54:
- Create External Dataset with Schema. 

Issue: https://issues.cask.co/browse/CDAP-14013

### DIFF
--- a/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
+++ b/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
@@ -24,16 +24,14 @@ import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
 import co.cask.cdap.api.dataset.lib.KeyValue;
-import co.cask.cdap.api.lineage.field.EndPoint;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.api.batch.BatchSourceContext;
-import co.cask.cdap.etl.api.lineage.field.FieldOperation;
-import co.cask.cdap.etl.api.lineage.field.FieldReadOperation;
 import co.cask.gcp.spanner.SpannerConstants;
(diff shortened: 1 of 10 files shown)


src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
```diff
@@ -133,17 +129,18 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
     configuration.set(SpannerConstants.SPANNER_BATCH_TRANSACTION_ID, getSerializedObjectString(batchTransactionId));
     configuration.set(SpannerConstants.PARTITIONS_LIST, getSerializedObjectString(partitions));
 
+    co.cask.hydrator.common.LineageRecorder lineageRecorder =
+      new co.cask.hydrator.common.LineageRecorder(batchSourceContext, config.referenceName);
+    lineageRecorder.createExternalDataset(config.getSchema());
```
rohitsinha54, line 134: @albertshau This does not seem right. We write the JSON string to the topic so we should have the schema as just body string. Do you agree?

src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
```diff
@@ -133,17 +129,18 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
     configuration.set(SpannerConstants.SPANNER_BATCH_TRANSACTION_ID, getSerializedObjectString(batchTransactionId));
     configuration.set(SpannerConstants.PARTITIONS_LIST, getSerializedObjectString(partitions));
 
+    co.cask.hydrator.common.LineageRecorder lineageRecorder =
```
albertshau: do you need the full classname?

src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
albertshau: was this comment meant for the PubSub publisher? JSON implies a schema, so I think it is still correct to register the full schema. It could be used by a PubSub source, for example, to read the message and turn it into a StructuredRecord with that schema.

rohitsinha54: Yes. Okay cool.

src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
rohitsinha54: Fixed.

albertshau (approved):
lgtm

PR #31, opened 2018-09-27
CDAP-14392 Validate Bigquery sink output schema

CuriousVini:
If field type does not match with big query column:
image

If schema has more columns than big query table:
image

If required field is missing:
image

Passes if nullable fields are missing.

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
+++ b/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
@@ -29,14 +29,22 @@ import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.gcp.common.GCPUtils;
 import co.cask.hydrator.common.LineageRecorder;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryOptions;
 import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.FieldList;
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.cloud.bigquery.Table;
+import com.google.cloud.bigquery.TableId;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryOutputConfiguration;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
 import com.google.cloud.hadoop.io.bigquery.output.IndirectBigQueryOutputFormat;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import org.apache.hadoop.conf.Configuration;
@@ -77,6 +85,24 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
   private static final Logger LOG = LoggerFactory.getLogger(BigQuerySink.class);
   private static final DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
 
+  private static final Map<Schema.Type, LegacySQLTypeName> TYPE_MAP = ImmutableMap.<Schema.Type,
+    LegacySQLTypeName>builder()
+    .put(Schema.Type.INT, LegacySQLTypeName.INTEGER)
+    .put(Schema.Type.LONG, LegacySQLTypeName.INTEGER)
+    .put(Schema.Type.STRING, LegacySQLTypeName.STRING)
+    .put(Schema.Type.FLOAT, LegacySQLTypeName.FLOAT)
+    .put(Schema.Type.DOUBLE, LegacySQLTypeName.FLOAT)
+    .put(Schema.Type.BOOLEAN, LegacySQLTypeName.BOOLEAN)
+    .put(Schema.Type.BYTES, LegacySQLTypeName.BYTES)
+    .build();
+
+  private static final Map<Schema.LogicalType, LegacySQLTypeName> LOGICAL_TYPE_MAP = ImmutableMap.of(
+    Schema.LogicalType.DATE, LegacySQLTypeName.DATE,
+    Schema.LogicalType.TIME_MILLIS, LegacySQLTypeName.TIME, Schema.LogicalType.TIME_MICROS, LegacySQLTypeName.TIME,
+    Schema.LogicalType.TIMESTAMP_MILLIS, LegacySQLTypeName.TIMESTAMP,
+    Schema.LogicalType.TIMESTAMP_MICROS, LegacySQLTypeName.TIMESTAMP
+  );
+
   private final BigQuerySinkConfig config;
   private Schema schema;
   private Configuration configuration;
@@ -89,6 +115,7 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
+    validateSchema();
     uuid = UUID.randomUUID();
     Job job = Job.getInstance();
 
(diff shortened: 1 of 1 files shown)


src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -65,6 +70,8 @@
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import static co.cask.gcp.common.GCPUtils.loadServiceAccountCredentials;
```
albertshau: our convention is to import the class and not have static imports

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -189,6 +196,8 @@ public String getOutputFormatClassName() {
                                                            .collect(Collectors.toList()));
       context.record(Collections.singletonList(operation));
     }
+
+    validateSchema(config);
```
rohitsinha54: this validation should happen as early as possible. Here for example we already tried to processed the schema and emitted the fll

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
     }
   }
+
+  /**
+   * Validates output schema against bigquery table schema. It throws {@link IllegalArgumentException}
+   * if the output schema has more fields than bigquery table or output schema field types does not match bigquery
+   * column types.
+   */
+  private void validateSchema(BigQuerySinkConfig config) throws IOException {
+    BigQuery bigquery;
```
albertshau: do you need to close this?

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
         break;
     }
   }
+
+  /**
+   * Validates output schema against bigquery table schema. It throws {@link IllegalArgumentException}
+   * if the output schema has more fields than bigquery table or output schema field types does not match bigquery
+   * column types.
+   */
+  private void validateSchema(BigQuerySinkConfig config) throws IOException {
```
rohitsinha54: We generally put validation of schema and other config properties in the config class itself

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
albertshau: I'm not sure if the lineage is recorded right when it's called above or if its recorded later. But in either case, it's safer to do this validation earlier to avoid any possible side-effects if the schema is invalid.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+    }
+
+    com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
+    if (bqSchema == null) {
+      // Table is created without schema, so no further validation is required.
+      return;
+    }
+
+    FieldList bqFields = bqSchema.getFields();
+    List<Schema.Field> outputSchemaFields = Schema.parseJson(config.schema).getFields();
```
albertshau: the config class should have a getSchema() method so that the plugin doesn't have to handle scenarios where the schema is malformed and can just operate on the Schema object.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+
+    if (bqFields.size() != outputSchemaFields.size()) {
```
albertshau: Is this too stringent? If the output schema has fewer fields, and the missing fields are nullable in the BQ table, are you able to insert?

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
rohitsinha54: @albertshau Can this validation also be done in configurePipeline so that we fail at deployment? Is that possible?

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+      throw new IllegalArgumentException(
+        "Output schema does not match BigQuery table schema. Output schema must be compatible with " +
```
albertshau: We should be able to give a more detailed message. If there is a required column in BQ that is not in the output schema, the message should look something like:

The output schema does not match the BigQuery table schema for '[dataset].[table]'. The table requires column 'xyz', which is not in the output schema.

If the output schema contains a field that is not in the BQ table:

The output schema does not match the BigQuery table schema for '[dataset].[table]'. The table does not contain the 'xyz' column.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+
+    // validate type of fields against bigquery column types
+    for (Schema.Field field : outputSchemaFields) {
+      Schema fieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
+      Schema.Type type = fieldSchema.getType();
+      Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+
+      if (logicalType != null) {
+        if (logicalTypeMap.get(logicalType) == null) {
+          throw new IllegalArgumentException(String.format("Type %s is not supported", logicalType));
```
albertshau: include the field name in the error too.

Field 'abc' is of unsupported type 'logical type'

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+        }
+        if (logicalTypeMap.get(logicalType) != bqFields.get(field.getName()).getType()) {
+          throw new IllegalArgumentException(String.format("Type of field %s does not match column type " +
```
albertshau: Field 'abc' of type 'type' is not compatible with column 'abc' in BigQuery table 'dataset.table' of type 'type'. It must be of type 'required type'.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+                                                             "%s of BigQuery table. ",
+                                                           fieldSchema.getLogicalType(), config.table));
+        }
+      }
+
+      if (typeMap.get(type) == null) {
+        throw new IllegalArgumentException(String.format("Type %s is not supported", type));
```
albertshau: Field 'abc' is of unsupported type 'type'

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -442,4 +451,104 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+      }
+
+      if (typeMap.get(type) != bqFields.get(field.getName()).getType()) {
+        throw new IllegalArgumentException(String.format("Type of field %s does not match column type " +
```
albertshau: similar comment here

albertshau (commented):
mostly comments about the error messages

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
rohitsinha54: It will be later. It is collected here but emitted to TMS in destroy of Workflow. Since the invalid schema will fail the pipeline we will not have emitted fll to tms

albertshau: We generally do not make calls with clients at configure time and only validate things that will never change. The reason is that we don't want to stop people from deploying pipelines before all their dependencies are setup.

rohitsinha54: 👍

CuriousVini:
Resolved conflicts.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
CuriousVini: I dont think so.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
CuriousVini: yea, it is too stringent. However it need not be that strict check. will fix it.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
CuriousVini: I have kept it in the same class because it uses BigQuery client to verify the schema. In future if client side requires some changes, it is there in one place. Also in hydrator-plugins we validate schema in the same class as plugin in some plugins: https://github.com/caskdata/hydrator-plugins/blob/develop/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java#L230

rohitsinha54: When this class calls validate it can query and pass the schema of actual table as para. The config validation logic should reside in the config IMO. In Hydrator where it is done outside of the config is not a good practice and should not be used as an example for new code.

albertshau: In general for new code we've been putting a validate() method in the config that can be used to do any validation. There are a few reasons it seems to be easier in the config:

Validation logic can be easily unit tested by simply creating the config rather than going through the pipeline creation
Properties required for validation are right there in the config and don't need to be accessed through getters
Which fields need containsMacro() checks is easier to see within the config where everything is annotated
If another config extends that config, they can inherit the validation logic.

These things are only applicable if everything you need for validation resides in the config. If that's the case, I would move it to the config.

CuriousVini: I agree @rohitsinha54 @albertshau that we should have validate methods in config. However, in this particular case, we are verifying the compatibility of the output schema fields with bigquery columns. This means the validation is against bigquery fields and does not use many config variables. It just performs some checks on provided outputSchemaFields and bqFields. Thats why the logic was part of the plugin.

However, we can still move some logic to config as below.
``` java
 private void validateSchema(BigQuerySinkConfig config) throws IOException {
    Table table = getBigqueryTable(config);
    if (table == null) {
      // Table does not exist, so no further validation is required.
      return;
    }

    com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
    if (bqSchema == null) {
      // Table is created without schema, so no further validation is required.
      return;
    }

    FieldList bqFields = bqSchema.getFields();
    List<Schema.Field> outputSchemaFields = config.getSchema().getFields();
    // Match compatibility of output schema fields with bigquery table fields
    config.matchSchemaCompatibility(outputSchemaFields, bqFields);
  }
``` 
and in config class
```java
/**
   * Matches the compatibility of schema fields with bigquery column types.
   *
   * @param schemaFields list of {@link Schema.Field}s to match compatibility with bigquery fields.
   * @param bqFields List of bigquery table fields
   * @throws IllegalArgumentException If output schema has fields that are not present in {@link FieldList} or
   * if one of the required fields is missing or type of {@link Schema.Field} does not match {@link Field} type.
   */
  public void matchSchemaCompatibility(List<Schema.Field> schemaFields, FieldList bqFields) {
}
```

what do you think?

albertshau: I don't think there's a need to split it up. Don't think this is a big deal let's keep it as is

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -89,6 +96,7 @@ public BigQuerySink(BigQuerySinkConfig config) {
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
+    validateSchema(config);
```
albertshau: this method shouldn't take the config as an argument, it's already accessible

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -423,4 +431,148 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+    }
+
+    FieldList bqFields = bqSchema.getFields();
+    List<Schema.Field> outputSchemaFields = config.getSchema().getFields();
+
+    // Output schema should not have fields that are not present in BigQuery table.
+    List<String> diff = getFieldDiff(bqFields, outputSchemaFields);
+    if (!diff.isEmpty()) {
+      throw new IllegalArgumentException(
+        String.format("Output schema does not match the BigQuery table schema for '%s.%s' table. " +
```
albertshau: for consistency, 'Output' -> 'The output'

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -423,4 +431,148 @@ private void decodeSimpleTypes(JsonObject json, String name, StructuredRecord in
...
+    }
+
+    for (Schema.Field field : outputSchemaFields) {
+      diff.remove(field.getName());
+    }
+    return diff;
+  }
+
+  private void matchSchema(BigQuerySinkConfig config, FieldList bqFields, List<Schema.Field> outputSchemaFields) {
+    Map<Schema.Type, LegacySQLTypeName> typeMap = new HashMap<>();
```
albertshau: these maps can be constants of the class instead of created in this method. You can use ImmutableMap, as guava is a direct dependency.

albertshau (commented):
minor comments

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -519,21 +520,22 @@ private Table getBigqueryTable(BigQuerySinkConfig config) throws IOException {
...
-    typeMap.put(Schema.Type.BOOLEAN, LegacySQLTypeName.BOOLEAN);
-    typeMap.put(Schema.Type.BYTES, LegacySQLTypeName.BYTES);
-
-    Map<Schema.LogicalType, LegacySQLTypeName> logicalTypeMap = new HashMap<>();
-    logicalTypeMap.put(Schema.LogicalType.DATE, LegacySQLTypeName.DATE);
-    logicalTypeMap.put(Schema.LogicalType.TIME_MILLIS, LegacySQLTypeName.TIME);
-    logicalTypeMap.put(Schema.LogicalType.TIME_MICROS, LegacySQLTypeName.TIME);
-    logicalTypeMap.put(Schema.LogicalType.TIMESTAMP_MILLIS, LegacySQLTypeName.TIMESTAMP);
-    logicalTypeMap.put(Schema.LogicalType.TIMESTAMP_MICROS, LegacySQLTypeName.TIMESTAMP);
+    Map<Schema.Type, LegacySQLTypeName> typeMap = ImmutableMap.<Schema.Type, LegacySQLTypeName>builder()
```
albertshau: By constant I meant these can be private static final in the class so that we don't need to build them when the method is called.

albertshau (approved):
just one minor comment, otherwise lgtm. fix and merge

CuriousVini:
Fixed, squashing and merging.

PR #32, opened 2018-09-28
feature/CDAP-14185 Add Google Cloud PubSub Streaming Source

rohitsinha54:
### Summary
- Adds Google PubSub Streaming source.
- It uses [Apache Bahir](https://github.com/apache/bahir/tree/master/streaming-pubsub) to create DStreams.
- It creates a `SparkGCPCredentials` from the service json file if one is provided and if not then it will use [Application Default Credentials](https://cloud.google.com/docs/authentication/production) Note: ADC part needs to be tested to work yet but since plugin is realtime its pending.
- FLL is not supported for Realtime Plugins so this does not include FLL.

Issue: https://issues.cask.co/browse/CDAP-14185

### DIFF
--- a/docs/GoogleSubscriber-streamingsource.md
+++ b/docs/GoogleSubscriber-streamingsource.md
@@ -1,0 +1,38 @@
+# Google Cloud PubSub Streaming Source
+
+Description
+-----------
+This sources reads from a Google Cloud Pub/Sub subscription in realtime.
+Cloud Pub/Sub brings the scalability, flexibility, and reliability of enterprise message-oriented
+middleware to the cloud. By providing many-to-many, asynchronous messaging that decouples senders and receivers,
+it allows for secure and highly available communication between independently written applications.
+
+Credentials
+-----------
+If the plugin is run on a Google Cloud Dataproc cluster, the service account key does not need to be
+provided and can be set to 'auto-detect'.
+Credentials will be automatically read from the cluster environment.
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access Google Cloud Pub/Sub.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**Project ID**: Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**Subscription**: Name of the Google Cloud PubSub subscription to subscribe.
+If the subscription needs to be created then the topic to which the subscription will belong must be provided.
+
+**Topic**: Name of the Google Cloud PubSub topic to subscribe to. If a topic is provided and the given subscriber
+does not exists it will be created. If a subscriber does not exists and is created only the messages arrived after
+the creation of subscriber will be received.
+
+**Service Account File Path**: Path on the local file system of the service account key used for
+authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
+When running on other clusters, the file must be present on every node in the cluster.
--- a/pom.xml
+++ b/pom.xml
@@ -396,6 +396,13 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- Start: dependency for Google PubSub Streaming Source -->
+    <dependency>
+      <groupId>org.apache.bahir</groupId>
+      <artifactId>spark-streaming-pubsub_2.11</artifactId>
+      <version>2.2.1</version>
+    </dependency>
+    <!-- End: dependency for Google PubSub Streaming Source -->
   </dependencies>
 
   <build>
@@ -445,6 +452,7 @@
                 org.apache.avro.mapreduce;
                 parquet.avro.*;
                 parquet.hadoop.*;
+                org.apache.spark.streaming.pubsub*;
               </_exportcontents>
             </instructions>
           </configuration>
--- a/src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
+++ b/src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.gcp.publisher.source;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.streaming.StreamingContext;
+import co.cask.cdap.etl.api.streaming.StreamingSource;
+import co.cask.gcp.common.GCPReferenceSourceConfig;
+import com.google.cloud.Timestamp;
+import org.apache.spark.api.java.function.Function;
+import org.apache.spark.storage.StorageLevel;
+import org.apache.spark.streaming.api.java.JavaDStream;
+import org.apache.spark.streaming.api.java.JavaReceiverInputDStream;
+import org.apache.spark.streaming.pubsub.PubsubUtils;
+import org.apache.spark.streaming.pubsub.SparkGCPCredentials;
+import org.apache.spark.streaming.pubsub.SparkGCPCredentials.Builder;
+import org.apache.spark.streaming.pubsub.SparkPubsubMessage;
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -445,6 +452,7 @@
                 org.apache.avro.mapreduce;
                 parquet.avro.*;
                 parquet.hadoop.*;
+                org.apache.spark.streaming.*;
```
albertshau: It doesn't seem like this would be needed. Were you getting errors without it?

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+    return pubSubMessages.map((Function<SparkPubsubMessage, StructuredRecord>) pubSubMessage ->
+      StructuredRecord.builder(DEFAULT_SCHEMA)
+        .set("message", Bytes.toString(pubSubMessage.getData()))
+        .set("id", pubSubMessage.getMessageId())
+        .set("timestamp", pubSubMessage.getPublishTime()).build());
+  }
+
+
+  public static class SubscriberConfig extends GCPReferenceSourceConfig {
+    @Name("subscriber")
```
albertshau: no need to have this annotation if it's the same as the variable name

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+    @Description("Cloud Pub/Sub subscriber to subscribe. If a subscriber does not exists and needs to be create then " +
+      "a topic must be specified to which the subscriber will belong.")
+    @Macro
+    private String subscriber;
```
albertshau: 'subscription' seems to be the correct term, based on the cloud console. Please update the reference doc and widget too.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+      StructuredRecord.builder(DEFAULT_SCHEMA)
+        .set("message", Bytes.toString(pubSubMessage.getData()))
+        .set("id", pubSubMessage.getMessageId())
+        .set("timestamp", pubSubMessage.getPublishTime()).build());
+  }
+
+
+  public static class SubscriberConfig extends GCPReferenceSourceConfig {
+    @Name("subscriber")
+    @Description("Cloud Pub/Sub subscriber to subscribe. If a subscriber does not exists and needs to be create then " +
```
albertshau: Cloud Pub/Sub subscription to read from. If the subscription does not exist it will be automatically created if a topic is specified. Messages published before the subscription was created will not be read.

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -0,0 +1,65 @@
...
+          "widget-type": "textbox",
+          "label": "Service Account File Path",
+          "name": "serviceFilePath",
+          "widget-attributes" : {
+            "default": "auto-detect"
+          }
+        }
+      ]
+    }
+  ],
```
albertshau, line 57: there is some non-editable schema widget that will force the schema to some value in the  UI. Check the twitter source in hydrator-plugins for an example.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+      "a topic must be specified to which the subscriber will belong.")
+    @Macro
+    private String subscriber;
+
+    @Description("Cloud Pub/Sub topic to subscribe messages from. If a topic is provided and the given subscriber " +
```
albertshau: Cloud Pub/Sub topic to create the subscription from. The topic is only used if the subscription does not already exist.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+      LOG.debug("Building credentials from specified service account file.");
+      credentials = new Builder().jsonServiceAccount(serviceAccountFilePath).build();
+    } else {
+      // this will use the application default credential which is available when running on google cloud
+      LOG.debug("No service account file specified. Building credentials using Application Default Credentials.");
+      credentials = new Builder().build();
+    }
+
+    JavaReceiverInputDStream<SparkPubsubMessage> pubSubMessages =
+      PubsubUtils.createStream(streamingContext.getSparkStreamingContext(), config.getProject(), config.topic,
```
albertshau, line 89: what happens if the topic is not actually the topic associated with the subscription? Is it simply ignored because the subscription already exists?

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+@Plugin(type = StreamingSource.PLUGIN_TYPE)
+@Name("GoogleSubscriber")
+@Description("Streaming Source to read messages from Google PubSub.")
+public class GoogleSubscriber extends StreamingSource<StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(GoogleSubscriber.class);
+  private static final Schema DEFAULT_SCHEMA =
+    Schema.recordOf("event",
+                    Schema.Field.of("message", Schema.of(Schema.Type.STRING)),
+                    Schema.Field.of("id", Schema.of(Schema.Type.STRING)),
+                    Schema.Field.of("timestamp", Schema.of(Schema.Type.STRING))
```
albertshau: you can use the timestamp logical type.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -0,0 +1,109 @@
...
+
+    JavaReceiverInputDStream<SparkPubsubMessage> pubSubMessages =
+      PubsubUtils.createStream(streamingContext.getSparkStreamingContext(), config.getProject(), config.topic,
+                               config.subscriber, credentials, StorageLevel.MEMORY_ONLY());
+
+    return pubSubMessages.map((Function<SparkPubsubMessage, StructuredRecord>) pubSubMessage ->
+      StructuredRecord.builder(DEFAULT_SCHEMA)
+        .set("message", Bytes.toString(pubSubMessage.getData()))
+        .set("id", pubSubMessage.getMessageId())
+        .set("timestamp", pubSubMessage.getPublishTime()).build());
```
albertshau: we should make use of the new timestamp logical types and use setTimestamp()

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -0,0 +1,65 @@
...
+          "widget-type": "textbox",
+          "label": "Project ID",
+          "name": "project",
+          "widget-attributes" : {
+            "default": "auto-detect"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Subscriber Name",
```
albertshau: Subscription

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -0,0 +1,65 @@
...
+          "name": "subscriber",
+          "widget-attributes" : {
+            "placeholder": "Subscriber to subscribe to"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "PubSub topic",
```
albertshau: Topic

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -0,0 +1,65 @@
...
+          "name": "topic",
+          "widget-attributes" : {
+            "placeholder": "Topic name for the subscriber"
```
albertshau: Topic to create the subscription from if it does not already exist

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -0,0 +1,65 @@
...
+          "widget-attributes" : {
+            "default": "auto-detect"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Subscriber Name",
+          "name": "subscriber",
+          "widget-attributes" : {
+            "placeholder": "Subscriber to subscribe to"
```
albertshau: Subscription to read from

docs/GoogleSubscriber-streamingsource.md
```diff
@@ -0,0 +1,38 @@
+# Google Cloud PubSub Streaming Source
+
+Description
+-----------
+This sources reads from a Google Cloud Pub/Sub topic in realtime.
```
albertshau: Pub/Sub subscription

albertshau (commented):
Can you also open a jira to add formats to this plugin (and also to the sink)? For example, if somebody uses the PubSub sink, they should be able to read it out using this PubSub source without an additional stage for parsing the json messages.

pom.xml
rohitsinha54: Yes. Without these I get class not found exception for org.apache.spark.streaming.pubsub.PubsubReceiver. 
I can be try being more specific to pubsub package level

widgets/GoogleSubscriber-streamingsource.json
rohitsinha54: Thanks.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
rohitsinha54: I updated with the earlier message which you suggested.

src/main/java/co/cask/gcp/publisher/source/GoogleSubscriber.java
rohitsinha54: Yes.

rohitsinha54: Yes.

rohitsinha54:
@albertshau Addressed all comments and opened https://issues.cask.co/browse/CDAP-14419 for format support. Please have a look.

The warn message changes which I discussed in person will be done another PR following right after.

albertshau (approved):
lgtm. Please open a jira for improving the performance of this and moving off of bahir.

rohitsinha54:
Opened a jira to use DirectStream: https://issues.cask.co/browse/CDAP-14438

PR #33, opened 2018-10-01
CDAP-14187 refactor bigquery source and sink

CuriousVini:
**Summary**
- Added output schema validation for bigquery source
- Added `BigQueryUtils` as a common class to reduce code redundancy. 
- Organized source and sink's prepareRun method to make it more readable. 
- Removed `RecordConverterException` to make exceptions consistent with other plugins.


src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -180,116 +114,121 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
-      lineageRecorder.recordWrite("Write", "Wrote to BigQuery table.",
-                                  fields.stream().map(BigQueryTableFieldSchema::getName).collect(Collectors.toList()));
     }
   }
 
-  @Override
-  public void initialize(BatchRuntimeContext context) throws Exception {
-    super.initialize(context);
-    schema = config.getSchema();
+  private String getBucket() {
```
albertshau: most people don't expect a getter to modify state. It would be better to name this 'setBucket' and pass in the Configuration object.

I also don't think that Configuration should be a class variable. It should only live in the prepareRun() method, otherwise it's confusing when it is null and when it is non-null.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -144,35 +100,17 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     AvroBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
     AvroBigQueryInputFormat.setEnableShardedExport(configuration, false);
     BigQueryConfiguration.configureBigQueryInput(configuration, config.getProject(), config.dataset, config.table);
+
+    JobID jobID = JobID.forName(String.format("job_%s-%s-%s_%s", context.getNamespace(),
```
albertshau: what is this and the next line for? Seems like it gets a temp path but never uses it?

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
-      bigqueryBuilder.setProjectId(project);
-      bigquery = bigqueryBuilder.build().getService();
 
-      TableId id = TableId.of(project, request.dataset, request.table);
-      Table table = bigquery.getTable(id);
       com.google.cloud.bigquery.Schema bgSchema = table.getDefinition().getSchema();
+      List<Schema.Field> fields = getSchemaFields(bgSchema);
+      return Schema.recordOf("output", fields);
+    } catch (Exception e) {
+      throw new Exception(e.getMessage(), e);
```
albertshau: doesn't look like there's a reason to catch this at all, just let it propagate.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+  /**
+   * Validate output schema. This is needed because its possible that output schema is set without using
+   * {@link #getSchema(BigQuerySourceConfig)} method.
+   */
+  private void validateOutputSchema() throws IOException {
+    Table table = BigQueryUtils.getBigqueryTable(config.getServiceAccountFilePath(), config.getProject(),
+                                                 config.dataset, config.table);
+    if (table == null) {
+      // Table does not exist
+      throw new IllegalArgumentException(String.format("BigQuery table %s does not exist.", config.table));
```
albertshau: Use 'dataset.table' in these error messages to make it consistent with the other plugin and to give more precise information.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+    com.google.cloud.bigquery.Schema bgSchema = table.getDefinition().getSchema();
+    if (bgSchema == null) {
+      throw new IllegalArgumentException(String.format("BigQuery table %s must not have null schema.", config.table));
+    }
+
+    // Output schema should not have more fields than BigQuery table
+    List<String> diff = BigQueryUtils.getOutputSchemaMinusBqFields(config.getSchema().getFields(),
+                                                                   bgSchema.getFields());
+    if (!diff.isEmpty()) {
+      throw new IllegalArgumentException(String.format("Output schema field(s) %s do(es) not exist in BigQuery " +
```
albertshau: Reformat this to be consistent with the error message in the other plugin

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+    Table table = BigQueryUtils.getBigqueryTable(config.getServiceAccountFilePath(), config.getProject(),
+                                                 config.dataset, config.table);
+    if (table == null) {
+      // Table does not exist
+      throw new IllegalArgumentException(String.format("BigQuery table %s does not exist.", config.table));
+    }
+
+    com.google.cloud.bigquery.Schema bgSchema = table.getDefinition().getSchema();
+    if (bgSchema == null) {
+      throw new IllegalArgumentException(String.format("BigQuery table %s must not have null schema.", config.table));
```
albertshau: what does it mean when a big query table has null schema? Does it mean there is data in it but that data has no structure? What does a select * query return in these cases?

A better error message would be 'Cannot read from table [dataset].[table] because it has no schema.' It avoids the "null" language and makes it clear that the table is valid big query table, just the plugin cannot read from it.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+        schema = Schema.of(Schema.Type.BYTES);
+      } else if (value == StandardSQLTypeName.TIME) {
+        schema = Schema.of(Schema.LogicalType.TIME_MICROS);
+      } else if (value == StandardSQLTypeName.DATE) {
+        schema = Schema.of(Schema.LogicalType.DATE);
+      } else if (value == StandardSQLTypeName.TIMESTAMP || value == StandardSQLTypeName.DATETIME) {
+        schema = Schema.of(Schema.LogicalType.TIMESTAMP_MICROS);
+      } else {
+        // this should never happen
+        throw new UnsupportedTypeException(String.format("Big Query Sql type %s is not supported", value));
```
albertshau: BigQuery column '%s' is of unsupported type '%s'.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+      }
+
+      if (field.getMode() == null || field.getMode() == Field.Mode.NULLABLE) {
+        fields.add(Schema.Field.of(field.getName(), Schema.nullableOf(schema)));
+      } else if (field.getMode() == Field.Mode.REQUIRED) {
+        fields.add(Schema.Field.of(field.getName(), schema));
+      } else if (field.getMode() == Field.Mode.REPEATED) {
+        throw new UnsupportedTypeException("Repeated type is not supported");
```
albertshau: BigQuery column '%s' is of unsupported mode 'repeated'.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -221,63 +159,128 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
+    return fields;
+  }
+
+  private String getBucket() {
+    String bucket = config.bucket;
+    if (config.bucket == null) {
+      bucket = uuid.toString();
+      // By default, this option is false, meaning the job can not delete the bucket. So enable it only when bucket name
+      // is not provided.
+      configuration.setBoolean("fs.gs.bucket.delete.enable", true);
```
albertshau: similar comment about renaming this method and passing in the Configuration

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -180,116 +114,121 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
-      public Map<String, String> getOutputFormatConfiguration() {
-        Map<String, String> config = new HashMap<>();
-        for (Map.Entry<String, String> entry : configuration) {
-          config.put(entry.getKey(), entry.getValue());
+  @Override
+  public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullWritable>> emitter) throws Exception {
+    List<Schema.Field> fields = config.getSchema().getFields();
+    JsonObject object = new JsonObject();
+    for (Schema.Field field : fields) {
+      BigQueryUtils.validateSimpleType(field);
```
albertshau: this validation is not required at transform time. It should happen once at configure time

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+
+
+  /**
+   * Validates if schema of the provided field is simple type. This is because BigQuery plugins only support simple
+   * types. It retrieves non-nullable type from field, if field is of nullable type.
+   *
+   * @param field field for which schema needs to be validated
+   * @throws IllegalArgumentException if field type is not simple type
+   */
+  static void validateSimpleType(Schema.Field field) {
```
albertshau: this is a poor validation. Other types are not supported right? I thought floats and ints don't exist in BigQuery?

albertshau: Though it does make sense for the sink to support floats and ints, it doesn't make sense for the source to have them. This indicates that each plugin should validate differently. This is the type of static validation that should be in a validate() method of the corresponding plugin config

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+   *
+   * @param serviceAccountFilePath service account file path
+   * @param project BigQuery project ID
+   * @param dataset dataset for the BigQuery table
+   * @param table BigQuery table
+   * @return returns BigQuery table
+   * @throws IOException if not able to load credentials
+   */
+  @Nullable
+  static Table getBigqueryTable(@Nullable String serviceAccountFilePath, String project,
```
albertshau: nit: Bigquery -> BigQuery

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+                                String dataset, String table) throws IOException {
+    BigQuery bigquery;
+    BigQueryOptions.Builder bigqueryBuilder = BigQueryOptions.newBuilder();
+    if (serviceAccountFilePath != null) {
+      bigqueryBuilder.setCredentials(loadServiceAccountCredentials(serviceAccountFilePath));
+    }
+    if (project == null) {
```
albertshau: project is not annotated as nullable. As long as plugins use the getProject() method, this should be handled by the logic in the plugin config.

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Match output schema with BigQuery table.
+   *
+   * @param bqFields bigquery table fields
+   * @param outputSchemaFields output schema fields
+   * @param dataset dataset name
+   * @param table table name
+   */
+  static void matchSchema(FieldList bqFields, List<Schema.Field> outputSchemaFields, String dataset, String table) {
```
albertshau: matchSchema -> validateSchemaMatches

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Get difference of output schema fields and big query table fields. The operation is equivalent to
+   * (Names of outputSchema fields - Names of bigQuery table fields).
+   *
+   * @param outputSchemaFields output schema fields
+   * @param bqFields bigquery table fields
+   * @return list of remaining field names
+   */
+  static List<String> getOutputSchemaMinusBqFields(List<Schema.Field> outputSchemaFields, FieldList bqFields) {
```
albertshau: outputSchema -> schema

src/main/java/co/cask/gcp/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Get difference of big query table fields and output schema fields. The operation is equivalent to
+   * (Names of bigQuery table fields - Names of outputSchema fields).
+   *
+   * @param bqFields bigquery table fields
+   * @param outputSchemaFields output schema fields
+   * @return list of remaining field names
+   */
+  static List<String> getBqFieldsMinusOutputSchema(FieldList bqFields, List<Schema.Field> outputSchemaFields) {
```
albertshau: outputSchema -> schema

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
CuriousVini: yea its not used. Removed.

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
CuriousVini: It can happen if the table is created without any schema and it does not have any data/schema yet.

albertshau (approved):
lgtm

CuriousVini:
Squashing and merging.

PR #36, opened 2018-10-02
CDAP-12852 use common file format classes

albertshau:
Switch to use the common file format classes from format-common.
This adds format as a property to the GCS Source and also adds
the workaround to support parse-as-csv with a header


src/main/java/co/cask/gcp/gcs/source/GCSSource.java
```diff
@@ -93,89 +69,153 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+    @Description("Regular expression that file paths must match in order to be included in the input. "
+      + "The full file path is compared, not just the file name."
+      + "If no value is given, no file filtering will be done. "
+      + "See https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html for more information about "
+      + "the regular expression syntax.")
+    private String fileRegex;
+
     @Nullable
-    @Description("Boolean value to determine if files are to be read recursively from the path. Default is false.")
+    @Description("Whether to recursively read directories within in the input directory. The default is false.")
```
rohitsinha54: extra in

src/main/java/co/cask/gcp/gcs/source/GCSSource.java
```diff
@@ -93,89 +69,153 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
     private Boolean recursive;
 
+    // this is a hidden property that only exists for wrangler's parse-as-csv that uses the header as the schema
+    // when this is true and the format is text, the header will be the first record returned by every record reader
+    @Nullable
+    private Boolean copyHeader;
+
     public GCSSourceConfig() {
-      this.maxSplitSize = 128L;
+      this.maxSplitSize = 128L * 1024 * 1024;
```
rohitsinha54, line 146: Was earlier the split size not being used in MB?

rohitsinha54 (approved):
One nit and a question. 

LGTM

src/main/java/co/cask/gcp/gcs/source/GCSSource.java
albertshau: it was, but that made it inconsistent with the file source, which uses bytes.

src/main/java/co/cask/gcp/gcs/source/GCSSource.java
albertshau: done

PR #41, opened 2018-10-05
CDAP-14473 macro enable schemas

albertshau:
Enabled macros for schemas and added appropriate checks during
validation to make sure schema is not validated if it contains
a macro.

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/BigQuerySourceConfig.java
+++ b/src/main/java/co/cask/gcp/bigquery/BigQuerySourceConfig.java
@@ -29,31 +29,27 @@ import javax.annotation.Nullable;
  * Holds configuration required for configuring {@link BigQuerySource}.
  */
 public final class BigQuerySourceConfig extends GCPReferenceSourceConfig {
-  @Name("dataset")
+  @Macro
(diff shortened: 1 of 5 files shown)


src/main/java/co/cask/gcp/bigquery/BigQuerySourceConfig.java
```diff
@@ -29,31 +29,28 @@
...
-  @Name("bucket")
   @Description("The Google Cloud Storage bucket to store temporary data in. "
     + "It will be automatically created if it does not exist, but will not be automatically deleted. "
     + "Temporary data will be deleted after it has been read. " +
     "If it is not provided, a unique bucket will be created and then deleted after the run finishes.")
-  @Macro
   public String bucket;
 
+  @Macro
   @Name("schema")
```
CuriousVini: remove this as well?

CuriousVini (approved):
lgtm 👍

src/main/java/co/cask/gcp/bigquery/BigQuerySourceConfig.java
albertshau: done

PR #42, opened 2018-10-05
Fix lineage in BQ plugins

CuriousVini:
Lineage should be emitted after `addOutput` or `setInput` because it creates an external dataset.

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
+++ b/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
@@ -121,8 +121,10 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
       BigQueryFileFormat.NEWLINE_DELIMITED_JSON,
       TextOutputFormat.class);
 
-    setOutputFormat(context);
+    // Both emitLineage and setOutputFormat internally try to create an external dataset if it does not already exists.
+    // We call emitLineage before since it creates the dataset with schema which .
     emitLineage(context, fields);
+    setOutputFormat(context);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -112,8 +112,8 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     job.setOutputKeyClass(LongWritable.class);
     job.setOutputKeyClass(Text.class);
 
-    setInputFormat(context);
     emitLineage(context);
+    setInputFormat(context);
```
rohitsinha54, line 118: Let's put a comment about the positioning so that later we can avoid this in refactoring.

rohitsinha54 (approved):

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -121,8 +121,8 @@ public void prepareRun(BatchSinkContext context) throws Exception {
       BigQueryFileFormat.NEWLINE_DELIMITED_JSON,
       TextOutputFormat.class);
 
-    setOutputFormat(context);
     emitLineage(context, fields);
+    setOutputFormat(context);
```
rohitsinha54, line 127: same here

rohitsinha54: Nit: can we add more detail
```
Both emitLineage and setOutputFormat internally try to create an external dataset if it does not already exists. We call emitLineage before since it creates the dataset with schema which .
```

rohitsinha54 (approved):
One nit other than that LGTM.

PR #43, opened 2018-10-05
Fix path macro in gcs sink

CuriousVini:
https://issues.cask.co/browse/CDAP-14488
https://issues.cask.co/browse/CDAP-14482

Before this change pipeline with macro in path was not getting deployed.

image

After the fix for allowing creation of non existing bq dataset:
image


albertshau (approved):
lgtm

albertshau (approved):
lgtm

PR #44, opened 2018-10-05
CDAP-14474 Update GCP Speech transform

rohitsinha54:
Update includes:
- Rename the plugin from SpeechTranslator to SpeechToText as it is not a translator
- Add description to configs
- Fix the check done during the deploy time on the inputSchema to include the audio field and it to be bytes. This was not happening before.
- The schema given by user is output schema and was being used as both input and output earlier
- Updated to pick credential file from environment
- Add Reference doc

Issue: https://issues.cask.co/browse/CDAP-14474


docs/SpeechToText-transform.md
```diff
@@ -0,0 +1,40 @@
...
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access Google Cloud Spanner.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Audio Field:** Name of the input which contains the raw audio data in bytes.
```
albertshau: of the input field

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -66,87 +63,29 @@
...
-          String.format(
-            "Field '%s' should be a byte array.", field.getName()
-          )
-        );
+    Schema inputSchema = configurer.getStageConfigurer().getInputSchema();
+    // If a inputSchema and audioField is available then verify that the schema does contain the given audioField
+    // and that the type is byte array. This will allow to fail fast i.e. during deployment time.
+    if (inputSchema != null && !config.containsMacro("audiofield")) {
+      Schema.Field field = inputSchema.getField(config.audioField);
+      if (field != null && field.getSchema().getType() != Schema.Type.BYTES) {
```
albertshau: if the field is null, that's also an error. File '%s' does not exist in the input schema.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -66,87 +63,29 @@
...
+        throw new IllegalArgumentException(String.format("Field '%s' must exists and should be a byte array.",
```
albertshau: Field '%s' must be of type 'bytes' but is of type '%s'.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
albertshau: this is incorrect, as it doesn't handle nullables correctly. If it's nullable bytes, it will be a union. Need to do something like:
```
Schema.Type fieldType = fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
```

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -218,54 +150,107 @@ private Schema getSchema(Schema outSchema) {
     fields.add(Schema.Field.of("transcript", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
     fields.add(Schema.Field.of("confidence", Schema.nullableOf(Schema.of(Schema.Type.FLOAT))));
     Schema.Field field = Schema.Field.of("speeches",
-                                         Schema.nullableOf(
-                      Schema.arrayOf(
-                        Schema.recordOf("speech", fields)
-                      )
-                    )
-    );
+                                         Schema.nullableOf(Schema.arrayOf(Schema.recordOf("speech", fields))));
```
albertshau: wait... what??? It outputs an array of records??? that's super weird.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -157,46 +96,39 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
         String transcript = alternative.getTranscript();
         StructuredRecord.Builder speech =
-          StructuredRecord.builder(schema.getField("speeches").getSchema().getNonNullable().getComponentSchema());
+          StructuredRecord.builder(recordSchema.getField("speeches").getSchema().getNonNullable().getComponentSchema());
         speech.set("confidence", confidence);
         speech.set("transcript", transcript);
         speechArray.add(speech.build());
       }
     }
     builder.set("speeches", speechArray);
```
albertshau: this should not be hardcoded, it should be a config property that can default to 'speeches'

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -157,46 +96,39 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
     Schema recordSchema = getSchema(schema);
     List<StructuredRecord> speechArray = new ArrayList<>();
```
albertshau: I don't think the output should contain an array of possibilities. It makes it very difficult to chain to anything else. Instead, let's just use the first (most likely) result. We can have optional 'transcriptField' and 'confidenceField' properties that default to 'transcript' and 'confidence'.

We should also have a Get Schema button that takes the input schema and adds these fields

albertshau (commented):
cleanup looks good. I have comments about the existing functionality, which I think should be changed.

rohitsinha54:
It will be good to emit FLL from this transform. I will add that in a separate PR after this one.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  private Schema.Field topConfidenceSchema(String transcriptTextField) {
+    return Schema.Field.of(transcriptTextField, Schema.nullableOf(Schema.of(Schema.Type.STRING)));
+  }
+
+  private static class SpeechTransformConfig extends PluginConfig {
+
+    @Macro
+    private String schema;
```
albertshau: why is this needed? Does it have something to do with the get schema button?

docs/SpeechToText-transform.md
```diff
@@ -0,0 +1,46 @@
...
+language tag. Example: "en-US". See [Language Support](https://cloud.google.com/speech/docs/languages) for a list of
+the currently supported language codes.
+
+**Transcript Parts Field**: The name of the field to store all the different chunks of transcription with all the
+different possibility and their confidence score. Defaults to 'parts'.
+
+**Transcription Text Field**: If a field name is specified then the transcription with highest confidence score will be
+stored as text.
+
+**Schema**: Output schema of the transcription
```
albertshau: don't include this, it's more confusing than it is helpful

docs/SpeechToText-transform.md
```diff
@@ -0,0 +1,46 @@
...
+16000 Hz. If that's not possible, use the native sample rate of the audio source (instead of re-sampling).
+
+**Profanity**: Whether to attempt filtering profanities, replacing all but the initial character in each filtered
+word with asterisks, e.g. "f***". If set to `false`, profanities won't be filtered out.
+
+**Language**: The language of the supplied audio as a [BCP-47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt)
+language tag. Example: "en-US". See [Language Support](https://cloud.google.com/speech/docs/languages) for a list of
+the currently supported language codes.
+
+**Transcript Parts Field**: The name of the field to store all the different chunks of transcription with all the
```
albertshau: Transcript -> Transcription for consistency

albertshau: The field to store the transcription parts. It will be an array of records. Each record in the array represents one part of the full audio data and will contain the transcription and confidence for that part.

docs/SpeechToText-transform.md
```diff
@@ -0,0 +1,46 @@
...
+different possibility and their confidence score. Defaults to 'parts'.
+
+**Transcription Text Field**: If a field name is specified then the transcription with highest confidence score will be
```
albertshau: The field to store the transcription of the full audio data. It is generated using the transcription for each part with the highest confidence.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+
+  @Override
+  public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgumentException {
+    super.configurePipeline(configurer);
+    Schema inputSchema = configurer.getStageConfigurer().getInputSchema();
+    // If a inputSchema and audioField is available then verify that the schema does contain the given audioField
+    // and that the type is byte array. This will allow to fail fast i.e. during deployment time.
+    if (inputSchema != null && !config.containsMacro("audiofield")) {
+      if (inputSchema.getField(config.audioField) == null) {
+        throw new IllegalArgumentException(String.format("Field %s must exists.", config.audioField));
```
albertshau: Field '%s' does not exist in the input schema.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+      builder.setCredentialsProvider(() -> GCPUtils.loadServiceAccountCredentials(config.getServiceAccountFilePath()));
+    }
+    return builder.build();
+  }
+
+  private void setRecognitionConfig() {
+    recognitionConfig = RecognitionConfig.newBuilder()
+      .setEncoding(getAudioEncoding())
+      .setSampleRateHertz(Integer.parseInt(config.sampleRate))
+      .setProfanityFilter(config.profanity.equals("true"))
```
albertshau: it's better to do this type of stuff in getters in the config class. For example, through shouldFilterProfanity() and getLanguage() methods.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+        serviceAccountFilePath.isEmpty() || GCPConfig.AUTO_DETECT.equals(serviceAccountFilePath)) {
+        return null;
+      }
+      return serviceAccountFilePath;
+    }
+
+    @Nullable
+    public String getTranscriptionTextField() {
+      if (containsMacro("transcriptionTextField")) {
+        return "";
```
albertshau: return null

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+      }
+      return transcriptionTextField;
+    }
+
+    @Nullable
+    public String getTranscriptionPartsField() {
+      if (containsMacro("transcriptionPartsField")) {
+        return "";
```
albertshau: return null

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+      topLevel.addAll(outSchema.getFields());
+    }
+    return Schema.recordOf("record", topLevel);
+  }
+
+  /**
+   * Request object from the Plugin REST call.
+   */
+  public static final class Request extends SpeechTransformConfig {
+    private Map<String, Schema> inputSchemas;
```
albertshau: why is this taking multiple input schemas?

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+  public void destroy() {
+    super.destroy();
+    try {
+      speech.close();
+    } catch (Exception e) {
+      // no-op
+    }
+  }
+
+  private Schema getSchema(Schema outSchema, @Nullable String transcriptPartsField,
```
albertshau: why does this generate a Schema from an output schema? Or is this really the input schema?

albertshau: Looking at the logic, this is a really weird method. It seems like it takes as input some potential output schema but also modifies it if it is not correct.

Change it to take an input schema. The input schema should not already have a parts field or a transcription field. Those should simply be added.

widgets/SpeechToText-transform.json
```diff
@@ -62,6 +69,23 @@
...
+            "placeholder" : "Field name to use to store all the different chunks of transcription",
+            "default" : "parts"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Transcription Text Field",
+          "name": "transcriptionTextField",
+          "widget-attributes" : {
+            "placeholder" : "Field name to store the transcription with highest confidence score."
```
albertshau: Can't comment below, but this is using the wrong widget. multiple-inputs should be false. Also, don't label it 'Extract Schema', nothing else calls it that, everything else uses 'Get Schema'

widgets/SpeechToText-transform.json
```diff
@@ -42,15 +42,22 @@
...
-          "name": "profanity",
-          "widget-attributes" : {
-            "values" : [
-              "True",
-              "False"
-            ],
-            "default" : "True"
+          "widget-type": "radio-group",
+          "name" : "profanity",
+          "label" : "Profanity",
```
albertshau: Profanity -> Mask Profanity

widgets/SpeechToText-transform.json
```diff
@@ -62,6 +69,23 @@
...
             ],
             "default" : "en-US"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Transcript Parts Field",
+          "name": "transcriptionPartsField",
+          "widget-attributes" : {
+            "placeholder" : "Field name to use to store all the different chunks of transcription",
```
albertshau: there is no need for a placeholder if there is a default. That being said, I think it makes more sense to have a default for the transcription text field and leave this one empty.

docs/SpeechToText-transform.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Audio Encoding**: Audio encoding of the data sent in the audio message. All encodings support only 1 channel (mono)
+audio. Only 'FLAC' and 'WAV' include a header that describes the bytes of audio that follow the header.
+The other encodings are raw audio bytes with no header.
+
+**Sampling Rate**: Sample rate in Hertz of the audio data sent in all 'RecognitionAudio' messages.
+Valid values are: 8000-48000. 16000 is optimal. For best results, set the sampling rate of the audio source to
+16000 Hz. If that's not possible, use the native sample rate of the audio source (instead of re-sampling).
+
+**Profanity**: Whether to attempt filtering profanities, replacing all but the initial character in each filtered
```
albertshau, line 33: Whether to mask profanity, replacing all but the initial character in each masked word with asterisks. For example, 'f***'.

Don't need the part about if false.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
albertshau: As part of validation, you should be checking that the parts and transcript fields don't exist in the input schema.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+    List<SpeechRecognitionResult> results = response.getResultsList();
+    Schema outputRecordSchema = getSchema(schema, config.getTranscriptionPartsField(),
+                                          config.getTranscriptionTextField());
+
+    StructuredRecord.Builder outputBuilder = StructuredRecord.builder(outputRecordSchema);
+
+    List<StructuredRecord> transcriptParts = new ArrayList<>();
+    StringBuilder completeTranscript = new StringBuilder();
+    for (SpeechRecognitionResult result : results) {
+      if (config.getTranscriptionPartsField() != null && !config.getTranscriptionPartsField().isEmpty()) {
```
albertshau: nit: these long method names make it hard to read, especially since they look very similar. Let's change to getPartsField() and getTextField()

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+        throw new IllegalArgumentException(String.format("Field '%s' must be of type 'bytes' but is of type " +
+                                                           "'%s'.", field.getName(), fieldType));
+      }
+    }
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    schema = config.getSchema();
```
albertshau: This transform should not use a user-supplied schema. That opens the door to all sorts of weirdness. What if they change the type of an input field?

Instead, the schema here should be the output schema, assuming the input schema is not null.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
+    ByteString audioBytes = ByteString.copyFrom((byte[]) input.get(config.audioField));
+    RecognitionAudio audio = RecognitionAudio.newBuilder()
+      .setContent(audioBytes)
+      .build();
+
+    RecognizeResponse response = speech.recognize(recognitionConfig, audio);
+    List<SpeechRecognitionResult> results = response.getResultsList();
+    Schema outputRecordSchema = getSchema(schema, config.getTranscriptionPartsField(),
```
albertshau: You should only need to generate this if the input schema is null, which means each record might have a different schema. Otherwise, you should use the output schema generated in the initialize() method.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
albertshau: The getter should also only return null, not null or empty.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+      speech.set("transcript", transcript);
+      speechArray.add(speech.build());
+    }
+  }
+
+  private Schema.Field transcriptWithConfidenceSchema(String transcriptPartsField) {
+    List<Schema.Field> fields = new ArrayList<>();
+    fields.add(Schema.Field.of("transcript", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    fields.add(Schema.Field.of("confidence", Schema.nullableOf(Schema.of(Schema.Type.FLOAT))));
+    return Schema.Field.of(transcriptPartsField, Schema.nullableOf(Schema.arrayOf(Schema.recordOf("speech", fields))));
```
albertshau: don't use a nullable array. The field should be an empty array if there was no data returned.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  private Schema.Field topConfidenceSchema(String transcriptTextField) {
```
albertshau: there's no need for this one line method, just create the field directly. This just adds an unneeded layer of indirection.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+      encoding = RecognitionConfig.AudioEncoding.FLAC;
+    } else if (config.encoding.equalsIgnoreCase("mulaw")) {
+      encoding = RecognitionConfig.AudioEncoding.MULAW;
+    } else if (config.encoding.equalsIgnoreCase("ogg_opus")) {
+      encoding = RecognitionConfig.AudioEncoding.OGG_OPUS;
+    }
+    return encoding;
+  }
+
+  private void addTranscriptWithConfidence(Schema recordSchema, List<StructuredRecord> speechArray,
```
albertshau: shouldn't need to pass in the record schema. The parts field schema should be known at initialize time. Set it and use it here

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
albertshau: validation should also be in its own method so that it can be called here at configure time, but also at prepare time, when all macros will be evaluated.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    schema = config.getSchema();
+    setRecognitionConfig();
+    speech = SpeechClient.create(getSettings());
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) throws Exception {
```
albertshau: don't throw Exception if you don't need to. It makes it hard to understand which methods can throw errors.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+    ByteString audioBytes = ByteString.copyFrom((byte[]) input.get(config.audioField));
+    RecognitionAudio audio = RecognitionAudio.newBuilder()
+      .setContent(audioBytes)
+      .build();
+
+    RecognizeResponse response = speech.recognize(recognitionConfig, audio);
```
albertshau, line 98: what types of exceptions can this throw? If an exception is due to bad input, it would be better to catch it, emit the record as an error, and move on. But if you get something like an unauthorized exception, then it makes sense to propagate it up so that the pipeline fails.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,332 @@
...
+    @Description("Audio encoding of the data sent in the audio message. All encodings support\n" +
+      "only 1 channel (mono) audio. Only `FLAC` and `WAV` include a header that\n" +
+      "describes the bytes of audio that follow the header. The other encodings\n" +
+      "are raw audio bytes with no header.")
+    private String encoding;
+
+    @Macro
+    @Name("samplerate")
+    @Description("Sample rate in Hertz of the audio data sent in all `RecognitionAudio` messages. Valid values are: " +
+      "8000-48000. 16000 is optimal. For best results, set the sampling rate of the audio source to 16000 Hz. If " +
```
albertshau, line 252: this value needs to be validated then.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
rohitsinha54: I don't think the transform was written initially with much thought lot of weird stuff going on. Have fixed the big ones.

Fixed this one.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
rohitsinha54: the exception is not documented in api. I will need to look into the implementation to see the possibility. I can add that in an independent PR.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,377 @@
...
+    StringBuilder completeTranscript = new StringBuilder();
+
+    Schema partsSchema = null;
+    if (config.getPartsField() != null && !config.getPartsField().isEmpty()) {
+      partsSchema = currentSchema.getField(config.getPartsField()).getSchema().getComponentSchema();
+    }
+
+    for (SpeechRecognitionResult result : results) {
+      if (config.getPartsField() != null && !config.getPartsField().isEmpty()) {
+        addTranscriptWithConfidence(partsSchema, transcriptParts, result.getAlternativesList());
```
rohitsinha54: Did it in this way since outputSchema might not be known in all cases.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -56,79 +56,72 @@
...
-      }
-    }
+    config.validate(configurer.getStageConfigurer().getInputSchema());
   }
 
   @Override
   public void initialize(TransformContext context) throws Exception {
     super.initialize(context);
-    schema = config.getSchema();
+    config.validate(context.getInputSchema());
```
albertshau: don't need to validate in initialize(), but should validate in prepareRun(), which is currently not overridden. prepareRun() is before the run starts and before an initialize() call is made

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
albertshau: the output schema might not be known, but the schema for the parts field will always be known. It should be created at initialize time. Instead of getting it from the outputschema, the getSchema() method should use the field's schema.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,377 @@
...
+    Schema partsSchema = null;
+    if (config.getPartsField() != null && !config.getPartsField().isEmpty()) {
+      partsSchema = currentSchema.getField(config.getPartsField()).getSchema().getComponentSchema();
+    }
+
+    for (SpeechRecognitionResult result : results) {
+      if (config.getPartsField() != null && !config.getPartsField().isEmpty()) {
+        addTranscriptWithConfidence(partsSchema, transcriptParts, result.getAlternativesList());
+      }
+      if (config.getTextField() != null && !config.getTextField().isEmpty()) {
```
albertshau: instead of always checking null or empty, have the getter return null if the value is empty.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  private void copyFields(StructuredRecord input, StructuredRecord.Builder outputBuilder) {
+    // copy other schema field values
+    List<Schema.Field> fields = input.getSchema().getFields();
+    if (fields != null) {
+      for (Schema.Field field : fields) {
+        String name = field.getName();
+        // if an output schema is known and the output schema does not contain the current input field then skip
+        if (outputSchema != null && outputSchema.getField(name) == null) {
```
albertshau: this can't happen right?

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,377 @@
...
+      topLevel.add(Schema.Field.of(transcriptTextField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+      hasTranscriptField = true;
+    }
+
+    if (!hasTranscriptField) {
+      throw new IllegalArgumentException("Either 'Transcript Parts Field' or 'Transcript Text Field' or both must be " +
+                                           "specified.");
+    }
+
+    if (inputSchema != null && inputSchema.getFields() != null) {
```
albertshau: inputSchema can't be null

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,377 @@
...
+      topLevel.addAll(inputSchema.getFields());
```
albertshau: nit: add the existing fields first, with the new fields at the end. This is generally what we've done for new fields.

rohitsinha54:
@albertshau Please have another look.

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,382 @@
...
+
+    @Nullable
+    public Integer getSampleRate() {
+      if (containsMacro("sampleRate")) {
+        return null;
+      }
+      return Integer.parseInt(sampleRate);
+    }
+
+    private void validate(Schema inputSchema) {
```
albertshau: annotate input schema as nullable

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,382 @@
...
+      if (inputSchema != null) {
+        // If a inputSchema and audioField is available then verify that the schema does contain the given audioField
+        // and that the type is byte array. This will allow to fail fast i.e. during deployment time.
+        if (getAudioField() != null && inputSchema.getField(getAudioField()) == null) {
```
albertshau: create an audioFieldName variable to avoid repeated null checks and make it easier to read

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,382 @@
...
+        Schema fieldSchema = field.getSchema();
+        Schema.Type fieldType =
+          fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+        if (fieldType != Schema.Type.BYTES) {
+          throw new IllegalArgumentException(String.format("Field '%s' must be of type 'bytes' but is of type " +
+                                                             "'%s'.", field.getName(), fieldType));
+        }
+        if ((getTextField() == null || getTextField().isEmpty()) &&
+          (getPartsField() == null || getPartsField().isEmpty())) {
+          throw new IllegalArgumentException("Either 'Transcript Parts Field' or 'Transcript Text Field' or both must" +
```
albertshau: At least one of '...' or '...' must be specified.

albertshau (commented):
minor comments

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    if (context.getInputSchema() != null) {
+      outputSchema = getSchema(context.getInputSchema(), config.getPartsField(), config.getTextField());
+    }
+    if (config.getPartsField() != null) {
+      transcriptionPartsSchema = Schema.recordOf("record", transcriptWithConfidenceSchema(config.getPartsField()));
```
albertshau: this shouldn't be a record, it should just be what you get from transcriptWithConfidenceSchema()

src/main/java/co/cask/gcp/speech/SpeechToTextTransform.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+  }
+
+  private void addTranscriptWithConfidence(List<StructuredRecord> speechArray,
+                                           List<SpeechRecognitionAlternative> alternatives) {
+    for (SpeechRecognitionAlternative alternative : alternatives) {
+      float confidence = alternative.getConfidence();
+      String transcript = alternative.getTranscript();
+      StructuredRecord.Builder speech =
+        StructuredRecord.builder(transcriptionPartsSchema.getField(config.getPartsField())
```
albertshau: that way, you don't do a lookup here, you just use the schema directly.

albertshau (approved):
minor comments, otherwise lgtm. fix and merge

albertshau (approved):
lgtm

PR #47, opened 2018-10-10
CDAP-14477 Remove bucket from GCS Create/Delete action

rohitsinha54:
Issue: https://issues.cask.co/browse/CDAP-14477

- Remove Bucket from GCS Create and Delete Action
- Support create delete across multiple buckets by just taking paths

### DIFF
--- a/src/main/java/co/cask/gcp/gcs/actions/GCSBucketCreate.java
+++ b/src/main/java/co/cask/gcp/gcs/actions/GCSBucketCreate.java
@@ -23,6 +23,7 @@ import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.etl.api.action.Action;
 import co.cask.cdap.etl.api.action.ActionContext;
 import co.cask.gcp.common.GCPConfig;
+import co.cask.gcp.gcs.GCSConfigHelper;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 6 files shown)


src/main/java/co/cask/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -126,20 +123,22 @@ public void run(ActionContext context) throws Exception {
...
     @Macro
-    public String failIfExists;
+    private String failIfExists;
+
+    public List<String> getPaths() {
+      return Arrays.stream(paths.split(",")).map(String::trim).collect(Collectors.toList());
+    }
+
+    public boolean getFailIfExists() {
+      return failIfExists.equalsIgnoreCase("yes");
```
albertshau: looks like there's a mismatch between the widget and this, where the widget has True and False but this... expects 'yes'. This is really weird, don't know why this plugin decided to be different. Let's change it from a String into a Boolean.

albertshau (commented):
logic looks good and cleaner. One comment about a pre-existing bug

src/main/java/co/cask/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -126,20 +123,22 @@ public void run(ActionContext context) throws Exception {
...
+    @Description("Comma separated list of objects to be created.")
     @Macro
-    public String paths;
+    private String paths;
 
     @Name("failIfExists")
     @Description("Fail if path exists.")
     @Macro
-    public String failIfExists;
+    private String failIfExists;
```
albertshau: this should be a Boolean now.

rohitsinha54: Oops!!

albertshau (approved):
lgtm

PR #49, opened 2018-10-16
CDAP-14490 auto create spanner table and database if does not exist

CuriousVini:
https://issues.cask.co/browse/CDAP-14490
Table created for the schema with composite primary key - city, zip:
image

image

### DIFF
--- a/docs/Spanner-batchsink.md
+++ b/docs/Spanner-batchsink.md
@@ -30,11 +30,15 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
  Instance is an allocation of resources that is used by Cloud Spanner databases created in that instance.
 
 **Database Name**: Database the Spanner table belongs to.
-Spanner database is contained within a specific Spanner instance.
+Spanner database is contained within a specific Spanner instance. If the database does not exist, it will get created.
 
 **Table Name**: Table to write to. A table contains individual records organized in rows.
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
+If the table does not exist, it will get created.
+
+**Primary Key**: If the table does not exist, a primary key must be provided in order to auto-create the table.
+The key can be a composite key of multiple fields in the schema. This is not required if the table already exists.
 
 **Service Account File Path**: Path on the local file system of the service account key used for
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
--- a/src/main/java/co/cask/gcp/spanner/common/SpannerUtil.java
+++ b/src/main/java/co/cask/gcp/spanner/common/SpannerUtil.java
@@ -26,7 +26,7 @@ import java.io.IOException;
 import java.util.Set;
 
 /**
- * Spanner utility classs to get spanner service
+ * Spanner utility class to get spanner service
  */
 public class SpannerUtil {
   // todo CDAP-14233 - add support for array
@@ -70,5 +70,85 @@ public class SpannerUtil {
         }
       }
     }
+  }
+
+  /**
+   * Converts schema to Spanner create statement
+   *
+   * @param tableName table name to be created
+   * @param primaryKeys comma seperated list of primary keys
+   * @param schema schema of the table
+   * @return Create statement
+   */
+  public static String convertSchemaToCreateStatement(String tableName, String primaryKeys, Schema schema) {
+    StringBuilder createStmt = new StringBuilder();
+    createStmt.append("CREATE TABLE ").append(tableName).append(" (");
+
+    for (Schema.Field field : schema.getFields()) {
+      String name = field.getName();
+      Schema fieldSchema = field.getSchema();
+      fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
+      Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+      String spannerType;
+
+      if (logicalType != null) {
+        switch (logicalType) {
+          case DATE:
+            spannerType = "DATE";
+            break;
+          case TIMESTAMP_MILLIS:
+          case TIMESTAMP_MICROS:
+            spannerType = "TIMESTAMP";
+            break;
+          default:
+            // this should not happen
+            throw new IllegalStateException("Logical type " + logicalType + " is not supported.");
+        }
+        addColumn(createStmt, name, field.getSchema().isNullable(), spannerType);
+        continue;
+      }
+
+      Schema.Type type = fieldSchema.getType();
+      switch (type) {
+        case BOOLEAN:
+          spannerType = "BOOLEAN";
+          break;
+        case STRING:
+          spannerType = "STRING(MAX)";
+          break;
+        case LONG:
+          spannerType = "INT64";
+          break;
+        case DOUBLE:
+          spannerType = "FLOAT64";
+          break;
+        case BYTES:
+          spannerType = "BYTES(MAX)";
+          break;
+        default:
+          throw new IllegalStateException(type.name() + " : Type currently not supported.");
+      }
+      addColumn(createStmt, name, field.getSchema().isNullable(), spannerType);
+    }
+
+    // remove trailing ", "
+    createStmt.deleteCharAt(createStmt.length() - 1)
+      .deleteCharAt(createStmt.length() - 1)
+      .append(")");
+
+    createStmt.append(" PRIMARY KEY (").append(primaryKeys).append(")");
+
+    return createStmt.toString();
+  }
+
+  /**
+   * Add column to create statement with appropriate type.
+   */
+  private static void addColumn(StringBuilder createStmt, String name, boolean isNullable, String spannerType) {
+    createStmt.append(name).append(" ").append(spannerType);
+    if (!isNullable) {
+      createStmt.append(" NOT NULL");
+    }
+    createStmt.append(", ");
   }
 }
--- a/src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
+++ b/src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
@@ -28,14 +28,33 @@ import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.gcp.spanner.common.SpannerUtil;
 import co.cask.hydrator.common.LineageRecorder;
 import co.cask.hydrator.common.ReferenceBatchSink;
 import co.cask.hydrator.common.batch.sink.SinkOutputFormatProvider;
+import com.google.cloud.spanner.Database;
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ErrorCode;
+import com.google.cloud.spanner.Operation;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.Statement;
+import com.google.spanner.admin.database.v1.CreateDatabaseMetadata;
+import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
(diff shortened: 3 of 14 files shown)


docs/Spanner-batchsink.md
```diff
@@ -30,11 +30,12 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
  Instance is an allocation of resources that is used by Cloud Spanner databases created in that instance.
 
 **Database Name**: Database the Spanner table belongs to.
-Spanner database is contained within a specific Spanner instance.
+Spanner database is contained within a specific Spanner instance. If database does not exist, it will get created.
```
albertshau: If the database

docs/Spanner-batchsink.md
```diff
@@ -30,11 +30,12 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
...
 
 **Table Name**: Table to write to. A table contains individual records organized in rows.
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
+Table will get created if it does not exist. The primary key will be the first column in the output schema.
```
albertshau: If the table does not exist, it will get created.

src/main/java/co/cask/gcp/spanner/common/SpannerUtil.java
```diff
@@ -71,4 +71,83 @@ public static void validateSchema(Schema schema) {
...
+      }
+      addColumn(createStmt, name, field.getSchema().isNullable(), spannerType);
+    }
+
+    // remove trailing ", "
+    createStmt.deleteCharAt(createStmt.length() - 1)
+      .deleteCharAt(createStmt.length() - 1)
+      .append(")");
+
+    // If table does not exist, by default first column becomes primary key
```
albertshau: The key should be specified by the user, rather than defaulted to the first field. Most of the time that won't be correct. 

I think we need to add a key property where the user can configure a list of fields for the key.

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
```diff
@@ -79,10 +114,79 @@ public void prepareRun(BatchSinkContext context) {
...
+    // information schema (metadata) of spanner database.
+    ResultSet resultSet = dbClient
+      .singleUse()
+      .executeQuery(Statement.of("SELECT\n" +
+                                   "    t.table_name\n" +
+                                   "FROM\n" +
+                                   "    information_schema.tables AS t\n" +
+                                   "WHERE\n" +
+                                   "    t.table_catalog = '' AND t.table_schema = '' AND\n" +
+                                   "    t.table_name = '" + config.table + "'"));
```
albertshau: there should be some way to do variable substitution with the spanner library rather than concatenating strings. This is subject to sql injection.

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
```diff
@@ -79,10 +114,79 @@ public void prepareRun(BatchSinkContext context) {
...
+      database = op.waitFor().getResult();
+    }
+
+    return database;
+  }
+
+  @Nullable
+  private Database getDatabaseIfPresent(DatabaseAdminClient dbAdminClient) {
+    Database database = null;
+    Page<Database> page = dbAdminClient.listDatabases(config.instance, Options.pageSize(100));
```
albertshau: there's a getDatabase() method that should be used instead.

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
```diff
@@ -68,6 +87,22 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+      Spanner spanner = SpannerUtil.getSpannerService(config.getServiceAccountFilePath(), config.getProject());
+      DatabaseId db = DatabaseId.of(config.getProject(), config.instance, config.database);
+      DatabaseClient dbClient = spanner.getDatabaseClient(db);
+      DatabaseAdminClient dbAdminClient = spanner.getDatabaseAdminClient();
+      // create database
+      Database database = getOrCreateDatabase(dbAdminClient);
+      // create table
+      createTableIfNotPresent(dbClient, database);
+      // Close spanner connection to flush all the changes.
+      spanner.close();
```
albertshau: close should be in a finally block

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
```diff
@@ -68,6 +87,22 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    } catch (IOException e) {
+      throw new RuntimeException("Exception while trying to get Spanner service. ", e);
```
albertshau, line 104: what throws an IOException? Is it just SpannerUtil.getSpannerService()?

If not, this message needs to be changed based on what can throw the exception

src/main/java/co/cask/gcp/spanner/common/SpannerUtil.java
```diff
@@ -71,4 +71,83 @@ public static void validateSchema(Schema schema) {
...
     }
   }
+
+  /**
+   * Converts schema to Spanner create statement
+   * @param tableName table name to be created
+   * @param schema schema of the table
+   * @return Create statement
+   */
+  public static String convertSchemaToCreateStatement(String tableName, Schema schema) {
```
albertshau: can you add some unit tests for this method

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
CuriousVini: `SpannerUtil.getSpannerService` throws it

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
CuriousVini: Good point

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
CuriousVini: It throws SpannerException: `Database does not exist`. We can catch that exception and then create a table. I thought this way would be cleaner. Please let me know if you think otherwise.

albertshau: Can you determine that it was a not found exception through the error code? I would hope that information is in the exception and you don't have to parse a message.

docs/Spanner-batchsink.md
```diff
@@ -30,12 +30,15 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
...
-Spanner database is contained within a specific Spanner instance. If database does not exist, it will get created.
+Spanner database is contained within a specific Spanner instance. If the database does not exist, it will get created.
 
 **Table Name**: Table to write to. A table contains individual records organized in rows.
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
-Table will get created if it does not exist. The primary key will be the first column in the output schema.
+If the table does not exist, it will get created.
+
+**Primary keys**: If the table does not exist, primary keys must be provided for the non existing spanner table. The
```
albertshau: Primary Key

If the table does not exist, a primary key must be provided in order to auto-create the table. The key can be a composite key of multiple fields in the schema. This is not required if the table already exists.

src/main/java/co/cask/gcp/spanner/sink/SpannerSinkConfig.java
```diff
@@ -66,6 +71,18 @@ public void validate() {
...
       throw new IllegalArgumentException("Spanner batch size for writes should be positive");
     }
+    if (!containsMacro("keys")) {
+      if (keys != null) {
+        String[] splitted = keys.split(",");
+        for (String key : splitted) {
+          Schema schema = getSchema();
+          if (schema.getField(key.trim()) == null) {
+            throw new IllegalArgumentException(
+              String.format("Spanner primary key %s must be present in output schema", key));
```
albertshau: nit: primary key '%s'

src/main/java/co/cask/gcp/spanner/sink/SpannerSinkConfig.java
```diff
@@ -66,6 +71,18 @@ public void validate() {
     if (!containsMacro("batchSize") && batchSize != null && batchSize < 1) {
       throw new IllegalArgumentException("Spanner batch size for writes should be positive");
     }
+    if (!containsMacro("keys")) {
+      if (keys != null) {
```
albertshau: this can be moved to the clause above to remove one layer of nesting

src/main/java/co/cask/gcp/spanner/sink/SpannerSinkConfig.java
```diff
@@ -66,6 +71,18 @@ public void validate() {
...
+        String[] splitted = keys.split(",");
+        for (String key : splitted) {
+          Schema schema = getSchema();
```
albertshau: the schema might be null if it contains a macro and needs to be checked:

```
if (keys != null && !containsMacro("keys") && !containsMacro("schema")) {
  Schema schema = getSchema();
  ...
```

src/test/java/co/cask/gcp/spanner/SpannerSinkConfigTest.java
```diff
@@ -0,0 +1,60 @@
...
+                                    Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
+                                    Schema.Field.of("name", Schema.of(Schema.Type.STRING)),
+                                    Schema.Field.of("price", Schema.of(Schema.Type.DOUBLE)),
+                                    Schema.Field.of("dt", Schema.nullableOf(Schema.of(Schema.LogicalType.DATE))),
+                                    Schema.Field.of("bytedata", Schema.of(Schema.Type.BYTES)),
+                                    Schema.Field.of("timestamp",
+                                                    Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS))));
+
+    SpannerSinkConfig config = new SpannerSinkConfig();
+    config.keys = "zip, name";
```
albertshau: This is bad practice from existing code to have these variables as public. Can you change them to private, and add a constructor for setting the fields. I imagine you'll need to add getter methods too.

widgets/Spanner-batchsink.json
```diff
@@ -46,6 +46,15 @@
           "widget-attributes" : {
             "placeholder": "Table name"
           }
+        },
+        {
+          "widget-type": "csv",
+          "label": "Primary keys",
```
albertshau: Primary Key

src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
CuriousVini: yea I think they do have error code for not_found. Will make that change.

src/main/java/co/cask/gcp/spanner/sink/SpannerSinkConfig.java
```diff
@@ -71,15 +99,13 @@ public void validate() {
...
-        for (String key : splitted) {
-          Schema schema = getSchema();
-          if (schema.getField(key.trim()) == null) {
-            throw new IllegalArgumentException(
-              String.format("Spanner primary key %s must be present in output schema", key));
-          }
+    if (!containsMacro("keys") && keys != null && !containsMacro("schema")) {
+      String[] splitted = keys.split(",");
+      for (String key : splitted) {
+        Schema schema = getSchema();
```
albertshau: should get this once outside of the for loop

widgets/Spanner-batchsink.json
```diff
@@ -49,7 +49,7 @@
         },
         {
           "widget-type": "csv",
-          "label": "Primary keys",
+          "label": "Primary key",
```
albertshau: Key.

For consistency, words in labels should be capitalized unless they are 'a', 'an', 'of', 'the'

docs/Spanner-batchsink.md
```diff
@@ -37,8 +37,8 @@ Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 If the table does not exist, it will get created.
 
-**Primary keys**: If the table does not exist, primary keys must be provided for the non existing spanner table. The
-keys must match the fields in output schema.
+**Primary key**: If the table does not exist, a primary key must be provided in order to auto-create the table.
```
albertshau: Key

albertshau (approved):
minor style comments, otherwise lgtm. fix and merge

PR #52, opened 2018-10-21
CDAP-14554 - google cloud plugin - support mutli files to gcs sink from plugin data-integrations/multi-table-plugins

yohayg:
This branch adds new sink that works in conjunction with https://github.com/data-integrations/multi-table-plugins/
given a database connection string and a list of tables. It will consume the tables and will write a file in GCS for each table.

This reduces the need for creating a different pipeline for each table as the current existing sink plugins.

I have tested avro, parquet, orc and text formats with codecs: snappy, deflate, gzip and lzo

I confirm that I have signed the ICLA  https://cla.developers.google.com/about/google-individual

### DIFF
--- a/src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
+++ b/src/main/java/co/cask/gcp/spanner/sink/SpannerSink.java
@@ -21,15 +21,25 @@ import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.batch.Output;
 import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.lineage.field.EndPoint;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldWriteOperation;
 import co.cask.hydrator.common.ReferenceBatchSink;
 import co.cask.hydrator.common.batch.sink.SinkOutputFormatProvider;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 
 /**
@@ -45,8 +55,7 @@ import org.apache.hadoop.io.NullWritable;
 @Description("Batch sink to write to Cloud Spanner. Cloud Spanner is a fully managed, mission-critical, " +
   "relational database service that offers transactional consistency at global scale, schemas, " +
   "SQL (ANSI 2011 with extensions), and automatic, synchronous replication for high availability.")
-public final class SpannerSink
-  extends ReferenceBatchSink<StructuredRecord, NullWritable, StructuredRecord> {
+public final class SpannerSink extends ReferenceBatchSink<StructuredRecord, NullWritable, StructuredRecord> {
   public static final String NAME = "Spanner";
   private final SpannerSinkConfig config;
 
@@ -72,6 +81,16 @@ public final class SpannerSink
     SpannerOutputFormat.configure(configuration, config);
     context.addOutput(Output.of(config.referenceName,
                                 new SinkOutputFormatProvider(SpannerOutputFormat.class, configuration)));
+
+    List<Schema.Field> fields = config.getSchema().getFields();
+    if (fields != null && !fields.isEmpty()) {
+        // Record the field level WriteOperation
+        FieldOperation operation = new FieldWriteOperation("Write", "Wrote to Spanner table.",
+                                                           EndPoint.of(context.getNamespace(), config.referenceName),
+                                                           fields.stream().map(Schema.Field::getName)
+                                                             .collect(Collectors.toList()));
+        context.record(Collections.singletonList(operation));
+    }
   }
 
   @Override
--- a/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
+++ b/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java
@@ -24,15 +24,18 @@ import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
 import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.lineage.field.EndPoint;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.api.batch.BatchSourceContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldReadOperation;
 import co.cask.gcp.common.GCPUtils;
-import co.cask.hydrator.common.SourceInputFormatProvider;
 import co.cask.gcp.spanner.SpannerConstants;
 import co.cask.gcp.spanner.common.SpannerUtil;
+import co.cask.hydrator.common.SourceInputFormatProvider;
 import com.google.cloud.ByteArray;
 import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
@@ -63,8 +66,10 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import javax.ws.rs.Path;
 
 /**
@@ -132,6 +137,16 @@ public class SpannerSource extends BatchSource<NullWritable, ResultSet, Structur
     // set input format and pass configuration
     batchSourceContext.setInput(Input.of(config.referenceName,
                                          new SourceInputFormatProvider(SpannerInputFormat.class, configuration)));
+    if (config.getSchema() != null) {
+      if (config.getSchema().getFields() != null) {
+        FieldOperation operation =
+          new FieldReadOperation("Read", "Read from Spanner table",
+                                 EndPoint.of(batchSourceContext.getNamespace(), config.referenceName),
+                                 config.getSchema().getFields().stream().map(Schema.Field::getName)
+                                   .collect(Collectors.toList()));
+        batchSourceContext.record(Collections.singletonList(operation));
+      }
+    }
   }
 
   @Override


src/main/java/co/cask/gcp/common/StructuredToOrcTransformer.java
```diff
@@ -0,0 +1,120 @@
...
+
+package co.cask.gcp.common;
+
+import co.cask.cdap.api.common.Bytes;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import co.cask.hydrator.common.HiveSchemaConverter;
+import co.cask.hydrator.common.RecordConverter;
+import org.apache.hadoop.io.*;
```
nitinmotgi: Can we import the exact class rather than wild card please.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+**Path Suffix:** The time format for the output directory that will be appended to the path.
+For example, the format 'yyyy-MM-dd-HH-mm' will result in a directory of the form '2015-01-01-20-42'.
+If not specified, nothing will be appended to the path."
+
+**Schema:** The schema of records to write.
+
+**Codec:** The codec to use when writing data. Must be 'none', 'snappy', or 'deflate'. Defaults to 'none'.
+
+
+**splitField:** The name of the field that will be used to determine which fileset to write to. Defaults to 'tablename'.
```
nitinmotgi: We are changing documentation to not include field name. Could we change this to "Split Field" as it shows in UI.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+
+**delimiter:** The delimiter used to separate record fields. Defaults to the tab character..
+
+**outputFormat** The output format Avro, Orc, Parquet or Text. Default value is Text
```
nitinmotgi: Same here - ```**Output Format**``

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
```diff
@@ -0,0 +1,153 @@
...
+package co.cask.gcp.common;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.hydrator.common.RecordConverter;
+import org.apache.avro.mapred.AvroKey;
+import org.apache.avro.mapreduce.AvroKeyOutputFormat;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.*;
```
nitinmotgi: Specific classes rather than wild card please.

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.orc.mapreduce.OrcOutputFormat;
+import parquet.avro.AvroParquetOutputFormat;
+
+import java.io.IOException;
+
+/**
+ * An OutputFormat that filters records before sending them to a delegate. Currently only supports TextOutputFormat
+ * as the delegate, but we may want to add other formats in the future.
+ */
+public class RecordFilterOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
```
nitinmotgi: Also, use 2 space tab instead of 4. I think that's the standard used in CDAP. If you are using Intellij, there is a profile that's available here - https://wiki.cask.co/display/CE/Coding+Standard

src/main/java/co/cask/gcp/common/StructuredToOrcTransformer.java
```diff
@@ -0,0 +1,120 @@
...
+                return new BooleanWritable((Boolean) fieldVal);
+            case INT:
+                return new IntWritable((Integer) fieldVal);
+            case LONG:
+                return new LongWritable((Long) fieldVal);
+            case FLOAT:
+                return new FloatWritable((Float) fieldVal);
+            case DOUBLE:
+                return new DoubleWritable((Double) fieldVal);
+            case BYTES:
```
nitinmotgi: Could we handle TIMESTAMP, DATETIME types too here.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,224 @@
...
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.batch.BatchSink;
+import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.gcp.common.*;
```
nitinmotgi: Specific classes rather than wildcard please.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * {@link GCSMultiBatchSink} that stores the data of the latest run of an adapter in S3.
```
nitinmotgi: It doesn't store in S3, it stores in GCS right ?

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,136 @@
...
+      "name": "schema",
+      "widget-type": "schema",
+      "widget-attributes": {
+        "schema-types": [
+          "boolean",
+          "int",
+          "long",
+          "float",
+          "double",
+          "bytes",
```
nitinmotgi: Might want to include Timestamp, datetime

nitinmotgi:
@yohayg Thanks for signing the ICLA.  Few minor comments for you to fix from me. Looking forward to this PR being merged.

yohayg:
fixed code review comments.
I'll open a new PR for supporting additional types such as Timestamp and datetime. This need more investigation

nitinmotgi:
LGTM 👍. @yohayg could you open an issue for the remaining items so that we don't lose track of it. 

@albertshau need your approval before we can merge this PR.

yohayg:
> LGTM 👍. @yohayg could you open an issue for the remaining items so that we don't lose track of it.
> 
> @albertshau need your approval before we can merge this PR.

Done: 
https://issues.cask.co/browse/CDAP-14554

yohayg:
@albertshau need your approval before we can merge this PR.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
+# Google Cloud Storage Multi Files Sink
+
+Description
+-----------
+
+
+This plugin is normally used in conjunction with the MultiTableDatabase batch source to write records from multiple
+databases into multiple filesets in text format. Each fileset it writes to will contain a single 'ingesttime' partition,
```
albertshau: text format -> various formats

albertshau: fileset is the wrong term to use here, it's writing to multiple directories.

albertshau: 'ingesttime' partition also has no meaning here, can probably just remove that sentence

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+which will contain the logical start time of the pipeline run. The plugin expects that the filsets it needs to write
```
albertshau: The plugin expects that the directories it needs to write to will be set as pipeline arguments, where the key is 'multisink.[directory]' and the value is the schema of the data.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+to will be set as pipeline arguments, where the key is 'multisink.[fileset]' and the value is the fileset schema.
+Normally, you rely on the MultiTableDatabase source to set those pipeline arguments, but they can also be manually
+set or set by an Action plugin in your pipeline. The sink will expect each record to contain a special split field
+that will be used to determine which records are written to each fileset. For example, suppose the
```
albertshau: which records are written to each directory.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+the split field is 'tablename'. A record whose 'tablename' field is set to 'activity' will be written to the 'activity'
+fileset.
```
albertshau: fileset -> directory

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+**Split Field:** The name of the field that will be used to determine which fileset to write to. Defaults to 'tablename'.
+
+**Field Delimiter:** The delimiter used to separate record fields. Defaults to the tab character..
+
+**Output Format** The output format Avro, Orc, Parquet or Text. Default value is Text
+
+Example
+-------
+
+This example uses a comma to delimit record fields:
```
albertshau: For new plugins, we're no longer putting in the json config, as it's hidden from most users. You can remove this first section

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+If not specified, nothing will be appended to the path."
+
+**Schema:** The schema of records to write.
+
+**Codec:** The codec to use when writing data. Must be 'none', 'snappy', or 'deflate'. Defaults to 'none'.
+
+
+**Split Field:** The name of the field that will be used to determine which fileset to write to. Defaults to 'tablename'.
+
+**Field Delimiter:** The delimiter used to separate record fields. Defaults to the tab character..
```
albertshau: extra period

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Output Format** The output format Avro, Orc, Parquet or Text. Default value is Text
```
albertshau: use 'Format' to be consistent with other plugins.

docs/GCSMultiFiles-batchsink.md
```diff
@@ -0,0 +1,120 @@
...
+    multisink.activity =
+     {
+       "type": "record",
+       "name": "activity",
+       "fields": [
+         { "name": "userid", "type": "long" } ,
+         { "name": "item", "type": "string" },
+         { "name": "action", "type": "string" }
+       ]
+     }
```
albertshau: the example should also include which records get written to which directories

pom.xml
```diff
@@ -84,6 +84,10 @@
       <id>sonatype-snapshots</id>
       <url>https://oss.sonatype.org/content/repositories/snapshots</url>
     </repository>
+      <repository>
+          <id>twitter-twttr</id>
```
albertshau: why is this needed?

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,136 @@
+
+{
+  "metadata": {
+    "spec-version": "1.5"
+  },
+  "display-name": "Multi Files Google Cloud Storage Sink",
+  "configuration-groups": [
+    {
+      "label" : "Governance",
```
albertshau: Follow the new convention, with a Basic, Credentials, and Advanced groups for the configuration properties, and the placeholders and defaults that are used in the other GCS plugins

pom.xml
```diff
@@ -407,6 +411,22 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-storage-api</artifactId>
```
albertshau: how come hive is being added?

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
```diff
@@ -0,0 +1,157 @@
...
+  public static final String FORMAT = "record.output.format";
+  public static final String ORIGINAL_SCHEMA = "record.original.schema";
+  public static final String AVRO = "avro";
+  public static final String ORC = "orc";
+  public static final String PARQUET = "parquet";
+
+
+  @Override
+  public RecordWriter<NullWritable, StructuredRecord> getRecordWriter(TaskAttemptContext context)
+          throws IOException, InterruptedException {
```
albertshau: style nit: indentation is off

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
```diff
@@ -0,0 +1,157 @@
...
+    Configuration conf = context.getConfiguration();
+
+    String filterField = conf.get(FILTER_FIELD);
+    String passthroughVal = conf.get(PASS_VALUE);
+    String delimiter = conf.get(DELIMITER);
+    String format = conf.get(FORMAT);
+    String originalSchema = conf.get(ORIGINAL_SCHEMA);
+
+    return new FilterRecordWriter(getOutputFormat(format).getRecordWriter(context),
+            filterField, passthroughVal, getStructuredRecordTransformer(format, delimiter),
```
albertshau: style nit: indentation is off

src/main/java/co/cask/gcp/common/StructuredToTextTransformer.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.hadoop.io.Text;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Transforms from structerd to text
+ */
+public class StructuredToTextTransformer extends RecordConverter<StructuredRecord, Text> {
```
albertshau: this class can be replaced with a call to StructuredRecordStringConverter.toDelimitedString()

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,136 @@
...
+          "widget-type": "textbox",
+          "label": "Split Field",
+          "name": "splitField",
+          "widget-attributes": {
+            "default": "tablename"
+          }
+        },
+        {
+          "widget-type": "select",
+          "label": "File Output Format",
```
albertshau: use 'Format' to be consistent with other plugins

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,136 @@
...
+              "avro",
+              "orc",
+              "parquet"
+            ],
+            "default": "text"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Field Delimiter",
```
albertshau: use 'Delimiter' to be consistent with other plugins

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Description("The prefix for the folder in GCS")
+    @Nullable
+    @Macro
+    protected String prefix;
+
+    @Name("codec")
+    @Description("The compression codec to use when writing data. Must be 'none', 'snappy', or 'deflated'.")
+    @Nullable
+    private String codec;
```
albertshau: Is this an important use case to you? It's not a setting in the other GCS plugins because it doesn't work out of the box in many environments, including on dataproc clusters.

Would like to be consistent and add it across they board at a later time if possible.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Nullable
+    @Description("The name of the field that will be used to determine which fileset to write to. " +
+            "Defaults to 'tablename'.")
```
albertshau: style nit: indentation

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+  GCSMultiBatchSinkConfig getConfig() {
+    return config;
+  }
+
+  /**
+   * Sink configuration.
+   */
+  public static class GCSMultiBatchSinkConfig extends GCSBatchSink.GCSBatchSinkConfig {
+
+    @Description("The prefix for the folder in GCS")
```
albertshau: What is this prefix for? It seems redundant, as it can just be included in the path.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    super.initialize(context);
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    config.validate();
+    super.configurePipeline(pipelineConfigurer);
```
albertshau: there should be additional validation done. For example, if pipelineConfigurer.getStageConfigurer().getInputSchema() is not null, the plugin should check that the configured splitField exists in the input schema and is of type string or nullable string.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+  private static final String TABLE_PREFIX = "multisink.";
+
+  private final GCSMultiBatchSinkConfig config;
+
+  public GCSMultiBatchSink(GCSMultiBatchSinkConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
```
albertshau: can remove this override, as it just calls super.initialize

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  @Override
+  public void transform(StructuredRecord input,
+                        Emitter<KeyValue<NullWritable, StructuredRecord>> emitter) {
+    emitter.emit(new KeyValue<>(org.apache.hadoop.io.NullWritable.get(), input));
+  }
+
+  @VisibleForTesting
+  GCSMultiBatchSinkConfig getConfig() {
```
albertshau: doesn't look like it's needed

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+
+      context.addOutput(Output.of(config.getReferenceName(),
+              new SinkOutputFormatProvider(RecordFilterOutputFormat.class.getName(), outputConfig)));
+    }
+  }
+
+  @Override
+  public void transform(StructuredRecord input,
+                        Emitter<KeyValue<NullWritable, StructuredRecord>> emitter) {
+    emitter.emit(new KeyValue<>(org.apache.hadoop.io.NullWritable.get(), input));
```
albertshau: don't need the package name

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+      } else {
+        // Encode the delimiter to base64 to support control characters. Otherwise serializing it in Cconf would result
+        // in an error
+        outputConfig.set(RecordFilterOutputFormat.DELIMITER, config.getDelimiter());
+      }
+
+
+      context.addOutput(Output.of(config.getReferenceName(),
+              new SinkOutputFormatProvider(RecordFilterOutputFormat.class.getName(), outputConfig)));
```
albertshau: style nit: indentation

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,191 @@
...
+          outputConfig.set(entry.getKey(), entry.getValue());
+        }
+      } else {
+        // Encode the delimiter to base64 to support control characters. Otherwise serializing it in Cconf would result
+        // in an error
+        outputConfig.set(RecordFilterOutputFormat.DELIMITER, config.getDelimiter());
+      }
+
+
+      context.addOutput(Output.of(config.getReferenceName(),
```
albertshau: does this work? I would expect it to add outputs of the same name and overwrite each other. Seems like it should be using the name from the argument and not the reference name

albertshau: reference name is also a funny thing to have for this plugin, as it's conceptually writing to many different datasets. I don't think it makes sense to have a reference name, but we could have a reference prefix, where the name from the argument is appended to that prefix.

example_avro_ingestion_gcs-cdap-data-pipeline.json
```diff
@@ -0,0 +1,101 @@
+{
```
albertshau: let's put this in an examples directory instead of at the top level,

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
```diff
@@ -0,0 +1,157 @@
...
+ * An OutputFormat that filters records before sending them to a delegate. Currently only supports TextOutputFormat
+ * as the delegate, but we may want to add other formats in the future.
+ */
+public class RecordFilterOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+  public static final String FILTER_FIELD = "record.filter.field";
+  public static final String PASS_VALUE = "record.filter.val";
+  public static final String DELIMITER = "record.filter.delimiter";
+  public static final String FORMAT = "record.output.format";
+  public static final String ORIGINAL_SCHEMA = "record.original.schema";
+  public static final String AVRO = "avro";
```
albertshau: fyi: much of the format related code will be changed soon, as we're moving formats into their own plugins. This has no bearing on this PR though.

albertshau (commented):
Thanks for the contribution! 

There are a few comments.

docs/GCSMultiFiles-batchsink.md
yohayg: fixed

docs/GCSMultiFiles-batchsink.md
yohayg: fixed

docs/GCSMultiFiles-batchsink.md
yohayg: fixed

docs/GCSMultiFiles-batchsink.md
yohayg: fixed

docs/GCSMultiFiles-batchsink.md
yohayg: fixed

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: fixed

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: fixed

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: changed to use name instead of reference name.
Both did not changed the functionality, it works with reference name and with name

example_avro_ingestion_gcs-cdap-data-pipeline.json
yohayg: done

pom.xml
yohayg: this is needed to support LZO codec

pom.xml
yohayg: needed to support orc format

src/main/java/co/cask/gcp/common/RecordFilterOutputFormat.java
yohayg: OK. I'll follow up on this

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: There is not input schema during the configuration stage

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: This is actually needed, since here I configure the different destination directories

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: The prefix is for the destination directories since they are created automatically. I need to to pass the the db name so I'll have the context from which DB I have consumed the tables

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: LZO might not be supported out of the box and will require apt-get install lzop
For building a data lake this is crucial. Users can always use none

widgets/GCSMultiFiles-batchsink.json
yohayg: Will open additional PR for that one

pom.xml
albertshau: Is this for https://github.com/twitter/hadoop-lzo? We can't use that because it's a GPL license.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
albertshau: Right, but why can't the db name just be in the base path to begin with? Unless I'm misunderstanding, it seems like the prefix always gets appended to the base path

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
albertshau: oh right because we expect various input schemas

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
albertshau: Hm ok CDAP must be generating some id for it somewhere then.

The name used here will be the one that is used for tracking lineage. If it's just the name, it can generate misleading lineage graphs if there are db tables that have the same name, but live in different databases. Let's just change it back to the reference name.

pom.xml
albertshau: Is this for https://github.com/twitter/hadoop-lzo? We can't use that because it's GPL licensed.

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
albertshau: I couldn't find anything that calls this. Where is it being called and why?

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,138 @@
...
+          }
+        }
+      ]
+    },
+    {
+      "label": "Advanced",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Base Path",
```
albertshau: The Advanced section generally contains properties that are not required and impact things like performance. I think we can put everything from here into the Basic section, except the compression codec.

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,138 @@
...
+          "widget-type": "textbox",
+          "label": "Service Account File Path",
+          "name": "serviceFilePath",
+          "widget-attributes" : {
+            "placeholder": "Path to the service account file on local file system"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Project ID",
```
albertshau: this should be in the basic section. Please follow the structure, defaults, and placeholders in GCS-batchsink.json so that there is consistency across the plugins.

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,138 @@
...
+          "name": "project",
+          "widget-attributes" : {
+            "placeholder": "The project id of the Cloud Storage bucket"
```
albertshau: instead of a placeholder, use a default like the other plugins:

"default": "auto-detect"

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,138 @@
...
+            ],
+            "default": "delimited"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Delimiter",
+          "name": "delimiter",
+          "widget-attributes": {
+            "default": "\t"
```
albertshau: "placeholder": "Delimiter if the format is 'delimited'"

widgets/GCSMultiFiles-batchsink.json
```diff
@@ -0,0 +1,138 @@
+
+{
+  "metadata": {
+    "spec-version": "1.5"
+  },
+  "display-name": "Multi Files Google Cloud Storage Sink",
```
albertshau: remove 'Sink', as the plugin will already show up in the sinks section in the UI.

albertshau (commented):
mostly small comments, with the bigger one around the LZO library. Is there another one that can be used with one of the 'notice' license types (apache 2, bsd being the most common of these)?

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: It implements the prepareRun from the BatchConfigurable interface, called by the cdap hydrator.

albertshau: This method is not part of the interface. Is it not showing the same line for you as for me? I'm talking about the getConfig() method that's annotated as visible for testing.

yohayg:
Hi @albertshau any thing else that should be fixed?

pom.xml
albertshau: orc-mapreduce is a project dependency already, was there something in here that is missing from orc-mapreduce?

albertshau (commented):
just one minor comment and one dependency question

pom.xml
yohayg: Without adding this dependency I get a runtime exception: java.lang.NoClassDefFoundError: org/apache/hadoop/hive/ql/exec/vector/ColumnVector

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+      } else {
+        // Encode the delimiter to base64 to support control characters. Otherwise serializing it in Cconf would result
+        // in an error
+        outputConfig.set(RecordFilterOutputFormat.DELIMITER, config.getDelimiter());
+      }
+
+
+      context.addOutput(
+          Output.of(name,
```
albertshau: My comment on this seems to have gotten lost.

Since it turns out this doesn't need to be a unique name, let's change it back to the reference name that you had earlier. Otherwise it can lead to misleading lineage graphs where tables with the same name but from different databases will be linked together.

albertshau:
I only have the one comment about reference name. After that's fixed, please squash your commits and we'll merge it. Thanks for the contribution!

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
yohayg: When I look at the equivalent sinks in the mutli-table-plugins repo they use the name and not the reference name:
https://github.com/data-integrations/multi-table-plugins/blob/develop/src/main/java/co/cask/plugin/DynamicMultiFilesetSink.java
 
What about concating the reference name and the table name? In this way the user will have more control on the created dataset

albertshau: That sounds good too.

yohayg: fixed

yohayg:
Hi @albertshau please let me know if there are additional items that needs a change.

Thanks,

Yohay

albertshau:
@yohayg changes look good. Please squash your commits and then we'll merge.

PR #53, opened 2018-10-22
CDAP-14531 Fix NPE when outputschema and input record schema do not m…

CuriousVini:
…atch

Validation for additional field in output schema which is not available in input record schema

image

Google plugins only support simple types or nullable simple types. So added simple type schema validation at deploy time to fail fast:
image

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
+++ b/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
@@ -89,12 +89,13 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    config.validate(pipelineConfigurer.getStageConfigurer().getInputSchema());
     super.configurePipeline(pipelineConfigurer);
-    config.validate();
   }
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
+    config.validate(context.getInputSchema());
     BigQuery bigquery = BigQueryUtils.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
     // create dataset if it does not exist
     if (bigquery.getDataset(config.getDataset()) == null) {
@@ -105,6 +106,7 @@ public final class BigQuerySink extends BatchSink<StructuredRecord, JsonObject, 
       }
     }
 
+    // schema validation against bigquery table schema
     validateSchema();
 
     uuid = UUID.randomUUID();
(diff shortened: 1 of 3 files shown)


src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -244,6 +248,12 @@ private void emitLineage(BatchSinkContext context, List<BigQueryTableFieldSchema
   }
 
   private static void decodeSimpleTypes(JsonObject json, String name, StructuredRecord input) {
+    // Avoid NPE when structured record schema does not contain field present in output schema.
+    if (input.getSchema().getField(name) == null) {
```
albertshau: this can be avoided if the method is changed so that instead of taking `String name, StructuredRecord input`, it takes `Schema.Field field, Object fieldVal`. The caller should not iterate through config.getSchema().getFields(), but through input.getSchema().getFields(). If the record field does not exist in the config schema, we can skip it. Otherwise it should call decodeSimpleTypes()

albertshau: also, the original bug was if the config schema contained extra non-nullable fields that do not exist in the input schema. Checking here will cause it to fail late, at runtime.

This can be checked at configure time if there are no macros, using pipelineConfigurer().getStageConfigurer().getInputSchema() to get the input schema if it's known. If the configured schema contains a macro, it can be checked against the input schema at prepare time using context.getInputSchema().

CuriousVini: [1.] Modified the code to iterate over record schema. The method arguments are still the same because the structured record is required to get logical type values. 
[2.] Added checks in schema validation to fail fast if required field in output schema is not present in input schema.

src/main/java/co/cask/gcp/bigquery/BigQuerySink.java
```diff
@@ -159,10 +157,13 @@ public void initialize(BatchRuntimeContext context) throws Exception {
 
   @Override
   public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullWritable>> emitter) throws Exception {
-    List<Schema.Field> fields = config.getSchema().getFields();
+    Schema outputSchemaFields = config.getSchema();
```
albertshau: it's more performant to call this once at initialize time. Generates less garbage in memory, as getSchema() actually creates a new object each time.

albertshau (approved):
one small comment, otherwise lgtm. fix and merge

PR #55, opened 2018-10-24
CDAP-14522 fix warn message in gcs plugins

CuriousVini:


src/main/java/co/cask/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -70,6 +70,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     String projectId = config.getProject();
     properties.put("fs.gs.project.id", projectId);
     properties.put("fs.gs.system.bucket", GCSConfigHelper.getBucket(config.path));
+    properties.put("fs.gs.working.dir", GCSConfigHelper.getBucket(config.path));
```
albertshau: the warning message says it defaults to gs://bucket/, which is different than this. Have we tested this to make sure it works?

In any case, I think it's safer to set it to gs://bucket/

CuriousVini: Yea, I did test these changes. The output path was same as provided in the plugin config and there were no warn messages in the pipeline. However the default value is root of the filesystem. https://github.com/GoogleCloudPlatform/bigdata-interop/blob/v1.9.1/gcs/conf/gcs-core-default.xml#L13

So I have changed it to be `/` which is relative to the `fs.gs.system.bucket` 

**Output Path** `gs://bq_cdap_batch_sink/path4/blah`

image

**Pipeline log levels:**
image

albertshau (approved):
i'm not really sure what the working dir is, but seems like it runs fine. lgtm

CuriousVini:
Without this change, the plugin uses bucket root as working directory. So, after these changes plugin behavior should remain same except no warn messages in logs.

PR #56, opened 2018-10-27
[CDAP-14539] Adds a BigQuery Execute Plugin

nitinmotgi:
In addition, also added BigQuery Delete Plugin.


src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryDelete.NAME)
+@Description("Delete a Google BigQuery dataset or table.")
+public final class BigQueryDelete extends Action {
+  private static final Logger LOG = LoggerFactory.getLogger(BigQueryDelete.class);
+  public static final String NAME = "BigQueryDelete";
+  private static final String TABLE = "table";
+  private static final String DATASET = "dataset";
+  private static final String RECORDS_OUT = "records.out";
```
albertshau, line 47: unused

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    // API request - starts the query.
+    BigQueryOptions.Builder bqBuilder = GCPUtils.getBigQuery(config.project, config.serviceAccountFilePath);
+    BigQuery bigquery = bqBuilder.build().getService();
+
+    // Delete a table.
+    if (config.type.equalsIgnoreCase(TABLE)) {
+      boolean delete = bigquery.delete(config.dataset, config.table);
```
albertshau: nit: rename to deleted

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+      if (!delete) {
+        LOG.info("Table %s was deleted", config.table);
```
albertshau: From the javadocs, it returns true if it was deleted, false if it doesn't exist. If anything needs to be logged it should be a message about the table not being found.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Name("type")
+    @Description("Delete table or dataset")
+    @Macro
+    public String type;
```
albertshau: private

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Name("dataset")
+    @Description("Name of dataset.")
+    @Macro
+    public String dataset;
```
albertshau: private

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Name("table")
+    @Description("Name of table.")
+    @Macro
+    @Nullable
+    public String table;
```
albertshau: private

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+      boolean delete = bigquery.delete(config.dataset, config.table);
+      if (!delete) {
+        LOG.info("Table %s was deleted", config.table);
+      }
+    }
+
+    // Delete a dataset
+    if (config.type.equalsIgnoreCase(DATASET)) {
+      boolean delete = false;
+      if (config.project == null) {
```
albertshau: this is supposed to be a private variable. It looks like release was never merged back into develop. Are you sure these changes shouldn't be going into the release/0.11 branch (for CDAP 5.1.1)?

With the latest code, you should be using getProject(), as it will look up the project id from the environment if project is not given or set to 'auto-detect'.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+        delete = bigquery.delete(DatasetId.of(config.dataset));
+      } else {
+        delete = bigquery.delete(DatasetId.of(config.project, config.dataset));
+      }
+      if (!delete) {
+        LOG.info("Dataset %s was deleted", config.dataset);
```
albertshau: similar comment about the log message

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+
+    context.getMetrics().gauge("records.out", 1);
```
albertshau: why record this? Nothing related to records is happening.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+import org.slf4j.LoggerFactory;
+import org.threeten.bp.Duration;
+
+import javax.annotation.Nullable;
+
+/**
+ * This class <code>BigQueryExecute</code> executes a single Cloud BigQuery SQL.
+ *
+ * The plugin provides the ability different options like choosing interactive or batch execution
+ * of sql query, setting of resulting dataset and table, enabling/disbaling cache, specifying whether
```
albertshau: typo: disabling

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+ * the query being executed is legacy or standard and reattempt strategy.
```
albertshau: reattempt -> retry

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryExecute.NAME)
+@Description("Execute a Google BigQuery SQL.")
```
albertshau, line 50: SQL query

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
+    }
+
+    TableResult queryResults = queryJob.getQueryResults();
+    long rows = queryResults.getTotalRows();
+
+    context.getMetrics().gauge(RECORDS_OUT, rows);
+    context.getArguments().set(context.getStageName() + ".query", config.sql);
+    context.getArguments().set(context.getStageName() + ".jobid", jobId.getJob());
+    context.getArguments().set(RECORDS_OUT, String.valueOf(rows));
```
albertshau: why not prefix with the stage name here? If you have multiple plugins they'll overwrite each other.

Also, how are we expecting these arguments to be used? It seems like another plugin will need to know the structure of the pipeline in order to lookup the correct value.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgumentException {
+    super.configurePipeline(configurer);
+
+    if (!config.containsMacro("sql")) {
+      if (config.sql == null || config.sql.isEmpty()) {
```
albertshau: null does not need to be checked because the variable is not annotated as Nullable. The platform will ensure it's not null (unless it contains a macro)

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Name("legacy")
+    @Description("Use Legacy SQL")
+    @Macro
+    public String legacy;
```
albertshau: all variables should be private

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Name("mode")
+    @Description("Batch or Interactive.")
+    @Macro
+    public String mode;
+
+    @Name("cache")
+    @Description("Use Cache")
+    @Macro
+    public String cache;
```
albertshau: Boolean useCache.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Name("type")
+    @Description("Delete table or dataset")
```
albertshau: Type of entity to delete. It must be either 'table' or 'dataset'.

albertshau: actually, this property shouldn't be needed. If a table name is not given, it can just delete the dataset.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+    @Macro
+    public String type;
+
+    @Name("dataset")
+    @Description("Name of dataset.")
+    @Macro
+    public String dataset;
+
+    @Name("table")
+    @Description("Name of table.")
```
albertshau: Name of the table to delete. Leave this empty if a dataset is being deleted.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+
+    // Delete a dataset
+    if (config.type.equalsIgnoreCase(DATASET)) {
+      boolean delete = false;
+      if (config.project == null) {
+        delete = bigquery.delete(DatasetId.of(config.dataset));
+      } else {
+        delete = bigquery.delete(DatasetId.of(config.project, config.dataset));
```
albertshau: What is the behavior if the dataset contains tables? Does it fail or are the tables deleted as well?

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    context.getMetrics().gauge("records.out", 1);
+  }
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Name("type")
```
albertshau: names are not needed when they are the same as the variable name

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
albertshau: Boolean useLegacy

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+        throw new IllegalArgumentException("SQL not specified. Please specify a SQL to execute");
+      }
+    }
+  }
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Name("legacy")
```
albertshau: names are not needed when they are the same as the variable name

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Description("Use Legacy SQL")
```
albertshau: Use legacy SQL. (include the period)

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Macro
+    public String legacy;
+
+    @Name("sql")
+    @Description("SQL query to execuute.")
```
albertshau: typo: execute

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Macro
+    public String sql;
+
+    @Name("mode")
+    @Description("Batch or Interactive.")
```
albertshau: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query .... An interactive query....

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Macro
+    public String mode;
+
+    @Name("cache")
+    @Description("Use Cache")
```
albertshau: Use the cache when executing the query.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Macro
+    public String cache;
+
+    @Name("dataset")
+    @Description("Permanent Dataset.")
```
albertshau: The dataset to store the query results in. If not specified, the results will not be stored.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    @Macro
+    @Nullable
+    public String dataset;
+
+    @Name("table")
+    @Description("Permanent Table.")
```
albertshau: The table to store the query results in. If not specified, the results will not be stored.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+
+  private Config config;
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    // Check if a SQL has been specified, if the SQL has not been specified,
+    // then there is nothing for this class to do.
+    if (config.sql == null || config.sql.isEmpty()) {
+      throw new IllegalArgumentException(
+        "SQL not specified. Please specify a SQL to execute"
```
albertshau: this looks like it can be on the previous line

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final String CACHE_ENABLED = "enabled";
+  private static final String LEGACY_ENABLED = "legacy-enabled";
+
+  private Config config;
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    // Check if a SQL has been specified, if the SQL has not been specified,
+    // then there is nothing for this class to do.
+    if (config.sql == null || config.sql.isEmpty()) {
```
albertshau: move validation logic into a separate method that is called at configure time as well as runtime. Errors like an empty query should cause the pipeline to fail to deploy

albertshau: sql can never be null (when it doesn't contain a macro) because it is not annotated as nullable

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+      throw new IllegalArgumentException(
+        "SQL not specified. Please specify a SQL to execute"
+      );
+    }
+
+    QueryJobConfiguration.Builder builder = QueryJobConfiguration.newBuilder(config.sql);
+
+    // Run at batch priority, which won't count toward concurrent rate limit.
+    if (config.mode.equalsIgnoreCase(MODE_BATCH)) {
```
albertshau: validation should ensure that the mode is either 'batch' or 'interactive'

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+      builder.setPriority(QueryJobConfiguration.Priority.BATCH);
+    } else {
+      builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
+    }
+
+    // Save the results of the query to a permanent table.
+    if (config.dataset != null && config.table != null
```
albertshau: validation should ensure that if a dataset is given, a table is also given. Similarly, if a table is given, a dataset must be given.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+      && !config.dataset.isEmpty() && !config.table.isEmpty()) {
+      builder.setDestinationTable(TableId.of(config.dataset, config.table));
+      context.getArguments().set(context.getStageName() + ".dataset", config.dataset);
+      context.getArguments().set(context.getStageName() + ".table", config.dataset);
+    }
+
+    // Enable or Disable the query cache to force live query evaluation.
+    if (config.cache.equalsIgnoreCase(CACHE_ENABLED)) {
+      builder.setUseQueryCache(true);
```
albertshau, line 79: this means the default is false if you don't call setUseQueryCache? It seems safer to always set this to whatever is configured.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    // Enable legacy SQL
+    if (config.legacy.equalsIgnoreCase(LEGACY_ENABLED)) {
+      builder.setUseLegacySql(true);
+    } else {
+      builder.setUseLegacySql(false);
+    }
+
+    QueryJobConfiguration queryConfig = builder.build();
+
+    // Location must match that of the dataset(s) referenced in the query.
```
albertshau, line 87: doesn't this mean the location must be a config property?

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    BigQuery bigquery = bqBuilder.build().getService();
+    Job queryJob = bigquery.create(
+      JobInfo.newBuilder(queryConfig)
+        .setJobId(jobId).build()
+    );
+
+    LOG.info("Executing query %s, Job id %s.", config.sql, jobId.getJob());
+
+    // Wait for the query to complete
+    RetryOption retryOption = RetryOption.totalTimeout(Duration.ofMinutes(10));
```
albertshau: this looks like it should be an advanced config property

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    queryJob.waitFor(retryOption);
```
albertshau, line 101: does this mean it will only wait for 10 minutes? Many queries will take longer than that

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+
+    // Check for errors
+    if (queryJob == null) {
```
albertshau: This can never be null at this point. If it can be null, it should be checked when it is created

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+      throw new RuntimeException("Job " + jobId.getJob() + " no longer exists");
+    } else if (queryJob.getStatus().getError() != null) {
+      // You can also look at queryJob.getStatus().getExecutionErrors() for all
+      // errors, not just the latest one.
+      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
```
albertshau, line 107: what does getExecutionErrors() return? A list? Does this result in a human readable error?

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
albertshau: Also, why set an argument for the number of records when a metric is emitted?

src/main/java/co/cask/gcp/bigquery/BigQuerySource.java
```diff
@@ -249,18 +247,9 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
...
     try {
-      BigQuery bigquery;
-      BigQueryOptions.Builder bigqueryBuilder = BigQueryOptions.newBuilder();
-      if (request.serviceFilePath != null) {
-        bigqueryBuilder.setCredentials(loadServiceAccountCredentials(request.serviceFilePath));
-      }
+      BigQueryOptions.Builder bigqueryBuilder = GCPUtils.getBigQuery(request.project, request.serviceFilePath);
+      BigQuery bigquery = bigqueryBuilder.build().getService();
       String project = request.project == null ? ServiceOptions.getDefaultProjectId() : request.project;
-      if (project == null) {
```
albertshau: Nothing here should change if you're using the latest code on release/0.11. If this is supposed to go to develop, please make sure the release changes get merged back to develop then rebase.

src/main/java/co/cask/gcp/pubsub/PubSubOutputFormat.java
```diff
@@ -177,7 +177,7 @@ public void close(TaskAttemptContext taskAttemptContext) throws IOException {
           publisher.shutdown();
         } catch (Exception e) {
           // if there is an exception while shutting down, we only log
-          LOG.debug("Exception while shutting down publisher ", e);
+          LOG.debug("Exception while shutting down pubsub ", e);
```
albertshau: why change this message? It is more accurate to say publisher.

widgets/BigQueryDelete-action.json
```diff
@@ -0,0 +1,70 @@
+{
+  "metadata": {
+    "spec-version": "1.5"
+  },
+  "display-name" : "Google BigQuery Delete",
+  "configuration-groups": [
+    {
+      "label" : "Project",
```
albertshau: Please be consistent with the configuration groups used in all the other plugins, with 'Basic', 'Credentials', and 'Advanced' groups.

albertshau (commented):
These plugins are missing reference docs.

They also have the same problem as https://issues.cask.co/browse/CDAP-14542, where they may not be able to execute queries against tables in another project.

src/main/java/co/cask/gcp/common/GCPUtils.java
```diff
@@ -48,4 +49,25 @@ public static String getProjectId(@Nullable String project) {
     }
     return projectId;
   }
+
+  /**
+   *
```
CuriousVini: Missing java doc?

CuriousVini: It would be good to first merge release/0.11 to develop and then make these changes.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+    } else {
+      builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
+    }
+
+    // Save the results of the query to a permanent table.
+    if (config.dataset != null && config.table != null
+      && !config.dataset.isEmpty() && !config.table.isEmpty()) {
+      builder.setDestinationTable(TableId.of(config.dataset, config.table));
+      context.getArguments().set(context.getStageName() + ".dataset", config.dataset);
+      context.getArguments().set(context.getStageName() + ".table", config.dataset);
```
CuriousVini: Shouldn't it be `config.table` instead of `config.dataset`

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,186 @@
...
+
+    // API request - starts the query.
+    BigQueryOptions.Builder bqBuilder = GCPUtils.getBigQuery(config.project, config.serviceAccountFilePath);
+    BigQuery bigquery = bqBuilder.build().getService();
+    Job queryJob = bigquery.create(
+      JobInfo.newBuilder(queryConfig)
+        .setJobId(jobId).build()
+    );
+
+    LOG.info("Executing query %s, Job id %s.", config.sql, jobId.getJob());
```
CuriousVini: I think its better to be specific: `Executing query '%s'. The Google BigQuery job id is '%s'.`

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.cloud.bigquery.DatasetId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+/**
+ * This class <code>BigQueryDelete</code> deletes a dataset or a table.
+ *
+ * If the table or dataset does not exist, it's still successful.
```
CuriousVini: `If the table or dataset does not exist, the delete action is still successful.`

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(BigQueryDelete.class);
+  public static final String NAME = "BigQueryDelete";
+  private static final String TABLE = "table";
+  private static final String DATASET = "dataset";
+  private static final String RECORDS_OUT = "records.out";
+  private Config config;
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    // API request - starts the query.
```
CuriousVini: unrelated?

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: used it.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: In action, there is no record out. Other than indicating that there was successful action.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: BigQuery SQL is different as per documentation, so tried to keep it in sync with that.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: If I am going to use the number of records processes in a condition, I should be able to use it right. E.g. if number of records affected > x ? do x : do y.

nitinmotgi: Fair point, fixed and added stage.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: changed to private. it's a radio button on UI and radio button only sends string to backend.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: It's radio-group-widget on frontend which sends string to backend.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: To make it clear, the user specifies the type using the radio-group widget.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: Using radio button to allow user to specify that explicitly. Rather than making him understand it. Will change the text - fixed.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: tables will also be deleted.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: by default it's false.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
nitinmotgi: yes, fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
nitinmotgi: fixed

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Added.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Configurable now. If user believe the query will take longer we expect them to provide a longer timeout. Default is 10 mins

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Fixed.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Yes, a list of BigQueryError class. 
Checked BigQueryError to have a toString() implementation.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Fixed.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
rohitsinha54: Done.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
rohitsinha54: Fixed.

src/main/java/co/cask/gcp/common/GCPUtils.java
```diff
@@ -53,4 +55,25 @@ public static ServiceAccountCredentials loadServiceAccountCredentials(String pat
     properties.put("fs.gs.impl.disable.cache", "true");
     return properties;
   }
+
+//  /**
```
bdmogal, line 59: remove if not required?

rohitsinha54:
I am going to remove the selection of dataset/table radio button from the BigQueryDelete plugin. This is confusing and not required. I just ended up deleting the whole dataset because I did not switch over the dataset radio button to table after specifying table name. Of course the code can have a guard against it but the user can still have inconsistency leading to failure. Will keep the config suggested here https://github.com/data-integrations/google-cloud/pull/56#discussion_r229033418
image

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
...
+Make sure the account key has permission to access BigQuery.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**Dataset Project ID**: Project the dataset or table belongs to. This is only required if the dataset or table is not
```
albertshau, line 25: I understand having two settings for the query action, but the delete action requires both as well?

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,47 @@
...
+**Mode**: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query is executed as
+soon as possible and count towards the concurrent rate limit and the daily rate limit. An interactive query is
+queued and started as soon as idle resources are available, usually within a few minutes. If the query hasn't
+started within 3 hours, its priority is changed to 'INTERACTIVE'.
+
+**Dataset Name**: The dataset to store the query results in. If not specified, the results will not be stored.
+
+**Table Name**: The table to store the query results in. If not specified, the results will not be stored.
+
+**Cache**: Specifies whether to use cache when executing the query.
```
albertshau, line 39: Seems like this should have more information. Is there standard BQ documentation on this?

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
...
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access BigQuery.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
```
albertshau, line 22: Remove 'The' to be consistent

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,47 @@
...
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access BigQuery.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
```
albertshau, line 22: For consistency, remove 'The' from the start of these descriptions.

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,47 @@
...
+
+**SQL**: The SQL command to execute.
+
+**Legacy**: Specified whether to use BigQuery's legacy SQL dialect for this query. By default this property is
+set to 'false'. If set to 'false', the query will use BigQuery's Standard SQL.
+
+**Mode**: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query is executed as
+soon as possible and count towards the concurrent rate limit and the daily rate limit. An interactive query is
+queued and started as soon as idle resources are available, usually within a few minutes. If the query hasn't
+started within 3 hours, its priority is changed to 'INTERACTIVE'.
```
albertshau, line 33: be consistent with casing, use 'interactive'.

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,47 @@
...
+
+**Dataset Name**: The dataset to store the query results in. If not specified, the results will not be stored.
+
+**Table Name**: The table to store the query results in. If not specified, the results will not be stored.
+
+**Cache**: Specifies whether to use cache when executing the query.
+
+**Job Location**: Location of the job. Must match the location of the dataset specified in the query. Defaults to 'US'
```
albertshau, line 41: Use complete sentences after the first fragment, like a javadoc.

Location of the job. It must match the location .... The default value is 'US'.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+                                  BigQuery.DatasetDeleteOption.deleteContents());
+      } else {
+        deleted = bigQuery.delete(DatasetId.of(config.getDatasetProject(), config.getDataset()));
+      }
+      if (!deleted) {
+        LOG.info("Failed to delete dataset '%s'. Dataset '%s' does not exist.", config.getDataset(),
+                 config.getDataset());
+      }
+    }
+    context.getMetrics().gauge(RECORDS_OUT, 1);
```
albertshau, line 73: I believe I asked this before, but what is the point of this metric? It seems confusing to talk about records in an action plugin.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+    @Macro
+    @Nullable
+    @Description("The project the dataset or table belongs to. This is only required if the dataset or table is not "
+      + "in the same project that the BigQuery job will run in. If no value is given, it will default to the " +
+      "configured project ID.")
+    private String datasetProject;
+
+    @Macro
+    @Description("Deletes a dataset even if non-empty. Defaults to 'false'")
+    private String deleteContents;
```
albertshau, line 98: I didn't see a description of this in the reference doc.

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
...
+**Dataset Project ID**: Project the dataset or table belongs to. This is only required if the dataset or table is not
+in the same project that the BigQuery query will run in. If no value is given,
+it will default to the configured Project ID.
+
+**Dataset Name**: Name of the dataset to delete or if a table needs to be deleted then the name of the dataset in
+which the table exists.
+
+**Dataset Name**: Name of the table to be deleted.
+
+**Delete Contents** Specifies whether to delete a dataset even when it is not empty. Defaults to false. This value is
```
albertshau, line 34: Use complete sentences after the first fragment. Also describe what happens if the dataset is not empty and this is false. I assume it results in the pipeline failing?

Whether to delete a dataset even when it is not empty. If this is false and the dataset is not empty .... The default value is false.

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
+# Google Cloud BigQuery Delete
+
+Description
+-----------
+This plugin deletes BigQuery dataset or tables.
```
albertshau, line 5: deletes a BigQuery dataset or table.

albertshau: There should be some description of what happens if the dataset or table does not exist. Does it fail? Does it continue on?

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
...
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**Dataset Project ID**: Project the dataset or table belongs to. This is only required if the dataset or table is not
+in the same project that the BigQuery query will run in. If no value is given,
+it will default to the configured Project ID.
+
+**Dataset Name**: Name of the dataset to delete or if a table needs to be deleted then the name of the dataset in
+which the table exists.
+
+**Dataset Name**: Name of the table to be deleted.
```
albertshau, line 32: to be deleted -> to delete

albertshau: Dataset Name -> Table Name

docs/BigQueryDelete-action.md
```diff
@@ -0,0 +1,39 @@
...
+Properties
+----------
+**Project ID**: The Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**Dataset Project ID**: Project the dataset or table belongs to. This is only required if the dataset or table is not
+in the same project that the BigQuery query will run in. If no value is given,
+it will default to the configured Project ID.
+
+**Dataset Name**: Name of the dataset to delete or if a table needs to be deleted then the name of the dataset in
```
albertshau, line 29: Name of the dataset that contains the table to delete. If no table is specified, the dataset will be deleted.

docs/BigQueryExecute-action.md
albertshau: Whether to use the cache when executing the query.

albertshau: If we name it 'Cache', it could be interpreted as "cache the results of the query", which doesn't sound like what this flag is controlling. Would 'Use Cache' be more descriptive?

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+  private Config config;
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    BigQuery bigQuery = BigQueryUtils.getBigQuery(config.getServiceAccountFilePath(), config.getDatasetProject());
+
+    // if a table is specified it is a delete a table command
+    if (config.getTable() != null) {
+      if (!bigQuery.delete(config.getDataset(), config.getTable())) {
+        LOG.info("Failed to delete table '%s'. Table '%s' does not exist in dataset '%s'.", config.getTable(),
```
albertshau, line 57: This is an incorrect message, as it uses the project as the dataset. I think it's better to have an error message like:

Table 'project:dataset.table' cannot be deleted because it does not exist

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+                 config.getTable(), config.getDatasetProject());
+      }
+    } else {
+      boolean deleted;
+      if (config.shouldDeleteContents()) {
+        deleted = bigQuery.delete(DatasetId.of(config.getDatasetProject(), config.getDataset()),
```
albertshau, line 63: nit: would be cleaner to have a DatasetId variable that is created at the start of the method and passed to whatever delete call is eventually used.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+                                  BigQuery.DatasetDeleteOption.deleteContents());
+      } else {
+        deleted = bigQuery.delete(DatasetId.of(config.getDatasetProject(), config.getDataset()));
+      }
+      if (!deleted) {
+        LOG.info("Failed to delete dataset '%s'. Dataset '%s' does not exist.", config.getDataset(),
```
albertshau, line 69: similar comment, think it would be better to say:

Dataset 'project:dataset' cannot be deleted because it does not exist

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+    @Macro
+    @Nullable
+    private String table;
+
+    @Macro
+    @Nullable
+    @Description("The project the dataset or table belongs to. This is only required if the dataset or table is not "
+      + "in the same project that the BigQuery job will run in. If no value is given, it will default to the " +
+      "configured project ID.")
+    private String datasetProject;
```
albertshau, line 94: doesn't seem like this is needed and we can just have a single projectId property. Unless we're using this someplace where the projectId should be used?

Can you try removing this and test deleting a dataset in another project?

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
albertshau: Is this the language used in official documentation? It sounds funny.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
albertshau: Ok but it still seems safer to always do `builder.setUseQueryCache(config.shouldUseCache())` in case the default in the library is ever changed.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,230 @@
...
+  private static final String MODE_BATCH = "batch";
+
+  private Config config;
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    config.validate();
+    QueryJobConfiguration.Builder builder = QueryJobConfiguration.newBuilder(config.getSql());
+    // Run at batch priority, which won't count toward concurrent rate limit.
+    if (config.getMode().equalsIgnoreCase(MODE_BATCH)) {
```
albertshau, line 64: would be cleaner if the config's getMode() method returned a QueryJobConfiguration.Priority

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,230 @@
...
+      builder.setPriority(QueryJobConfiguration.Priority.BATCH);
+    } else {
+      builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
+    }
+
+    // Save the results of the query to a permanent table.
+    if (config.getDataset() != null  && config.getTable() != null) {
+      builder.setDestinationTable(TableId.of(config.getDataset(), config.getTable()));
+      context.getArguments().set(context.getStageName() + ".dataset", config.getDataset());
```
albertshau, line 73: what is the point of this? I'm guessing this is done to allow some other plugin to read this argument? If so, it should be removed because it is introducing a different pattern to try to solve a problem that macros already solve.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,230 @@
...
+    JobId jobId = JobId.newBuilder().setRandomJob().setLocation(config.getLocation()).build();
+
+    // API request - starts the query.
+    BigQuery bigQuery = BigQueryUtils.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
+    Job queryJob = bigQuery.create(
+      JobInfo.newBuilder(queryConfig)
+        .setJobId(jobId).build()
+    );
+
+    LOG.info("Executing query '%s'. The Google BigQuery job id is '%s'.", config.getSql(), jobId.getJob());
```
albertshau, line 97: I don't think it's a good idea to include the full query in the log message at info level as the query can be very big. I think logging 'Executing SQL as job '%s'.' is good enough. I would log the sql statement at debug level.

src/main/java/co/cask/gcp/bigquery/BigQueryExecute.java
```diff
@@ -0,0 +1,230 @@
...
+
+  /**
+   * Config for the plugin.
+   */
+  public final class Config extends GCPConfig {
+    @Description("Use Legacy SQL.")
+    @Macro
+    private String legacy;
+
+    @Description("SQL query to execute.")
```
albertshau, line 134: descriptions and log messages are inconsistent about their use 'sql' or 'sql query'. Please figure out which is correct and use it everywhere.

src/main/java/co/cask/gcp/common/GCPUtils.java
albertshau: +1

widgets/BigQueryDelete-action.json
```diff
@@ -0,0 +1,78 @@
...
+          "widget-type": "textbox",
+          "label": "Table Name",
+          "name": "table",
+          "widget-attributes": {
+            "placeholder": "Name of Table"
+          }
+        },
+        {
+          "name": "deleteContents",
+          "label": "Delete Contents",
```
albertshau, line 44: 'Delete Dataset Contents' would be more descriptive. Otherwise it can be confusing what this means when a table is specified.

widgets/BigQueryDelete-action.json
```diff
@@ -0,0 +1,78 @@
...
+          "widget-attributes": {
+            "placeholder": "Name of Dataset"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Table Name",
+          "name": "table",
+          "widget-attributes": {
+            "placeholder": "Name of Table"
```
albertshau, line 39: Name of Table. Leave this empty if the entire dataset should be deleted.

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,139 @@
...
+          "widget-attributes": {
+            "default": "auto-detect"
+          }
+        },
+        {
+          "widget-type": "textarea",
+          "label": "SQL",
+          "name": "sql",
+          "widget-attributes": {
+            "placeholder": "Specifies the query to be executed."
```
albertshau, line 23: similar comment about use of 'query'

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,139 @@
...
+            "placeholder": "Location of the job.",
+            "default": "US"
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Timeout",
+          "name": "timeout",
+          "widget-attributes": {
+            "placeholder": "Query timeout in minutes. Defaults to 10.",
```
albertshau, line 118: is there a point in having both a placeholder and a default? Does the placeholder get displayed if the default value is deleted? Don't need to mention the default in the placeholder when there is a default for the widget

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,139 @@
...
+            "default": "10"
```
albertshau, line 119: I think we should pick defaults that are less likely to result in errors. 10 minutes seems really short. What is the default for the BigQuery client if no timeout is specified?
It may be better to allow this to be null, which means to never time out.

albertshau: Also, I think caching, legacy, and timeout would fit better under an 'Advanced' section, as they are all optional and require some understanding of BigQuery.

docs/BigQueryDelete-action.md
albertshau: Delete Contents -> Delete Dataset Contents.

If you just look at the label without reading the description, it can be confusing what 'delete contents' means in a table delete scenario.

src/main/java/co/cask/gcp/bigquery/BigQueryDelete.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * This class <code>BigQueryDelete</code> deletes a dataset or a table.
+ * <p>
+ * If the table or dataset does not exist, the delete action is still successful.
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryDelete.NAME)
+@Description("Delete a Google BigQuery dataset or table.")
+public final class BigQueryDelete extends Action {
```
bdmogal, line 42: let's rename this to BigQueryDrop instead of BigQueryDelete. Feedback from users is that in the SQL world, delete gives the impression that only the data will be removed, but the table will stay (``DELETE from``). However, this deletes the table itself, which is equivalent to the ``DROP`` statement in SQL

PR #59, opened 2018-11-02
CDAP-14442 remove avro, parquet, and orc dependencies

albertshau:
Removing format dependencies, as formats have been changed to
plugins


CuriousVini (approved):
lgtm

PR #65, opened 2018-11-08
Drop snapshot

rohitsinha54:


pom.xml
```diff
@@ -509,8 +509,8 @@
         <version>1.0.1</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[5.0.0,6.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[5.0.0,6.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline[5.1.1,6.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-streams[5.1.1,6.0.0-SNAPSHOT)</parent>
```
elfenheart: I think you don't need to change this parent artifact min version. Was there a platform change that required the gcp plugins to only work with 5.1.1?

albertshau: It should be 5.1.0 (doesn't need to be 5.1.1). It uses the new time/date types introduced in 5.1.0.

albertshau (approved):
lgtm

PR #68, opened 2018-12-04
CDAP-14583 refactor multi gcs sink to use format plugins

albertshau:
Refactor the GCSMultiBatchSink to use output format plugins just
like the normal GCS Sink. This change makes it so that the plugin
automatically gets the other formats supported by the GCS sink,
like json, csv, and tsv.


albertshau:
this also should fix the build failures preventing the merge of hydrator-plugins release to developer (https://builds.cask.co/browse/HYP-BAD519)

src/main/java/co/cask/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -65,68 +63,48 @@ public GCSMultiBatchSink(GCSMultiBatchSinkConfig config) {
...
+                                                       format.name().toLowerCase()));
+    }
   }
 
   @Override
-  public void prepareRun(BatchSinkContext context) throws IOException, UnsupportedTypeException {
+  public void prepareRun(BatchSinkContext context) throws IOException, InstantiationException {
     config.validate();
-    for (Map.Entry<String, String> argument : context.getArguments()) {
+    Map<String, String> baseProperties = new HashMap<>();
```
rohitsinha54: This can be replaced with:
```
Map<String, String> baseProperties = new HashMap<>(GCPUtils.getFileSystemProperties(config));
```

rohitsinha54 (approved):
One minor comment. Besides that LGTM

PR #71, opened 2018-12-18
CDAP-14655 consider bigquery datetime as string

CuriousVini:
BigQuery type DATETIME is exported as string value. So convert DATETIME type to string when data is read from big query. 
Also format TIME field to specific ISO 8601 format `HH:mm:ss.SSSSSS`. This change is needed because LocalTime.toString() supports `HH:mm` ISO 8601 format, whereas bigquery does not support it for some reason. 

image

image


albertshau (approved):
lgtm

PR #75, opened 2019-02-04
CDAP-14740: Add Cloud Datastore Source and Sink plugins

aielchiieva:
1. Implemented Datastore Source plugin.
2. Implemented Datastore Sink plugin.
3. Added common utility methods for both plugins.
4. Added appropriate unit tests.

PR #78, opened 2019-02-26
Fix docs build, similar to other projects, such as https://github.com/cdapio/hydrator-plugins/pull/744.

anwar6953:
Example of failing build: https://builds.cask.co/browse/CDAP-BUT-1707/log
Caused by merging of https://github.com/data-integrations/google-cloud/pull/75 (only fails with `-Prelease`)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,7 @@
   </issueManagement>
 
   <properties>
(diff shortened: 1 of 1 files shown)


albertshau (approved):
lgtm

PR #79, opened 2019-02-26
CDAP-14895: Support array / complex union for Cloud Datastore Batch Source and Sink

aielchiieva:
1. Support array / complex union for Cloud Datastore Batch Source
2. Support array / complex union for Cloud Datastore Batch Sink
3. Add unit tests

### DIFF
--- a/src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
+++ b/src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
@@ -42,6 +42,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -329,11 +331,15 @@ public class RecordToEntityTransformer {
           .build();
       case ARRAY:
         Schema elementSchema = Schema.recordOf("arrayElementSchema",
-                                               Schema.Field.of(fieldName, fieldSchema.getComponentSchema()
-                                               ));
-        List<?> arrayValues = getValue(record::get, fieldName, fieldType.toString(), List.class);
+                                               Schema.Field.of(fieldName, fieldSchema.getComponentSchema()));
+
+        Collection<Object> arrayValues = toCollection(fieldName, fieldType, record.get(fieldName));
         List<Value<?>> values = arrayValues.stream()
           .map(value -> {
+            // Wrap array element into record to be able to re-use convertToValue method type conversion logic
+            // because StructuredRecord has different methods to extract value based on schema type.
+            // For example: record.get(fieldName) is used for most of the values,
+            // record.getTimestamp(fieldName) is used for timestamp values.
             StructuredRecord structuredRecord = StructuredRecord.builder(elementSchema)
               .set(fieldName, value)
               .build();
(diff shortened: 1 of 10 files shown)


src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
```diff
@@ -268,77 +272,107 @@ private Key transformKeyLiteralToKey(Schema.Field field, KeyFactory keyFactory,
...
           .build();
-        break;
+      case ARRAY:
+        Schema elementSchema = Schema.recordOf("arrayElementSchema",
+                                               Schema.Field.of(fieldName, fieldSchema.getComponentSchema()
+                                               ));
+        List<?> arrayValues = getValue(record::get, fieldName, fieldType.toString(), List.class);
+        List<Value<?>> values = arrayValues.stream()
+          .map(value -> {
+            StructuredRecord structuredRecord = StructuredRecord.builder(elementSchema)
```
albertshau, line 343: It looks like a StructuredRecord is only being created so that convertToValue() can be called? Is it possible to refactor things so that the method looks something like:

`convertToValue(Schema.Field field, Object value, boolean excludeFromIndex)`

That way you can just convert the array values directly instead of wrapping it in a StructuredRecord

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
```diff
@@ -268,77 +272,107 @@ private Key transformKeyLiteralToKey(Schema.Field field, KeyFactory keyFactory,
...
-        outValue = EntityValue.newBuilder(nestedBuilder.build())
+        return EntityValue.newBuilder(nestedBuilder.build())
           .setExcludeFromIndexes(excludeFromIndex)
           .build();
-        break;
+      case ARRAY:
+        Schema elementSchema = Schema.recordOf("arrayElementSchema",
+                                               Schema.Field.of(fieldName, fieldSchema.getComponentSchema()
+                                               ));
+        List<?> arrayValues = getValue(record::get, fieldName, fieldType.toString(), List.class);
```
albertshau: The underlying value can also be an array. See https://github.com/cdapio/hydrator-plugins/blob/develop/hydrator-common/src/main/java/co/cask/hydrator/common/RecordConverter.java#L63.

src/main/java/co/cask/gcp/datastore/source/DatastoreSource.java
```diff
@@ -209,44 +221,73 @@ Schema constructSchema(Entity entity, boolean isIncludeKey, String keyName) {
...
-        return Schema.Field.of(name, Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS)));
-      case BLOB:
-        return Schema.Field.of(name, Schema.nullableOf(Schema.of(Schema.Type.BYTES)));
       case NULL:
         // Datastore stores null values in Null Type unlike regular databases which have nullable type definition
         // set default String type, instead of skipping the field
-        LOG.debug("Unable to determine type, setting default type '{}' for the field '{}'", Schema.Type.STRING, name);
-        return Schema.Field.of(name, Schema.nullableOf(Schema.of(Schema.Type.STRING)));
+        LOG.debug("Field '{}' is null, setting default type '{}'", name, Schema.Type.STRING);
+        return Schema.of(Schema.Type.STRING);
```
albertshau: why not use the Schema.Type.NULL type for this?

albertshau (commented):
minor comments

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
aielchiieva: According to Datastore rules array cannot contain another array (https://cloud.google.com/datastore/docs/concepts/entities#array). Added this validation checks for Source and Sink schemas + unit tests.

src/main/java/co/cask/gcp/datastore/source/DatastoreSource.java
aielchiieva: Agree, changed to NULL.

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
aielchiieva: There is no way to provide Object value instead of record without duplicate checks because we use two different StructuredRecord methods based on schema types:
1. `record.get(fieldName)` for most of the values
2. `record.getTimestamp(fieldName)` for timestamp values

aielchiieva:
@albertshau addressed code review comments, please take a look.

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
albertshau: I meant the object returned by StructuredRecord.get() can be an array, it is not necessarily a List.

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
albertshau: Hm I see. This is a deficiency in the design of these logical types in StructuredRecord. Can you leave a comment here then about why this wrapping is needed?

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
aielchiieva: Thanks for the clarification, updated the code and added unit test.

src/main/java/co/cask/gcp/datastore/sink/RecordToEntityTransformer.java
aielchiieva: Done.

aielchiieva:
@albertshau addressed changes after second round of code review, please take a look.

albertshau (approved):
lgtm

PR #82, opened 2019-03-09
CDAP-14844 remove plugin endpoint usage

albertshau:
Removed usage of plugin endpoints since they are replaced by the
pipeline system service. Instead, moving related logic to the
configurePipeline() method and setting the output schema directly
there. Made the schema property nullable so that the schema is
fetched from underlying storage if it is not provided.
The 'Get Schema' button is configured to not pass schema to the
plugin.

### DIFF
--- a/widgets/Datastore-batchsource.json
+++ b/widgets/Datastore-batchsource.json
@@ -33,10 +33,14 @@
           "label": "Kind",
           "name": "kind",
           "plugin-function": {
-            "method": "POST",
+            "label": "Get Schema",
             "widget": "outputSchema",
             "output-property": "schema",
-            "plugin-method": "getSchema"
+            "omit-properties": [
+              {
+                "name": "schema"
+              }
+            ]
           }
         },
         {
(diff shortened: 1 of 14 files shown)


CuriousVini (approved):
lgtm 👍

CuriousVini:
Resolved conflicts in BigQuery source because of name change from BigQueryUtils to BigQueryUtil in [this](https://github.com/data-integrations/google-cloud/pull/84/files/caf0c68c7197a447a289dbafe6d072d6e55f8173#discussion_r269267439) PR.

CuriousVini:
Fixed widget json and tested on cluster
image

widgets/Datastore-batchsource.json
```diff
@@ -33,10 +33,14 @@
           "label": "Kind",
           "name": "kind",
           "plugin-function": {
-            "method": "POST",
-            "widget": "outputSchema",
+            "label": "Get Schema",
+            "widget": "getSchema",
```
elfenheart: should be `outputSchema`

elfenheart (approved):
👍 LGTM

PR #83, opened 2019-03-22
CDAP-15105 do not create external resources in preview

yaojiefeng:
We should not create external tables in preview, otherwise on k8s this can fail due to lack of permissions.


CuriousVini (approved):
lgtm 👍

PR #84, opened 2019-03-26
CDAP-15128: Improve BigQuery package structure

aielchiieva:
This PR creates three the sub-packages in `co.cask.gcp.bigquery` package: 
- source
- sink
- util

and moves appropriate classes into these packages.

### DIFF
--- a/src/main/java/co/cask/gcp/bigquery/util/BigQueryUtil.java
+++ b/src/main/java/co/cask/gcp/bigquery/util/BigQueryUtil.java
@@ -14,9 +14,11 @@
  *  the License.
  */
 
-package co.cask.gcp.bigquery;
+package co.cask.gcp.bigquery.util;
 
 import co.cask.cdap.api.data.schema.Schema;
+import co.cask.gcp.bigquery.sink.BigQuerySink;
+import co.cask.gcp.bigquery.source.BigQuerySource;
 import co.cask.gcp.gcs.GCSPath;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryOptions;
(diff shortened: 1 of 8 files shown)


src/main/java/co/cask/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -45,7 +47,7 @@
 /**
  * Common Util class for big query plugins such as {@link BigQuerySource} and {@link BigQuerySink}
  */
-final class BigQueryUtils {
+public final class BigQueryUtil {
```
jayfjin, line 50: why the change from BigQueryUtils to BigQueryUtil? I think most other places use the plural form

jayfjin (commented):
Just to confirm, these will still be in the same jar?

src/main/java/co/cask/gcp/bigquery/util/BigQueryUtil.java
aielchiieva: Looked through the `google-cloud` project, we have the following:
`common` package: FileSetUtil (actually unused in the project), GCPUtils
`datastore` package: DatastoreUtil + two util packages in source and sink sub-packages
`spanner` package: SpannerUtil

Looks like `util` is slightly more used than `utils`. Personally I also prefer `util` naming.
If you want, I can also rename `GCPUtils` to `GCPUtil` for consistency.

aielchiieva:
@jayfjin 
> Just to confirm, these will still be in the same jar?

Yes, this does not affect jar packaging.

src/main/java/co/cask/gcp/bigquery/util/BigQueryUtil.java
jayfjin: I see, seems like a different naming convention inside google-cloud vs cdap. I'm ok with either then

PR #85, opened 2019-04-02
CDAP-15129: Big Query multi table sink plugin

aielchiieva:
1. Create AbstractBigQuerySinkConfig which holds common config properties for Big Query Sink and Multi Sink plugins.
2. Create AbstractBigQuerySink which contains common logic of data output into Big Query tables and coordinates overall output process, including pre and post run actions.
3. Add BigQueryMultiSink and MultiSinkOutputFormatProvider classes which provide core implementation for Big Query Multi Sink plugin allowing to write data into multiple Big Query tables.


docs/BigQueryMultiTable-batchsink.md
```diff
@@ -0,0 +1,68 @@
...
+
+Credentials
+-----------
+If the plugin is run on a Google Cloud Dataproc cluster, the service account key does not need to be
+provided and can be set to 'auto-detect'.
+Credentials will be automatically read from the cluster environment.
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access BigQuery and Google Cloud Storage.
```
jayfjin, line 17: Great documentation!

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -0,0 +1,419 @@
...
+public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, JsonObject, NullWritable> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractBigQuerySink.class);
+
+  private static final String gcsPathFormat = "gs://%s";
+  private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
+  private static final DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSS");
+
+  // UUID for the run. Will be used as bucket name if bucket is not provided.
+  private final UUID uuid = UUID.randomUUID();
```
jayfjin, line 68: Let's add as a comment here that the default bucket name needs to be a UUID since GCS bucket names must be globally unique.

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -0,0 +1,419 @@
...
+                                                                  Schema tableSchema);
+
+  /**
+   * If provided dataset name does not exists in Big Query, creates it.
+   * This operation only takes place, if pipeline does not run in preview mode.
+   *
+   * @param previewEnabled indicates whether the pipeline is running in preview.
+   */
+  private void createDataset(boolean previewEnabled) throws IOException {
+    if (previewEnabled) {
```
jayfjin, line 236: If the bucket does not exist then this will fail in preview? Is this intended? I think this may be confusing for the user

jayfjin: Also I think this is backwards. if (previewEnabled) == true means that the pipeline is running in preview mode, where as it seems you want the opposite condition.

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -0,0 +1,419 @@
...
+
+  /**
+   * Generates full path to temporary bucket based on given bucket and table names.
+   *
+   * @param bucket bucket name
+   * @param tableName table name
+   * @return full path to temporary bucket
+   */
+  private String getTemporaryGcsPath(String bucket, String tableName) {
+    return String.format(gcsPathFormat + "/hadoop/input/%s-%s", bucket, tableName, uuid);
```
jayfjin: 3 params given but only two %s?

src/main/java/co/cask/gcp/bigquery/sink/MultiSinkOutputFormatProvider.java
```diff
@@ -0,0 +1,145 @@
...
+
+      delegate.write(object, value);
+    }
+
+    @Override
+    public void close(TaskAttemptContext context) throws IOException, InterruptedException {
+      delegate.close(context);
+    }
+  }
+
```
jayfjin: nit: extra newline

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
bdmogal: Also, why do we need ``hadoop`` in the path? can we remove it?

docs/BigQueryMultiTable-batchsink.md
aielchiieva: Thanks, mostly copied from existing Batch Sink plugin though :)

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
aielchiieva: Done.

src/main/java/co/cask/gcp/bigquery/sink/MultiSinkOutputFormatProvider.java
aielchiieva: Removed.

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
aielchiieva: 1. Other param is in the constant: `private static final String gcsPathFormat = "gs://%s"`. Created `temporaryBucketFormat` to avoid confusion: `private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s"`.
2. Was present in the original implementation but we can remove it. Removed.

src/main/java/co/cask/gcp/bigquery/sink/AbstractBigQuerySink.java
aielchiieva: Yes, my fault. We should only create bucket is we are not in preview mode. Updated the code.
In preview mode actual run is not performed thus absent dataset has no effect.

aielchiieva:
@jayfjin thanks for the code review, updated the PR based on code review comments.

jayfjin:
lgtm

PR #90, opened 2019-04-18
[CDAP-15257] support int and float types in spanner sink

CuriousVini:
image

image

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.12.0</version>
+  <version>0.12.1</version>
   <name>Google Cloud Plugins</name>
   <packaging>jar</packaging>
   <description>Plugins for Google Big Query</description>
@@ -59,7 +59,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <avro.version>1.7.7</avro.version>
     <cdap.version>6.0.0-SNAPSHOT</cdap.version>
-    <cdap.plugin.version>2.2.0</cdap.plugin.version>
+    <cdap.plugin.version>2.2.0-SNAPSHOT</cdap.plugin.version>
     <guava.version>20.0</guava.version>
     <hadoop.version>2.8.0</hadoop.version>
     <jackson.core.version>2.8.11.1</jackson.core.version>
--- a/src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
@@ -29,11 +29,6 @@ import java.util.Set;
  * Spanner utility class to get spanner service
  */
 public class SpannerUtil {
-  // todo CDAP-14233 - add support for array
-  private static final Set<Schema.Type> SUPPORTED_TYPES =
-    ImmutableSet.of(Schema.Type.BOOLEAN, Schema.Type.STRING, Schema.Type.LONG, Schema.Type.DOUBLE,
-                    Schema.Type.BYTES);
-
   private static final Set<Schema.LogicalType> SUPPORTED_LOGICAL_TYPES =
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -19,7 +19,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.12.0</version>
+  <version>0.12.1</version>
```
bdmogal, line 22: are any more fixes planned? Until we release, let's make this 0.12.1-SNAPSHOT?

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
```diff
@@ -116,9 +111,11 @@ public static String convertSchemaToCreateStatement(String tableName, String pri
         case STRING:
           spannerType = "STRING(MAX)";
           break;
+        case INT:
```
bdmogal, line 114: does this account for all types in ``Schema.Type`` now?

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -143,9 +143,15 @@ public void write(NullWritable nullWritable, StructuredRecord record) throws IOE
           case STRING:
             builder.set(name).to(record.<String>get(name));
             break;
+          case INT:
```
bdmogal, line 146: similar question here.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
```diff
@@ -24,12 +25,16 @@
...
 import java.io.IOException;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
  * Spanner source config
  */
 public class SpannerSourceConfig extends GCPReferenceSourceConfig {
+  // todo CDAP-14233 - add support for array
+  private static final Set<Schema.Type> SUPPORTED_TYPES =
```
bdmogal, line 36: can't this be re-used with SpannerSinkConfig?

bdmogal (changes requested):
some minor questions about version + reusing code. Other than that, as long as it is tested and all primitive types are supported, this looks good.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
CuriousVini: This is because both of them support different types. Sources does not allow importing int and float.

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
CuriousVini: The sink supports all primitive types - boolean, string, int, long, float, double, bytes, date and timestamp. However, source schema can only have boolean, string, long, double, bytes, date and timestamp. This is because java int and float are 4 bytes and spanner float and int are 8 bytes
https://cloud.google.com/spanner/docs/data-types

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
CuriousVini: same as above.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
bdmogal: why is that?

pom.xml
bdmogal: per our discussion, this is ok presuming no more fixes will be pushed to 0.12.1

bdmogal (approved):

chtyim (approved):
LGTM

CuriousVini:
squashing and merging.

PR #91, opened 2019-04-26
[CDAP-15289] Improve error message from big query sink

CuriousVini:
image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -1,0 +1,290 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sink;
+
+import com.google.api.client.json.JsonParser;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.util.BackOff;
+import com.google.api.client.util.ExponentialBackOff;
+import com.google.api.client.util.Sleeper;
+import com.google.api.services.bigquery.Bigquery;
+import com.google.api.services.bigquery.model.Dataset;
+import com.google.api.services.bigquery.model.EncryptionConfiguration;
+import com.google.api.services.bigquery.model.ErrorProto;
+import com.google.api.services.bigquery.model.Job;
+import com.google.api.services.bigquery.model.JobConfiguration;
+import com.google.api.services.bigquery.model.JobConfigurationLoad;
+import com.google.api.services.bigquery.model.JobReference;
+import com.google.api.services.bigquery.model.TableFieldSchema;
+import com.google.api.services.bigquery.model.TableReference;
+import com.google.api.services.bigquery.model.TableSchema;
+import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegateBigQueryOutputFormat.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
albertshau: The spacing in the copyright looks funny, is there an extra space in front?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegateBigQueryOutputFormat.java
```diff
@@ -0,0 +1,107 @@
...
+      delegate.abortTask(taskContext);
+    }
+
+    @Override
+    public void commitJob(JobContext jobContext) throws IOException {
+      try {
+        // CDAP-15289 - add specific error message.
+        delegate.commitJob(jobContext);
+      } catch (IOException e) {
+        throw new IOException("Error occurred while writing to BigQuery. Please see BigQuery job logs for more " +
```
albertshau: the error message should contain the job id, otherwise the user won't know which logs to examine.

CuriousVini: The BigQuery import job is submitted by output committer. I do not think that is exposed.

albertshau: Doesn't the plugin know the job id? Without the id, this error message is not really much better than the previous one.

CuriousVini: Had an offline discussion with @albertshau . BigQuery outputformat `IndirectBigQueryOutputFormat` does not provide a way to get job Id. Therefore, we decided to  override logic of `IndirectBigQueryOutputCommitter#commitJob` for useful log message.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -0,0 +1,280 @@
...
+          IOException.class,
+          sleeper);
+
+        elapsedTime = System.currentTimeMillis() - startTime;
+        LOG.debug("Job status ({} ms) {}: {}", elapsedTime, jobReference.getJobId(),
+                  pollJob.getStatus().getState());
+        if (pollJob.getStatus().getState().equals("DONE")) {
+          notDone = false;
+          if (pollJob.getStatus().getErrorResult() != null) {
+            // Only add first error message in the exception. For other errors user should look at BigQuery job logs.
```
albertshau, line 227: which one is the first error message? getErrorResult().getMessage()?

CuriousVini: Fixed
image

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -0,0 +1,282 @@
...
+          RetryDeterminer.RATE_LIMIT_ERRORS,
+          IOException.class,
+          sleeper);
+
+        elapsedTime = System.currentTimeMillis() - startTime;
+        LOG.debug("Job status ({} ms) {}: {}", elapsedTime, jobReference.getJobId(),
+                  pollJob.getStatus().getState());
+        if (pollJob.getStatus().getState().equals("DONE")) {
+          notDone = false;
+          if (pollJob.getStatus().getErrorResult() != null) {
```
albertshau, line 216: Is it possible for getErrorResult() to be non-null, but for errors to be empty or null? 

It's probably best to check for these scenarios to be safe. I think this check should be for the errors list, and we can have an 'else if' when getErrorResult() is not null

CuriousVini: As per javadoc`getErrorResult() != null` indicates an unsuccessful job. So this check should be here. However, the javadoc does not indicate any contract between these 2 methods. So we should do a null check for `getErrors` method before using it.

albertshau (approved):
lgtm

PR #92, opened 2019-05-01
[CDAP-14579] Allow spaces in gcs plugins

CuriousVini:
Reference for the change:
https://github.com/GoogleCloudPlatform/bigdata-interop/issues/31

image

image

image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
@@ -16,9 +16,11 @@
 
 package io.cdap.plugin.gcp.gcs;
 
+import com.google.common.net.UrlEscapers;
+
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.Objects;
+import java.util.regex.Pattern;
 
 /**
  * A path on GCS. Contains information about the bucket and blob name (if applicable).
@@ -26,11 +28,12 @@ import java.util.Objects;
  */
 public class GCSPath {
   public static final String ROOT_DIR = "/";
+  private static final String SCHEME = "gs://";
   private final URI uri;
   private final String bucket;
   private final String name;
 
-  GCSPath(URI uri, String bucket, String name) {
+  private GCSPath(URI uri, String bucket, String name) {
     this.uri = uri;
     this.bucket = bucket;
     this.name = name;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
```diff
@@ -93,13 +93,14 @@ public static GCSPath from(String path) {
 
   private static URI getURI(String path) {
     try {
-      URI uri = new URI(path);
+      String modifiedPath = path.replaceAll("\\s", "%20");
```
chtyim: Don't do this manual replacement. Use `URLEncoder.encode(path, "UTF-8")` instead.
Also, `\\s` includes all white spaces (e.g. tab), which should be encoded as `%09`.

chtyim: Correction, please use `UrlEscapers.urlPathSegmentEscaper().escape(path)` instead.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -120,7 +120,7 @@ public void validate() {
 
     @Override
     public String getPath() {
-      return GCSPath.from(path).getUri().toString();
+      return GCSPath.from(path).getUri().toString().replaceAll("%20", " ");
```
chtyim: Why need to reverse the encoding?

src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
```diff
@@ -83,28 +86,31 @@ public int hashCode() {
...
-      throw new IllegalArgumentException(String.format("Invalid path '%s'. The path must be of form " +
-                                                         "'gs://<bucket-name>/path'.", path), e);
+    if (!Pattern.matches("[a-zA-Z0-9-_.]+", bucket)) {
+      throw new IllegalArgumentException(String.format("Invalid bucket name in path '%s'. Bucket name should only " +
+                                                         "contain alphanumeric, '-'. '_' and '.'.", path));
     }
+
+    String file = idx > 0 ? path.substring(idx) : "";
+    URI uri = URI.create(SCHEME + bucket + UrlEscapers.urlFragmentEscaper().escape(file));
+    return new GCSPath(uri, bucket, file.replaceAll("^/", ""));
```
chtyim: Why need to do the `replaceAll` with regex? If the file is `"/abc/def"`, then it will become `"abcdec"`.. is it desired? It won't match with what's in the URI. I think ideally the `GCPath` constructor should only takes `URI`, as both `bucket` and `file` should be derived from that.

CuriousVini: This regex only replaces leading `/` because current behavior expects name without leading `/`.
image

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -67,13 +68,15 @@ public void run(ActionContext context) throws Exception {
...
+    configuration.set("fs.gs.path.encoding", "uri-path");
 
     configuration.setBoolean("fs.gs.impl.disable.cache", true);
     configuration.setBoolean("fs.gs.metadata.cache.enable", false);
 
     List<Path> gcsPaths = new ArrayList<>();
     for (String path : config.getPaths()) {
-      gcsPaths.add(new Path(GCSPath.from(path).getUri()));
+      GCSPath gcsPath = GCSPath.from(path);
+      gcsPaths.add(new Path(SCHEME + gcsPath.getBucket() + "/" + gcsPath.getUri().getPath()));
```
chtyim: Why not just do `new Path(gcsPath.getUri())`?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -67,12 +68,14 @@ public void run(ActionContext context) throws Exception {
...
     configuration.set("fs.gs.project.id", projectId);
+    configuration.set("fs.gs.path.encoding", "uri-path");
 
     configuration.setBoolean("fs.gs.impl.disable.cache", true);
 
     List<Path> gcsPaths = new ArrayList<>();
     for (String path : config.getPaths()) {
-      gcsPaths.add(new Path(GCSPath.from(path).getUri()));
+      GCSPath gcsPath = GCSPath.from(path);
+      gcsPaths.add(new Path(SCHEME + gcsPath.getBucket() + "/" + gcsPath.getUri().getPath()));
```
chtyim: Same here

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -122,7 +122,7 @@ public void validate() {
     @Override
     public String getPath() {
       GCSPath gcsPath = GCSPath.from(path);
-      return SCHEME + gcsPath.getBucket() + "/" + gcsPath.getUri().getPath();
+      return SCHEME + gcsPath.getBucket() + gcsPath.getUri().getPath();
```
CuriousVini, line 125: uri.getPath() returns path starting with `/`

chtyim (approved):

PR #93, opened 2019-05-02
Allow gs:// scheme in bucket name

CuriousVini:
image

image

image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -267,7 +267,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
    */
   private String configureBucket() {
     String bucket = getConfig().getBucket();
-    if (getConfig().getBucket() == null) {
+    if (bucket == null) {
       bucket = uuid.toString();
       // By default, this option is false, meaning the job can not delete the bucket.
       // So enable it only when bucket name is not provided.
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -17,14 +17,17 @@ package io.cdap.plugin.gcp.bigquery.sink;
 
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
 import io.cdap.plugin.gcp.common.GCPReferenceSinkConfig;
 
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 /**
  * Base class for Big Query batch sink configs.
  */
 public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig {
+  private static final String SCHEME = "gs://";
 
   @Macro
   @Description("The dataset to write to. A dataset is contained within a specific project. "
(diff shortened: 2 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -267,12 +267,14 @@ private String getTemporaryGcsPath(String bucket, String tableName) {
...
     String bucket = getConfig().getBucket();
-    if (getConfig().getBucket() == null) {
+    if (bucket == null || bucket.trim().isEmpty()) {
       bucket = uuid.toString();
       // By default, this option is false, meaning the job can not delete the bucket.
       // So enable it only when bucket name is not provided.
       baseConfiguration.setBoolean("fs.gs.bucket.delete.enable", true);
     }
+
+    bucket = BigQueryUtil.getBucketNameWithoutScheme(bucket);
```
albertshau: If this is invalid, we should throw an InvalidConfigPropertyException that includes the name of the property that is invalid.

albertshau (commented):
Should be sure to validate the bucket at configure time too

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -48,4 +51,21 @@ public String getBucket() {
     return bucket;
   }
 
+  @Override
+  public void validate() {
+    super.validate();
+    if (!containsMacro("bucket") && bucket != null) {
+      String bucketName = BigQueryUtil.getBucketNameWithoutScheme(bucket);
+      if (bucketName.isEmpty()) {
+        throw new InvalidConfigPropertyException("Bucket name must be of format gs://<bucket> or <bucket>.",
```
albertshau: In prepareRun, this is actually allowed, with the bucket getting auto-created. I think we can just skip validation if it is empty.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -48,4 +51,21 @@ public String getBucket() {
...
+                                                 "bucket");
+      }
+      // Basic validation for allowed characters as per https://cloud.google.com/storage/docs/naming
+      Pattern p = Pattern.compile("[a-z0-9._-]+");
+      if (!p.matcher(bucket).matches()) {
+        throw new InvalidConfigPropertyException("Bucket name must be a valid bucket name of format " +
+                                                   "gs://<bucket> or <bucket>", "bucket");
```
albertshau: This happens if the bucket contains unsupported characters. A better error message would be something like:
"Bucket names can only contain lowercase characters, numbers, '.', '_', and '-'."

Also, is there a length limit for buckets?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -48,4 +51,21 @@ public String getBucket() {
     return bucket;
   }
 
+  @Override
+  public void validate() {
+    super.validate();
+    if (!containsMacro("bucket") && bucket != null) {
+      String bucketName = BigQueryUtil.getBucketNameWithoutScheme(bucket);
```
albertshau: it would be better if the getBucket() methods remove the scheme and also return null if the bucket is empty. Then you just use getBucket() everywhere instead of forcing every caller to check for empty and remove the scheme.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
CuriousVini: Wouldn't bucket be null in that case? If its not specified in the config?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
CuriousVini: For length there is a limit to number of characters. 
However, This check is just a basic allowed character check, which is just a subset of all the constraints applied to gcs bucket name. This is done to fail fast if there are any characters in the bucket name which are not allowed.
https://cloud.google.com/storage/docs/naming

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
albertshau: Actually I'm not sure how this can happen through the UI, probably only if the pipeline is created through REST, if somebody used an empty string instead of not including the property.

CuriousVini: This is because UI allows spaces in the bucket name. `getBucketNameWithoutScheme` method trims the spaces and can return an empty string. However, with latest commit this would be allowed and a temporary bucket will get created with uuid.
image

albertshau (approved):
lgtm

CuriousVini:
Squashing and merging.

PR #95, opened 2019-05-03
CDAP-14951 fix bug when dataset is in a different region

albertshau:
Fixed a bug that occurs when the bigquery dataset that is being
written to is in a different region than the dataproc cluster.
This was a bug in the bigquery library that the project currently
depends on, where the wrong get job request was not setting the
job location.


CuriousVini (approved):

anwar6953:
Let's also commit this to the dependent bigquery library?

albertshau:
This is fixed in the latest version (https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/BigQueryUtils.java#L90), but updating libraries in this project is a huge hassle as it requires updating all the various unrelated libraries due to conflicting dependencies of dependencies.

anwar6953:
Ah ok. It still seems like a good idea to spend the effort on (later on, perhaps), since we could be missing out on other bug fixes as well.
Definitely will require additional validation, since it's a bigger change, though.

PR #96, opened 2019-05-04
[CDAP-15256] support array type in big query plugins

CuriousVini:
https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#array-type

**Source**
image

**Pipeline**
image

**Output Table**
image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -26,8 +26,6 @@ import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryOutputConfiguration;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
-import com.google.gson.JsonNull;
-import com.google.gson.JsonObject;
 import io.cdap.cdap.api.data.batch.Output;
 import io.cdap.cdap.api.data.batch.OutputFormatProvider;
 import io.cdap.cdap.api.data.format.StructuredRecord;
@@ -41,12 +39,12 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.time.format.DateTimeFormatter;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -57,14 +55,12 @@ import javax.annotation.Nullable;
 /**
  * Base class for Big Query batch sink plugins.
  */
-public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, JsonObject, NullWritable> {
+public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, Text, NullWritable> {
 
   private static final Logger LOG = LoggerFactory.getLogger(AbstractBigQuerySink.class);
 
   private static final String gcsPathFormat = "gs://%s";
   private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s";
-  private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
-  private static final DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSS");
 
   // UUID for the run. Will be used as bucket name if bucket is not provided.
   // UUID is used since GCS bucket names must be globally unique.
@@ -130,65 +126,6 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
       .collect(Collectors.toList());
     recordLineage(context, outputName, tableSchema, fieldNames);
     context.addOutput(Output.of(outputName, getOutputFormatProvider(configuration, tableName, tableSchema)));
-  }
-
-  protected final void decodeSimpleTypes(JsonObject json, String name, StructuredRecord input) {
-    Object object = input.get(name);
-    Schema.Field field = input.getSchema().getField(name);
-
-    if (field == null) {
-      throw new IllegalStateException(String.format("Field '%s' is absent in input record", name));
-    }
-
-    Schema schema = BigQueryUtil.getNonNullableSchema(field.getSchema());
-
-    if (object == null) {
-      json.add(name, JsonNull.INSTANCE);
-      return;
-    }
-
-    Schema.LogicalType logicalType = schema.getLogicalType();
-    if (logicalType != null) {
-      switch (logicalType) {
-        case DATE:
-          json.addProperty(name, Objects.requireNonNull(input.getDate(name)).toString());
-          break;
-        case TIME_MILLIS:
-        case TIME_MICROS:
-          json.addProperty(name, timeFormatter.format(Objects.requireNonNull(input.getTime(name))));
-          break;
-        case TIMESTAMP_MILLIS:
-        case TIMESTAMP_MICROS:
-          //timestamp for json input should be in this format yyyy-MM-dd HH:mm:ss.SSSSSS
-          json.addProperty(name, dateTimeFormatter.format(Objects.requireNonNull(input.getTimestamp(name))));
-          break;
-        default:
-          throw new IllegalStateException(
-            String.format("Field '%s' is of unsupported type '%s'", name, logicalType.getToken()));
-      }
-      return;
-    }
-
-    switch (schema.getType()) {
-      case NULL:
-        json.add(name, JsonNull.INSTANCE); // nothing much to do here.
-        break;
-      case INT:
-      case LONG:
-      case FLOAT:
-      case DOUBLE:
-        json.addProperty(name, (Number) object);
-        break;
-      case BOOLEAN:
-        json.addProperty(name, (Boolean) object);
-        break;
-      case STRING:
-        json.addProperty(name, object.toString());
-        break;
-      default:
-        throw new IllegalStateException(
-          String.format("Field '%s' is of unsupported type '%s'", name, schema.getType()));
-    }
   }
 
   /**
@@ -396,9 +333,18 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
     return inputFields.stream()
       .map(field -> new BigQueryTableFieldSchema()
         .setName(field.getName())
-        .setType(getTableDataType(BigQueryUtil.getNonNullableSchema(field.getSchema())).name())
-        .setMode(field.getSchema().isNullable() ? Field.Mode.NULLABLE.name() : Field.Mode.REQUIRED.name()))
+        .setType(getTableDataType(field.getSchema()).name())
+        .setMode(getMode(field.getSchema()).name()))
       .collect(Collectors.toList());
+  }
+
+  private Field.Mode getMode(Schema schema) {
+    if (schema.getType() == Schema.Type.ARRAY) {
+      return Field.Mode.REPEATED;
+    } else if (schema.isNullable()) {
+      return Field.Mode.NULLABLE;
+    }
+    return Field.Mode.REQUIRED;
   }
 
   /**
@@ -443,6 +389,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
   }
 
   private LegacySQLTypeName getTableDataType(Schema schema) {
+    schema = BigQueryUtil.getNonNullableSchema(schema);
     Schema.LogicalType logicalType = schema.getLogicalType();
 
     if (logicalType != null) {
@@ -474,9 +421,10 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
         return LegacySQLTypeName.BOOLEAN;
       case BYTES:
         return LegacySQLTypeName.BYTES;
+      case ARRAY:
+        return getTableDataType(schema.getComponentSchema());
       default:
         throw new IllegalStateException("Unsupported type " + type);
     }
   }
-
 }
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sink;
+
+import com.google.gson.stream.JsonWriter;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
+
+import java.io.IOException;
+import java.lang.reflect.Array;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ * Util class to convert structured record into json.
+ */
+public final class BigQueryRecordToJson {
+  private static final DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSS");
+  private static final DateTimeFormatter TIME_FORMATTER = DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSS");
+
+  /**
+   * Writes object and writes to json writer.
+   * @param writer json writer to write the object to
+   * @param name name of the field to be written. If its null, the name of the field is not included in the json writer
+   * @param object object to be written
+   * @param fieldSchema field schema to be written
+   */
+  public static void write(JsonWriter writer, @Nullable String name, Object object,
+                           Schema fieldSchema) throws IOException {
+    Schema schema = BigQueryUtil.getNonNullableSchema(fieldSchema);
+    switch (schema.getType()) {
+      case NULL:
+      case INT:
+      case LONG:
+      case FLOAT:
+      case DOUBLE:
+      case BOOLEAN:
+      case STRING:
+      case BYTES:
+        writeSimpleTypes(writer, name, object, schema);
+        break;
+      case ARRAY:
+        writeArray(writer, name, object, schema);
+        break;
+      default:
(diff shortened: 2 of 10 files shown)


nitinmotgi:
Can u test for a file from GCS into BQ with arrays please.

CuriousVini:
yes, tested with gcs -> bq.
image

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -168,26 +212,72 @@ protected final void decodeSimpleTypes(JsonObject json, String name, StructuredR
...
+    }
+
+    if (!(value instanceof Collection) && !value.getClass().isArray()) {
+      throw new IllegalArgumentException("The value should be of type collection or array. Got: " + value.getClass());
+    }
+
+    Schema componentSchema = BigQueryUtil.getNonNullableSchema(fieldSchema.getComponentSchema());
+    // Arrays within arrays are not allowed in big query
+    if (componentSchema.getType() == Schema.Type.ARRAY) {
+      throw new IllegalArgumentException("Nested Arrays is not a valid type in big query.");
```
albertshau: should mention which field this is. "Field '%s' is a nested array, which is not a valid BigQuery type."

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -168,26 +212,72 @@ protected final void decodeSimpleTypes(JsonObject json, String name, StructuredR
...
+      // TODO CDAP-15256 write byte type to json writer
       default:
-        throw new IllegalStateException(
-          String.format("Field '%s' is of unsupported type '%s'", name, schema.getType()));
+        throw new IllegalStateException(String.format("Field '%s' is of unsupported type '%s'",
+                                                      name, schema.getType()));
     }
   }
 
+  private void decodeArray(JsonWriter writer, String name, Object value, Schema fieldSchema) throws IOException {
```
albertshau: 'decode' isn't a great name for these methods, as nothing is being read or decoded. 'write' would be more accurate.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -129,35 +139,69 @@ protected final void initOutput(BatchSinkContext context, String outputName,
...
+        decodeArray(writer, name, object, schema);
+        break;
+      default:
+        throw new IllegalStateException(
+          String.format("Field '%s' is of unsupported type '%s'", name, fieldSchema.getType()));
     }
+  }
 
-    Schema schema = BigQueryUtil.getNonNullableSchema(field.getSchema());
+  private void decodeSimpleTypes(JsonWriter writer, @Nullable String name, Object object,
```
albertshau: can you include a javadoc describing when the name is null.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
albertshau: also, is this checked at configure time? Other checks should be done, like checking for other types of invalid component types. I assume arrays of maps are not supported. Are arrays of structs supported? Arrays of bytes?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -168,26 +212,72 @@ protected final void decodeSimpleTypes(JsonObject json, String name, StructuredR
...
+    if (componentSchema.getType() == Schema.Type.ARRAY) {
+      throw new IllegalArgumentException("Nested Arrays is not a valid type in big query.");
+    }
+
+    writer.name(name);
+    writer.beginArray();
+    if (value instanceof Collection) {
+      for (Object element : (Collection) value) {
+        // big query does not allow null values in array items
+        if (element != null) {
```
albertshau: this will silently drop null elements, which seems bad because it's modifying the input data. I think this should cause a failure, with an error message describing which field is causing the problem.

I also wonder if we should be checking the schema to make sure it's not an array of nullable elements, as that would be a deploy time failure instead of runtime.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -342,11 +432,20 @@ private void validateSchema(String tableName, Schema tableSchema) throws IOExcep
...
         .setName(field.getName())
-        .setType(getTableDataType(BigQueryUtil.getNonNullableSchema(field.getSchema())).name())
-        .setMode(Field.Mode.NULLABLE.name()))
+        .setType(getTableDataType(field.getSchema()).name())
+        .setMode(getMode(field.getSchema()).name()))
       .collect(Collectors.toList());
   }
 
+  private Field.Mode getMode(Schema schema) {
+    if (BigQueryUtil.getNonNullableSchema(schema).getComponentSchema() != null) {
```
albertshau: it's better if this is a check that the schema type is array. Though these two are logically equivalent today, it is more clear if this is a type check on array.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -384,6 +483,7 @@ private void recordLineage(BatchSinkContext context,
   }
 
   private LegacySQLTypeName getTableDataType(Schema schema) {
+    schema = BigQueryUtil.getNonNullableSchema(schema);
```
albertshau, line 392: does this mean there was a bug with nullable schemas before this change?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -129,35 +139,69 @@ protected final void initOutput(BatchSinkContext context, String outputName,
...
-    if (field == null) {
-      throw new IllegalStateException(String.format("Field '%s' is absent in input record", name));
+  /**
+   * Decodes object and writes to json writer.
+   * @param writer json writer to write the object to
+   * @param name name of the field to be decoded
+   * @param object object to be decoded
+   * @param fieldSchema field schema to be decoded
+   */
+  protected void decode(JsonWriter writer, @Nullable String name, Object object,
```
albertshau: these decode methods are really about transforming a StructuredRecord into json object. It would be better if we pull out this logic from the plugins into a separate class (ex: RecordJsonTransformer). We can then have comprehensive unit tests for the transformation logic.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -103,15 +105,21 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
-  public void transform(StructuredRecord input, Emitter<KeyValue<JsonObject, NullWritable>> emitter) {
-    JsonObject object = new JsonObject();
-    for (Schema.Field recordField : Objects.requireNonNull(input.getSchema().getFields())) {
-      // From all the fields in input record, decode only those fields that are present in output schema
-      if (schema.getField(recordField.getName()) != null) {
-        decodeSimpleTypes(object, recordField.getName(), input);
+  public void transform(StructuredRecord input, Emitter<KeyValue<Text, NullWritable>> emitter) {
+    StringWriter strWriter = new StringWriter();
+    try (JsonWriter writer = new JsonWriter(strWriter)) {
+      writer.beginObject();
```
albertshau, line 113: I assume structs are coming in a later PR then? Otherwise we could just call the method to write a struct here.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -228,18 +228,18 @@ private void validateOutputSchema() throws IOException {
...
-  private void validateSimpleTypes(Schema.Field field) {
+  private void validateSupportedTypes(Schema.Field field) {
     String name = field.getName();
     Schema fieldSchema = BigQueryUtil.getNonNullableSchema(field.getSchema());
     Schema.Type type = fieldSchema.getType();
 
-    // Complex types like arrays, maps and unions are not supported in BigQuery plugins.
-    if (!type.isSimpleType()) {
+    // Complex types like maps and unions are not supported in BigQuery plugins.
+    if ((!type.isSimpleType() && type != Schema.Type.ARRAY)) {
```
albertshau: ok I see this is where the deploy time validation is.

There needs to be a bit more checking done, checking that the component schema of the array is also valid.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
CuriousVini: No. Before we were calling `BigQueryUtil.getNonNullableSchema` before  `getTableDataType ` method was called. However now with array types, we are using component schema so calling `getTableDataType ` method recursively. Thats why we need to check here now.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
CuriousVini: structs are not part of this PR.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
CuriousVini: Since this is BQ source, we do not expect invalid arrays coming in from BQInputFormat.

albertshau: I thought the source lets the user modify the schema or use a macro?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
CuriousVini: Arrays of bytes and structs are allowed while arrays of arrays and arrays of maps are not allowed. Added maps as unsupported type

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
CuriousVini: Done, also Filed https://issues.cask.co/browse/CDAP-15345

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
```diff
@@ -0,0 +1,195 @@
...
+        String.format("Field '%s' is of value null, which is not a valid value for BigQuery type array.", name));
+    }
+
+    if (!(value instanceof Collection) && !value.getClass().isArray()) {
+      throw new IllegalArgumentException("The value should be of type collection or array. Got: " + value.getClass());
+    }
+
+    Schema componentSchema = BigQueryUtil.getNonNullableSchema(fieldSchema.getComponentSchema());
+    if (BigQueryUtil.UNSUPPORTED_ARRAY_TYPES.contains(componentSchema.getType())) {
+      throw new IllegalArgumentException(String.format("Field '%s' is of %s type within array, " +
```
albertshau: Field '%s' is an array of '%s', which is not a valid BigQuery type.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
```diff
@@ -0,0 +1,195 @@
...
+                                                       name, componentSchema));
+    }
+
+    writer.name(name);
+    writer.beginArray();
+    if (value instanceof Collection) {
+      for (Object element : (Collection) value) {
+        // big query does not allow null values in array items
+        if (element == null) {
+          throw new IllegalArgumentException("");
```
albertshau: should have an error message, something like:

Field '%s' contains null values in its array, which is not allowed by BigQuery.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        write(writer, null, element, componentSchema);
+      }
+    } else {
+      for (int i = 0; i < Array.getLength(value); i++) {
+        // big query does not allow null values in array items
```
albertshau, line 175: nulls are handled differently in the Collection than here.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -86,10 +86,31 @@ public void validate(@Nullable Schema inputSchema) {
...
+        if (type == Schema.Type.ARRAY) {
+          Schema componentSchema = fieldSchema.getComponentSchema();
+
+          if (field.getSchema().isNullable()) {
+            throw new IllegalArgumentException(String.format("Field '%s' is of type array. " +
+                                                               "BigQuery does not allow nullable arrays.", name));
+          }
+
+          if (componentSchema.isNullable()) {
+            throw new IllegalArgumentException(String.format("Field '%s' has nullable array items, " +
```
albertshau: Field '%s' is an array of nullable values, but BigQuery does not allow null values in arrays

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -86,10 +86,31 @@ public void validate(@Nullable Schema inputSchema) {
...
+                                                               "BigQuery does not allow nullable array items.", name));
+          }
+
+          if (BigQueryUtil.UNSUPPORTED_ARRAY_TYPES.contains(componentSchema.getType())) {
+            throw new IllegalArgumentException(String.format("Field '%s' is of %s type within array, " +
```
albertshau: Field '%s' is an array of '%s', which is not supported by BigQuery.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -180,11 +184,35 @@ public static void validateFieldSchemaMatches(Field bqField, Schema.Field field,
       return;
     }
 
-    if (TYPE_MAP.get(type) == null) {
+    if (TYPE_MAP.get(type) == null && !SUPPORTED_COMPLEX_TYPES.contains(type)) {
```
albertshau: seems simpler to have a SUPPORTED_TYPES set that contains all the supported types, rather than having multiple places that check for simple type or supported complex type. Unless they are distinguished for some reason?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -180,11 +184,35 @@ public static void validateFieldSchemaMatches(Field bqField, Schema.Field field,
...
       throw new IllegalArgumentException(String.format("Field '%s' is of unsupported type '%s'",
                                                        field.getName(), type));
     }
 
+    if (type == Schema.Type.ARRAY) {
+      String name = field.getName();
+      if (field.getSchema().isNullable()) {
+        throw new IllegalArgumentException(String.format("Field '%s' is of type array. " +
```
albertshau: similar comments about error messages. Is there no way to consolidate the similar logic?

src/main/java/io/cdap/plugin/gcp/common/Schemas.java
```diff
@@ -59,8 +59,18 @@ public static void validateFieldsMatch(Schema trueSchema, Schema providedSchema)
                         field.getName(), providedFieldType.name(), trueFieldType.name()));
       }
       if (isTrueFieldNullable && !isProvidedFieldNullable) {
+        throw new IllegalArgumentException(String.format("Field '%s' should be nullable.", field.getName()));
```
albertshau: I don't think we need to throw an exception here. This is if the field in BigQuery is nullable but the data we're writing is not nullable, which should be fine.

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
```diff
@@ -0,0 +1,66 @@
...
+      Schema.Field.of("double", Schema.of(Schema.Type.DOUBLE)),
+      Schema.Field.of("array", Schema.arrayOf(Schema.of(Schema.Type.STRING)))
+    );
+
+    StructuredRecord record = StructuredRecord.builder(schema)
+      .set("int", 1)
+      .set("double", 1.1d)
+      .set("array", ImmutableList.of("1", "2", "3")).build();
+
+    StringWriter strWriter = new StringWriter();
```
albertshau: should not be comparing strings, as it is difficult to read and maintain. For example, if more fields are added to the test. 

It looks like gson has a JsonTreeWriter that you can use to get JsonObjects and compare things more programmatically.

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
```diff
@@ -0,0 +1,66 @@
...
+      writer.beginObject();
+      for (Schema.Field recordField : Objects.requireNonNull(record.getSchema().getFields())) {
+        if (schema.getField(recordField.getName()) != null) {
+          BigQueryRecordToJson.write(writer, recordField.getName(), record.get(recordField.getName()),
+                                     recordField.getSchema());
+        }
+      }
+      writer.endObject();
+    } catch (IOException e) {
+      throw new RuntimeException("Exception while converting structured record to json.", e);
```
albertshau: don't need to run and throw, can just make the test throw exception

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
CuriousVini: We have a separate set of supported complex types because there is no corresponding `LegacySQLTypeName` type mapping for array type.

src/main/java/io/cdap/plugin/gcp/common/Schemas.java
CuriousVini: In that case, are assuming that users know their data and there are no null values present for that field?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
CuriousVini: Consolidated this part.

CuriousVini:
Updated test and resolved conflicts.

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
```diff
@@ -0,0 +1,83 @@
...
+          BigQueryRecordToJson.write(writer, recordField.getName(), record.get(recordField.getName()),
+                                     recordField.getSchema());
+        }
+      }
+      writer.endObject();
+    }
+    String actual = strWriter.toString();
+
+    ObjectMapper mapper = new ObjectMapper();
+    JsonNode node = mapper.readTree(actual);
```
albertshau: Instead of this writing to a string then reading it back with jackson, I think it's better to stick with gson classes and use JsonTreeWriter as the JsonWriter, and use it to get the result:

```
JsonTreeWriter writer = new JsonTreeWriter();
writer.beginObject();
...
writer.close();
JsonObject actual = writer.get().getAsJsonObject();
```

I haven't actually tried this, but seems like this would work based on a quick look at the code.

CuriousVini: changed implementation to use `JsonTreeWriter`

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
```diff
@@ -0,0 +1,77 @@
...
+      writer.beginObject();
+      for (Schema.Field recordField : Objects.requireNonNull(record.getSchema().getFields())) {
+        if (schema.getField(recordField.getName()) != null) {
+          BigQueryRecordToJson.write(writer, recordField.getName(), record.get(recordField.getName()),
+                                     recordField.getSchema());
+        }
+      }
+      writer.endObject();
+
+      JsonObject actual = ((JsonTreeWriter) writer).get().getAsJsonObject();
```
albertshau: it's a little cleaner to declare writer as a JsonTreeWriter so you don't have to cast here.

albertshau (approved):
one minor comment, otherwise lgtm. fix, squash, and merge

PR #98, opened 2019-05-06
[CDAP-15256] Support bigquery numeric type

CuriousVini:
Dependent PR: https://github.com/cdapio/cdap/pull/11325

image

image

BQ sink validation
image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
@@ -24,6 +24,7 @@ import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import org.apache.avro.generic.GenericRecord;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.util.concurrent.TimeUnit;
@@ -76,6 +77,13 @@ public class BigQueryAvroToStructuredTransformer extends RecordConverter<Generic
           case TIMESTAMP_MILLIS:
           case TIMESTAMP_MICROS:
             return field;
+          case DECIMAL:
+            ByteBuffer value = (ByteBuffer) field;
+            byte[] bytes = new byte[value.remaining()];
+            int pos = value.position();
+            value.get(bytes);
+            value.position(pos);
+            return bytes;
           default:
             throw new UnexpectedFormatException("Field type " + fieldType + " is not supported.");
         }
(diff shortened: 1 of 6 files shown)


CuriousVini:
Resolving conflicts in `AbstractBigQuerySink`

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -76,6 +77,11 @@ protected Object convertField(Object field, Schema fieldSchema) throws IOExcepti
           case TIMESTAMP_MILLIS:
           case TIMESTAMP_MICROS:
             return field;
+          case DECIMAL:
+            ByteBuffer value = (ByteBuffer) field;
+            byte[] bytes = new byte[value.remaining()];
+            value.get(bytes);
```
albertshau, line 84: need to reset the position in the ByteBuffer, otherwise something else that uses it will get messed up:

```
    int pos = value.position();
    value.get(bytes);
    value.position(pos);
```

albertshau (commented):
how does the BigQueryOutputFormat behave if given numeric values that BigQuery doesn't support, such as precision > 38, scale > 9?

CuriousVini:
Tested with higher precision and scale for BigDecimal by writing a mock transform plugin to emit higher precision/scale records. The big query job fails with `Invalid NUMERIC value: `. Will add validation for that in the bq sink.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -183,6 +183,17 @@ public static void validateFieldSchemaMatches(Field bqField, Schema.Field field,
                         field.getName(), logicalType, bqField.getName(), dataset, table,
                         bqField.getType(), bqField.getType()));
       }
+
+      // BigQuery requires schema precision to be 38 and scale 9
+      if (logicalType == Schema.LogicalType.DECIMAL) {
+        if (fieldSchema.getPrecision() != 38 || fieldSchema.getScale() != 9) {
```
albertshau: so it doesn't allow lower precision and lower scale? If it's lower the plugin can adjust it up can't it?

albertshau: To start out with, I think the plugin should support whatever the outputformat supports.

If it has to be exactly 38 and 9, then let's keep this check but change the error message to: 
Numeric field '%s' has invalid precision '%d' and scale '%d'. Precision must be 38 and scale must be 9.

If it allows smaller values, then this should be changed to > checks and the error message should be:
Numeric field '%s' has invalid precision '%d' and scale '%d'. Precision must be at most 38 scale must be at most 9.

albertshau (approved):
one remaining comment. Depending on what you find, fix then merge

CuriousVini:
Big query allows smaller precision and scale. So we should allow that in the plugins as well.

CuriousVini:
squashing and merging.

PR #100, opened 2019-05-15
Update Subscriber source's configurations to remove ambiguity.

anwar6953:
Update Subscriber source's configurations to remove ambiguity. Otherwise, a user may think that "subscription does not exist" means that the "subscription field is empty/unspecified".

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
@@ -118,15 +118,15 @@ public class GoogleSubscriber extends StreamingSource<StructuredRecord> {
    */
   public static class SubscriberConfig extends GCPReferenceSourceConfig {
 
-    @Description("Cloud Pub/Sub subscription to read from. If the subscription does not exist it will be " +
-      "automatically created if a topic is specified. Messages published before the subscription was created will " +
-      "not be read.")
+    @Description("Cloud Pub/Sub subscription to read from. If a subscription with the specified name does not " +
+      "exist, it will be automatically created if a topic is specified. Messages published before the subscription " +
+      "was created will not be read.")
     @Macro
     private String subscription;
 
-    @Description("Cloud Pub/Sub topic to subscribe messages from. If a topic is provided and the given subscription " +
-      "does not exists it will be created. Note: If a subscription does not exists and is created only the messages " +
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -118,9 +118,9 @@ private ZonedDateTime getTimestamp(String publishTime) {
    */
   public static class SubscriberConfig extends GCPReferenceSourceConfig {
 
-    @Description("Cloud Pub/Sub subscription to read from. If the subscription does not exist it will be " +
-      "automatically created if a topic is specified. Messages published before the subscription was created will " +
-      "not be read.")
+    @Description("Cloud Pub/Sub subscription to read from. If a subscription with the configured name does not " +
+      "exist it will be automatically created if a topic is specified. Messages published before the subscription " +
```
albertshau: We use 'configured' and 'specified' here, as well as 'provided' and 'given' in the topic description. Let's use a single word. My feeling is that 'specified' is most common in our descriptions.

Let's also fix the grammar and consistency of the topic description. How about we have the following for subscription:

Cloud Pub/Sub subscription to read from. If the specified subscription does not exist, it will automatically be created. Messages published before the subscription was created will not be read.

and for the topic:

Cloud Pub/Sub topic to subscribe to. If the specified topic does not exist, it will automatically be created. (remove the note, as that belongs in the subscription description)

anwar6953: I think the description for topic you suggested implies that the topic would be created, instead of a subscription to that topic.

PTAL

albertshau: OK this thing is super confusing. Was wondering why the topic was nullable, and it's because the topic is only used when the subscription needs to be auto-created and is otherwise never used. How about:

Cloud Pub/Sub topic to create a subscription on. This is only used when the specified subscription does not already exist and needs to be automatically created. If the specified subscription already exists, this value is ignored.

anwar6953: sgtm

albertshau (approved):
lgtm

anwar6953:
Squashed commits and merging.

PR #101, opened 2019-05-15
[CDAP-15386] Fix issue when Pub/Sub event does not have attributes on…

nitinmotgi:
… event


anwar6953:
Change the base branch to release/0.12.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -93,7 +93,9 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
     return pubSubMessages.map(pubSubMessage -> {
       // Convert to a HashMap because com.google.api.client.util.ArrayMap is not serializable.
       HashMap<String, String> hashMap = new HashMap<>();
-      pubSubMessage.getAttributes().forEach((k, v) -> hashMap.put(k, v));
+      if (pubSubMessage.getAttributes() != null) {
+        pubSubMessage.getAttributes().forEach((k, v) -> hashMap.put(k, v));
```
chtyim: No need to be this fancy. Just do

```java
hashMap.putAll(pubSubMessage.getAttributes());
```

PR #111, opened 2019-05-25
[CDAP-15438] Added BigQuery Create and Delete bucket actions

nitinmotgi:
. Also, standardized documentation and JSON for all Google plugins.

pr


docs/BigQueryMultiTable-batchsink.md
```diff
@@ -1,66 +1,77 @@
...
+    * the pipeline schema does not contain fields that exist in the BigQuery schema.
+
+  In this case, those fields in the BigQuery schema will be modified to become nullable. Incompatible schema
+  changes will result in pipeline failure.
+
+#### **Project and Credentials**
+
+If the plugin is run in GCP environment, the service account file path does not need to be
+specified and can be set to 'auto-detect'. Credentials will be automatically read from the GCP environment.
+A path to a service account key must be provided when not running in GCP. The service account
```
albertshau, line 59: Why change the description about when not running on GCP?

docs/BigQueryMultiTable-batchsink.md
```diff
@@ -1,66 +1,77 @@
...
+key can be found on the Dashboard in the Cloud Platform Console. Ensure that the account key has permission
+to access resource.
+
+* **Project Id**
```
albertshau, line 63: How come these are in a separate section now? They are still properties that somebody needs to fill out when configuring the plugin.

albertshau: Also, why group this together with credentials? The project is most often used to identify whatever is being read from/written to. To me it makes more sense to be grouped together with dataset and table name or their equivalents.

widgets/GCSMove-action.json
```diff
@@ -2,19 +2,11 @@
   "metadata": {
     "spec-version": "1.5"
   },
-  "display-name" : "Google Cloud Storage Move",
+  "display-name" : "GCS Move",
```
albertshau, line 5: From what I understand, these names were chosen because these are the actual product names. Is that not wanted?

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+import io.cdap.plugin.gcp.common.GCPConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * This class <code>BigQueryCreateDataset</code> drops a BigQuery dataset.
```
albertshau, line 40: fix the javadoc.

pom.xml
```diff
@@ -58,7 +58,7 @@
     <jee.version>7</jee.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <avro.version>1.7.7</avro.version>
-    <cdap.version>6.1.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.0.1-SNAPSHOT</cdap.version>
```
albertshau: not needed

pom.xml
```diff
@@ -19,7 +19,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.12.2</version>
+  <version>0.12.3-SNAPSHOT</version>
```
albertshau: develop is supposed to be on 0.13.0-SNAPSHOT, with release/0.12 on 0.12.3 already.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+      );
+    }
+  }
+
+  /**
+   * This class <code>Config</code> provides the configuration for deleting a dataset from a project.
+   */
+  private static final class Config extends GCPConfig {
+    @Name("dataset")
+    @Description("Dataset to be drop from the project")
```
albertshau, line 110: Dataset to create

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+      throw new Exception(
+        String.format(
+          "Failed to create dataset '%s'. Reason : %s", config.getDatasetName(), e.getMessage()
+        )
+      );
+    }
+  }
+
+  /**
+   * This class <code>Config</code> provides the configuration for deleting a dataset from a project.
```
albertshau, line 106: update

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+   */
+  private static final class Config extends GCPConfig {
+    @Name("dataset")
+    @Description("Dataset to be drop from the project")
+    @Macro
+    private String dataset;
+
+    @Name("description")
+    @Description("Provide a description for dataset")
```
albertshau, line 115: Dataset description

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
+
+#### **Description**
+
+Create a BigQuery Dataset. Use this plugin to create a BigQuery dataset.
```
albertshau, line 4: We have been using complete sentences everywhere in the documentation. The only exception is the first part of a property description, which is a fragment.

Also, these sentences are redundant. Should just be "This plugin creates a BigQuery dataset"

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+
+#### **Usage**
```
albertshau, line 6: Why separate sections for description and usage? Shouldn't the description describe what the plugin is, how to use it, and relevant details about how it works?

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+
+User can configure the pipeline to fail if BigQuery dataset already exists.
```
albertshau, line 8: We shouldn't refer to the user as 'User' anywhere in the documentation. They know they are a user, they are reading the doc.

albertshau: The plugin can be configured to fail the pipeline if the dataset already exists.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+To create a dataset in another project provide the target project id and make
```
albertshau, line 9: 'Another' project assumes this plugin is being run on GCP. This also gives the impression that this will always work if running on GCP and the dataset is in the same project. In actuality, everything is tied to the service account. It is more accurate to say something like:

The service account must have permission to create BigQuery datasets.

docs/BigQueryCreateDataset-action.md
albertshau: What happens if the dataset already exists but with a different description and labels? Does it get updated?

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+sure right permission are provided.
+
+#### **Properties**
+
+Following are properties used to configure this plugin
```
albertshau, line 14: This looks like filler text. IMO it doesn't convey any useful information and should be removed.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+
+* **Name**
+
+  Name of dataset to be created. May contain up to 1,024 characters and
```
albertshau, line 18: Please use complete sentences after the first fragment and use active voice when possible.

Name of the dataset to create. The name may contain up to 1,024 characters and can only contain letters, numbers, and underscores.

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,33 @@
...
+
+Drop a BigQuery dataset.
+
+#### **Properties**
+
+Following are properties used to configure this plugin
+
+* **Dataset**
+
+  Name of dataset to be dropped. Name may contain up to 1,024
```
albertshau, line 12: Name of the dataset to drop.

I'm not sure this needs the description about valid characters, as the dataset should already exist, which means it's already valid.

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,33 @@
+
+#### **Description**
+
+Drop a BigQuery dataset.
```
albertshau, line 4: Drops a BigQuery dataset. All tables in the dataset are dropped.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+   */
+  @Override
+  public void run(ActionContext context) throws Exception {
+    BigQuery bq = BigQueryUtil.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
+    try {
+      Dataset dataset = bq.getDataset(config.getDatasetId());
+      if (dataset == null) {
+        dataset = bq.create(config.getDatasetInfo());
+        try {
+          Dataset.Builder builder = dataset.toBuilder();
```
albertshau, line 69: There are several layers of nesting that can be removed. This section for building the Dataset should happen at the beginning, outside the try-catch. 

Does this throw exceptions if invalid values are set? If so, it should also be built in the configurePipeline() method, assuming none of the required fields contain macros. That way if anything is invalid, pipeline deployment will fail.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+    @Macro
+    private String dataset;
+
+    @Name("description")
+    @Description("Provide a description for dataset")
+    @Nullable
+    @Macro
+    private String description;
+
+    @Name("fail-if-exists")
```
albertshau, line 120: why use dashes? These names are invisible to UI users and nothing else uses dashes instead of camel case.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+    @Description("Fails pipeline if dataset already exists or skip silently")
+    @Macro
+    private String failIfExists;
```
albertshau, line 123: this should be a Boolean, not a string.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+  can contain letters (upper or lower case), numbers, and underscores
+
+* **Description**
+
+  Description to be associated with dataset.
+
+* **Labels**
+
+  GCP labels to be associated with the dataset.
```
albertshau, line 27: Labels have restrictions and should be mentioned. It's also redundant to say 'GCP' labels, it should just be 'Labels'.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+
+* **Fail if exists**
+
+  Configuration specifying if pipeline should fail when dataset already exists.
```
albertshau, line 31: Whether the pipeline should fail if the dataset already exists.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+
+* **Description**
+
+  Description to be associated with dataset.
+
+* **Labels**
+
+  GCP labels to be associated with the dataset.
+
+* **Fail if exists**
```
albertshau, line 29: Convention has been to capitalize these in the docs and widget. 'Fail If Exists'

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,51 @@
...
+Following are properties used to configure this plugin
+
+* **Name**
+
+  Name of dataset to be created. May contain up to 1,024 characters and
+  can contain letters (upper or lower case), numbers, and underscores
+
+* **Description**
+
+  Description to be associated with dataset.
```
albertshau, line 23: with the dataset

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+     * @return a <code>Map</code> specifing the key value of labels.
+     */
+    public Map<String, String> getLabels() {
+      Map<String, String> labelMap = new HashMap<>();
+      if (labels == null || labels.isEmpty()) {
+        return labelMap;
+      }
+      String[] kvPairs = labels.split(",");
+      for (String kvPair : kvPairs) {
+        String[] kv = kvPair.split(":");
```
albertshau, line 178: are these characters allowed in labels? If so this logic will fail.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    public DatasetId getDatasetId() {
+      return DatasetId.of(getProject(), dataset);
+    }
+
+    /**
+     * @return a instance of <code>DatasetInfo</code> based on project and dataset name provided by user.
+     */
+    public DatasetInfo getDatasetInfo() {
+      return DatasetInfo.newBuilder(dataset).build();
```
albertshau, line 141: no description or labels set?

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+          Dataset updateDataset = builder.build();
+          updateDataset.update();
+        } catch (BigQueryException e) {
+          // we don't fail if we cannot update dataset.
+          LOG.warn(
+            String.format("Dataset '%s' was created, but failed to update metadata for dataset.",
+                          config.getDatasetName())
+          );
+        }
+        context.getMetrics().count(RECORDS_OUT, 1);
```
albertshau, line 86: this doesn't do anything and should be removed

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+      } else {
+        if (config.failIfExists()) {
+          throw new Exception(
+            String.format(
+              "Dataset '%s' already exists. Failing pipeline as per user request.", config.getDatasetName()
```
albertshau, line 91: include the project in the error message, something like:

Failing the pipeline because dataset '%s' in project '%s' already exists.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+            )
+          );
+        }
+      }
+    } catch (BigQueryException e) {
+      throw new Exception(
+        String.format(
+          "Failed to create dataset '%s'. Reason : %s", config.getDatasetName(), e.getMessage()
```
albertshau, line 99: include the project in the message as well

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+        if (config.failIfExists()) {
+          throw new Exception(
+            String.format(
+              "Dataset '%s' already exists. Failing pipeline as per user request.", config.getDatasetName()
+            )
+          );
+        }
+      }
+    } catch (BigQueryException e) {
+      throw new Exception(
```
albertshau, line 97: Can also get here if the request failed because the dataset already exists. Should be able to tell based on the BigQueryException. If it failed because it already exists and failIfExists is false, it should not throw an exception.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+  public static final String NAME = "BigQueryCreateDataset";
+  public static final String DESC = "Create a Cloud BigQuery Dataset";
+  private static final String RECORDS_OUT = "records.out";
+  private Config config;
+
+  /**
+   * Creates a dataset.
+   *
+   * @param context a <code>ActionContext</code> instance during runtime.
+   * @throws Exception thrown when delete of dataset fails.
```
albertshau, line 59: fix javadoc

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+ *  <code>BigQueryCreateDataset</code> can be used to drop BQ dataset in other projects.
+ * </p>
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryCreateDataset.NAME)
+@Description(BigQueryCreateDataset.DESC)
+public class BigQueryCreateDataset extends Action {
+  private static final Logger LOG = LoggerFactory.getLogger(BigQueryCreateDataset.class);
+  public static final String NAME = "BigQueryCreateDataset";
+  public static final String DESC = "Create a Cloud BigQuery Dataset";
```
albertshau, line 51: description won't be referenced from anywhere else and should just be hardcoded in the description annotation. This is different than name, which might be used in a unit test.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+ * <p>
+ *  <code>BigQueryDropDataset</code> can be used to drop BQ dataset in other projects.
+ * </p>
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryDropDataset.NAME)
+@Description(BigQueryDropDataset.DESC)
+public class BigQueryDropDataset extends Action {
+  public static final String NAME = "BigQueryDropDataset";
+  public static final String DESC = "Drops a Cloud BigQuery Dataset";
```
albertshau, line 44: description should be hardcoded in the description annotation

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+   * @throws Exception thrown when delete of dataset fails.
+   */
+  @Override
+  public void run(ActionContext context) throws Exception {
+    BigQuery bq = BigQueryUtil.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
+    try {
+      Boolean deleted = bq.delete(config.getDatasetId());
+      if (deleted) {
+        // the dataset was deleted
+        context.getMetrics().gauge(RECORDS_OUT, 1);
```
albertshau, line 61: this doesn't do anything and should be removed

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+  private static final class Config extends GCPConfig {
+    @Name("dataset")
+    @Description("Dataset to be drop from the project")
+    @Macro
+    private String dataset;
+
+    @Name("not-found")
+    @Description("Choice to fail job if BigQuery dataset is not found.")
+    @Macro
+    private String failOnNotFound;
```
albertshau, line 94: should be a Boolean

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+    }
+  }
+
+  /**
+   * This class <code>Config</code> provides the configuration for deleting a dataset from a project.
+   */
+  private static final class Config extends GCPConfig {
+    @Name("dataset")
+    @Description("Dataset to be drop from the project")
```
albertshau, line 87: Dataset to drop from the project

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+    @Macro
+    private String dataset;
+
+    @Name("not-found")
+    @Description("Choice to fail job if BigQuery dataset is not found.")
```
albertshau, line 92: Whether the pipeline should fail if the dataset is not found.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+        if (config.failIfNotFound()) {
+          throw new NotFoundException(
+            String.format("Failed to delete dataset '%s'. Failing pipeline as configured by user.",
+                          config.getDatasetName())
+          );
+        }
+      }
+    } catch (BigQueryException e) {
+      // the dataset delete had an issue.
+      if (config.failIfNotFound()) {
```
albertshau, line 73: We don't know that the failure is because the dataset does not exist. It can be permissions, the BQ API is down, etc.

If this is the desired behavior, the property name and documentation should be changed.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+   * Drops a dataset.
+   *
+   * @param context a <code>ActionContext</code> instance during runtime.
+   * @throws Exception thrown when delete of dataset fails.
+   */
+  @Override
+  public void run(ActionContext context) throws Exception {
+    BigQuery bq = BigQueryUtil.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
+    try {
+      Boolean deleted = bq.delete(config.getDatasetId());
```
albertshau, line 58: Boolean -> boolean

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+      if (deleted) {
+        // the dataset was deleted
+        context.getMetrics().gauge(RECORDS_OUT, 1);
+      } else {
+        // the dataset set was not found.
+        if (config.failIfNotFound()) {
+          throw new NotFoundException(
+            String.format("Failed to delete dataset '%s'. Failing pipeline as configured by user.",
```
albertshau, line 66: include the project in the message

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+          if (config.getLabels().size() > 0) {
+            builder.setLabels(config.getLabels());
+          }
+          // update dataset.
+          Dataset updateDataset = builder.build();
+          updateDataset.update();
+        } catch (BigQueryException e) {
+          // we don't fail if we cannot update dataset.
+          LOG.warn(
+            String.format("Dataset '%s' was created, but failed to update metadata for dataset.",
```
albertshau, line 82: Unable to update the metadata for dataset '%s' in project '%s'

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+   * @param context a <code>ActionContext</code> instance during runtime.
+   * @throws Exception thrown when delete of dataset fails.
+   */
+  @Override
+  public void run(ActionContext context) throws Exception {
+    BigQuery bq = BigQueryUtil.getBigQuery(config.getServiceAccountFilePath(), config.getProject());
+    try {
+      Dataset dataset = bq.getDataset(config.getDatasetId());
+      if (dataset == null) {
+        dataset = bq.create(config.getDatasetInfo());
```
albertshau, line 67: You should be able to create the dataset with labels and description already set.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+          Dataset.Builder builder = dataset.toBuilder();
+          if (config.getDescription() != null) {
+            builder.setDescription(config.getDescription());
+          }
+          if (config.getLabels().size() > 0) {
+            builder.setLabels(config.getLabels());
+          }
+          // update dataset.
+          Dataset updateDataset = builder.build();
+          updateDataset.update();
```
albertshau, line 78: update should only be attempted if the dataset already exists and its description or labels are different.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+        } catch (BigQueryException e) {
+          // we don't fail if we cannot update dataset.
+          LOG.warn(
+            String.format("Dataset '%s' was created, but failed to update metadata for dataset.",
+                          config.getDatasetName())
+          );
```
albertshau, line 84: The exception should always be included in the log, otherwise the root cause is lost.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,185 @@
...
+              "Dataset '%s' already exists. Failing pipeline as per user request.", config.getDatasetName()
+            )
+          );
+        }
+      }
+    } catch (BigQueryException e) {
+      throw new Exception(
+        String.format(
+          "Failed to create dataset '%s'. Reason : %s", config.getDatasetName(), e.getMessage()
+        )
```
albertshau, line 100: the exception should always be included as the cause, otherwise the root cause is lost.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryDropDataset.java
```diff
@@ -0,0 +1,121 @@
...
+                          config.getDatasetName())
+          );
+        }
+      }
+    } catch (BigQueryException e) {
+      // the dataset delete had an issue.
+      if (config.failIfNotFound()) {
+        throw new NotFoundException(
+          String.format("User configured pipeline to fail if dataset '%s' was not found. Reason : %s",
+                        config.getDatasetName(), e.getMessage())
```
albertshau, line 76: the exception should always be included as the cause, otherwise the root cause is lost.

widgets/BigQueryCreateDataset-action.json
```diff
@@ -0,0 +1,98 @@
...
+          "label": "Service Account File Path",
+          "name": "serviceFilePath",
+          "widget-attributes" : {
+            "default": "auto-detect"
+          }
+        }
+      ]
+    }
+  ],
+  "outputs": [
```
albertshau, line 73: this should not be here

albertshau (commented):
Please keep PRs isolated to a single change. If you would like to change the structure of the documentation and property grouping, that should be in a different PR.

docs/BigQueryMultiTable-batchsink.md
nitinmotgi: Why not ? I have seen customers when they confused when it's along with other configuration. Even though they are properties that need to be filled grouping them is easier and in future we will turn them into tabs.

docs/BigQueryMultiTable-batchsink.md
nitinmotgi: What is wrong with this ?

widgets/GCSMove-action.json
nitinmotgi: Customers don't know long form of GCS, so they don't relate that to GCS. They know GCS but not Google Cloud Storage or Cloud Storage.

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
nitinmotgi: Will do

pom.xml
nitinmotgi: Will do

pom.xml
nitinmotgi: done

pom.xml
nitinmotgi: done

docs/BigQueryCreateDataset-action.md
nitinmotgi: Description is short and conveys what the plugin will do. Usage is for things that talk about behavior. We shouldn't mix them together IMHO.

docs/BigQueryCreateDataset-action.md
nitinmotgi: +1

docs/BigQueryCreateDataset-action.md
nitinmotgi: +1

docs/BigQueryDropDataset-action.md
nitinmotgi: fair

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
nitinmotgi: Not sure your assertion is true.

docs/BigQueryMultiTable-batchsink.md
albertshau: Because the project is used to identify *what* you are reading from and not *who* is reading. It's not related to credentials or auth at all. I can use a service account from a different project. Why should they be together?

docs/BigQueryMultiTable-batchsink.md
albertshau: There was information removed about how you need to set things up when not running on dataproc (file needs to be on every node in the cluster, etc). Was that intentional?

docs/BigQueryCreateDataset-action.md
albertshau: It just seems like descriptions become useless at that point. They basically just say writes to XYZ, reads from ABC

PR #115, opened 2019-06-08
[CDAP-15438] Working version of Create and Delete dataset

nitinmotgi:
Separated into different PR. Closing PR: https://github.com/data-integrations/google-cloud/pull/111 and open other separate PRs.


docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+Creates a BigQuery Dataset.
+
+#### **Usage**
+
+Dataset names must be unique per project. The dataset name contain up to 1,024 characters and can
+contain letters (upper or lower case), numbers and underscore. Dataset names cannot contain special characters
+such as '-', '&', '*', or '%'.
+
+If dataset name starts with '_' (underscore) then the dataset would be hidden and not visible in
```
albertshau, line 13: This paragraph is showing up italicized in this PR for some reason. Are they actually italics?

albertshau: What does 'in the view' mean? Does this mean it's invisible in the cloud console? Is this only when listing datasets in a project or are you unable to directly look it up too? Assuming it's completely invisible:

If the dataset name start with an underscore, it will not be visible in the Google Cloud console. Dataset names are case sensitive. For example, 'dataset' and 'Dataset' are two different datasets that can co-exist in a project.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+#### **Description**
+
+Creates a BigQuery Dataset.
+
+#### **Usage**
+
+Dataset names must be unique per project. The dataset name contain up to 1,024 characters and can
+contain letters (upper or lower case), numbers and underscore. Dataset names cannot contain special characters
+such as '-', '&', '*', or '%'.
```
albertshau, line 11: part of this sentence looks italicized (', or '%'). Is this just a PR formatting thing (maybe because of the *)?

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+the view and dataset names are case sensitive (E.g. 'dataset' and 'Dataset' are two different
+datasets that can co-exists in a project.
+
+#### **Basic**
+
+Following are properties used to configure this plugin
+
+* **Name**
+
+  Name of dataset to be created. May contain up to 1,024 characters and
```
albertshau, line 23: By convention, everything except the first fragment should be a complete sentence, with full punctuation, like a period at the end. 

Name of the dataset to create. The name may contain up to 1,024 characters and must consists of letters (upper or lower case), numbers, and underscores.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+  Description to be associated with dataset.
+
+* **Labels**
+
+  GCP labels to be associated with the dataset.
+
+* **Fail if exists**
+
+  Configuration specifying if pipeline should fail when dataset already exists.
```
albertshau, line 36: Mentioning that it's configuration is redundant:

Whether the pipeline should fail if the dataset already exists.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+#### **Advanced**
+
+If the plugin is run in GCP environment, the service account file path does not need to be
```
albertshau, line 40: This looks like it should be in the description for Service Account File Path instead of for the entire section.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+* **Project Id**
+
+  Google Cloud Project Id, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console. More information
+can be found [here](https://cloud.google.com/resource-manager/docs/creating-managing-projects#identifying_projects)
+
+* **Service Account File Path**
+
+  Path on the local file system of the service account key used for
+authorization. Can be set to 'auto-detect' when running in GCP. When running on outside GCP,
```
albertshau, line 55: What does 'auto-detect' mean when running in GCP? I think this description should be merged with the paragraph above to be something like:

Path on the local file system of the service account key to be used for authorization. The service account must have permission to create a BigQuery dataset. If the pipeline is being run on a Dataproc cluster, this can be set to 'auto-detect', which will use the credentials of the cluster. If the pipeline is not run on a Dataproc cluster, the file must be present on every node in the cluster.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+If the plugin is run in GCP environment, the service account file path does not need to be
+specified and can be set to 'auto-detect'. Credentials will be automatically read from the GCP environment.
+A path to a service account key must be provided when not running in GCP. The service account
+key can be found on the Dashboard in the Cloud Platform Console. Ensure that the account key has permission
+to access resource.
+
+* **Project Id**
+
+  Google Cloud Project Id, which uniquely identifies a project.
```
albertshau, line 48: I don't like our existing project descriptions for these plugins. What is this the project of? My data fusion instance? The VM that is running CDAP? My dataproc cluster? I think it's more descriptive to say something like:

Project to create the BigQuery dataset in. If the pipeline is being run on a Dataproc cluster, this can be set to 'auto-detect', which will resolve to the project that the Dataproc cluster is in.

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+contain letters (upper or lower case), numbers and underscore. Dataset names cannot contain special characters
+such as '-', '&', '*', or '%'.
+
+If dataset name starts with '_' (underscore) then the dataset would be hidden and not visible in
+the view and dataset names are case sensitive (E.g. 'dataset' and 'Dataset' are two different
+datasets that can co-exists in a project.
+
+#### **Basic**
+
+Following are properties used to configure this plugin
```
albertshau, line 19: Does this sentence fragment really convey anything useful? IMO it doesn't and can be removed. If we really want to keep it, we would want it to be exactly the same in every reference doc and it should be:

The following properties can be used to configure this plugin:

docs/BigQueryCreateDataset-action.md
```diff
@@ -0,0 +1,56 @@
...
+#### **Usage**
+
+Dataset names must be unique per project. The dataset name contain up to 1,024 characters and can
+contain letters (upper or lower case), numbers and underscore. Dataset names cannot contain special characters
+such as '-', '&', '*', or '%'.
+
+If dataset name starts with '_' (underscore) then the dataset would be hidden and not visible in
+the view and dataset names are case sensitive (E.g. 'dataset' and 'Dataset' are two different
+datasets that can co-exists in a project.
+
```
albertshau, line 16: As mentioned in the original PR, there should be a description about what happens if the dataset already exists. Do the description and labels get updated? Do they get ignored for whatever currently exists?

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,34 @@
...
+
+Drop a BigQuery dataset.
+
+#### **Basic**
+
+Following are properties used to configure this plugin
+
+* **Dataset**
+
+  Name of dataset to be dropped. Name may contain up to 1,024
```
albertshau, line 13: Name of the dataset to drop. The name may contain up to 1,024 characters and must consist of letters (upper or lower case), numbers, or underscores.

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,34 @@
...
+characters and can contain letters (upper or lower case), numbers, and underscores
+
+#### **Advanced**
+
+If the plugin is run in GCP environment, the service account file path does not need to be
```
albertshau, line 18: similar questions/comments about this section

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,34 @@
+# BigQuery Drop Dataset
+
+#### **Description**
+
+Drop a BigQuery dataset.
+
+#### **Basic**
+
+Following are properties used to configure this plugin
```
albertshau, line 9: remove or change to:

The following properties can be used to configure this plugin:

docs/BigQueryDropDataset-action.md
```diff
@@ -0,0 +1,34 @@
+# BigQuery Drop Dataset
+
+#### **Description**
+
+Drop a BigQuery dataset.
+
```
albertshau, line 6: No Usage section for this one?

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryCreateDataset.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * This class <code>BigQueryCreateDataset</code> drops a BigQuery dataset.
+ * <p>
+ *  <code>BigQueryCreateDataset</code> can be used to drop BQ dataset in other projects.
+ * </p>
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(BigQueryCreateDataset.NAME)
+@Description(BigQueryCreateDataset.DESC)
```
albertshau, line 48: there's no need for the description to be a variable since it will never be used by anything else. It can be inline here.

albertshau (commented):
I had left a bunch of comments about the code on the previous PR that don't seem to have been addressed. Please see those.

PR #117, opened 2019-06-17
CDAP-15535 Avoid using database and '.' in the table name in BigQuery.

anwar6953:
In https://github.com/data-integrations/multi-table-plugins/pull/7, the table name variable was modified to include the database name as well, resulting in an error when using oracle db as the db in MultiTableDatabase.
This removes that prefix, in order to fix that error.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
@@ -81,6 +81,11 @@ public class BigQueryMultiSink extends AbstractBigQuerySink {
         continue;
       }
       String tableName = key.substring(TABLE_PREFIX.length());
+      // remove the database prefix, as BigQuery doesn't allow dots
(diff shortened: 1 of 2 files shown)


albertshau (commented):
Does this mean the source can output table data from multiple databases? If so then there can be multiple tables with the same name?

anwar6953:
In one pipeline run, the source will only operate within a single database. Would that resolve the concern?
(I imagine that the dataset field of the sink is analogous to the database field of the source.)

albertshau:
Yes, though I'm curious why it attaches the db name then (perhaps it wasn't intentional?)

anwar6953:
Looking at the PR where the change was introduced (https://github.com/data-integrations/multi-table-plugins/pull/7/files), it seems like it was done to combine the two fields (db and tablename) into a single field for code simplicity, rather than the intention being to namespace the tables in the sink with the database name.

albertshau (approved):

albertshau (approved):
lgtm

PR #119, opened 2019-06-20
Updates GCS, Spanner and BigQuery source plugin widget jsons to have …

ajainarayanan:
…browse widget associated with projectId

### DIFF
--- a/widgets/BigQueryTable-batchsource.json
+++ b/widgets/BigQueryTable-batchsource.json
@@ -21,6 +21,13 @@
           "name": "project",
           "widget-attributes" : {
             "default": "auto-detect"
+          },
+          "plugin-function": {
+            "label": "Browse",
+            "widget": "connection-browser",
+            "widget-attributes": {
+              "connectionType": "BIGQUERY"
+            }
           }
         },
         {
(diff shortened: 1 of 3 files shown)


widgets/BigQueryTable-batchsource.json
```diff
@@ -21,6 +21,13 @@
           "name": "project",
           "widget-attributes" : {
             "default": "auto-detect"
+          },
+          "plugin-function": {
```
albertshau, line 25: does this get ignored by earlier versions of the UI?

albertshau (approved):
lgtm assuming it doesn't break earlier versions of the UI

widgets/BigQueryTable-batchsource.json
ajainarayanan: The plugin function as a concept got removed in the backend. I don't believe we changed the spec to reflect that. In plugins like [Database source](https://github.com/cdapio/hydrator-plugins/blob/develop/database-plugins/widgets/Database-batchsource.json#L47) it still exists.

ajainarayanan:
@albertshau Good point. Right now it does throw an exception but this will not be visible to the user nor will it be blocking the user flow. It should be fine if the plugin is used in earlier versions of CDAP (6.0). Have filed a [JIRA-CDAP-15592](https://issues.cask.co/browse/CDAP-15592)  anyways to fix this in 6.0 and 6.1

PR #120, opened 2019-06-21
[CDAP-15568] fix bigquery multi sink

CuriousVini:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -120,10 +118,9 @@ public void transform(StructuredRecord input, Emitter<KeyValue<Text, NullWritabl
         }
       }
       writer.endObject();
+      emitter.emit(new KeyValue<>(writer.get().getAsJsonObject(), NullWritable.get()));
```
albertshau, line 121: Just curious, how come the sink works regardless of whether you send a JsonObject or Text? Is the underlying output format handling both?

CuriousVini: yes, underlying TextOutputFormat handles both `JsonObject` and `Text` objects.
image

albertshau (approved):
lgtm

PR #122, opened 2019-06-26

mlozbin-cybervisiontech:
WIKI: https://wiki.cask.co/display/CE/Bigtable+Source

In scope of this PR:
- implemented Bigtable Source Plugin


mlozbin-cybervisiontech:

docs/Bigtable-batchsource.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Table**: Database table name.
+
+**Key Alias**: Name of the field for row key.
+
+**Scan Row Start**: Scan start row.
+
+**Scan Row Stop**: Scan stop row.
+
+**Scan Time Range Start**: The starting timestamp used to filter columns.
```
albertshau: The starting -> Starting.
Is it inclusive?

docs/Bigtable-batchsource.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Scan Time Range Stop**: The ending timestamp used to filter columns.
```
albertshau: The ending -> ending.
Is it exclusive?

docs/Bigtable-batchsource.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Service Account File Path**: Path on the local file system of the service account key used for
+authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
+When running on other clusters, the file must be present on every node in the cluster.
+
+**Schema**: Specifies the schema that has to be output. 
+Only columns defined in schema will be included into output record. Field name should be in form 
+"<family>__<column>".
```
albertshau, line 48: I checked with UI and the restriction on ':' is due to an avro schema library. Avro doesn't allow ':' in field names (https://avro.apache.org/docs/1.8.2/spec.html#names). I don't think that restriction is going away any time soon.

I don't imagine most people have double underscores in the family so this is probably ok for most use cases, though ugly. This also means the plugin cannot handle other characters in the family or qualifier either though...

Perhaps we could support a 'Column Mapping' advanced property, where a user can specify that 'family:qualifier' maps to output field 'xyz'. If the user has unsupported characters in their qualifier or family they can use that mapping property. This can be done in a follow up PR to make it easier to review.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
```
albertshau: update year

pom.xml
```diff
@@ -155,6 +159,27 @@
...
+          <artifactId>hbase-server</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-shaded-server</artifactId>
+      <version>${hbase-shaded-server.version}</version>
+    </dependency>
+    <dependency>
```
albertshau, line 184: can you add a comment about what requires this dependency?

pom.xml
```diff
@@ -74,6 +76,8 @@
     <google.protobuf.java.version>3.4.0</google.protobuf.java.version>
     <slf4j.version>1.7.5</slf4j.version>
     <spark.version>1.6.1</spark.version>
+    <hbase-shaded-server.version>1.4.10</hbase-shaded-server.version>
```
albertshau: keep the properties alphabetically sorted so that it's easier to find one.

pom.xml
```diff
@@ -155,6 +159,27 @@
...
       <version>${bigquery.connector.hadoop2.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-hbase-1.x-mapreduce</artifactId>
+      <version>${google.cloud.bigtable.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.hbase</groupId>
+          <artifactId>hbase-server</artifactId>
```
albertshau, line 170: It's unfortunate that this requires server code, it seems like it should just be client logic. Do you know how much bigger this makes the jar?

Out of curiosity, what fails if we just exclude this without adding the shaded server dependency?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+   * @param input   input record
+   * @param emitter emitting the transformed record into downstream nodes.
+   */
+  @Override
+  public void transform(KeyValue<ImmutableBytesWritable, Result> input, Emitter<StructuredRecord> emitter) {
+    try {
+      StructuredRecord record = convertInputToRecord(input);
+      emitter.emit(record);
+    } catch (Exception e) {
+      switch (config.getErrorHandling()) {
```
albertshau, line 134: I didn't see this property in the reference doc.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+        case FAIL_PIPELINE:
+          throw new RuntimeException("Failed to process message", e);
+        default:
+          throw new IllegalStateException(String.format("Unknown error handling strategy '%s'",
+                                                            config.getErrorHandling()));
+      }
+    }
+  }
+
+  private StructuredRecord convertInputToRecord(KeyValue<ImmutableBytesWritable, Result> input) {
```
albertshau: As a general pattern, it's better to have StructuredRecord conversion logic in its own class. This makes it re-usable and also much easier to unit test. This PR, for example, should have unit tests for converting ImmutableBytesWritable and Result into StructuredRecord.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+      emitter.emit(record);
+    } catch (Exception e) {
+      switch (config.getErrorHandling()) {
+        case SKIP:
+          LOG.warn("Failed to process message, skipping it", e);
+          break;
+        case FAIL_PIPELINE:
+          throw new RuntimeException("Failed to process message", e);
+        default:
+          throw new IllegalStateException(String.format("Unknown error handling strategy '%s'",
```
albertshau, line 142: add a comment that this should never happen because it is validated at configure and prepare time

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+      case LONG:
+        return Bytes.toLong(bytes);
+      case FLOAT:
+        return Bytes.toFloat(bytes);
+      case DOUBLE:
+        return Bytes.toDouble(bytes);
+      case BOOLEAN:
+        return Bytes.toBoolean(bytes);
+      default:
+        throw new UnexpectedFormatException("Field type '" + fieldSchema.getType() + "' is not supported");
```
albertshau: message should include which field has the problem, otherwise the user won't know what to do.

Field '%s' is of unsupported type '%s'.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+  private void emitLineage(BatchSourceContext context) {
+    LineageRecorder lineageRecorder = new LineageRecorder(context, config.referenceName);
+    lineageRecorder.createExternalDataset(config.getSchema());
+
+    List<Schema.Field> fields = Objects.requireNonNull(config.getSchema()).getFields();
+    if (null != fields) {
+      List<String> fieldNames = fields.stream()
+        .map(Schema.Field::getName)
+        .collect(Collectors.toList());
+      lineageRecorder.recordRead("Read", "Read from Bigtable.", fieldNames);
```
albertshau: would be good to include the project, instance, and table in the description.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+  private Scan getConfiguredScanForJob() throws IOException {
+    Scan s = new Scan();
+    if (null != config.scanTimeRangeStart && null != config.scanTimeRangeStop) {
+      s.setTimeRange(config.scanTimeRangeStart, config.scanTimeRangeStop);
+    }
+    s.setCacheBlocks(false);
+    if (null != config.scanRowStart) {
+      s.withStartRow(Bytes.toBytes(config.scanRowStart));
+    }
+    if (null != config.scanTimeRangeStop) {
```
albertshau: should this be scanRowStop?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+      s.withStartRow(Bytes.toBytes(config.scanTimeRangeStop));
```
albertshau: shouldn't this be withStopRow?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -0,0 +1,249 @@
...
+      List<String> fieldNames = fields.stream()
+        .map(Schema.Field::getName)
+        .collect(Collectors.toList());
+      lineageRecorder.recordRead("Read", "Read from Bigtable.", fieldNames);
+    }
+  }
+
+  private Scan getConfiguredScanForJob() throws IOException {
+    Scan s = new Scan();
+    if (null != config.scanTimeRangeStart && null != config.scanTimeRangeStop) {
```
albertshau: does setTimeRange require non-null values for both the time range need both start and end? Logically it seems like only one should be required. If start is not given, there is no start time (0). If end is not given, there is no end time (long max value)

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+    this.scanRowStop = scanRowStop;
+    this.scanTimeRangeStart = scanTimeRangeStart;
+    this.scanTimeRangeStop = scanTimeRangeStop;
+    this.onError = onError;
+    this.schema = schema;
+  }
+
+  public void validate() {
+    super.validate();
+    if (!containsMacro(PROJECT) && null == project) {
```
albertshau: this is not correct, it should call getProject() to actually try and auto-detect.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+      throw new InvalidConfigPropertyException("Could not detect Google Cloud project id from the environment. " +
+                                                 "Please specify a project id.", PROJECT);
+    }
+    if (!containsMacro(INSTANCE) && Strings.isNullOrEmpty(instance)) {
+      throw new InvalidConfigPropertyException("Instance ID cannot be null or empty", INSTANCE);
```
albertshau: we're trying to avoid "technical" language like "null" in error messages that can appear in the UI. We can just say:

An Instance ID must be specified.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+    String serviceAccountFilePath = getServiceAccountFilePath();
+    if (!containsMacro(SERVICE_ACCOUNT_FILE_PATH) && null != serviceAccountFilePath) {
+      File serviceAccountFile = new File(serviceAccountFilePath);
+      if (!serviceAccountFile.exists()) {
+        throw new InvalidConfigPropertyException(String.format("File '%s' does not exist", serviceAccountFilePath),
```
albertshau: File -> Service account file

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+                                                 SERVICE_ACCOUNT_FILE_PATH);
+      }
+    }
+    if (!containsMacro(ON_ERROR) && null != onError) {
+      if (Strings.isNullOrEmpty(onError)) {
+        throw new InvalidConfigPropertyException("On Record Error cannot be null or empty", ON_ERROR);
```
albertshau: Error handling must be specified.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+      }
+      if (null == ErrorHandling.fromDisplayName(onError)) {
+        throw new InvalidConfigPropertyException("Invalid record error handling strategy name", ON_ERROR);
+      }
+    }
+    if (!containsMacro(SCHEMA)) {
+      Schema parsedSchema = getSchema();
+      if (null == parsedSchema) {
+        throw new InvalidConfigPropertyException("Schema cannot be null", SCHEMA);
```
albertshau: Schema must be specified.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+      }
+      List<Schema.Field> fields = parsedSchema.getFields();
+      if (null == fields) {
+        throw new InvalidConfigPropertyException("Schema should contain fields to map", SCHEMA);
```
albertshau, line 173: This can only happen if the schema is not a Record. We generally don't check this, though it's not wrong to have it. The message should be:

Schema is of invalid type '%s'. The schema must be a record.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+      }
+      for (Schema.Field field : fields) {
+        if (field.getName().equals(keyAlias)) {
+          // key should be defined in schema but will not be retrieved from column families.
+          continue;
```
albertshau, line 189: should make sure it's a supported type. Array, Map, Record, and a union that doesn't represent a nullable don't make sense.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        if (!HBaseColumn.isValidName(field.getName())) {
+          throw new InvalidConfigPropertyException("Wrong field name for schema. " +
+                                                     "Required format is '<family>__<qualifier>'", SCHEMA);
+        }
```
albertshau, line 197: should check that the type is supported

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseColumn.java
```diff
@@ -0,0 +1,38 @@
...
+  public static HBaseColumn fromFullName(String fullName) {
+    String[] parts = PATTERN.split(fullName, 2);
+    String family = parts[0];
+    String qualifier = parts[1];
+    return new HBaseColumn(family, qualifier);
+  }
+
+  public static boolean isValidName(String fullName) {
+    String[] parts = PATTERN.split(fullName);
+    return 2 == parts.length;
```
albertshau: it is valid to have '__' in the qualifier, which will result in more than 2 parts. This should check that there are at least 2 parts.

docs/Bigtable-batchsource.md
mlozbin-cybervisiontech: Was copied from TableInputFormat documentation. Fixed.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: There is a method in Schema class that creates a Schema with record type and 'null' fields. Added both checks.

albertshau:
Not all the comments were addressed, please address them.

For example, moving conversion logic to its own class and unit testing that class.

pom.xml
mlozbin-cybervisiontech: Unfortunately, TableInputFormat and TableOutputFormat are both parts of server library code

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: This is a name of the enum class. Property is called onError. This getter just parses user-defined property and returns enum.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: fixed but had to introduce new method to GCPConfig.

docs/Bigtable-batchsource.md
mlozbin-cybervisiontech: fixed

docs/Bigtable-batchsource.md
mlozbin-cybervisiontech: left without changes for now

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

pom.xml
mlozbin-cybervisiontech: added

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: added check

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: added check

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseColumn.java
mlozbin-cybervisiontech: fixed

pom.xml
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: included

mlozbin-cybervisiontech:
> Not all the comments were addressed, please address them.
> 
> For example, moving conversion logic to its own class and unit testing that class.

@albertshau just pushed commit with all comments being addressed. Sorry for missing some of them last time. For some reason Github hidden them and I missed "Load more messages" link.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
albertshau: Can you make sure it's included in the reference doc with a description?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,238 @@
...
+      }
+      for (Schema.Field field : fields) {
+        Schema.Type fieldType = field.getSchema().isNullable() ?
+          field.getSchema().getNonNullable().getType() : field.getSchema().getType();
+        if (!SUPPORTED_FIELD_TYPES.contains(fieldType)) {
+          String supportedTypes = SUPPORTED_FIELD_TYPES.stream()
+            .map(Enum::name)
+            .map(String::toLowerCase)
+            .collect(Collectors.joining(", "));
+          String errorMessage = String.format("Wrong type '%s' for field '%s'. Supported types are: '%s'",
```
albertshau: nit: Field '%s' is of unsupported type '%s'. Supported types are: %s

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseColumn.java
```diff
@@ -0,0 +1,50 @@
...
+  private final String family;
+  private final String qualifier;
+
+  private HBaseColumn(String family, String qualifier) {
+    this.family = family;
+    this.qualifier = qualifier;
+  }
+
+  public static HBaseColumn fromFullName(String fullName) {
+    int delimiterIndex = fullName.indexOf(FAMILY_QUALIFIER_DELIMITER);
```
albertshau, line 36: I think isValidName() is always called before this, but it is safer to check that this is not negative and throw an exception with a descriptive message if it is invalid. This makes it impossible for the caller to use this class incorrectly and it also isolates the logic about what is a valid field name to just this class.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -0,0 +1,238 @@
...
+                                              fieldType, field.getName(), supportedTypes);
+          throw new InvalidConfigPropertyException(errorMessage, SCHEMA);
+        }
+        if (field.getName().equals(keyAlias)) {
+          // key should be defined in schema but will not be retrieved from column families.
+          continue;
+        }
+        if (!HBaseColumn.isValidName(field.getName())) {
+          String errorMessage = String.format("Wrong name for field '%s'. Required format is '<family>__<qualifier>'",
```
albertshau: this is a little dirty because both this config class and HBaseColumn now need to know what the expected format is. It's cleaner to have HBaseColumn throw an exception with this message and to have this class catch the exception and pass the message to InvalidConfigPropertyException.

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseResultToRecordTransformer.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.util.Bytes;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Transforms Google Cloud Bigtable {@link Result} to {@link StructuredRecord}.
+ */
+public class HBaseResultToRecordTransformer {
```
albertshau, line 30: please add unit tests for this class

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseColumn.java
mlozbin-cybervisiontech: done

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseResultToRecordTransformer.java
mlozbin-cybervisiontech: all unit and integration tests will be published in next PR for easier review

mlozbin-cybervisiontech:
@albertshau addressed all recent comments

src/main/java/io/cdap/plugin/gcp/bigtable/source/HBaseColumn.java
```diff
@@ -30,13 +32,22 @@ private HBaseColumn(String family, String qualifier) {
   }
 
   public static HBaseColumn fromFullName(String fullName) {
+    if (!isValidName(fullName)) {
```
albertshau: this can be replaced with a call to validateName()

albertshau (commented):
just one minor comment otherwise lgtm. Please fix and squash commits and i'll merge.

PR #123, opened 2019-06-26
CDAP-15587 BigQuery: Partitioning

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/CDAP-15587

In scope of this PR:

- Added partitions functionality for BigQuery Sink plugin during create new table.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -17,6 +17,7 @@ package io.cdap.plugin.gcp.bigquery.sink;
 
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
 import io.cdap.plugin.gcp.common.GCPReferenceSinkConfig;
 
@@ -46,6 +47,25 @@ public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig 
   @Description("Whether to modify the BigQuery table schema if it differs from the input schema.")
   protected boolean allowSchemaRelaxation;
 
+  @Macro
+  @Nullable
+  @Description("Whether to create the BigQuery table with time partitioning. This value is ignored if the table " +
+    "already exists.")
+  protected Boolean createPartitionedTable;
+
+  @Name("partitionByField")
+  @Macro
+  @Nullable
+  @Description("Partitioning column for the BigQuery table. This should be left empty if the BigQuery table is an " +
+    "ingestion-time partitioned table.")
+  protected String partitionByField;
+
+  @Macro
+  @Nullable
+  @Description("Whether to create a table that requires a partition filter. This value is ignored if the table " +
+    "already exists.")
+  protected Boolean partitionFilterRequired;
+
   public String getDataset() {
     return dataset;
   }
@@ -67,6 +87,19 @@ public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig 
 
   public boolean isAllowSchemaRelaxation() {
     return allowSchemaRelaxation;
+  }
+
+  public boolean shouldCreatePartitionedTable() {
+    return createPartitionedTable == null ? false : createPartitionedTable;
+  }
+
+  @Nullable
+  public String getPartitionByField() {
+    return partitionByField;
+  }
+
+  public boolean isPartitionFilterRequired() {
+    return partitionFilterRequired == null ? false : partitionFilterRequired;
   }
 
   @Override
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -32,6 +32,7 @@ import com.google.api.services.bigquery.model.JobReference;
 import com.google.api.services.bigquery.model.TableFieldSchema;
 import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableSchema;
+import com.google.api.services.bigquery.model.TimePartitioning;
 import com.google.cloud.bigquery.JobInfo;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import com.google.cloud.hadoop.io.bigquery.BigQueryFactory;
@@ -115,10 +116,17 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
 
       boolean allowSchemaRelaxation = conf.getBoolean(BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION, false);
       LOG.debug("Allow schema relaxation: '{}'", allowSchemaRelaxation);
+      boolean createPartitionedTable = conf.getBoolean(BigQueryConstants.CONFIG_CREATE_PARTITIONED_TABLE, false);
+      LOG.debug("Create Partitioned Table: '{}'", createPartitionedTable);
+      String partitionByField = conf.get(BigQueryConstants.CONFIG_PARTITION_BY_FIELD, null);
+      LOG.debug("Partition Field: '{}'", partitionByField);
+      boolean requirePartitionFilter = conf.getBoolean(BigQueryConstants.CONFIG_REQUIRE_PARTITION_FILTER, false);
+      LOG.debug("Require partition filter: '{}'", requirePartitionFilter);
 
       try {
         importFromGcs(destProjectId, destTable, destSchema.orElse(null), kmsKeyName, outputFileFormat,
-                      writeDisposition, sourceUris, allowSchemaRelaxation);
+                      writeDisposition, sourceUris, allowSchemaRelaxation, createPartitionedTable, partitionByField,
+                      requirePartitionFilter);
       } catch (InterruptedException e) {
         throw new IOException("Failed to import GCS into BigQuery", e);
       }
@@ -138,7 +146,8 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
      */
     private void importFromGcs(String projectId, TableReference tableRef, @Nullable TableSchema schema,
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
-                               List<String> gcsPaths, boolean allowSchemaRelaxation)
+                               List<String> gcsPaths, boolean allowSchemaRelaxation, boolean createPartitionedTable,
+                               @Nullable String partitionByField, boolean requirePartitionFilter)
       throws IOException, InterruptedException {
       LOG.info("Importing into table '{}' from {} paths; path[0] is '{}'; awaitCompletion: {}",
                BigQueryStrings.toString(tableRef), gcsPaths.size(), gcsPaths.isEmpty() ? "(empty)" : gcsPaths.get(0),
@@ -151,6 +160,16 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       loadConfig.setSourceUris(gcsPaths);
       loadConfig.setDestinationTable(tableRef);
       loadConfig.setWriteDisposition(writeDisposition);
+      if (createPartitionedTable) {
+        TimePartitioning timePartitioning = new TimePartitioning();
+        timePartitioning.setType("DAY");
+        if (partitionByField != null) {
+          timePartitioning.setField(partitionByField);
+        }
+        timePartitioning.setRequirePartitionFilter(requirePartitionFilter);
+        loadConfig.setTimePartitioning(timePartitioning);
+      }
+
       if (!Strings.isNullOrEmpty(kmsKeyName)) {
         loadConfig.setDestinationEncryptionConfiguration(new EncryptionConfiguration().setKmsKeyName(kmsKeyName));
       }
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
@@ -16,11 +16,18 @@
 
 package io.cdap.plugin.gcp.bigquery.sink;
 
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.Table;
+import com.google.cloud.bigquery.TimePartitioning;
 import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
+import io.cdap.cdap.etl.api.validation.InvalidStageException;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import javax.annotation.Nullable;
@@ -30,6 +37,7 @@ import javax.annotation.Nullable;
  * configuring the <code>BigQuerySink</code> plugin.
  */
 public final class BigQuerySinkConfig extends AbstractBigQuerySinkConfig {
+  private static final Logger LOG = LoggerFactory.getLogger(BigQuerySinkConfig.class);
 
   @Macro
   @Description("The table to write to. A table contains individual records organized in rows. "
@@ -80,6 +88,8 @@ public final class BigQuerySinkConfig extends AbstractBigQuerySinkConfig {
     super.validate();
     if (!containsMacro("schema")) {
       Schema outputSchema = getSchema();
+      Schema schema = outputSchema != null ? outputSchema : inputSchema;
+      validatePartitionProperties(schema);
       if (outputSchema == null) {
         return;
       }
@@ -113,6 +123,53 @@ public final class BigQuerySinkConfig extends AbstractBigQuerySinkConfig {
           BigQueryUtil.validateArraySchema(field.getSchema(), name);
         }
       }
+    }
+  }
+
+  private void validatePartitionProperties(@Nullable Schema schema) {
+    Table table = BigQueryUtil.getBigQueryTable(getProject(), getDataset(), getTable(), getServiceAccountFilePath());
+    if (table != null) {
+      StandardTableDefinition tableDefinition = table.getDefinition();
+      TimePartitioning timePartitioning = tableDefinition.getTimePartitioning();
+      if (timePartitioning == null && createPartitionedTable != null && createPartitionedTable) {
+        LOG.warn(String.format("The plugin is configured to auto-create a partitioned table, but table '%s' already " +
+                                 "exists without partitioning. Please verify the partitioning configuration.",
+                               table.getTableId().getTable()));
+      }
+      if (timePartitioning != null && timePartitioning.getField() != null
+        && !timePartitioning.getField().equals(partitionByField)) {
+        throw new InvalidConfigPropertyException(String.format("Destination table '%s' is partitioned by column '%s'." +
+                                                                 " Please set the partition field to '%s'.",
+                                                               table.getTableId().getTable(),
+                                                               timePartitioning.getField(),
+                                                               timePartitioning.getField()), "partitionByField");
+      }
+      validateColumnForPartition(partitionByField, schema);
+    }
+    if (createPartitionedTable == null || !createPartitionedTable) {
+      return;
+    }
+    validateColumnForPartition(partitionByField, schema);
+  }
+
+  private void validateColumnForPartition(@Nullable String columnName, @Nullable Schema schema) {
+    if (columnName == null || schema == null) {
+      return;
+    }
+    Schema.Field field = schema.getField(columnName);
+    if (field == null) {
+      throw new InvalidConfigPropertyException(
+        String.format("Partition column '%s' is missing from the table schema", columnName), "Partition Field");
+    }
+    Schema fieldSchema = field.getSchema();
+    fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
+    Schema.LogicalType logicalType = fieldSchema.getLogicalType();
+    if (logicalType != Schema.LogicalType.DATE && logicalType != Schema.LogicalType.TIMESTAMP_MICROS
+      && logicalType != Schema.LogicalType.TIMESTAMP_MILLIS) {
+      String type = logicalType != null ? logicalType.getToken() : fieldSchema.getType().name();
+      throw new InvalidStageException(String.format("Partition column '%s' is of invalid type '%s'. " +
+                                                      "Please change it to a date or timestamp.",
+                                                    columnName, type));
     }
   }
 }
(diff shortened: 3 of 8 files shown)


AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -46,6 +46,19 @@
   @Description("Whether to modify the BigQuery table schema if it differs from the input schema.")
   protected boolean allowSchemaRelaxation;
 
+  @Macro
+  @Description("Enable partitioning during creating BigQuery table.")
```
albertshau: Whether to create the BigQuery table with time partitioning. This value is ignored if the table already exists.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -46,6 +46,19 @@
...
+  protected boolean timePartitioningAllowed;
```
albertshau: this should be a Boolean, as it will be null in the configurePipeline() method if it has a macro. It should also be Nullable, as this only matters if we are automatically creating the table vs writing to an existing one.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -46,6 +46,19 @@
...
+
+  @Macro
+  @Nullable
+  @Description("Column name for partitioning during creating BigQuery table.")
```
albertshau: Partitioning column for the BigQuery table. This should be left empty if the BigQuery table is an ingestion-time partitioned table.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
albertshau: I think 'createPartitionedTable' would be a more descriptive name for this variable.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -46,6 +46,19 @@
...
+  protected String partitionByField;
+
+  @Macro
+  @Description("Enable required partition filter for BigQuery table during creating table.")
```
albertshau: Whether to create a table that requires a partition filter. This value is ignored if the table already exists.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -80,6 +80,9 @@ public void validate(@Nullable Schema inputSchema) {
     super.validate();
     if (!containsMacro("schema")) {
       Schema outputSchema = getSchema();
+      if (timePartitioningAllowed) {
+        BigQueryUtil.validateColumnForPartition(partitionByField, outputSchema);
```
albertshau: it's better to have the validation logic as another method in this class instead of in a util class since nothing else is calling this method.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
```diff
@@ -21,4 +21,7 @@
 public interface BigQueryConstants {
 
   String CONFIG_ALLOW_SCHEMA_RELAXATION = "mapred.bg.sink.allow_schema_relaxation";
+  String CONFIG_ALLOW_TIME_PARTITIONING = "mapred.bg.sink.allow_time_partitioning";
```
albertshau: It seems like 'bg' in the original constant was a typo. Should be using 'bq' everywhere. A property starting with 'mapred' is also usually one from the mapreduce framework so it would be better if these started with 'cdap.bq'.
Another nitpick is that they should use dots everywhere instead of a mix of dots and underscores (cdap.bq.sink.allow.schema.relaxation) 

This doesn't impact functionality at all, but it might make debugging easier in situations where we might need to investigate a failing job and have access to the configuration file with all these properties.

widgets/BigQueryTable-batchsink.json
```diff
@@ -47,6 +47,52 @@
             "placeholder": "Google Cloud Storage bucket for temporary data"
           }
         },
+        {
+          "widget-type": "radio-group",
+          "name": "timePartitioningAllowed",
+          "label": "Allow Time Partitioning",
```
albertshau: Create Partitioned Table

widgets/BigQueryTable-batchsink.json
```diff
@@ -47,6 +47,52 @@
...
+              {
+                "id": "false",
+                "label": "False"
+              }
+            ]
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Partition By Field",
```
albertshau: Partition Field

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
albertshau: these are also properties specific to the outputformat and nothing else. They should be static final variables in that class and this interface deleted.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -151,6 +160,16 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
       loadConfig.setSourceUris(gcsPaths);
       loadConfig.setDestinationTable(tableRef);
       loadConfig.setWriteDisposition(writeDisposition);
+      if (allowTimePartitioning) {
```
albertshau: Is this needed if the table already exists or does the load job always require this?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
                                                        name, componentSchema));
     }
   }
+
+  public static void validateColumnForPartition(String columnName, Schema outputSchema) {
```
albertshau: annotate arguments as nullable

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+    if (columnName == null) {
+      return;
+    }
+    if (outputSchema == null) {
```
albertshau: If the schema contains a macro, it will be null during the configurePipeline() call. In that scenario, this should not throw an exception. This is one reason this method fits better in the config class rather than here, as there you can check with containsMacro("schema")

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+      throw new InvalidStageException(String.format("Can't validate column '%s' because output schema not specified",
+                                                       columnName));
+    }
+    for (Schema.Field field : outputSchema.getFields()) {
+      if (field.getName().equals(columnName)) {
```
albertshau: Don't need to loop, can just do:

```
Schema.Field field = outputSchema.getField(columnName);
if (field == null) {
  throw ...
}
```

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+        if (!field.getSchema().getType().equals(Schema.Type.UNION)) {
```
albertshau: this can be simplified to:

```
fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
Schema.LogicalType logicalType = fieldSchema.getLogicalType();
if (logicalType != ...) {
  throw ...
}
```

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+          ? field.getSchema().getLogicalType()
+          : field.getSchema().getUnionSchema(0).getLogicalType();
+        throw new InvalidStageException(String.format("Column '%s' is of type '%s' (logical type '%s'), " +
+                                                           "expected logical type is '%s' or '%s'",
+                                                      columnName, type, fieldType,
+                                                      Schema.LogicalType.DATE,
+                                                      Schema.LogicalType.TIMESTAMP_MICROS));
+      }
+    }
+    throw new InvalidStageException(String.format("Column '%s' missing in output table schema", columnName));
```
albertshau: This should throw an InvalidConfigPropertyException where it sets the partition column property as the invalid property, with message:

Partition column '%s' is missing from the table schema.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+          ? field.getSchema().getType()
+          : field.getSchema().getUnionSchema(0).getType();
+        Schema.LogicalType fieldType = !Schema.Type.UNION.equals(field.getSchema().getType())
+          ? field.getSchema().getLogicalType()
+          : field.getSchema().getUnionSchema(0).getLogicalType();
+        throw new InvalidStageException(String.format("Column '%s' is of type '%s' (logical type '%s'), " +
+                                                           "expected logical type is '%s' or '%s'",
+                                                      columnName, type, fieldType,
+                                                      Schema.LogicalType.DATE,
+                                                      Schema.LogicalType.TIMESTAMP_MICROS));
```
albertshau: use the .token() method for the error messages.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+    if (outputSchema == null) {
+      throw new InvalidStageException(String.format("Can't validate column '%s' because output schema not specified",
+                                                       columnName));
+    }
+    for (Schema.Field field : outputSchema.getFields()) {
+      if (field.getName().equals(columnName)) {
+        if (!field.getSchema().getType().equals(Schema.Type.UNION)) {
+          if (field.getSchema().getLogicalType() != null
+            && (field.getSchema().getLogicalType().equals(Schema.LogicalType.DATE)
+            || field.getSchema().getLogicalType().equals(Schema.LogicalType.TIMESTAMP_MICROS))) {
```
albertshau: Even though most plugins don't use it, the sink should also allow TIMESTAMP_MILLIS in case somebody has written their own plugin that uses it.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -341,4 +342,45 @@ public static void validateArraySchema(Schema arraySchema, String name) {
...
+            }
+          }
+        }
+        Schema.Type type = !Schema.Type.UNION.equals(field.getSchema().getType())
+          ? field.getSchema().getType()
+          : field.getSchema().getUnionSchema(0).getType();
+        Schema.LogicalType fieldType = !Schema.Type.UNION.equals(field.getSchema().getType())
+          ? field.getSchema().getLogicalType()
+          : field.getSchema().getUnionSchema(0).getLogicalType();
+        throw new InvalidStageException(String.format("Column '%s' is of type '%s' (logical type '%s'), " +
```
albertshau: logical types are not a concept exposed to UI users. They see logical types 'timestamp' or 'date' in the same dropdown as the physical types. The message should be something like:

Partition column '%s' is of invalid type '%s'. Please change it to a date or timestamp.

where the type mentioned in the message is the logical type if it's not null and the physical type if there is no logical type.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -80,6 +80,9 @@ public void validate(@Nullable Schema inputSchema) {
...
+      }
```
albertshau: we also need to handle the scenario where the table exists but it doesn't match the partitioning values that have been configured. For example:

1. Table is not partitioned but the user has specified a partition column
2. Table is partitioned but on a column that doesn't match the user provided column.
3. Table is requires a partition filter and the user did not provide a partition column

albertshau (commented):
please also update the reference doc (BigQueryTable-batchsink.md)

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
AntonGensitskiy: This properties specific to the BigQueryOutputFormat and AbstractBigQuerySink classes.

widgets/BigQueryTable-batchsink.json
AntonGensitskiy: Done.

widgets/BigQueryTable-batchsink.json
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: This need only when new table should be created.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

AntonGensitskiy:
@Sree allocate reviewer.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -69,6 +87,21 @@ public boolean isAllowSchemaRelaxation() {
     return allowSchemaRelaxation;
   }
 
+  @Nullable
+  public Boolean isCreatePartitionedTable() {
```
albertshau: nit: rename to shouldCreatePartitionedTable

albertshau: I think it's better if this getter returns a boolean so that callers don't need to check for null. If the value is null, it should return false.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -69,6 +87,21 @@ public boolean isAllowSchemaRelaxation() {
...
+    return createPartitionedTable;
+  }
+
+  @Nullable
+  public String getPartitionByField() {
+    return partitionByField;
+  }
+
+  @Nullable
+  public Boolean isPartitionFilterRequired() {
```
albertshau: similar comment, would be simpler if this returns a boolean, returning false if it's null.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -115,4 +122,49 @@ public void validate(@Nullable Schema inputSchema) {
       }
     }
   }
+
+  private void validatePartitionProperties(@Nullable Schema schema) {
+    Table table = BigQueryUtil.getBigQueryTable(getProject(), getDataset(), getTable(), getServiceAccountFilePath());
+    if (table != null) {
+      StandardTableDefinition tableDefinition = table.getDefinition();
+      TimePartitioning timePartitioning = tableDefinition.getTimePartitioning();
+      if (timePartitioning == null && createPartitionedTable != null && createPartitionedTable) {
```
albertshau, line 134: We documented that this property is ignored if the table already exists, so we should not through an exception here. It does seem like a weird configuration though, so it may make sense to log a warning. Something like:

The plugin is configured to auto-create a partitioned table, but table '%s' already exists without partitioning. Please verify the partitioning configuration.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -115,4 +122,49 @@ public void validate(@Nullable Schema inputSchema) {
...
+        throw new InvalidConfigPropertyException(String.format("Destination table '%s' does not support partitioning",
+                                                               table.getTableId().getTable()),
+                                                 "Create Partitioned Table");
+      }
+      if (timePartitioning != null && timePartitioning.getField() != null
+        && !timePartitioning.getField().equals(partitionByField)) {
+        throw new InvalidConfigPropertyException(String.format("Destination table '%s' is partitioned by column '%s'",
```
albertshau: Destination table '%s' is partitioned by column '%s'. Please set the partition field to '%s'.

widgets/BigQueryTable-batchsink.json
```diff
@@ -47,6 +47,52 @@
             "placeholder": "Google Cloud Storage bucket for temporary data"
           }
         },
+        {
+          "widget-type": "radio-group",
```
albertshau: New plugins should use the 'toggle' type for boolean properties (https://docs.cdap.io/cdap/6.0.0/en/developer-manual/pipelines/developing-plugins/presentation-plugins.html#H2293).

albertshau (commented):
Small comments.

The sink also supports an 'Update Schema' property. Have we checked if this works properly with partitioned tables? For example, suppose we create a partitioned table with fields x, y. Then we create a new pipeline that writes to the same table with fields x, y, z. Does the table get updated correctly as a partitioned table?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

widgets/BigQueryTable-batchsink.json
AntonGensitskiy: Done.

AntonGensitskiy:
> Small comments.
> 
> The sink also supports an 'Update Schema' property. Have we checked if this works properly with partitioned tables? For example, suppose we create a partitioned table with fields x, y. Then we create a new pipeline that writes to the same table with fields x, y, z. Does the table get updated correctly as a partitioned table?

This functionality correct work with partitioned tables.

AntonGensitskiy:

widgets/BigQueryTable-batchsink.json
```diff
@@ -48,22 +48,59 @@
...
         },
         {
-          "widget-type": "radio-group",
+          "name": "createPartitionedTable",
+          "widget-type": "toggle",
+          "label": "Create Partitioned Table",
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "On"
```
albertshau: It sounds a bit weird to say 'Create Partitioned Table' is 'On'. I believe other places just use capitalized 'True' and 'False' as the label.

albertshau (commented):
just one comment about the toggle labels, otherwise lgtm. Please fix and squash commits and i'll merge.

widgets/BigQueryTable-batchsink.json
AntonGensitskiy: Labels was changed.

AntonGensitskiy:

PR #125, opened 2019-07-11
CDAP-15587 BigQuery: Partitioning

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/CDAP-15587

In scope of this PR:
* Added partitions functionality for BigQuery Source plugin.

### DIFF
--- a/docs/BigQueryTable-batchsource.md
+++ b/docs/BigQueryTable-batchsource.md
@@ -39,6 +39,12 @@ Datasets are top-level containers that are used to organize and control access t
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 
+**Partition Start Date**: Inclusive partition start date, specified as 'yyyy-MM-dd'. For example, '2019-01-01'. 
+If no value is given, all partitions up to the partition end date will be read.
+
+**Partition End Date**: Exclusive partition end date, specified as 'yyyy-MM-dd'. For example, '2019-01-01'. 
+If no value is given, all partitions up from the partition start date will be read.
+
 **Temporary Bucket Name**: Google Cloud Storage bucket to store temporary data in.
 It will be automatically created if it does not exist, but will not be automatically deleted.
 Temporary data will be deleted after it has been read. If it is not provided, a unique bucket will be
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
@@ -23,9 +23,10 @@ import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.StandardSQLTypeName;
+import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
-import com.google.cloud.hadoop.io.bigquery.AvroBigQueryInputFormat;
+import com.google.cloud.bigquery.TimePartitioning;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
@@ -43,6 +44,7 @@ import io.cdap.cdap.etl.api.batch.BatchSourceContext;
 import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
 import io.cdap.cdap.etl.api.validation.InvalidStageException;
 import io.cdap.plugin.common.LineageRecorder;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.GCPUtils;
 import io.cdap.plugin.gcp.common.Schemas;
@@ -56,9 +58,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.time.DateTimeException;
+import java.time.LocalDate;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -114,6 +119,7 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
       null : GCPUtils.loadServiceAccountCredentials(serviceAccountPath);
     BigQuery bigQuery = GCPUtils.getBigQuery(config.getDatasetProject(), credentials);
     validateOutputSchema(bigQuery);
+    validatePartitionProperties();
 
     uuid = UUID.randomUUID();
     configuration = BigQueryUtil.getBigQueryConfig(config.getServiceAccountFilePath(), config.getProject());
@@ -126,18 +132,26 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
       configuration.setBoolean("fs.gs.bucket.delete.enable", true);
     }
 
-    if (!context.isPreviewEnabled()) {
-      BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(config.getDatasetProject(), credentials),
-                                   config.getDataset(), bucket);
-    }
+    BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(config.getDatasetProject(), credentials),
+                                 config.getDataset(), bucket);
 
     configuration.set("fs.gs.system.bucket", bucket);
     configuration.setBoolean("fs.gs.impl.disable.cache", true);
     configuration.setBoolean("fs.gs.metadata.cache.enable", false);
 
+    if (config.getServiceAccountFilePath() != null) {
+      configuration.set(BigQueryConstants.CONFIG_SERVICE_ACCOUNT_FILE_PATH, config.getServiceAccountFilePath());
+    }
+    if (config.getPartitionFrom() != null) {
+      configuration.set(BigQueryConstants.CONFIG_PARTITION_FROM_DATE, config.getPartitionFrom());
+    }
+    if (config.getPartitionTo() != null) {
+      configuration.set(BigQueryConstants.CONFIG_PARTITION_TO_DATE, config.getPartitionTo());
+    }
+
     String temporaryGcsPath = String.format("gs://%s/hadoop/input/%s", bucket, uuid);
-    AvroBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
-    AvroBigQueryInputFormat.setEnableShardedExport(configuration, false);
+    PartitionedBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
+    PartitionedBigQueryInputFormat.setEnableShardedExport(configuration, false);
     BigQueryConfiguration.configureBigQueryInput(configuration, config.getDatasetProject(),
                                                  config.getDataset(), config.getTable());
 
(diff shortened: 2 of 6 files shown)


AntonGensitskiy:

docs/BigQueryTable-batchsource.md
```diff
@@ -39,6 +39,10 @@ Datasets are top-level containers that are used to organize and control access t
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 
+**Partition From**: Partition start date. This value is ignored if the table does not support partitioning.
```
albertshau: Doc should mention how this is expected to be specified with an example. Timestamp? String with a specific format?

albertshau: Is the start date inclusive? exclusive?

docs/BigQueryTable-batchsource.md
```diff
@@ -39,6 +39,10 @@ Datasets are top-level containers that are used to organize and control access t
...
+
+**Partition To**: Partition end date. This value is ignored if the table does not support partitioning.
```
albertshau: same comments as start date

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -136,11 +146,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     configuration.setBoolean("fs.gs.impl.disable.cache", true);
     configuration.setBoolean("fs.gs.metadata.cache.enable", false);
 
+    String query = generateQuery();
```
albertshau: This logic should be moved into the InputFormat. The InputFormat should support additional partition start and end settings, with the query generation logic and temp table logic entirely in the InputFormat. The source should only be responsible for setting the partition start and end settings.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -136,11 +146,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+
+    if (query != null) {
+      configuration.set(BigQueryConfiguration.INPUT_QUERY_KEY, query);
+    }
+
     String temporaryGcsPath = String.format("gs://%s/hadoop/input/%s", bucket, uuid);
-    AvroBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
-    AvroBigQueryInputFormat.setEnableShardedExport(configuration, false);
+    CdapBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
```
albertshau: Cdap isn't really a descriptive name. PartitionedBigQueryInputFormat would be better.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -285,6 +322,92 @@ private void validateSupportedTypes(Schema.Field field) {
...
+    if (partitionFromDate == null && partitionToDate == null) {
+      return;
+    }
+    LocalDate fromDate = null;
+    if (partitionFromDate != null) {
+      try {
+        fromDate = LocalDate.parse(partitionFromDate);
+      } catch (DateTimeException ex) {
+        throw new InvalidConfigPropertyException("Invalid Partition From Date format. Partition From Date should be " +
+                                                   "in format yyyy-MM-dd", ex, "partitionFrom");
```
albertshau: it's better to create 'public static final String' constants in the config class and reference them here.

src/main/java/io/cdap/plugin/gcp/bigquery/source/CdapBigQueryInputFormat.java
```diff
@@ -0,0 +1,173 @@
...
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * BigQuery input format, splits query from the configuration into list of queries
+ * in order to create input splits.
+ */
+public class CdapBigQueryInputFormat extends AbstractBigQueryInputFormat<LongWritable, GenericData.Record> {
```
albertshau: It looks like there's a lot of copied code here and it's hard for me to tell what is copied and what is ours.

It's simpler to decorate the bigquery inputformat rather than trying to hook into the Export stuff. From our perspective, all we're doing is running a query that reads relevant partitions into a temporary table, then configuring the bigquery inputformat to read from that temporary table. So it should look something like:

```
public class CdapBigQueryInputFormat extends InputFormat<LongWritable, GenericData.Record> {
  
  public List<InputSplit> getSplits(JobContext context) {
    // code to execute query that writes to temporary table
    ...
    // set tablename configuration setting to temporary table
    return delegate.getSplits(context);
  }
}
```

docs/BigQueryTable-batchsource.md
albertshau: These names should match the labels used in the widget, which has 'Partition From Date'.

AntonGensitskiy: Done.

docs/BigQueryTable-batchsource.md
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/CdapBigQueryInputFormat.java
AntonGensitskiy: InputFormat class was refactored.

AntonGensitskiy:

docs/BigQueryTable-batchsource.md
```diff
@@ -39,6 +39,12 @@ Datasets are top-level containers that are used to organize and control access t
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 
+**Partition From Date**: It's inclusive partition start date. It should be a String with format "yyyy-MM-dd".
```
albertshau: Inclusive partition start date, specified as 'yyyy-MM-dd'. For example, '2019-01-01'. If no value is given, all partitions up to the partition end date will be read.

docs/BigQueryTable-batchsource.md
```diff
@@ -39,6 +39,12 @@ Datasets are top-level containers that are used to organize and control access t
 Each record is composed of columns (also called fields).
 Every table is defined by a schema that describes the column names, data types, and other information.
 
+**Partition From Date**: It's inclusive partition start date. It should be a String with format "yyyy-MM-dd". 
+This value is ignored if the table does not support partitioning.
+
+**Partition To Date**: It's inclusive partition end date. It should be a String with format "yyyy-MM-dd".
```
albertshau: Most of the other plugins that have a start and end time have the start as inclusive and end as exclusive. Let's do the same for this plugin.

docs/BigQueryTable-batchsource.md
albertshau: In the descriptions we call them 'start' and 'end' vs 'from' and 'to'. I think that is more natural, so let's change the labels as well.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -136,9 +146,20 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+    if (config.getServiceAccountFilePath() != null) {
+      configuration.set(BigQueryConstants.CONFIG_SERVICE_ACCOUNT_FILE_PATH, config.getServiceAccountFilePath());
+    }
+    if (config.getPartitionFrom() != null) {
+      configuration.set(BigQueryConstants.CONFIG_PARTITION_FROM_DATE, config.getPartitionFrom());
+    }
+    if (config.getPartitionTo() != null) {
+      configuration.set(BigQueryConstants.CONFIG_PARTITION_TO_DATE, config.getPartitionTo());
+    }
+    configuration.set(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME, uuid.toString().replaceAll("-", "_"));
```
albertshau: the temporary table name should be an implementation detail that only the InputFormat should know about. Allowing the plugin to set it would actually open up the possibility of the plugin using a non-unique name, resulting in failures.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -184,6 +205,26 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     } catch (IOException e) {
       LOG.warn("Failed to delete bucket " + gcsPath.toUri().getPath() + ", " + e.getMessage());
     }
+    if (configuration.get(BigQueryConstants.CONFIG_PARTITION_FROM_DATE) != null
+      || configuration.get(BigQueryConstants.CONFIG_PARTITION_TO_DATE) != null) {
+      try {
+        BigQueryFactory factory = new BigQueryFactory();
+        BigQueryHelper bigQueryHelper = factory.getBigQueryHelper(configuration);
+        Bigquery.Tables tables = bigQueryHelper.getRawBigquery().tables();
+        Bigquery.Tables.Delete delete = tables.delete(
```
albertshau: BigQuery has temporary tables that get cleaned up automatically after some expiry (BQ queries actually get stored in temporary tables). We should be relying on that functionality, as it's possible this logic never gets to run if the pipeline is killed.

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -0,0 +1,192 @@
...
+    // Create anonymous Progressable object
+    Progressable progressable = new Progressable() {
+      @Override
+      public void progress() {
+        // TODO(user): ensure task doesn't time out
+      }
+    };
+
+    // Poll until job is complete.
+    BigQueryUtils.waitForJobCompletion(
```
albertshau, line 196: we've had bugs in the past with datasets in non-default regions. Let's be sure to test with datasets in different regions.

docs/BigQueryTable-batchsource.md
AntonGensitskiy: Done.

docs/BigQueryTable-batchsource.md
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
AntonGensitskiy: Logic was changed. Currently temporary table is created with expiration time 1 day. Automatic deleting of table was removed.

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
AntonGensitskiy: I will check this case.

AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
AntonGensitskiy: @albertshau I checked with non-default regions and you was right I got error:
com.google.api.client.googleapis.json.GoogleJsonResponseException: 404 Not Found
{
  "code" : 404,
  "errors" : [ {
    "domain" : "global",
    "message" : "Not found: Job cdap-cybervision:querybasedexport-eff333bb-3784-4e1d-ac2b-ac67683c24bd",
    "reason" : "notFound"
  } ],
  "message" : "Not found: Job cdap-cybervision:querybasedexport-eff333bb-3784-4e1d-ac2b-ac67683c24bd",
  "status" : "NOT_FOUND"
}

If bugs with non-standard regions are known, are there any solutions that can be used?

albertshau: I've merged release to develop, can you check if the 404 problem still exists?

For the problem with the GCS bucket being created in the wrong region, was that only a problem in preview? I think it's because of:

```
if (!context.isPreviewEnabled()) {
      BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(config.getDatasetProject(), credentials),
                                   config.getDataset(), bucket);
    }
```

The if condition should be removed, since the GCS bucket is required whether it's in preview or not.

albertshau (commented):
small comment, otherwise lgtm. Please fix and squash, then I'll merge.

AntonGensitskiy:

PR #126, opened 2019-07-11

mlozbin-cybervisiontech:
WIKI: https://wiki.cask.co/display/CE/Bigtable+Sink

In scope of this PR:
- implemented batch sink plugin
- additional configuration for column mappings that will allow to use multiple column families will be submitted in separate PR for both source and sink plugins


mlozbin-cybervisiontech:

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
```diff
@@ -0,0 +1,95 @@
...
+                            String columnFamily) {
+    this.table = table;
+    this.instance = instance;
+    this.keyAlias = keyAlias;
+    this.columnFamily = columnFamily;
+  }
+
+  public void validate() {
+    super.validate();
+    if (!containsMacro(PROJECT) && null == project) {
```
albertshau: this won't actually try to detect the project. The getProject() method will try to detect and throw an exception if it can't read it. This should be:

```
if (!containsMacro(PROJECT)) {
  getProject();
}
```

I think I left a similar comment for the other PR, did we fix it there?

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
```diff
@@ -0,0 +1,88 @@
...
+  private final String columnFamily;
+
+  public RecordToHBaseMutationTransformer(String keyAlias, String columnFamily) {
+    this.keyAlias = keyAlias;
+    this.columnFamily = columnFamily;
+  }
+
+  public Mutation transform(StructuredRecord record) {
+    List<Schema.Field> fields = Objects.requireNonNull(record.getSchema().getFields(), "Schema fields cannot be empty");
+    String key = Objects.requireNonNull(record.get(keyAlias), "Key cannot be null");
```
albertshau: We should not be assuming the key is a string. The plugin should allow any of the supported types as a key.

albertshau: Include the field name in the error message:

"Key field '%s' contained a null value. Keys cannot be null."

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
```diff
@@ -0,0 +1,88 @@
...
+      case LONG:
+        return Bytes.toBytes((long) value);
+      case FLOAT:
+        return Bytes.toBytes((float) value);
+      case DOUBLE:
+        return Bytes.toBytes((double) value);
+      case BOOLEAN:
+        return Bytes.toBytes((boolean) value);
+      default:
+        throw new UnexpectedFormatException("Field type '" + fieldSchema.getType() + "' is not supported");
```
albertshau, line 99: should pass in a Schema.Field into the method so that we can include the field name in this error message:

"Field '%s' is of unsupported type '%s'".

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -0,0 +1,223 @@
...
+  public BigtableSink(BigtableSinkConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer configurer) {
+    super.configurePipeline(configurer);
+    config.validate();
+    Schema inputSchema = configurer.getStageConfigurer().getInputSchema();
+    if (null == inputSchema) {
```
albertshau: schema can be null if the previous stage did not set an output schema because it doesn't know at configure time, or because it might output multiple schemas. This plugin should be able to handle these scenarios.

Just skip the schema validation if it's null.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -0,0 +1,223 @@
...
+      }
+    }
+
+    configurer.getStageConfigurer().setOutputSchema(inputSchema);
+  }
+
+  @Override
+  public void prepareRun(BatchSinkContext context) {
+    config.validate();
+    Schema inputSchema = Objects.requireNonNull(context.getInputSchema());
```
albertshau: it's ok for the input schema to be null. Also, currently this method just returns whatever getInputSchema() returned in the configure method, so there isn't a need to validate it again here.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -0,0 +1,223 @@
...
+        );
+      }
+    } catch (IOException e) {
+      throw new InvalidStageException("Failed to connect to Bigtable", e);
+    }
+  }
+
+  private void emitLineage(BatchSinkContext context) {
+    Schema inputSchema = context.getInputSchema();
+    LineageRecorder lineageRecorder = new LineageRecorder(context, config.referenceName);
```
albertshau, line 204: should only do this if the input schema is not null

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -0,0 +1,223 @@
...
+        TableName tableName = TableName.valueOf(config.table);
+        if (admin.tableExists(tableName)) {
+          validateExistingTable(connection, tableName);
+        }
+      } catch (IOException e) {
+        throw new InvalidStageException("Failed to connect to Bigtable", e);
+      }
+    }
+
+    configurer.getStageConfigurer().setOutputSchema(inputSchema);
```
albertshau: there's no need to call this, sinks don't need to set an output schema.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
mlozbin-cybervisiontech: yes, that was fixed in source. Missed this update when moved fixes to sink part. Thanks.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
mlozbin-cybervisiontech: Changed logic to allow key to be of any supported type.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
mlozbin-cybervisiontech: Field name is provided in the caller method. I wrapped convertFieldValueToBytes to handle possible ClassCastExceptions.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
mlozbin-cybervisiontech: Thanks. Looks like some guide about validation best-practices (where and what to validate). I thought "configurePipeline" method is only called when starting the pipeline, is it right? If so, how can we validate input schema for all sink plugins?

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
albertshau: configurePipeline is called when the pipeline is deployed, prepareRun is called at the start of a pipeline run.

Plugins should validate schemas when they are known and constant. Null means the schema was not known at configure time, or it could be multiple schemas. For a schema-less storage system like Bigtable, the plugin should handle situations where it might get records with different schemas. It is fine as long as the row key is always there.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -92,10 +91,9 @@ public void configurePipeline(PipelineConfigurer configurer) {
     super.configurePipeline(configurer);
     config.validate();
     Schema inputSchema = configurer.getStageConfigurer().getInputSchema();
-    if (null == inputSchema) {
-      throw new InvalidStageException("Input schema was not provided");
+    if (null != inputSchema) {
```
albertshau: nit: we usually do null checks with the variable first  (inputSchema != null). Let's change the order to make it consistent with the rest of the projects. Same for the other places

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
```diff
@@ -41,32 +42,51 @@ public RecordToHBaseMutationTransformer(String keyAlias, String columnFamily) {
...
+  private static byte[] convertFieldValueToBytes(@Nullable Object value, Schema fieldSchema) {
     if (fieldSchema.isNullable()) {
       if (null == value) {
         return null;
       }
       return convertFieldValueToBytes(value, fieldSchema.getNonNullable());
     }
-    switch (fieldSchema.getType()) {
-      case NULL:
+    if (null == value) {
```
albertshau: don't think you need this if block, StructuredRecord.Builder will enforce that nulls cannot be set for a non-nullable field.

albertshau (commented):
minor comments, otherwise lgtm. Please fix, squash, and I'll merge.

Let's also be sure to add an integration test after this is merged.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
mlozbin-cybervisiontech: fixed

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
mlozbin-cybervisiontech: fixed

mlozbin-cybervisiontech:
@albertshau fixed and squashed

mlozbin-cybervisiontech:
@albertshau fixed and squashed

PR #127, opened 2019-07-12
CDAP-15627: Added support of clustering in BigQuery Sink plugin.

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/CDAP-15627

In scope of this PR:
* Added support of clustering in BigQuery Sink plugin.

### DIFF
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -53,6 +53,11 @@ is ignored if the table already exists.
 * When this is set to true, table will be created with required partition filter. 
 * When this is set to false, table will be created without required partition filter.
 
+**Clustering Order**: List of fields that determines the sort order of the data. Fields must be of type 
+INT, LONG, STRING, DATE, TIMESTAMP, BOOLEAN or DECIMAL. Tables cannot be clustered on more than 4 fields.
+ This value is only used when the BigQuery table is automatically created and ignored if the table 
+ already exists.
+
 **Update Table Schema**: Whether the BigQuery table schema should be modified 
 when it does not match the schema expected by the pipeline. 
 * When this is set to false, any mismatches between the schema expected by the pipeline 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -87,6 +87,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
     BigQuery bigQuery = GCPUtils.getBigQuery(project, credentials);
     baseConfiguration = getBaseConfiguration();
     String bucket = configureBucket();
+    configureTable();
     if (!context.isPreviewEnabled()) {
       BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(project, credentials), config.getDataset(), bucket);
     }
@@ -194,6 +195,9 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
     }
     baseConfiguration.setBoolean(BigQueryConstants.CONFIG_REQUIRE_PARTITION_FILTER,
                                  getConfig().isPartitionFilterRequired());
+    if (getConfig().getClusteringOrder() != null) {
+      baseConfiguration.set(BigQueryConstants.CONFIG_CLUSTERING_ORDER, getConfig().getClusteringOrder());
+    }
     return baseConfiguration;
   }
 
@@ -226,6 +230,14 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
     baseConfiguration.setBoolean("fs.gs.impl.disable.cache", true);
     baseConfiguration.setBoolean("fs.gs.metadata.cache.enable", false);
     return bucket;
+  }
+
+  private void configureTable() {
+    AbstractBigQuerySinkConfig config = getConfig();
+    Table table = BigQueryUtil.getBigQueryTable(config.getProject(), config.getDataset(),
+                                                config.getTable(),
+                                                config.getServiceAccountFilePath());
+    baseConfiguration.setBoolean(BigQueryConstants.CONFIG_DESTINATION_TABLE_EXISTS, table != null);
   }
 
   /**
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -22,6 +22,7 @@ import com.google.api.client.util.BackOff;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.Sleeper;
 import com.google.api.services.bigquery.Bigquery;
+import com.google.api.services.bigquery.model.Clustering;
 import com.google.api.services.bigquery.model.Dataset;
 import com.google.api.services.bigquery.model.EncryptionConfiguration;
 import com.google.api.services.bigquery.model.ErrorProto;
@@ -65,6 +66,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
@@ -122,11 +124,16 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       LOG.debug("Partition Field: '{}'", partitionByField);
       boolean requirePartitionFilter = conf.getBoolean(BigQueryConstants.CONFIG_REQUIRE_PARTITION_FILTER, false);
       LOG.debug("Require partition filter: '{}'", requirePartitionFilter);
+      String clusteringOrder = conf.get(BigQueryConstants.CONFIG_CLUSTERING_ORDER, null);
+      List<String> clusteringOrderList = Arrays.stream(
+        clusteringOrder != null ? clusteringOrder.split(",") : new String[0]).map(String::trim)
+        .collect(Collectors.toList());
+      boolean tableExists = conf.getBoolean(BigQueryConstants.CONFIG_DESTINATION_TABLE_EXISTS, false);
 
       try {
         importFromGcs(destProjectId, destTable, destSchema.orElse(null), kmsKeyName, outputFileFormat,
                       writeDisposition, sourceUris, allowSchemaRelaxation, createPartitionedTable, partitionByField,
-                      requirePartitionFilter);
+                      requirePartitionFilter, clusteringOrderList, tableExists);
       } catch (InterruptedException e) {
         throw new IOException("Failed to import GCS into BigQuery", e);
       }
@@ -147,7 +154,8 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
     private void importFromGcs(String projectId, TableReference tableRef, @Nullable TableSchema schema,
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
                                List<String> gcsPaths, boolean allowSchemaRelaxation, boolean createPartitionedTable,
-                               @Nullable String partitionByField, boolean requirePartitionFilter)
+                               @Nullable String partitionByField, boolean requirePartitionFilter,
+                               List<String> clusteringOrderList, boolean tableExists)
       throws IOException, InterruptedException {
       LOG.info("Importing into table '{}' from {} paths; path[0] is '{}'; awaitCompletion: {}",
                BigQueryStrings.toString(tableRef), gcsPaths.size(), gcsPaths.isEmpty() ? "(empty)" : gcsPaths.get(0),
@@ -160,7 +168,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       loadConfig.setSourceUris(gcsPaths);
       loadConfig.setDestinationTable(tableRef);
       loadConfig.setWriteDisposition(writeDisposition);
-      if (createPartitionedTable) {
+      if (!tableExists && createPartitionedTable) {
         TimePartitioning timePartitioning = new TimePartitioning();
         timePartitioning.setType("DAY");
         if (partitionByField != null) {
@@ -168,6 +176,11 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
         }
         timePartitioning.setRequirePartitionFilter(requirePartitionFilter);
         loadConfig.setTimePartitioning(timePartitioning);
+        if (!clusteringOrderList.isEmpty()) {
+          Clustering clustering = new Clustering();
+          clustering.setFields(clusteringOrderList);
+          loadConfig.setClustering(clustering);
+        }
       }
 
       if (!Strings.isNullOrEmpty(kmsKeyName)) {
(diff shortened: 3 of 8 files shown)


AntonGensitskiy:

nitinmotgi (changes requested):
Shouldn't be a text box we should add a ordering list. Let's not make users type please.

albertshau:
@nitinmotgi what do you mean by an ordering list? The csv widget? Users would still have to type the field names though.

I see in https://docs.cdap.io/cdap/6.0.0/en/developer-manual/pipelines/developing-plugins/presentation-plugins.html#H2651 that there's an input-field-selector, but not one for selecting multiple input fields. @ajainarayanan is there a widget for this that isn't documented?

docs/BigQueryTable-batchsink.md
```diff
@@ -53,6 +53,11 @@ is ignored if the table already exists.
 * When this is set to true, table will be created with required partition filter. 
 * When this is set to false, table will be created without required partition filter.
 
+**Clustering order**: Comma separated list of fields that determines the sort order of the data.
```
albertshau: Clustering Order

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -228,6 +232,18 @@ private String configureBucket() {
     return bucket;
   }
 
+  private void configureTable() {
+    AbstractBigQuerySinkConfig config = getConfig();
+    if (!(config instanceof BigQuerySinkConfig)) {
```
albertshau: changing behavior based on the type of config is not a good pattern. Instead, this can be a protected method that only the sink overrides.

albertshau: I'm also not sure I understand why this is required. Is the sink not already setting these properties? How is it currently working if it is not?

docs/BigQueryTable-batchsink.md
albertshau: Should use the csv widget so that users don't need to worry about a delimiter.

albertshau: Should document any restrictions, such as the supported data types and if there is a max number of fields. Are they allowed to be nullable?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -122,11 +124,13 @@ public void commitJob(JobContext jobContext) throws IOException {
       LOG.debug("Partition Field: '{}'", partitionByField);
       boolean requirePartitionFilter = conf.getBoolean(BigQueryConstants.CONFIG_REQUIRE_PARTITION_FILTER, false);
       LOG.debug("Require partition filter: '{}'", requirePartitionFilter);
+      String clusteringOrder = conf.get(BigQueryConstants.CONFIG_CLUSTERING_ORDER, null);
+      boolean tabelExists = conf.getBoolean(BigQueryConstants.CONFIG_DESTINATION_TABLE_EXISTS, false);
```
albertshau: typo: tabel -> table

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -147,8 +151,8 @@ public void abortJob(JobContext context, JobStatus.State state) throws IOExcepti
     private void importFromGcs(String projectId, TableReference tableRef, @Nullable TableSchema schema,
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
                                List<String> gcsPaths, boolean allowSchemaRelaxation, boolean createPartitionedTable,
-                               @Nullable String partitionByField, boolean requirePartitionFilter)
-      throws IOException, InterruptedException {
+                               @Nullable String partitionByField, boolean requirePartitionFilter,
+                               String clusteringOrder, boolean tabelExists) throws IOException, InterruptedException {
```
albertshau: if something needs to be parsed, it's better to do it right away when it is read. So this should take a List instead of a String, with the other method reading the value from the Configuration and splitting it.

docs/BigQueryTable-batchsink.md
albertshau: Also, should make it clear that this is only used when the BigQuery table is automatically created. It is ignored if the table already exists.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,44 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+    if (columnsNames.size() > MAX_NUMBER_OF_COLUMNS) {
+      throw new InvalidConfigPropertyException(
+        String.format("Expected no more than '%d' clustering fields, found '%d'.", MAX_NUMBER_OF_COLUMNS,
+                      columnsNames.size()), NAME_CLUSTERING_ORDER);
+    }
+    for (String column : columnsNames) {
+      Schema.Field field = schema.getField(column);
+      if (field == null) {
+        throw new InvalidConfigPropertyException(
+          String.format("Clustering column '%s' is missing from the table schema", column),
```
albertshau, line 198: is this the table schema or the input schema?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,44 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+          NAME_CLUSTERING_ORDER);
+      }
+      Schema.Type type  = field.getSchema().isNullable()
```
albertshau: remove extra space after 'type'

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,44 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+        : field.getSchema().getType();
+      Schema.LogicalType logicalType = field.getSchema().isNullable()
+        ? field.getSchema().getNonNullable().getLogicalType()
+        : field.getSchema().getLogicalType();
+      if (!BigQueryUtil.SUPPORTED_CLUSTERING_TYPES.contains(type)) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' has type '%s', which is not supported for clustering.", column, type),
+          NAME_CLUSTERING_ORDER);
+      }
+      if ((Schema.Type.INT.equals(type) && !Schema.LogicalType.DATE.equals(logicalType))
```
albertshau: don't need to check the type here, just the logical type.

ajainarayanan:
@albertshau @nitinmotgi We don't have a way to allow users to select multiple fields from input schema. Right now we only allow users to select one field (input-field-selector). We would need to build this component if we need it.

docs/BigQueryTable-batchsink.md
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
AntonGensitskiy: Output schema will be used if it not null. In other case - will be used input schema.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
AntonGensitskiy: For example: 
we have types that didn't contained Logical part - Schema.Type.STRING, Schema.Type.BOOLEAN. For this type is not necessary check LogicalType.
Other case Schema.Type.BYTES with Schema.LogicalType.DECIMAL that related to BQ type NUMBER. In this case need check Schema.Type.BYTES has Schema.LogicalType.DECIMAL and this check specific only for Schema.Type.BYTES.
If Schema.Type check will be remove system can't correct validate avalible types for clustering fields.

AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
AntonGensitskiy: No this property not setting. Current property need for set clustering configuration only when new table will be created.
Example: destination table exists and not support clustering, in this case if clustering propertyes be set, system will get exeption during connection to BQ table.

nitinmotgi:
My point is we should not have them to type input fields. We should have a
widget that allows them to select and order

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
albertshau: let's say 'input schema' instead of 'table schema' then, since it's coming from the pipeline and not from bigquery.

albertshau:
@nitinmotgi @ajainarayanan https://issues.cask.co/browse/CDAP-15646. It's not possible today, the plugin can change when there is support

docs/BigQueryTable-batchsink.md
```diff
@@ -53,6 +53,11 @@ is ignored if the table already exists.
 * When this is set to true, table will be created with required partition filter. 
 * When this is set to false, table will be created without required partition filter.
 
+**Clustering Order**: List of fields that determines the sort order of the data. Fields must be of type 
+INT, LONG, STRING, DATE, TIMESTAMP, BOOLEAN or DECIMAL. Maximum number of fields for clustering order
```
albertshau: change "Maximum number of fields for clustering order should not be more that 4" to:

"Tables cannot be clustered on more than 4 fields."

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
albertshau: The check that happens above on the physical type should remain. What I meant is that this if condition does not need to check the physical type, it only needs to check the logical type. This is because a logical type always has the same physical type. This:

```
Schema.Type.INT == type && Schema.LogicalType.DATE == logicalType
```

is logically equivalent to:

```
Schema.LogicalType.DATE == logicalType
```

albertshau (commented):
a couple small comments, otherwise lgtm. Please fix and squash, then I'll merge.

AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,43 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+        : field.getSchema().getType();
+      Schema.LogicalType logicalType = field.getSchema().isNullable()
+        ? field.getSchema().getNonNullable().getLogicalType()
+        : field.getSchema().getLogicalType();
+      if (!BigQueryUtil.SUPPORTED_CLUSTERING_TYPES.contains(type)) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' has type '%s', which is not supported for clustering.", column, type),
+          NAME_CLUSTERING_ORDER);
+      }
+      if ((Schema.Type.INT == type && logicalType != null && Schema.LogicalType.DATE != logicalType)
```
albertshau: this still hasn't been fixed. This is equivalent to:

```
if (logicalType != Schema.LogicalType.DATE || logicalType != Schema.LogicalType.TIMESTAMP_MICROS || logicalType != Schema.LogicalType.DECIMAL) {
  ...
}
```

albertshau: this still hasn't been fixed. It is logically equivalent to:

```
if (logicalType != Schema.LogicalType.DATE && 
      logicalType != Schema.LogicalType.TIMESTAMP_MICROS &&
      logicalType != Schema.LogicalType.DECIMAL) {
  ...
} 
```

albertshau: TIMESTAMP_MILLIS should also be supported, even though the UI currently doesn't display that as a 'timestamp', it's still a supported type that can appear in a custom transform.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,42 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+        : field.getSchema().getLogicalType();
+      if (!BigQueryUtil.SUPPORTED_CLUSTERING_TYPES.contains(type)) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' has type '%s', which is not supported for clustering.", column, type),
+          NAME_CLUSTERING_ORDER);
+      }
+      if (logicalType != Schema.LogicalType.DATE && logicalType != Schema.LogicalType.TIMESTAMP_MICROS
+        && logicalType != Schema.LogicalType.TIMESTAMP_MILLIS && logicalType !=  Schema.LogicalType.DECIMAL) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' has type '%s', which is not supported for clustering.", column, type),
```
albertshau: one more thing -- this should use logicalType.getToken(), as it's a problem with the logical type and not the physical type.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -172,4 +179,42 @@ private void validateColumnForPartition(@Nullable String columnName, @Nullable S
...
+      Schema.LogicalType logicalType = field.getSchema().isNullable()
+        ? field.getSchema().getNonNullable().getLogicalType()
+        : field.getSchema().getLogicalType();
+      if (!BigQueryUtil.SUPPORTED_CLUSTERING_TYPES.contains(type)) {
+        throw new InvalidConfigPropertyException(
+          String.format("Field '%s' has type '%s', which is not supported for clustering.", column, type),
+          NAME_CLUSTERING_ORDER);
+      }
+      if (logicalType != Schema.LogicalType.DATE && logicalType != Schema.LogicalType.TIMESTAMP_MICROS
+        && logicalType != Schema.LogicalType.TIMESTAMP_MILLIS && logicalType !=  Schema.LogicalType.DECIMAL) {
```
albertshau: remove extra space in front of Schema.LogicalType.DECIMAL

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

albertshau (approved):
lgtm

PR #129, opened 2019-07-18

mlozbin-cybervisiontech:
WIKI: https://wiki.cask.co/display/CE/Bigtable+Source https://wiki.cask.co/display/CE/Bigtable+Sink

In scope of this PR:
- implemented column mappings

### DIFF
--- a/docs/Bigtable-batchsink.md
+++ b/docs/Bigtable-batchsink.md
@@ -32,9 +32,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 **Key Alias**: Name of the field for row key.
 
 **Column Mappings**: Mappings from record field to Bigtable column name. 
-Column names must be formatted as <family>:<qualifier>.
+Column names must be formatted as 'family:qualifier'.
 
 **Bigtable Options**: Additional connection properties for Bigtable.
+Full list of allowed properties: https://cloud.google.com/bigtable/docs/hbase-client/javadoc/constant-values.
 
 **Service Account File Path**: Path on the local file system of the service account key used for
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
--- a/docs/Bigtable-batchsink.md
+++ b/docs/Bigtable-batchsink.md
@@ -31,7 +31,7 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
-**Column Family**: Column family to use for all inserted rows.
+**Column Mappings**: Mappings from record field to Bigtable column name. Column name format: <family>:<qualifier>.
 
 **Service Account File Path**: Path on the local file system of the service account key used for
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
--- a/docs/Bigtable-batchsink.md
+++ b/docs/Bigtable-batchsink.md
@@ -31,7 +31,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
-**Column Mappings**: Mappings from record field to Bigtable column name. Column name format: <family>:<qualifier>.
+**Column Mappings**: Mappings from record field to Bigtable column name. 
+Column names must be formatted as <family>:<qualifier>.
+
+**Bigtable Options**: Additional connection properties for Bigtable.
 
 **Service Account File Path**: Path on the local file system of the service account key used for
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
--- a/docs/Bigtable-batchsource.md
+++ b/docs/Bigtable-batchsource.md
@@ -31,6 +31,8 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
+**Column Mappings**: Mappings from Bigtable column name to record field. Column name format: <family>:<qualifier>.
+
 **Scan Row Start**: Scan start row.
 
 **Scan Row Stop**: Scan stop row.
--- a/docs/Bigtable-batchsource.md
+++ b/docs/Bigtable-batchsource.md
@@ -31,7 +31,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
-**Column Mappings**: Mappings from Bigtable column name to record field. Column name format: <family>:<qualifier>.
+**Column Mappings**: Mappings from Bigtable column name to record field. 
+Column names must be formatted as <family>:<qualifier>.
+
+**Bigtable Options**: Additional connection properties for Bigtable.
 
 **Scan Row Start**: Scan start row.
 
--- a/docs/Bigtable-batchsource.md
+++ b/docs/Bigtable-batchsource.md
@@ -32,9 +32,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 **Key Alias**: Name of the field for row key.
 
 **Column Mappings**: Mappings from Bigtable column name to record field. 
-Column names must be formatted as <family>:<qualifier>.
+Column names must be formatted as 'family:qualifier'.
 
-**Bigtable Options**: Additional connection properties for Bigtable.
+**Bigtable Options**: Additional connection properties for Bigtable. 
+Full list of allowed properties: https://cloud.google.com/bigtable/docs/hbase-client/javadoc/constant-values.
 
 **Scan Row Start**: Scan start row.
 
@@ -49,8 +50,7 @@ authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 
 **Schema**: Specifies the schema that has to be output. 
-Only columns defined in schema will be included into output record. Field name should be in form 
-"<family>__<column>".
+Only columns defined in schema will be included into output record.
 
 **onError**: Strategy used to handle errors during transformation of a text entry to record. Possible values are:
 - **Skip error** - Ignores erroneous records.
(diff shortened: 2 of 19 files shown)


docs/Bigtable-batchsink.md
```diff
@@ -31,7 +31,7 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
-**Column Family**: Column family to use for all inserted rows.
+**Column Mappings**: Mappings from record field to Bigtable column name. Column name format: <family>:<qualifier>.
```
albertshau: The ':' might be a little confusing since it also separates family and qualifier. 

Column names must be formatted as <family>:<qualifier>.

docs/Bigtable-batchsource.md
```diff
@@ -31,6 +31,8 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
+**Column Mappings**: Mappings from Bigtable column name to record field. Column name format: <family>:<qualifier>.
```
albertshau: The ':' might be a little confusing since it also separates family and qualifier.

Column names must be formatted as :.

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
```diff
@@ -86,6 +89,11 @@ public void validate() {
     }
   }
 
+  @Nullable
```
albertshau: its best to avoid null collections when possible. Can this return an empty map?

src/main/java/io/cdap/plugin/gcp/bigtable/sink/RecordToHBaseMutationTransformer.java
```diff
@@ -53,11 +55,14 @@ public Mutation transform(StructuredRecord record) {
                                                         keyAlias));
     }
     Put put = new Put(rowKeyBytes);
-    byte[] columnFamilyBytes = Bytes.toBytes(columnFamily);
     for (Schema.Field field : fields) {
       String fieldName = field.getName();
+      if (fieldName.equals(keyAlias)) {
+        continue;
+      }
+      HBaseColumn column = HBaseColumn.fromFullName(columnMappings.get(fieldName));
```
albertshau: rather than do this for every record, it would make more sense to have columnMappings be a `Map<String, HBaseColumn>`.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -88,6 +88,7 @@ public void configurePipeline(PipelineConfigurer configurer) {
 
   @Override
   public void prepareRun(BatchSourceContext context) {
+    LOG.warn("Mappings: {}", config.columnMappings);
```
albertshau: left in from debugging?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
```diff
@@ -146,6 +133,21 @@ public void transform(KeyValue<ImmutableBytesWritable, Result> input, Emitter<St
     }
   }
 
+  private Configuration getConfiguration() {
+    try {
+      Configuration conf = HBaseConfiguration.create();
+      BigtableConfiguration.configure(conf, config.getProject(), config.instance);
+      conf.set(BigtableOptionsFactory.BIGTABLE_RPC_TIMEOUT_MS_KEY, "60000");
```
albertshau: Didn't notice this before, but this seems like it should be a configuration property instead of hardcoded. Are there other common options that should be configurable? This can be done in a separate PR.

docs/Bigtable-batchsink.md
albertshau: This can be annoying if the user is writing a bunch of fields that all go to a single family. I think it's better if we allow the user to specify either a family or mapping.

docs/Bigtable-batchsource.md
albertshau: I think it makes sense for the source to mirror the sink and also have a family property. The user needs to specify either a family or a mapping. Can be done in a separate PR.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -173,6 +199,27 @@ public void validate() {
...
+      if (columnMappings != null) {
+        Set<String> mappedFieldNames = Sets.newHashSet(columnMappings.values());
+        Set<String> schemaFieldNames = fields.stream()
+          .map(Schema.Field::getName)
+          .filter(name -> !name.equals(keyAlias))
+          .collect(Collectors.toSet());
+        Sets.SetView<String> nonDescribedFields = Sets.difference(schemaFieldNames, mappedFieldNames);
+        if (!nonDescribedFields.isEmpty()) {
+          String nonDescribedFieldsString = String.join(", ", nonDescribedFields);
+          String errorMessage = String.format("Some mapped columns are not described in schema: %s. ",
```
albertshau: Isn't this the reverse? There are schema fields that do not have a mapping?

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
```diff
@@ -173,6 +199,27 @@ public void validate() {
...
+                                              nonDescribedFieldsString);
+          throw new InvalidConfigPropertyException(errorMessage, SCHEMA);
+        }
+        Sets.SetView<String> nonMappedColumns = Sets.difference(mappedFieldNames, schemaFieldNames);
+        if (!nonMappedColumns.isEmpty()) {
+          String nonMappedColumnsString = String.join(", ", nonDescribedFields);
+          String errorMessage = String.format("Some schema fields do not have corresponding column mappings: %s. ",
```
albertshau, line 224: I don't think we need to error here, it just means there are extra mapping that will never apply. Maybe just log a warning in this case.

docs/Bigtable-batchsource.md
albertshau: This also means the doc for the schema needs to be updated, since it talks about the family being in the field name.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
mlozbin-cybervisiontech: actually, there are a lot of advanced connection properties. I'm not sure end user needs some of them (see https://cloud.google.com/bigtable/docs/hbase-client/javadoc/com/google/cloud/bigtable/hbase/BigtableOptionsFactory). Maybe I should just remove this because timeouts are currently experimental. What do you think?

docs/Bigtable-batchsource.md
mlozbin-cybervisiontech: As I said on last weekly call, I think it could be confusing for user to have both columnFamily and mappings without some select on UI that will affect what properties to set. For some reason I understood your answer as keep mappings for now until UI will support such conditional properties. So should this be delayed or is it better to have both options now?

albertshau: the UI is currently building this out (design at https://wiki.cask.co/display/CE/Dynamic+plugin+property+filters+for+CDAP+Pipelines).

We can revisit this in a future PR once it's been added and only focus on the column mapping for now. Please open a jira so it's not lost.

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSource.java
albertshau: let's remove the timeout, but instead add a 'Bigtable Options' property that is just a map of key-values. Advanced users can add whatever they want there, which we will set in the Configuration.

docs/Bigtable-batchsource.md
mlozbin-cybervisiontech: created Jira issue https://issues.cask.co/browse/CDAP-15662

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
mlozbin-cybervisiontech: I can hardly imagine the use case when it was intended by the user. Are you sure I should replace it with warning? User may miss some fields.

albertshau: In general, I think it's better to try not to make assumptions about user intent, as some users will end up trying to use it differently than we think.

For example, this can happen if the mapping is a macro that is a common value shared across multiple pipelines. The value could be a superset of everything needed, in order to managing one property per pipeline. Or maybe they want to debug a pipeline so they have cloned it and they want to delete a bunch of fields from the output schema, but leave the mapping because they intend to add most of them back after debugging.

mlozbin-cybervisiontech:
@albertshau addressed all comments

docs/Bigtable-batchsink.md
```diff
@@ -31,7 +31,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Key Alias**: Name of the field for row key.
 
-**Column Family**: Column family to use for all inserted rows.
+**Column Mappings**: Mappings from record field to Bigtable column name. 
+Column names must be formatted as <family>:<qualifier>.
+
+**Bigtable Options**: Additional connection properties for Bigtable.
```
albertshau, line 37: Can you include a link to the page that specifies all the options

src/main/java/io/cdap/plugin/gcp/bigtable/source/BigtableSourceConfig.java
albertshau: this still needs to be addressed.

mlozbin-cybervisiontech:
@albertshau addressed last comments and added unit tests

albertshau (approved):
lgtm

PR #130, opened 2019-08-05
CDAP-15712 GCS Create and Delete actions are broken

vladhlinskiy:
1. Fixed wrong call of `getPaths` method
2. Refactored validation flow

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
@@ -131,7 +131,7 @@ public final class GCSBucketCreate extends Action {
   /**
    * Config for the plugin.
    */
-  public final class Config extends GCPConfig {
+  public static final class Config extends GCPConfig {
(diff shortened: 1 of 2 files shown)


vladhlinskiy:
@albertshau could you, please, take a look?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -151,11 +151,10 @@ public boolean getFailIfExists() {
     }
 
     void validate() {
-      if (containsMacro("paths")) {
-        return;
-      }
-      for (String path : config.getPaths()) {
```
albertshau, line 157: it's not good that it was possible to call 'config.getPaths()'. Can you also make the Config class static so that this is impossible

vladhlinskiy: Thanks for comment. Done.

albertshau (approved):
lgtm

PR #131, opened 2019-08-08
CDAP-15645: Added possibility choose operation type - Insert, Update, Upsert.

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/CDAP-15645

In scope of this PR:

* Added possibility choose operation type (Insert, Update, Upsert) in BigQuery Sink plugin.
* For operation type Update and Upsert was added possibility choose list of fields for relation between tables.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -30,11 +30,20 @@ import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.JobConfiguration;
 import com.google.api.services.bigquery.model.JobConfigurationLoad;
 import com.google.api.services.bigquery.model.JobReference;
+import com.google.api.services.bigquery.model.Table;
 import com.google.api.services.bigquery.model.TableFieldSchema;
 import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableSchema;
 import com.google.api.services.bigquery.model.TimePartitioning;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.FieldList;
+import com.google.cloud.bigquery.JobId;
 import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.TableId;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import com.google.cloud.hadoop.io.bigquery.BigQueryFactory;
 import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
@@ -57,15 +66,19 @@ import org.apache.hadoop.mapreduce.JobStatus;
 import org.apache.hadoop.mapreduce.OutputCommitter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.util.Progressable;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.math.BigInteger;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -76,6 +89,10 @@ import javax.annotation.Nullable;
  */
 public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<JsonObject, NullWritable> {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryOutputFormat.class);
+
+  private static final String UPDATE_QUERY = "UPDATE %s T SET %s FROM %s S WHERE %s";
+  private static final String UPSERT_QUERY = "MERGE %s T USING %s S ON %s WHEN MATCHED THEN UPDATE SET %s " +
+    "WHEN NOT MATCHED THEN INSERT (%s) VALUES(%s)";
 
   @Override
   public OutputCommitter createCommitter(TaskAttemptContext context) throws IOException {
@@ -89,6 +106,13 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
    */
   public static class BigQueryOutputCommitter extends ForwardingBigQueryFileOutputCommitter {
     private BigQueryHelper bigQueryHelper;
+
+    private Operation operation;
+    private TableReference temporaryTableReference;
+    private List<String> tableKeyList;
+    private List<String> tableFieldsList;
+
+    private boolean allowSchemaRelaxation;
 
     BigQueryOutputCommitter(TaskAttemptContext context, OutputCommitter delegate) throws IOException {
       super(context, delegate);
@@ -116,7 +140,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       BigQueryFileFormat outputFileFormat = BigQueryOutputConfiguration.getFileFormat(conf);
       List<String> sourceUris = getOutputFileURIs();
 
-      boolean allowSchemaRelaxation = conf.getBoolean(BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION, false);
+      allowSchemaRelaxation = conf.getBoolean(BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION, false);
       LOG.debug("Allow schema relaxation: '{}'", allowSchemaRelaxation);
       boolean createPartitionedTable = conf.getBoolean(BigQueryConstants.CONFIG_CREATE_PARTITIONED_TABLE, false);
       LOG.debug("Create Partitioned Table: '{}'", createPartitionedTable);
@@ -124,16 +148,26 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       LOG.debug("Partition Field: '{}'", partitionByField);
       boolean requirePartitionFilter = conf.getBoolean(BigQueryConstants.CONFIG_REQUIRE_PARTITION_FILTER, false);
       LOG.debug("Require partition filter: '{}'", requirePartitionFilter);
+      operation = Operation.valueOf(conf.get(BigQueryConstants.CONFIG_OPERATION));
       String clusteringOrder = conf.get(BigQueryConstants.CONFIG_CLUSTERING_ORDER, null);
       List<String> clusteringOrderList = Arrays.stream(
         clusteringOrder != null ? clusteringOrder.split(",") : new String[0]).map(String::trim)
         .collect(Collectors.toList());
+      String tableKey = conf.get(BigQueryConstants.CONFIG_TABLE_KEY, null);
+      tableKeyList = Arrays.stream(tableKey != null ? tableKey.split(",") : new String[0]).map(String::trim)
+        .collect(Collectors.toList());
+      String tableFields = conf.get(BigQueryConstants.CONFIG_TABLE_FIELDS, null);
+      tableFieldsList = Arrays.stream(tableFields != null ? tableFields.split(",") : new String[0])
+        .map(String::trim).collect(Collectors.toList());
       boolean tableExists = conf.getBoolean(BigQueryConstants.CONFIG_DESTINATION_TABLE_EXISTS, false);
 
       try {
         importFromGcs(destProjectId, destTable, destSchema.orElse(null), kmsKeyName, outputFileFormat,
-                      writeDisposition, sourceUris, allowSchemaRelaxation, createPartitionedTable, partitionByField,
+                      writeDisposition, sourceUris, createPartitionedTable, partitionByField,
                       requirePartitionFilter, clusteringOrderList, tableExists);
+        if (temporaryTableReference != null) {
+          operationAction(destTable);
+        }
       } catch (InterruptedException e) {
         throw new IOException("Failed to import GCS into BigQuery", e);
       }
@@ -153,7 +187,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
      */
     private void importFromGcs(String projectId, TableReference tableRef, @Nullable TableSchema schema,
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
-                               List<String> gcsPaths, boolean allowSchemaRelaxation, boolean createPartitionedTable,
+                               List<String> gcsPaths, boolean createPartitionedTable,
                                @Nullable String partitionByField, boolean requirePartitionFilter,
                                List<String> clusteringOrderList, boolean tableExists)
       throws IOException, InterruptedException {
@@ -166,7 +200,6 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Jso
       loadConfig.setSchema(schema);
       loadConfig.setSourceFormat(sourceFormat.getFormatIdentifier());
       loadConfig.setSourceUris(gcsPaths);
-      loadConfig.setDestinationTable(tableRef);
       loadConfig.setWriteDisposition(writeDisposition);
       if (!tableExists && createPartitionedTable) {
         TimePartitioning timePartitioning = new TimePartitioning();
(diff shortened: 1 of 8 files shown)


AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -325,11 +370,50 @@ private static TableReference getTableReference(Configuration conf) throws IOExc
...
     private static TableSchema createTableSchemaFromFields(String fieldsJson) throws IOException {
       List<TableFieldSchema> fields = new ArrayList<>();
       JsonParser parser = JacksonFactory.getDefaultInstance().createJsonParser(fieldsJson);
       parser.parseArrayAndClose(fields, TableFieldSchema.class);
       return new TableSchema().setFields(fields);
     }
+
+    private boolean isTableEmpty(TableReference tableRef) throws IOException {
+      Table table = bigQueryHelper.getTable(tableRef);
+      return table.getNumRows().compareTo(BigInteger.ZERO) <= 0 && table.getNumBytes() == 0
```
albertshau, line 456: does getNumRows() result in a BQ query or is it just metadata? If it has to run through all the data to do a count then it could be more expensive than just running the update query.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -183,14 +209,25 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
+      if (tableExists && !isTableEmpty(tableRef) && !Operation.INSERT.equals(operation)) {
+        String temporaryTableName = UUID.randomUUID().toString().replaceAll("-", "_");
+        temporaryTableReference = new TableReference()
+          .setDatasetId(tableRef.getDatasetId())
+          .setProjectId(tableRef.getProjectId())
+          .setTableId(temporaryTableName);
+        loadConfig.setDestinationTable(temporaryTableReference);
+      } else {
+        loadConfig.setDestinationTable(tableRef);
+        if (!Strings.isNullOrEmpty(kmsKeyName)) {
```
albertshau: this should be done regardless of the operation type. If they want the final data to be encrypted, we should also encrypt the intermediate data.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -219,6 +256,14 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
 
       // Poll until job is complete.
       waitForJobCompletion(bigQueryHelper.getRawBigquery(), projectId, jobReference);
+
+      if (temporaryTableReference != null && bigQueryHelper.tableExists(temporaryTableReference)) {
+        long expirationMillis = DateTime.now().plusDays(1).getMillis();
+        Table table = bigQueryHelper.getTable(temporaryTableReference).setExpirationTime(expirationMillis);
+        bigQueryHelper.getRawBigquery().tables().update(temporaryTableReference.getProjectId(),
```
albertshau, line 271: is there a way to specify this in the load config rather than updating it later? Otherwise a pipeline failure before this point will cause the temporary table to not have the expiration set.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -218,4 +220,31 @@ private void validateClusteringOrder(@Nullable Schema schema) {
...
+
+  private void validateOperationProperties(@Nullable Schema schema) {
+    if (Arrays.stream(Operation.values()).map(Enum::name).noneMatch(operation.toUpperCase()::equals)) {
+      throw new InvalidConfigPropertyException(String.format("'%s' is incorrect value for field 'Operation'. " +
+                                                               "This field should contain one of the next values: " +
+                                                               "INSERT, UPDATE or UPSERT.", operation), NAME_OPERATION);
+    }
+    if (!Operation.INSERT.equals(getOperation()) && getRelationTableKey() == null) {
+      throw new InvalidConfigPropertyException(
+        "Table key value should be set in case UPDATE or UPSERT operations selected.", NAME_TABLE_KEY);
```
albertshau: Table key must be set if the operation is 'UPDATE' or 'UPSERT'

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -218,4 +220,31 @@ private void validateClusteringOrder(@Nullable Schema schema) {
...
     }
   }
+
+  private void validateOperationProperties(@Nullable Schema schema) {
+    if (Arrays.stream(Operation.values()).map(Enum::name).noneMatch(operation.toUpperCase()::equals)) {
+      throw new InvalidConfigPropertyException(String.format("'%s' is incorrect value for field 'Operation'. " +
+                                                               "This field should contain one of the next values: " +
+                                                               "INSERT, UPDATE or UPSERT.", operation), NAME_OPERATION);
+    }
+    if (!Operation.INSERT.equals(getOperation()) && getRelationTableKey() == null) {
```
albertshau: though it's logically equivalent now, it's safer for this check to be on equality with UPDATE or UPSERT, in case new types are added in the future.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -218,4 +220,31 @@ private void validateClusteringOrder(@Nullable Schema schema) {
...
+    if (schema == null) {
+      return;
+    }
+    List<String> fields = Objects.requireNonNull(schema.getFields()).stream().map(Schema.Field::getName)
+      .collect(Collectors.toList());
+    List<String> keyFields = Arrays.stream(Objects.requireNonNull(getRelationTableKey()).split(","))
+      .map(String::trim).collect(Collectors.toList());
+    if (!Operation.INSERT.equals(getOperation()) &&  keyFields.size() >= fields.size()) {
+      throw new InvalidConfigPropertyException(
+        "Count of Table key value should be less then table columns count.", NAME_TABLE_KEY);
```
albertshau: this is not a useful error message. Should validate that all fields in the table key are also in the schema. If not, the error message should say something like:

Fields x, y, and z are in the table key, but not in the input schema.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -218,4 +220,31 @@ private void validateClusteringOrder(@Nullable Schema schema) {
       }
     }
   }
+
+  private void validateOperationProperties(@Nullable Schema schema) {
+    if (Arrays.stream(Operation.values()).map(Enum::name).noneMatch(operation.toUpperCase()::equals)) {
+      throw new InvalidConfigPropertyException(String.format("'%s' is incorrect value for field 'Operation'. " +
+                                                               "This field should contain one of the next values: " +
+                                                               "INSERT, UPDATE or UPSERT.", operation), NAME_OPERATION);
```
albertshau: these show up to the user as 'Insert', 'Update', or 'Upsert', so we should use that instead of all-caps.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -183,14 +209,25 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
+        loadConfig.setDestinationTable(tableRef);
+        if (!Strings.isNullOrEmpty(kmsKeyName)) {
+          loadConfig.setDestinationEncryptionConfiguration(new EncryptionConfiguration().setKmsKeyName(kmsKeyName));
+        }
 
-      if (allowSchemaRelaxation) {
-        loadConfig.setSchemaUpdateOptions(Arrays.asList(
-          JobInfo.SchemaUpdateOption.ALLOW_FIELD_ADDITION.name(),
-          JobInfo.SchemaUpdateOption.ALLOW_FIELD_RELAXATION.name()));
+        if (allowSchemaRelaxation) {
```
albertshau, line 230: this means schema relaxation only applies if the the operation is 'insert'. There should be a way to update the schema for the other operation types as well.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
nitinmotgi: Also there is cost implication. Scans cost on BigQuery.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
nitinmotgi: We can show with Uppercase as first letter and then still pass All CAPS if a drop down is used.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: getNumRows() is it just metadata.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: Load config does not have possibility to configure expiration time.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: Added possibility update table schema during Update/Upsert operation.

AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -383,6 +394,35 @@ private void operationAction(TableReference tableRef) throws InterruptedExceptio
...
+      List<String> fieldsName = destinationFields.stream().map(Field::getName).collect(Collectors.toList());
+
+      List<Field> newFields = sourceFields.stream().filter(field -> !fieldsName.contains(field.getName()))
+        .collect(Collectors.toList());
+
+      if (!newFields.isEmpty()) {
+        newFields.addAll(destinationFields);
+        Schema newSchema = Schema.of(newFields);
+        bigquery.update(
+          destinationTable.toBuilder().setDefinition(destinationTable.getDefinition().toBuilder()
```
albertshau: this is hard to read, can you break it up into a couple lines

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -383,6 +394,35 @@ private void operationAction(TableReference tableRef) throws InterruptedExceptio
...
+      FieldList sourceFields = sourceTable.getDefinition().getSchema().getFields();
+      tableFieldsList = sourceFields.stream().map(Field::getName).collect(Collectors.toList());
+      FieldList destinationFields = destinationTable.getDefinition().getSchema().getFields();
+
+      List<String> fieldsName = destinationFields.stream().map(Field::getName).collect(Collectors.toList());
+
+      List<Field> newFields = sourceFields.stream().filter(field -> !fieldsName.contains(field.getName()))
+        .collect(Collectors.toList());
+
+      if (!newFields.isEmpty()) {
```
albertshau: table schema can also be updated to change a non-nullable field to a nullable one, which will not satisfy this condition.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: Added possibility change a non-nullable field to a nullable one.

AntonGensitskiy:
@albertshau I added addition change and addressed all your comments, could you, please, take a look.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
       FieldList sourceFields = sourceTable.getDefinition().getSchema().getFields();
       tableFieldsList = sourceFields.stream().map(Field::getName).collect(Collectors.toList());
       FieldList destinationFields = destinationTable.getDefinition().getSchema().getFields();
+      Map<String, Field> destinationFieldMap = destinationFields.stream()
+        .collect(Collectors.toMap(Field::getName, x -> x));
 
-      List<String> fieldsName = destinationFields.stream().map(Field::getName).collect(Collectors.toList());
-
-      List<Field> newFields = sourceFields.stream().filter(field -> !fieldsName.contains(field.getName()))
+      List<Field> newFields = sourceFields.stream().filter(field -> destinationFieldMap.get(field.getName()) == null)
```
albertshau: nit: use containsKey() instead of checking null

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
...
         .collect(Collectors.toList());
+      List<Field> updateFields = new ArrayList<>();
+      sourceFields.stream()
+        .filter(sourceField -> sourceField.getMode() == null
```
albertshau: When is the mode is null? Apparently it's the same as if the mode is there and it's set to nullable? 

can you add a comment?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
...
+          || (sourceField.getMode() != null && sourceField.getMode().equals(Field.Mode.NULLABLE)))
```
albertshau: this can be simplified to `sourceField.getMode() == Field.Mode.NULLABLE`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
...
+          }
+        });
+
+      if (!newFields.isEmpty() || !updateFields.isEmpty()) {
+        List<Field> resultFieldsList = destinationFields.stream()
+          .filter(field ->
+                    updateFields.stream().noneMatch(x -> x.getName().equals(field.getName())))
+          .collect(Collectors.toList());
+        resultFieldsList.addAll(updateFields);
+        resultFieldsList.addAll(newFields);
```
albertshau: isn't this the same as using the sourceFields?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
...
-      List<String> fieldsName = destinationFields.stream().map(Field::getName).collect(Collectors.toList());
-
-      List<Field> newFields = sourceFields.stream().filter(field -> !fieldsName.contains(field.getName()))
+      List<Field> newFields = sourceFields.stream().filter(field -> destinationFieldMap.get(field.getName()) == null)
         .collect(Collectors.toList());
+      List<Field> updateFields = new ArrayList<>();
+      sourceFields.stream()
+        .filter(sourceField -> sourceField.getMode() == null
+          || (sourceField.getMode() != null && sourceField.getMode().equals(Field.Mode.NULLABLE)))
+        .filter(sourceField -> destinationFieldMap.get(sourceField.getName()) != null)
```
albertshau: it would be better for this to be a map to get the destinationField, followed by a filter on null and the field mode, followed by a collect() instead of a forEach.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -406,19 +408,36 @@ private void updateTableSchema(TableReference tableRef) {
...
+        .forEach(sourceField -> {
+          Field destinationField = destinationFieldMap.get(sourceField.getName());
+          if (destinationField.getMode() != null && destinationField.getMode().equals(Field.Mode.REQUIRED)) {
```
albertshau: this can be simplified to `destinationField.getMode() == Field.Mode.REQUIRED`

albertshau:
style comments, otherwise lgtm. Please fix and squash.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: Yes. For BigQuery the mode is null the same as mode set to nullable.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: You are right.
Logic was simplified.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

AntonGensitskiy:

albertshau (approved):
lgtm. 

Can you also open a follow-up PR where we add some examples to the documentation about what happens when the operation is insert vs update vs upsert?

albertshau:
As well as a follow-up PR to make the temporary table name of the form [actual table]_[id], with a description about what it is. 

If the source has similar logic where it uses just an UUID as the temporary table name for partitioned queries, we should also fix that as well.

PR #134, opened 2019-08-13
fix bq in preview

yaojiefeng:
this was fixed in develop in #125 but not in the release branch.
The source will need to create the GCS bucket no matter it is in preview or not. Sink does not matter since preview will not write to the sink


albertshau (approved):
lgtm

PR #138, opened 2019-08-22
CDAP-15736: Added BigQuery Merge Action plugin.

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/CDAP-15736

In scope of this PR:

* Added BigQuery action plugin for merging two BigQuery tables.


AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/actions/BigQueryMergeAction.java
```diff
@@ -0,0 +1,118 @@
...
+    if (config.isAllowSchemaRelaxation()) {
+      updateTableSchema();
+    }
+    String query = generateQuery();
+    setQuery(query);
+    super.run(context);
+    LOG.debug("Complete merge action.");
+  }
+
+  private String generateQuery() {
```
albertshau, line 65: this plugin assumes that the source and destination tables have the same columns, which will not always be true, and which probably is not true in most situations. This plugin needs to go through a design before we implement it. I would imagine the user would need to provide a more generic condition to merge on and select fields from the source need to be updated/inserted in the destination (kind of similar to what is given in the joiner)

@nitinmotgi can you provide more information about the use case? Is it only merging on equality? Do we need this now that there is 'upsert' in the BQ sink and given that there will soon be a generic BigQuery action?

albertshau:
I talked with @nitinmotgi and @bdmogal about this, there isn't a specific type of merge that we're targeting, so any plugin we do would essentially make the user provide the same information they would in a merge query. So we're not going to add this, as users will be able to use the generic BQ action plugin when it's implemented.

PR #140, opened 2019-08-22
Feature/support write disposition

dli357:
Modified the BigQuery Sink (singular and multi) to support WriteDisposition when writing to BigQuery.

Ref: https://cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/WriteDisposition

This modification supports three options:
1. Append - Appends the data to the BigQuery tables.
2. Empty - Only write the data to empty BigQuery tables, otherwise fail.
3. Truncate - Truncate the BigQuery table before writing to it.

Additionally, resolved a bug causing NullPointerExceptions when using the BigQuery Multi Sink.

Tested using CDAP Sandbox and Cloud Data Fusion. Test Cases:
- Single Sink Append
- Single Sink Truncate
- Single Sink Fail to Create
- Multi Sink Append
- Multi Sink Truncate
- Multi Sink Fail to Create

Resolves:
https://issues.cask.co/browse/CDAP-15784
https://issues.cask.co/browse/CDAP-15616

### DIFF
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -36,9 +36,8 @@ It will be automatically created if it does not exist, but will not be automatic
 Temporary data will be deleted after it is loaded into BigQuery. If it is not provided, a unique
 bucket will be created and then deleted after the run finishes.
 
-**Truncate Table**: Whether to truncate the output table before writing to it.
-* True - Will delete all rows in the table before writing to it.
-* False - Will append the new rows to the table.
+**Truncate Table:** Whether or not to truncate the table before writing to it.
+Should only be used with the Insert operation.
 
 **Split Field:** The name of the field that will be used to determine which table to write to.
 
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -36,6 +36,11 @@ It will be automatically created if it does not exist, but will not be automatic
 Temporary data will be deleted after it is loaded into BigQuery. If it is not provided, a unique
 bucket will be created and then deleted after the run finishes.
 
+**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
+* Empty - Job should only write to empty tables, otherwise an error is thrown.
+* Truncate - Job will truncate an existing table before writing to it.
+* Append - Job will append to an existing table.
+
 **Split Field:** The name of the field that will be used to determine which table to write to.
 
 **Update Table Schema**: Whether the BigQuery table schema should be modified 
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -36,10 +36,9 @@ It will be automatically created if it does not exist, but will not be automatic
 Temporary data will be deleted after it is loaded into BigQuery. If it is not provided, a unique
 bucket will be created and then deleted after the run finishes.
 
-**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
-* Empty - Job should only write to empty tables, otherwise an error is thrown.
-* Truncate - Job will truncate an existing table before writing to it.
-* Append - Job will append to an existing table.
+**Truncate Table**: Whether to truncate the output table before writing to it.
+* True - Will delete all rows in the table before writing to it.
+* False - Will append the new rows to the table.
 
 **Split Field:** The name of the field that will be used to determine which table to write to.
 
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -46,10 +46,9 @@ bucket will be created and then deleted after the run finishes.
 will be dropped.
 * Upsert - records that match on Table Key will be updated. Records that do not match will be inserted.
 
-**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
-* Empty - Job should only write to empty tables, otherwise an error is thrown.
-* Truncate - Job will truncate an existing table before writing to it.
-* Append - Job will append to an existing table.
+**Truncate Table**: Whether to truncate the output table before writing to it.
+* True - Will delete all rows in the table before writing to it.
+* False - Will append the new rows to the table.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -46,6 +46,11 @@ bucket will be created and then deleted after the run finishes.
 will be dropped.
 * Upsert - records that match on Table Key will be updated. Records that do not match will be inserted.
 
+**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
+* Empty - Job should only write to empty tables, otherwise an error is thrown.
+* Truncate - Job will truncate an existing table before writing to it.
+* Append - Job will append to an existing table.
+
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
 **Create Partitioned Table**: Whether to create the BigQuery table with time partitioning. This value 
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -46,9 +46,8 @@ bucket will be created and then deleted after the run finishes.
 will be dropped.
 * Upsert - records that match on Table Key will be updated. Records that do not match will be inserted.
 
-**Truncate Table**: Whether to truncate the output table before writing to it.
-* True - Will delete all rows in the table before writing to it.
-* False - Will append the new rows to the table.
+**Truncate Table**: Whether or not to truncate the table before writing to it.
+Should only be used with the Insert operation.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -67,7 +67,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, J
   // UUID for the run. Will be used as bucket name if bucket is not provided.
   // UUID is used since GCS bucket names must be globally unique.
   private final UUID uuid = UUID.randomUUID();
-  private Configuration baseConfiguration;
+  protected Configuration baseConfiguration;
 
   /**
    * Executes main prepare run logic. Child classes cannot override this method,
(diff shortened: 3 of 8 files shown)


docs/BigQueryMultiTable-batchsink.md
```diff
@@ -36,6 +36,11 @@ It will be automatically created if it does not exist, but will not be automatic
 Temporary data will be deleted after it is loaded into BigQuery. If it is not provided, a unique
 bucket will be created and then deleted after the run finishes.
 
+**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
```
nitinmotgi: Can we keep it simple. Like normal English that non tech users can understand

docs/BigQueryTable-batchsink.md
```diff
@@ -46,6 +46,11 @@ bucket will be created and then deleted after the run finishes.
 will be dropped.
 * Upsert - records that match on Table Key will be updated. Records that do not match will be inserted.
 
+**WriteDisposition**: Type of write disposition to use. This can be set to Empty, Truncate, or Append.
```
nitinmotgi: Same here

widgets/BigQueryMultiTable-batchsink.json
```diff
@@ -30,6 +30,29 @@
           "widget-attributes" : {
             "placeholder": "Dataset the tables belong to"
           }
+        },
+        {
+          "widget-type": "radio-group",
+          "name": "writeDisposition",
```
nitinmotgi: Same here

nitinmotgi (changes requested):
Minor comments

docs/BigQueryMultiTable-batchsink.md
dli357: Hi, could you clarify which part needs to be simplified? This is currently a direct mapping to the doc here:

https://cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/WriteDisposition

albertshau: Write Disposition corresponds to the BigQuery concept, but the BigQuery concept isn't understandable without reading the doc. I think all we really need is a "Truncate Table" option. If true, the plugin sets write disposition to 'truncate'. Not sure we really need 'Empty'.

dli357: That makes sense, I've gone ahead and simplified it to "Truncate Table" with a true-false radio group.

docs/BigQueryTable-batchsink.md
```diff
@@ -46,6 +46,10 @@ bucket will be created and then deleted after the run finishes.
 will be dropped.
 * Upsert - records that match on Table Key will be updated. Records that do not match will be inserted.
 
+**Truncate Table**: Whether to truncate the output table before writing to it.
+* True - Will delete all rows in the table before writing to it.
+* False - Will append the new rows to the table.
```
albertshau: It doesn't necessarily append, depending on what is set for 'Operation'. I think we can just remove the true false lines, it seems clear enough with the single sentence description.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -199,6 +199,8 @@ private Configuration getBaseConfiguration() throws IOException {
       baseConfiguration.set(BigQueryConstants.CONFIG_CLUSTERING_ORDER, getConfig().getClusteringOrder());
     }
     baseConfiguration.setStrings(BigQueryConstants.CONFIG_OPERATION, getConfig().getOperation().name());
+    baseConfiguration.setStrings(BigQueryConfiguration.OUTPUT_TABLE_WRITE_DISPOSITION_KEY,
+        getConfig().getWriteDisposition().name());
```
albertshau: nit: use indentation consistent with the rest of the class (aligned with BigQueryConfiguration.xxx on the previous line)

widgets/BigQueryTable-batchsink.json
```diff
@@ -70,6 +70,25 @@
             ]
           }
         },
+        {
+          "widget-type": "radio-group",
```
albertshau: for booleans, there is a 'toggle' widget that we can use

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -70,6 +72,11 @@
   @Description("Type of write operation to perform. This can be set to Insert, Update or Upsert.")
   protected String operation;
 
+  @Name(NAME_WRITE_DISPOSITION)
+  @Macro
+  @Description("Type of write disposition to use. This can be set to Empty, Truncate, or Append.")
```
albertshau: this description should match the one given in the documentation

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -70,6 +72,11 @@
...
+  protected String writeDisposition;
```
albertshau: this should become a Boolean for whether to truncate the table or not, which means 'Empty' is no longer supported. The use cases we have are all truncate, we can add another "Fail if not Empty" property in the future if anybody asks for it.

docs/BigQueryMultiTable-batchsink.md
dli357: Additionally, should we need to change the mode in the future, all we would need to do is change the widget JSON and the documentation.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -140,6 +147,10 @@ public Operation getOperation() {
     return Operation.valueOf(operation.toUpperCase());
   }
 
+  public WriteDisposition getWriteDisposition() {
+    return WriteDisposition.valueOf(writeDisposition.toUpperCase());
```
albertshau: with the change to use a boolean, this just returns truncate if true, append if false or null.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
albertshau: this should also be annotated as Nullable. If null, it should default to false. 

It won't ever be null if the pipeline is created through the UI, but it can be if somebody creates it through the REST endpoint.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -251,4 +253,14 @@ private void validateOperationProperties(@Nullable Schema schema) {
         "Fields %s are in the table key, but not in the input schema.", result), NAME_TABLE_KEY);
     }
   }
+
+  private void validateWriteDisposition() {
+    if (Arrays.stream(WriteDisposition.values()).map(Enum::name).noneMatch(writeDisposition.toUpperCase()::equals)) {
```
albertshau: after the change to boolean, this won't need to happen. But we should add some validation to make sure truncate is false when the operation is 'update'.

albertshau (commented):
Can you check if this works when the operation is 'upsert'? 

When upsert is used, we use the output format to write to a temporary staging table, then run a 'merge' query to merge the data into the actual output table. I'm guessing the current behavior will set write disposition on the temporary staging table, which won't actually truncate the output.

When truncate is true and the operation is 'upsert', it's functionally equivalent to truncate is true and the operation is 'insert'. So we can avoid using the temporary table in that scenario.

widgets/BigQueryMultiTable-batchsink.json
dli357: Fixed!

docs/BigQueryTable-batchsink.md
dli357: Fixed!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
dli357: Fixed!

widgets/BigQueryTable-batchsink.json
dli357: Fixed!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
dli357: Fixed!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
dli357: Fixed!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
dli357: Fixed!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
dli357: It won't be nullable if it uses the primitive boolean rather than the Boolean wrapper class, right? Or should I change it to the wrapper class?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
dli357: I moved the validation to the abstract class since it applies to both BigQuery sinks. Now it throws an error during deployment if somebody tries to use the truncate option without using the Insert operation.

docs/BigQueryTable-batchsink.md
dli357: Fixed by shortening the doc to a single sentence.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
albertshau: should change it to a Boolean. The nullable annotation also tells CDAP that the user doesn't have to provide this property. If it's not annotated as nullable, CDAP will complain if the user doesn't provide it.

albertshau (commented):
minor comment about using a Boolean and handling null for Truncate Table.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
dli357: Gotcha, fixed

albertshau (approved):
lgtm, please squash commits before merging

PR #141, opened 2019-08-23
CDAP-14233 Add support for array in Spanner source

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-14233
In scope of this PR:
* Added support of Array in Spanner plugin

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
@@ -97,7 +97,7 @@ public class SpannerUtil {
             break;
           default:
             // this should not happen
-            throw new IllegalStateException("Logical type " + logicalType + " is not supported");
+            throw new IllegalStateException("Logical type '" + logicalType + "' is not supported");
         }
         addColumn(createStmt, name, field.getSchema().isNullable(), spannerType);
         continue;
@@ -122,6 +122,13 @@ public class SpannerUtil {
         case BYTES:
           spannerType = "BYTES(MAX)";
           break;
+        case ARRAY:
+          Schema componentSchema = fieldSchema.getComponentSchema();
+          if (componentSchema == null) {
+            throw new IllegalStateException("Component schema of field '" + name + "' is null");
+          }
+          spannerType = getArrayType(componentSchema);
+          break;
         default:
           throw new IllegalStateException(type.name() + " : Type currently not supported.");
       }
@@ -136,6 +143,42 @@ public class SpannerUtil {
     createStmt.append(" PRIMARY KEY (").append(primaryKeys).append(")");
 
     return createStmt.toString();
+  }
+
+  private static String getArrayType(Schema schema) {
+    Schema componentSchema = schema.isNullable() ? schema.getNonNullable() : schema;
+    Schema.LogicalType logicalType = componentSchema.getLogicalType();
+
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DATE:
+          return "ARRAY<DATE>";
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          return "ARRAY<TIMESTAMP>";
+        default:
+          // this should not happen
+          throw new IllegalStateException("Array of '" + logicalType + "' logical type currently not supported.");
+      }
+    }
+
+    Schema.Type type = componentSchema.getType();
+    switch (type) {
+      case BOOLEAN:
+        return "ARRAY<BOOL>";
+      case STRING:
+        return "ARRAY<STRING(MAX)>";
+      case INT:
+      case LONG:
+        return "ARRAY<INT64>";
+      case FLOAT:
+      case DOUBLE:
+        return "ARRAY<FLOAT64>";
+      case BYTES:
+        return "ARRAY<BYTES(MAX)>";
+      default:
+        throw new IllegalStateException("Array of '" + type.name() + "' type currently not supported.");
+    }
   }
 
   /**
--- a/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
@@ -16,15 +16,10 @@
 
 package io.cdap.plugin.gcp.spanner.sink;
 
-import com.google.cloud.ByteArray;
-import com.google.cloud.Date;
-import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.Spanner;
-import io.cdap.cdap.api.data.format.StructuredRecord;
-import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.gcp.spanner.SpannerConstants;
 import io.cdap.plugin.gcp.spanner.common.SpannerUtil;
 import org.apache.hadoop.conf.Configuration;
@@ -36,21 +31,19 @@ import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 
 import java.io.IOException;
-import java.time.LocalDate;
-import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
  * Spanner output format
  */
-public class SpannerOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+public class SpannerOutputFormat extends OutputFormat<NullWritable, Mutation> {
 
   /**
    * Get properties from SparkSinkConfig and store them as properties in Configuration
    *
    * @param configuration the Hadoop configuration to set the properties in
-   * @param config the spanner configuration
+   * @param config        the spanner configuration
    */
   public static void configure(Configuration configuration, SpannerSinkConfig config) {
     String projectId = config.getProject();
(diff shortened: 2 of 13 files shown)


vladhlinskiy:
@albertshau could you, please, take a look?

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
```diff
@@ -138,6 +145,46 @@ public static String convertSchemaToCreateStatement(String tableName, String pri
     return createStmt.toString();
   }
 
+  private static String getArrayType(Schema componentSchema) {
+    Schema unionSchema = componentSchema.getUnionSchema(0);
```
albertshau: Shouldn't assume the component schema is a union. instead, do:

```
componentSchema = componentSchema.isNullable() ? componentSchema.getNonNullable() : componentSchema;
```

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
       }
     };
   }
+
+  private static void writeArray(String name, Schema componentSchema, Object values, Mutation.WriteBuilder builder) {
+    Schema unionSchema = componentSchema.getUnionSchema(0);
```
albertshau: similar comment here, shouldn't assume the component is a union

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
...
+      throw new UnexpectedFormatException(
+        String.format("Field '%s' of type '%s' has unexpected value '%s'", name, fieldType, value));
+    };
+
+    if (value instanceof Collection) {
+      valueExtractor = name -> (Collection<?>) value;
+    } else if (value.getClass().isArray()) {
+      valueExtractor = name -> convertToObjectCollection(value);
+    }
+    return getValue(valueExtractor, fieldName, fieldType, Collection.class);
```
albertshau: this seems really complicated, why not just return the collections in the if-else above?

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
...
+    Class ofArray = array.getClass().getComponentType();
+    if (ofArray.isPrimitive()) {
+      List<Object> list = new ArrayList<>();
+      int length = Array.getLength(array);
+      for (int i = 0; i < length; i++) {
+        list.add(Array.get(array, i));
+      }
+      return list;
+    } else {
+      return Arrays.asList((Object[]) array);
```
albertshau: how come we need to check if the component is primitive? Does it work if we always do this?

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
...
+          Collection<Float> floatList = toCollection(name, type.toString(), values);
+          builder.set(name)
+            .toFloat64Array(
+              floatList.stream()
+                .map(Float::doubleValue)
+                .collect(Collectors.toList())
+            );
+          break;
+        case BYTES:
+          Collection<byte[]> bytesList = toCollection(name, type.toString(), values);
```
albertshau: CDAP bytes can be either a byte[] or a ByteBuffer. If it's a ByteBuffer it needs to be converted somewhere, either in toCollection or in the stream below

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
...
+                .collect(Collectors.toList())
+            );
+          break;
+        default:
+          throw new IllegalStateException("Array of '" + type.name() + "' type currently not supported.");
+      }
+    }
+  }
+
+  private static <T> Collection<T> toCollection(String fieldName, String fieldType, Object value) {
```
albertshau: I assume this means Spanner doesn't allow arrays of arrays (otherwise this needs to recursively convert the collection contents).

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -231,4 +243,125 @@ public void abortTask(TaskAttemptContext taskAttemptContext) {
...
+          builder.set(name)
+            .toDateArray(
+              dateList.stream()
+                .map(date -> Date.fromYearMonthDay(date.getYear(), date.getMonthValue(), date.getDayOfMonth()))
+                .collect(Collectors.toList())
+            );
+          break;
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          Collection<ZonedDateTime> timestampList = toCollection(name, logicalType.toString(), values);
```
albertshau: values should only ever contain the physical representation of the value, which means you'll get back a collection of longs here, not a collection of ZonedDateTime. Similarly, date will give you a collection of ints.

albertshau: This conversion logic should all be unit tested. Please extract out the conversion logic into a new class that converts StructuredRecord into Mutation and add unit tests for it.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -224,6 +230,44 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     }
   }
 
+  private List<?> transformArrayToList(ResultSet resultSet, String fieldName, Type arrayElementType) {
```
albertshau: there should also be unit tests for the conversion logic from a ResultSet into a StructuredRecord. Please extract out the conversion logic into a new class and add unit tests for it.

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: Changed to return Collection, moved to RecordToEntityTransformer class

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: If we try to convert array of primitives to Object[] we will get an error.

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: Changed to use transformation from realization for BYTES type, made it similar to Datastore plugin and moved to RecordToEntityTransformer class.

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: Yes, spanner doesn’t support Array of Arrays https://cloud.google.com/spanner/docs/data-types#array-type

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
vladhlinskiy: Changed to use transformation from realization for TIMESTAMP type, made it similar to Datastore plugin and moved to RecordToEntityTransformer class. Added tests in RecordToEntityTransformerTest class

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
vladhlinskiy: Moved logic to ResultSetToRecordTransformer class, covered with tests in  ResultSetToRecordTransformerTest class

vladhlinskiy:
@albertshau could you, please, take a look?

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
```diff
@@ -138,6 +145,42 @@ public static String convertSchemaToCreateStatement(String tableName, String pri
...
+    Schema componentSchema = schema.isNullable() ? schema.getNonNullable() : schema;
+    Schema.LogicalType logicalType = componentSchema.getLogicalType();
+
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DATE:
+          return "ARRAY<DATE>";
+        case TIMESTAMP_MILLIS:
+        case TIMESTAMP_MICROS:
+          return  "ARRAY<TIMESTAMP>";
```
albertshau: remove extra space after 'return'

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Transforms CDAP {@link StructuredRecord} to Google Spanner {@link Mutation}.
+ */
+public class RecordToEntityTransformer {
```
albertshau: nit: let's rename to RecordToMutationTransformer

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+        Float floatValue = record.<Float>get(fieldName);
+        if (floatValue == null) {
+          return Value.float64(null);
+        } else {
+          return Value.float64(floatValue.doubleValue());
+        }
+      case DOUBLE:
+        return Value.float64(record.<Double>get(fieldName));
+      case BYTES:
+        byte[] byteArray = record.get(fieldName);
```
albertshau: this can be a byte[] or a ByteBuffer, you have to check with instanceof

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+    this.schema = schema;
+  }
+
+  public Mutation transform(StructuredRecord record) {
+    Mutation.WriteBuilder builder = com.google.cloud.spanner.Mutation.newInsertOrUpdateBuilder(tableName);
+    List<Schema.Field> fields = schema.getFields();
+    for (Schema.Field field : fields) {
+      String name = field.getName();
+      Schema fieldSchema = field.getSchema();
+      fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
```
albertshau: convertToValue() does this already, no need to do it here.

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+          return Value.bytes(ByteArray.copyFrom(byteArray));
+        }
+      case ARRAY:
+        Schema componentSchema = schema.getComponentSchema();
+        if (componentSchema == null) {
+          throw new IllegalStateException("Component schema of field '" + fieldName + "' is null");
+        }
+        return transformCollectionToValue(fieldName, componentSchema, record.get(fieldName));
+      default:
+        throw new IllegalStateException(type.name() + " : Type currently not supported.");
```
albertshau: Should include the field name in the error message.

Field '%s' is of unsupported type '%s'

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+        byte[] byteArray = record.get(fieldName);
+        if (byteArray == null) {
+          return Value.bytes(null);
+        } else {
+          return Value.bytes(ByteArray.copyFrom(byteArray));
+        }
+      case ARRAY:
+        Schema componentSchema = schema.getComponentSchema();
+        if (componentSchema == null) {
+          throw new IllegalStateException("Component schema of field '" + fieldName + "' is null");
```
albertshau: Component schema of array field

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+        return transformCollectionToValue(fieldName, componentSchema, record.get(fieldName));
+      default:
+        throw new IllegalStateException(type.name() + " : Type currently not supported.");
+    }
+  }
+
+  private Value transformCollectionToValue(String name, Schema schema, Object object) {
+    Schema componentSchema = schema.isNullable() ? schema.getNonNullable() : schema;
+    Schema elementSchema = Schema.recordOf("arrayElementSchema", Schema.Field.of(name, componentSchema));
```
albertshau: componentSchema -> schema

otherwise this will fail if the collection contains null values. Please make sure this situation is covered in unit tests.

Also, does spanner allow null values in arrays? If not we have to fail with a descriptive message

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+            .collect(Collectors.toList())
+        );
+      case BYTES:
+        return Value.bytesArray(
+          values.stream()
+            .map(Value::getBytes)
+            .collect(Collectors.toList())
+        );
+      default:
+        throw new IllegalStateException("Array of '" + type.name() + "' type currently not supported.");
```
albertshau: Field '%s' is an array of '%s', which is not supported.

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
```diff
@@ -0,0 +1,229 @@
...
+    Schema componentSchema = schema.isNullable() ? schema.getNonNullable() : schema;
+    Schema elementSchema = Schema.recordOf("arrayElementSchema", Schema.Field.of(name, componentSchema));
+
+    Collection<Object> arrayValues = toCollection(name, componentSchema.getType().toString(), object);
+    List<Value> values = arrayValues.stream()
+      .map(value -> {
+        // Wrap array element into record to be able to re-use convertToValue method type conversion logic
+        // because StructuredRecord has different methods to extract value based on schema type.
+        // For example: record.get(fieldName) is used for most of the values,
+        // record.getTimestamp(fieldName) is used for timestamp values.
```
albertshau: FYI at some point we'll get around to https://issues.cask.co/browse/CDAP-15727 which should avoid the need for hacks like this.

src/main/java/io/cdap/plugin/gcp/spanner/common/SpannerUtil.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: Added support of null values in collections and covered with tests

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: Added todo to remove this hack after CDAP-15727 will be implemented

src/main/java/io/cdap/plugin/gcp/spanner/sink/RecordToEntityTransformer.java
vladhlinskiy: done

vladhlinskiy:
@albertshau could you, please, take a look?

albertshau (approved):
lgtm, squash and i'll merge

PR #142, opened 2019-08-23
add failure collector to speech to text transform

CuriousVini:
Adding failure collector api usage to speech to text transform.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
+++ b/src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
@@ -32,6 +32,7 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.StageSubmitterContext;
 import io.cdap.cdap.etl.api.Transform;
@@ -69,14 +70,15 @@ public class SpeechToTextTransform extends Transform<StructuredRecord, Structure
   public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgumentException {
     super.configurePipeline(configurer);
     Schema inputSchema = configurer.getStageConfigurer().getInputSchema();
-    config.validate(inputSchema);
+    FailureCollector collector = configurer.getStageConfigurer().getFailureCollector();
+    config.validate(inputSchema, collector);
     configurer.getStageConfigurer().setOutputSchema(getSchema(inputSchema));
   }
 
   @Override
   public void prepareRun(StageSubmitterContext context) throws Exception {
     super.prepareRun(context);
-    config.validate(context.getInputSchema());
+    config.validate(context.getInputSchema(), context.getFailureCollector());
   }
 
   @Override
@@ -140,7 +142,11 @@ public class SpeechToTextTransform extends Transform<StructuredRecord, Structure
     }
   }
 
-  private Schema getSchema(Schema inputSchema) {
+  @Nullable
+  private Schema getSchema(@Nullable Schema inputSchema) {
+    if (inputSchema == null) {
+      return null;
+    }
     List<Schema.Field> fields = new ArrayList<>();
     if (inputSchema.getFields() != null) {
       fields.addAll(inputSchema.getFields());
@@ -220,6 +226,11 @@ public class SpeechToTextTransform extends Transform<StructuredRecord, Structure
   }
 
   private static class SpeechTransformConfig extends PluginConfig {
+    private static final String NAME_AUDIOFIELD = "audiofield";
+    private static final String NAME_TRANS_PART = "transcriptionPartsField";
+    private static final String NAME_TRANS_TEXT = "transcriptionTextField";
+    private static final String NAME_RATE = "sampleRate";
+
     @Macro
     @Name("audiofield")
     @Description("Name of field containing binary audio file data")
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -53,6 +55,11 @@
 @Name(SpeechToTextTransform.NAME)
 @Description(SpeechToTextTransform.DESCRIPTION)
 public class SpeechToTextTransform extends Transform<StructuredRecord, StructuredRecord> {
+  private static final String NAME_AUDIOFIELD = "audiofield";
```
albertshau: these fit better in the SpeechTransformConfig class

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -68,15 +75,16 @@
...
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    config.validate(inputSchema, stageConfigurer);
     configurer.getStageConfigurer().setOutputSchema(getSchema(inputSchema));
   }
 
   @Override
   public void prepareRun(StageSubmitterContext context) throws Exception {
     super.prepareRun(context);
-    config.validate(context.getInputSchema());
+    config.validate(context.getInputSchema(), context.getFailureCollector());
```
albertshau, line 81: it seems a bit odd that here we do getFailureCollector(), but at configure time the StageConfigurer is a failure collector. What do you think about modifying the API such that PipelineConfigurer has a getFailureCollector() method?

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -140,7 +148,10 @@ public void destroy() {
     }
   }
 
-  private Schema getSchema(Schema inputSchema) {
+  private Schema getSchema(@Nullable Schema inputSchema) {
```
albertshau, line 146: annotate method as nullable

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
     }
 
     @Nullable
-    public Integer getSampleRate() {
+    public Integer getSampleRate() throws IllegalArgumentException {
```
albertshau: methods generally do not declare the runtime exceptions that they throw and instead just document it in the javadoc.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
-        }
-        Schema.Field field = inputSchema.getField(audioFieldName);
-        Schema fieldSchema = field.getSchema();
-        Schema.Type fieldType =
-          fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
-        if (fieldType != Schema.Type.BYTES) {
-          throw new IllegalArgumentException(String.format("Field '%s' must be of type 'bytes' but is of type " +
-                                                             "'%s'.", field.getName(), fieldType));
+          collector.addFailure(
+            String.format("Field '%s' does not exists in the input schema.", audioFieldName),
```
albertshau: exists -> exist

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
+            String.format("Make sure field '%s' exists in the input schema.", audioFieldName))
+            .withConfigProperty(NAME_AUDIOFIELD);
+        } else {
+          Schema.Field field = inputSchema.getField(audioFieldName);
```
albertshau: it's better to get the field once, before the if-else

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
+          Schema fieldSchema = field.getSchema();
+          Schema.Type fieldType =
+            fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+          if (fieldType != Schema.Type.BYTES) {
+            collector.addFailure(
+              String.format("Field '%s' must be of type 'bytes' but is of type '%s'.", field.getName(), fieldType),
```
albertshau: one thing we've run into several times with the error message is that there isn't an easy way to get a user friendly string for the schema. Here, it will use the physical type, but it's better to use the logical type if the schema has one. We should consider adding a method to the Schema class, or adding a Schemas utility class to hydrator-common for getting a display name or something.

Also, if the physical type is bytes but the logical type is decimal, we should still fail.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
+            "Make sure one of 'Transcript Parts Field' or 'Transcript Text Field' are specified.")
+            .withConfigProperty(NAME_TRANS_PART).withConfigProperty(NAME_TRANS_TEXT);
         }
         Set<String> fields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toSet());
         if (getTextField() != null && fields.contains(getTextField())) {
-          throw new IllegalArgumentException(String.format("Input schema contains given Transcription Text Field " +
-                                                             "'%s'. Please provide a different name.", getTextField()));
+          collector.addFailure(
+            String.format("Input schema contains given Transcription Text Field '%s'.", getTextField()),
+            String.format("Make sure Transcription Text Field '%s' is not in input schema.", getTextField()))
```
albertshau: in -> in the

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
+            .withConfigProperty(NAME_TRANS_TEXT).withInputSchemaField(getTextField(), null);
         }
         if (getPartsField() != null && fields.contains(getPartsField())) {
-          throw new IllegalArgumentException(String.format("Input schema contains given Transcription Parts Field " +
-                                                             "'%s'. Please provide a different name.",
-                                                           getPartsField()));
+          collector.addFailure(
+            String.format("Input schema contains given Transcription Parts Field '%s'.", getPartsField()),
+            String.format("Make sure Transcription Parts Field '%s' is not in input schema.", getPartsField()))
```
albertshau: in -> in the

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
       }
 
-      if (getSampleRate() != null && (getSampleRate() < 8000 || getSampleRate() > 48000)) {
-        throw new IllegalArgumentException("Sample Rate value must be between 8000 to 48000");
+      try {
+        Integer sampleRate = getSampleRate();
+        if (sampleRate != null && (sampleRate < 8000 || sampleRate > 48000)) {
+          collector.addFailure(
+            "Sample Rate value must be between 8000 to 48000",
+            "Make sure sample is in valid range")
```
albertshau: sample -> Sample Rate

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -314,49 +325,78 @@ public String getLanguage() {
...
+            .withConfigProperty(NAME_RATE);
+        }
+      } catch (IllegalArgumentException e) {
+        collector.addFailure(
+          "Sample Rate value must be a number between 8000 to 48000",
+          "Make sure sample is a number and is in valid range")
```
albertshau: same here

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
CuriousVini: yea makes sense.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
CuriousVini: yes that should still fail. I used existing conditions as is. 
I agree that we should have a method in Schema class to provide the name of the schema type. Filed https://issues.cask.co/browse/CDAP-15788

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -313,50 +324,80 @@ public String getLanguage() {
...
-        Schema fieldSchema = field.getSchema();
-        Schema.Type fieldType =
-          fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
-        if (fieldType != Schema.Type.BYTES) {
-          throw new IllegalArgumentException(String.format("Field '%s' must be of type 'bytes' but is of type " +
-                                                             "'%s'.", field.getName(), fieldType));
+        if (audioFieldName != null && field == null) {
+          collector.addFailure(
+            String.format("Field '%s' does not exist in the input schema.", audioFieldName),
+            String.format("Field '%s' must exist in the input schema.", audioFieldName))
```
albertshau: This error message and corrective action is basically the same. In those situations, I think it's better to just have an error message and no corrective action, otherwise it's basically repeating the same thing.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -313,50 +324,80 @@ public String getLanguage() {
...
+            "At least one of 'Transcript Parts Field' or 'Transcript Text Field' must be specified.", null)
+            .withConfigProperty(NAME_TRANS_PART).withConfigProperty(NAME_TRANS_TEXT);
         }
         Set<String> fields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toSet());
         if (getTextField() != null && fields.contains(getTextField())) {
-          throw new IllegalArgumentException(String.format("Input schema contains given Transcription Text Field " +
-                                                             "'%s'. Please provide a different name.", getTextField()));
+          collector.addFailure(
+            String.format("Transcription Text Field '%s' exists in the input schema.", getTextField()),
+            String.format("Transcription Text Field '%s' must not exist in the input schema.", getTextField()))
```
albertshau: Specify a field that does not already exist.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -313,50 +324,80 @@ public String getLanguage() {
...
+            .withConfigProperty(NAME_TRANS_TEXT).withInputSchemaField(getTextField(), null);
         }
         if (getPartsField() != null && fields.contains(getPartsField())) {
-          throw new IllegalArgumentException(String.format("Input schema contains given Transcription Parts Field " +
-                                                             "'%s'. Please provide a different name.",
-                                                           getPartsField()));
+          collector.addFailure(
+            String.format("Transcription Parts Field '%s' exists in the input schema.", getPartsField()),
+            String.format("Transcription Parts Field '%s' does not exist in the input schema.", getPartsField()))
```
albertshau: Specify a field that does not already exist.

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -313,50 +324,80 @@ public String getLanguage() {
...
         }
       }
 
-      if (getSampleRate() != null && (getSampleRate() < 8000 || getSampleRate() > 48000)) {
-        throw new IllegalArgumentException("Sample Rate value must be between 8000 to 48000");
+      try {
+        Integer sampleRate = getSampleRate();
+        if (sampleRate != null && (sampleRate < 8000 || sampleRate > 48000)) {
+          collector.addFailure(
+            "Sample Rate value must be between 8000 to 48000", null)
```
albertshau: between 8000 and 48000

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -313,50 +324,80 @@ public String getLanguage() {
...
+            .withConfigProperty(NAME_RATE);
+        }
+      } catch (IllegalArgumentException e) {
+        collector.addFailure(
+          "Sample Rate value must be a number between 8000 to 48000", null)
```
albertshau: between 8000 and 48000

albertshau (approved):
small comments otherwise lgtm

PR #149, opened 2019-09-16
add BQ action plugin

yaojiefeng:


src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    public void validate() {
+      // check the mode is valid
+      if (!containsMacro("mode")) {
+        getMode();
+      }
+
+      if (!containsMacro("sql") && (sql == null || sql.isEmpty())) {
+        throw new IllegalArgumentException("SQL not specified. Please specify a SQL to execute");
```
CuriousVini: we should use new failure collector apis for validation

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,45 @@
...
+
+Properties
+----------
+**Project ID**: Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+
+**SQL**: The SQL command to execute.
+
+**Legacy**: Specified whether to use BigQuery's legacy SQL dialect for this query. By default this property is
+set to 'false'. If set to 'false', the query will use BigQuery's Standard SQL.
```
albertshau: It's better to have a 'Dialect' property with values 'Legacy' and 'Standard', with it defaulting to Standard. This is more extensible in case a new dialect is ever added, and also makes it easier to understand.

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Mode**: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query is executed as
+soon as possible and count towards the concurrent rate limit and the daily rate limit. An interactive query is
```
albertshau: counts roward

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,45 @@
...
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access BigQuery.
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Project ID**: Google Cloud Project ID, which uniquely identifies a project.
```
albertshau, line 22: This is the project in which the query job is executed right?

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,45 @@
...
+**Mode**: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query is executed as
+soon as possible and count towards the concurrent rate limit and the daily rate limit. An interactive query is
+queued and started as soon as idle resources are available, usually within a few minutes. If the query hasn't
+started within 3 hours, its priority is changed to 'interactive'.
+
+**Dataset Name**: The dataset to store the query results in. If not specified, the results will not be stored.
+
+**Table Name**: The table to store the query results in. If not specified, the results will not be stored.
+
+**UseCache**: Specifies whether to use cache when executing the query.
```
albertshau: Insert space between words 'Use Cache'.

Also, what does this mean? I don't know what this does from reading this doc

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Job Location**: Location of the job. It must match the location of the dataset specified in the query. Defaults to 'US'
```
albertshau: Don't need to mention the default here, just set a default in the widget json

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+    if (queryJob.getStatus().getError() != null) {
+      // You can also look at queryJob.getStatus().getExecutionErrors() for all
+      // errors, not just the latest one.
+      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
+    }
+
+    TableResult queryResults = queryJob.getQueryResults();
+    long rows = queryResults.getTotalRows();
+
+    context.getMetrics().gauge(context.getStageName() + RECORDS_OUT, rows);
```
albertshau: this should not be done. records.out is the metric emitted by the app indicating records flowing from one stage to another. 'records.processed' would be more accurate.

This is also doing something weird with concatenation, as there's no separator between stage name and records.out. You should use context.getStageMetrics() instead of prefixing. Also, is there a reason gauge is used instead of increment?

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+    QueryJobConfiguration.Builder builder = QueryJobConfiguration.newBuilder(config.getSql());
+    // Run at batch priority, which won't count toward concurrent rate limit.
+    if (config.getMode().equals(QueryJobConfiguration.Priority.BATCH)) {
+      builder.setPriority(QueryJobConfiguration.Priority.BATCH);
+    } else {
+      builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
+    }
+
+    // Save the results of the query to a permanent table.
+    if (config.getDataset() != null  && config.getTable() != null) {
```
albertshau: remove extra space

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
albertshau: +1

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,130 @@
...
+          "widget-attributes": {
+            "default": "auto-detect"
+          }
+        },
+        {
+          "widget-type": "textarea",
+          "label": "SQL",
+          "name": "sql",
+          "widget-attributes": {
+            "placeholder": "Specifies the BigQuery SQL to be executed."
```
albertshau: this should be a better placeholder, like

SELECT * from [dataset].[table]

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,130 @@
+{
+  "metadata": {
+    "spec-version": "1.5"
+  },
+  "display-name": "Google BigQuery Execute",
```
albertshau: source and sink use 'BigQuery' not 'Google BigQuery', let's do the same here: 'BigQuery Execute'

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+    queryJob.waitFor();
+
+    // Check for errors
+    if (queryJob.getStatus().getError() != null) {
+      // You can also look at queryJob.getStatus().getExecutionErrors() for all
+      // errors, not just the latest one.
+      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
+    }
+
+    TableResult queryResults = queryJob.getQueryResults();
```
albertshau, line 112: one use case I've heard multiple times is to have the ability to store the query results in an argument that is reference further down in the pipeline. For example, they want to do something like 'select max(updated_time) as max_updated from [dataset].[table]', then use it in multiple future queries.

I think it makes sense to have a 'Set Results as Arguments' property. When set to true, the results of the first row are set as arguments through context.getArguments().set(), where the key is the column name and the value is the string value of the result.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+      if (!containsMacro("mode")) {
+        getMode();
+      }
+
+      if (!containsMacro("sql") && (sql == null || sql.isEmpty())) {
+        throw new IllegalArgumentException("SQL not specified. Please specify a SQL to execute");
+      }
+
+      // validates that either they are null together or not null together
+      if ((dataset == null && table != null) || (table == null && dataset != null)) {
```
CuriousVini: Option string config properties should use null and empty checks.`Strings.isNullOrEmpty`

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Nullable
+    public String getTable() {
+      return table;
+    }
+
+    public void validate() {
+      // check the mode is valid
+      if (!containsMacro("mode")) {
```
CuriousVini: should use static final constant.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+        getMode();
+      }
+
+      if (!containsMacro("sql") && (sql == null || sql.isEmpty())) {
```
CuriousVini: should use static final constant.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,209 @@
...
+        throw new IllegalArgumentException("SQL not specified. Please specify a SQL to execute");
+      }
+
+      // validates that either they are null together or not null together
+      if ((dataset == null && table != null) || (table == null && dataset != null)) {
+        throw new IllegalArgumentException("Dataset and table must be specified together.");
+      }
+
+      if (dataset != null) {
```
CuriousVini: same here.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
CuriousVini: this should also check for `containsMacro`.

docs/BigQueryExecute-action.md
yaojiefeng: Yes, updated the doc

docs/BigQueryExecute-action.md
yaojiefeng: Added more description to the doc using https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: gauge and increment should be the same in this case, since we are only adding the metrics once. I have fixed the metrics emit

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: This is the name of a variable, I see in gcp plugin, we do not create a constant for these, so I will keep it as is

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: Same as above

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
CuriousVini: This is the name of the config property. can't we do something like this?
https://github.com/data-integrations/google-cloud/blob/release/0.13/src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java#L229

yaojiefeng: Done

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,252 @@
...
+
+      if (!containsMacro(SQL) && Strings.isNullOrEmpty(sql)) {
+        failureCollector.addFailure("SQL not specified. Please specify a SQL to execute",
+                                    null).withConfigProperty(SQL);
+      }
+
+      // validates that either they are null together or not null together
+      if ((!containsMacro(DATASET) && !containsMacro(TABLE)) &&
+            (Strings.isNullOrEmpty(dataset) != Strings.isNullOrEmpty(table))) {
+        failureCollector.addFailure("Dataset and table must be specified together.", null);
```
CuriousVini: `. withConfigProperty(DATASET).withConfigProperty(TABLE)` since these 2 are associated properties.

yaojiefeng: Fixed

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,50 @@
...
+The service account key file needs to be available on every node in your cluster and
+must be readable by all users running the job.
+
+Properties
+----------
+**Project ID**: Google Cloud Project ID, which uniquely identifies a project.
+It can be found on the Dashboard in the Google Cloud Platform Console.
+This is the project that the BigQuery SQL will run in.
+
+**SQL**: The SQL command to execute.
```
albertshau: SQL command to execute.

We don't use 'The' as the first word, otherwise we would have to put it there for almost every description in order to be consistent.

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Dialect**: Specified the dialect for the BigQuery SQL. The value must be 'legacy' or 'standard'. If set to 'standard',
```
albertshau: Dialect of the SQL command.

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,50 @@
...
+the query will use BigQuery's standard SQL: https://cloud.google.com/bigquery/sql-reference/.
+If set to 'legacy', BigQuery's legacy SQL dialect will be used for this query.
+
+**Mode**: Mode to execute the query in. The value must be 'batch' or 'interactive'. A batch query is executed as
+soon as possible and counts towards the concurrent rate limit and the daily rate limit. An interactive query is
+queued and started as soon as idle resources are available, usually within a few minutes. If the query hasn't
+started within 3 hours, its priority is changed to 'interactive'.
+
+**Dataset Name**: The dataset to store the query results in. If not specified, the results will not be stored.
```
albertshau: Dataset to store ...

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Table Name**: The table to store the query results in. If not specified, the results will not be stored.
```
albertshau: Table to store ...

docs/BigQueryExecute-action.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Use Cache**: Specifies whether to look for the result in the query cache. The query cache is a best-effort
+cache that will be flushed whenever tables in the query are modified.
+
+**Job Location**: Location of the job. It must match the location of the dataset specified in the query.
+
+**Set As Arguments**: Whether to set the results of the first row as arguments.
```
albertshau: row as arguments. For example, if the query is 'select min(id) as min_id, max(id) as max_id from my_dataset.my_table', an arguments for 'min_id' and 'max_id' will be set based on the query results. Plugins further down the pipeline can then reference these values with macros ${min_id} and ${max_id}.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,253 @@
...
+    if (config.shouldSetAsArguments()) {
+      if (rows == 0 || queryResults.getSchema() == null) {
+        LOG.warn("The query result does not contain any row or schema, will not save the results in the arguments");
+      } else {
+        Schema schema = queryResults.getSchema();
+        FieldValueList firstRow = queryResults.iterateAll().iterator().next();
+        for (int i = 0; i < schema.getFields().size(); i++) {
+          Field field = schema.getFields().get(i);
+          String name = field.getName();
+          if (field.getType().equals(LegacySQLTypeName.RECORD)) {
```
albertshau, line 128: what about arrays?

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -0,0 +1,253 @@
...
+            LOG.warn("Field {} is a record type with nested schema, will not save the value in the argument", name);
+            continue;
+          }
+          context.getArguments().set(name, firstRow.get(name).getStringValue());
+        }
+      }
+    }
+
+    context.getMetrics().gauge(RECORDS_PROCESSED, rows);
```
albertshau, line 137: use context.getStageMetrics(), which will automatically prefix the stage name. Otherwise this will clobber metrics from other stages in the pipeline.

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,149 @@
...
+          "name": "location",
+          "widget-attributes": {
+            "placeholder": "Location of the job.",
+            "default": "US"
+          }
+        },
+        {
+          "name": "setAsArguments",
+          "label": "Set As Arguments",
+          "widget-type": "radio-group",
```
albertshau: use the 'toggle' widget type for booleans.

widgets/BigQueryExecute-action.json
```diff
@@ -0,0 +1,149 @@
...
+          "label": "Table Name",
+          "name": "table",
+          "widget-attributes": {
+            "placeholder": "Name of table to store the results."
+          }
+        },
+        {
+          "name": "useCache",
+          "label": "Use Cache",
+          "widget-type": "radio-group",
```
albertshau: use the 'toggle' widget type for booleans

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: They don't have array as the type, from https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/LegacySQLTypeName.java#L89, the only nested type is RECORD

CuriousVini: bigquery array is represented as repeated mode of some type.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: `context.getMetrics` already returns a `StageMetrics`, which I believe will prefix with stage name

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
yaojiefeng: Updated the check for mode

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
albertshau: ah I see, nevermind then

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -120,8 +121,9 @@ public void run(ActionContext context) throws Exception {
         for (int i = 0; i < schema.getFields().size(); i++) {
           Field field = schema.getFields().get(i);
           String name = field.getName();
-          if (field.getType().equals(LegacySQLTypeName.RECORD)) {
-            LOG.warn("Field {} is a record type with nested schema, will not save the value in the argument", name);
+          if (field.getMode().equals(Field.Mode.REPEATED) || field.getType().equals(LegacySQLTypeName.RECORD)) {
```
albertshau: use 2 separate if statements so that we can give a more specific warning

albertshau (approved):
one comment, otherwise lgtm

PR #153, opened 2019-09-19
Fix speech transform to handle macros correctly

CuriousVini:


src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -39,6 +39,7 @@
 import io.cdap.cdap.etl.api.TransformContext;
 import io.cdap.plugin.gcp.common.GCPConfig;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import scala.annotation.meta.field;
```
albertshau: accidental?

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
```diff
@@ -226,9 +228,9 @@ private void addTranscriptWithConfidence(List<StructuredRecord> speechArray,
   }
 
   private static class SpeechTransformConfig extends PluginConfig {
+    static final String NAME_TRANS_PART = "transcriptionPartsField";
+    static final String NAME_TRANS_TEXT = "transcriptionTextField";
```
albertshau: these can still be private right?

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
CuriousVini: yea, fixed

src/main/java/io/cdap/plugin/gcp/speech/SpeechToTextTransform.java
CuriousVini: right

albertshau (approved):

PR #154, opened 2019-09-19
CDAP-15896 skip validation when cant autodetect

albertshau:
Skip certain validation checks when the project or service account
cannot be auto-detected. This is done because auto-detection may
work at runtime when the pipeline is run on dataproc, so the
user needs to be able to deploy pipelines like these.


CuriousVini (approved):
lgtm 👍

PR #155, opened 2019-09-19
[CDAP-15787] add validation apis for gcs source and sink

CuriousVini:
Requires https://github.com/cdapio/hydrator-plugins/pull/947 changes to be merged.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -22,6 +22,7 @@ import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
@@ -56,7 +57,6 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
-    config.validate();
   }
 
   @Override
@@ -74,8 +74,13 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
    */
   @SuppressWarnings("unused")
   public static class GCSBatchSinkConfig extends GCPReferenceSinkConfig implements FileSinkProperties {
+    private static final String NAME_PATH = "path";
+    private static final String NAME_SUFFIX = "suffix";
+    private static final String NAME_FORMAT = "format";
+    private static final String NAME_SCHEMA = "schema";
+
     private static final String SCHEME = "gs://";
-    @Name("path")
+    @Name(NAME_PATH)
     @Description("The path to write to. For example, gs://<bucket>/path/to/directory")
     @Macro
     private String path;
@@ -89,7 +94,6 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
 
     @Description("The format to write in. The format must be one of 'json', 'avro', 'parquet', 'csv', 'tsv', "
       + "or 'delimited'.")
-    @Macro
     protected String format;
 
     @Description("The delimiter to use if the format is 'delimited'. The delimiter will be ignored if the format "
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -106,19 +111,41 @@ protected void recordLineage(LineageRecorder lineageRecorder, List<String> outpu
...
+          collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_PATH).withStacktrace(e.getStackTrace());
+        }
       }
-      if (suffix != null && !containsMacro("suffix")) {
-        new SimpleDateFormat(suffix);
+      if (suffix != null && !containsMacro(NAME_SUFFIX)) {
+        try {
+          new SimpleDateFormat(suffix);
+        } catch (IllegalArgumentException e) {
+          collector.addFailure("Invalid suffix.", "Ensure provided suffix is valid.")
```
albertshau: should we include the message from the exception? Otherwise it won't be apparent why it's invalid.

Also don't think we need the corrective action if we include the exception message

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -62,24 +63,37 @@ public GCSMultiBatchSink(GCSMultiBatchSinkConfig config) {
...
       .addAll(config.getProperties().getProperties())
       .add("schema", String.format("${%s}", SCHEMA_MACRO)).build();
-    OutputFormatProvider outputFormatProvider =
-      pipelineConfigurer.usePlugin(BatchSink.FORMAT_PLUGIN_TYPE, format.name().toLowerCase(),
-                                   FORMAT_PLUGIN_ID, formatProperties);
-    if (outputFormatProvider == null) {
-      throw new IllegalArgumentException(String.format("Could not find the '%s' output format plugin.",
-                                                       format.name().toLowerCase()));
+
+    if (format != null) {
```
albertshau: don't think this check is needed. Instead, if the format couldn't be fetched above, we can just throw the exception there since everything below it requires that format is not null.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -129,18 +128,10 @@ public void validate(FailureCollector collector) {
...
           new SimpleDateFormat(suffix);
         } catch (IllegalArgumentException e) {
-          collector.addFailure("Invalid suffix.", "Ensure provided suffix is valid.")
+          collector.addFailure("Invalid suffix : " + e.getMessage(), null)
             .withConfigProperty(NAME_SUFFIX).withStacktrace(e.getStackTrace());
         }
       }
-      if (!containsMacro(NAME_FORMAT)) {
-        try {
-          getFormat();
```
albertshau: doesn't this still need to be called?

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -203,13 +201,6 @@ public void validate(FailureCollector collector) {
             .withConfigProperty(NAME_FILE_REGEX).withStacktrace(e.getStackTrace());
         }
       }
-      if (!containsMacro(NAME_FORMAT)) {
-        try {
-          getFormat();
```
albertshau: doesn't this still need to be called?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
CuriousVini: When would it be invalid? since its not macro enabled.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -204,7 +229,7 @@ public Pattern getFilePattern() {
       try {
         return fileRegex == null ? null : Pattern.compile(fileRegex);
       } catch (RuntimeException e) {
-        throw new IllegalArgumentException("Invalid file regular expression: " + e.getMessage(), e);
+        throw new IllegalArgumentException("Invalid file regular expression.", e);
```
albertshau: why remove the message? It will be hard to tell what to change without it.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -169,18 +170,42 @@ public GCSSourceConfig() {
...
+        } catch (Exception e) {
+          collector.addFailure("File system properties must be a valid json.", null)
+            .withConfigProperty(NAME_FILE_SYSTEM_PROPERTIES).withStacktrace(e.getStackTrace());
+        }
+      }
+      if (!containsMacro(NAME_FILE_REGEX)) {
+        try {
+          getFilePattern();
+        } catch (IllegalArgumentException e) {
+          collector.addFailure("Provide a valid regular expression.", null)
```
albertshau: this won't say what is wrong with the pattern, so the user won't be able to adjust accordingly

CuriousVini:
I removed exception messages because of this:
image

For file system properties, since it must be a valid json, not adding exception message for `getFileSystemProperties` call. Others I have added error message.

albertshau (approved):

CuriousVini:
squashing and merging.

PR #157, opened 2019-09-25
add cmek to some gcs plugins

yaojiefeng:
Add CMEK support for PubSub, GCS and bigquery, these plugins still pending test

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -89,10 +89,12 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
       null : GCPUtils.loadServiceAccountCredentials(serviceAccountFilePath);
     String project = config.getProject();
     BigQuery bigQuery = GCPUtils.getBigQuery(project, credentials);
-    baseConfiguration = getBaseConfiguration();
+    String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
+    baseConfiguration = getBaseConfiguration(cmekKey);
     String bucket = configureBucket();
     if (!context.isPreviewEnabled()) {
-      BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(project, credentials), config.getDataset(), bucket);
+      BigQueryUtil.createResources(bigQuery, GCPUtils.getStorage(project, credentials), config.getDataset(), bucket,
+                                   cmekKey);
     }
 
     prepareRunInternal(context, bigQuery, bucket);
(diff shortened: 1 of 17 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -187,9 +189,9 @@ protected abstract OutputFormatProvider getOutputFormatProvider(Configuration co
    *
    * @return base configuration
    */
-  private Configuration getBaseConfiguration() throws IOException {
+  private Configuration getBaseConfiguration(String cmekKey) throws IOException {
```
albertshau: key is nullable right?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -381,15 +381,20 @@ private static TableReference getTableReference(Configuration conf) throws IOExc
...
       String query = generateQuery(tableRef);
       LOG.debug("Update/Upsert query: " + query);
 
       BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
-      QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).setUseLegacySql(false).build();
+      QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder(query)
+          .setUseLegacySql(false)
+          .setDestinationEncryptionConfiguration(
+            com.google.cloud.bigquery.EncryptionConfiguration.newBuilder().setKmsKeyName(cmekKey).build())
```
albertshau, line 396: is the full class name required?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -156,16 +160,18 @@ public static Configuration getBigQueryConfig(@Nullable String serviceAccountFil
...
-                                     String datasetName, String bucketName) throws IOException {
+                                     String datasetName, String bucketName,
+                                     @Nullable String cmekKey) throws IOException {
     Dataset dataset = bigQuery.getDataset(datasetName);
     Bucket bucket = storage.get(bucketName);
 
     if (dataset == null && bucket == null) {
       createBucket(storage, bucketName, null,
-                   () -> String.format("Unable to create Cloud Storage bucket '%s'", bucketName));
+                   () -> String.format("Unable to create Cloud Storage bucket '%s'", bucketName), cmekKey);
```
albertshau: it's easier to read these method calls if the cmekKey comes before the Supplier, so that the closure is the last argument and not in the middle

albertshau (commented):
minor comments.

Don't the GCS move and copy plugins need to be modified too?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
yaojiefeng: Fixed

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
yaojiefeng: Yes, the load config uses a different `EncrptionConfiguration`, https://github.com/data-integrations/google-cloud/blob/77e98bc7f7d7630f2e8f113c986b140799458934/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java#L241

yaojiefeng:
GCSCopy and GCSMove must have the destination exist, https://github.com/data-integrations/google-cloud/blob/release/0.13/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java#L91, so it is not required to add cmek for them

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/publisher/source/GCPCredentialsProvider.java
```diff
@@ -0,0 +1,86 @@
...
+      this.serviceAccountFilePath = serviceAccountFilePath;
+      return this;
+    }
+
+    public GCPCredentialsProvider build() {
+      return new GCPCredentialsProvider(serviceAccountFilePath);
+    }
+  }
+
+}
```
yaojiefeng: Missing new line

albertshau:
tested all the sources and sink with local runs as well as dataproc runs

PR #162, opened 2019-10-09
[CDAP-15978] Validate dataset and table names for BQ plugins

CuriousVini:
image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -22,10 +22,10 @@ import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.GCPReferenceSinkConfig;
 
 import java.util.Set;
-import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -104,12 +105,24 @@ public boolean isAllowSchemaRelaxation() {
...
     super.validate(collector);
     String bucket = getBucket();
-    if (!containsMacro(NAME_BUCKET) && bucket != null) {
-      // Basic validation for allowed characters as per https://cloud.google.com/storage/docs/naming
-      Pattern p = Pattern.compile("[a-z0-9._-]+");
-      if (!p.matcher(bucket).matches()) {
-        collector.addFailure("Bucket must only contain lowercase characters, numbers, '.', '_', and '-'", null)
-          .withConfigProperty(NAME_BUCKET);
+    // Allowed character validation for bucket name as per https://cloud.google.com/storage/docs/naming
+    String errorMessage = "Bucket name can contain following characters: lowercase letters, numbers, " +
```
albertshau: It's sounds more natural to remove 'following characters:' and say:

Bucket name can only contain lowercase letters, numbers, ...

similar comment for the other messages.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -104,12 +105,24 @@ public boolean isAllowSchemaRelaxation() {
...
+    // Allowed character validation for dataset name as per https://cloud.google.com/bigquery/docs/datasets
+    errorMessage = "Dataset name can contain following characters: letters (lower or uppercase), numbers or '_'.";
+    match(dataset, "[A-Za-z0-9_]+", collector, errorMessage, NAME_DATASET);
+  }
+
+  /**
+   * Matches text with provided pattern. If the text does not match the pattern, add a new failure to failure collector.
+   *
+   */
+  protected void match(String text, String pattern, FailureCollector collector, String errorMessage, String property) {
```
albertshau: it's a little better to pass in a Pattern instead of a String and have constants for the Patterns instead of re-using hardcoded strings.

This method can also be moved to some common place instead of being implemented in 2 plugins.

CuriousVini:
@albertshau Addressed comments. PTAL, thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -133,14 +133,23 @@ public String getDatasetProject() {
...
+
+    if (!containsMacro(dataset)) {
+      // Allowed character validation for dataset name as per https://cloud.google.com/bigquery/docs/datasets
+      String errorMessage = "Dataset name can only contain letters (lower or uppercase), numbers and '_'.";
+      BigQueryUtil.match(dataset, NAME_DATASET, BigQueryUtil.DATASET_PATTERN, collector, errorMessage);
+    }
+
+    if (!containsMacro(table)) {
+      // Allowed character validation for table name as per https://cloud.google.com/bigquery/docs/tables
+      String errorMessage = "Table name can only contain letters (lower or uppercase), numbers and '_'.";
```
albertshau: actually, since there multiple places this is called, I think it makes sense to have separate validateBucket(), validateDataset(), validateTable() methods and also move the error messages into the util class as constants.

albertshau (approved):
one comment about moving things around to reduce code duplication, otherwise lgtm. fix and merge.

PR #163, opened 2019-10-10
[CDAP-16026] Fix google-cloud plugins to assume non-null values for n…

CuriousVini:
…ullable properties.
Also scanned through the plugins to spot similar issue through intellij.

Screen Shot 2019-10-10 at 4 28 40 PM

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
@@ -166,13 +166,10 @@ public final class SpannerSink extends BatchSink<StructuredRecord, NullWritable,
 
     ResultSet resultSet = dbClient.singleUse().executeQuery(statement);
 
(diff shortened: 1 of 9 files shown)


src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
```diff
@@ -166,13 +166,13 @@ private boolean isTablePresent(DatabaseClient dbClient) {
 
     ResultSet resultSet = dbClient.singleUse().executeQuery(statement);
 
-    boolean tableExists = false;
-    while (resultSet.next()) {
+    boolean tableExists = resultSet.next();
+    if (resultSet.next()) {
```
albertshau: this if isn't needed right?

CuriousVini: yea. Fixed

albertshau (approved):

PR #165, opened 2019-10-14
[CDAP-16039] Fix macros to use propery name instead of value

CuriousVini:


src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreSinkConfig.java
```diff
@@ -287,7 +287,7 @@ private void validateSinkFieldSchema(String fieldName, Schema fieldSchema, Failu
   }
 
   private void validateIndexStrategy(Schema schema, FailureCollector collector) {
-    if (containsMacro(indexStrategy)) {
+    if (containsMacro(DatastoreSinkConstants.PROPERTY_INDEX_STRATEGY)) {
```
albertshau, line 290: :)

albertshau (approved):
lgtm

PR #166, opened 2019-10-15
CDAP-16037 add location to gcp sinks

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-16037

This pr adds the location to the gcs and bq sinks that create resources.
Pubsub does not expose location.
Bigtable and Spanner only supports location at the instance level, and we do not create instance in these sinks.
DataStore we do not create non-existing instances.

### DIFF
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -39,6 +39,9 @@ bucket will be created and then deleted after the run finishes.
 **Truncate Table:** Whether or not to truncate the table before writing to it.
 Should only be used with the Insert operation.
 
+**Location:** The location where the big query datasets will get created. This value is ignored
+if the dataset or temporary bucket already exist.
+
 **Split Field:** The name of the field that will be used to determine which table to write to.
 
 **Update Table Schema**: Whether the BigQuery table schema should be modified 
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -51,6 +51,9 @@ Should only be used with the Insert operation.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
+**Location:** The location where the big query dataset will get created. This value is ignored
+if the dataset or temporary bucket already exist.
+
 **Create Partitioned Table**: Whether to create the BigQuery table with time partitioning. This value 
 is ignored if the table already exists.
 * When this is set to true, table will be created with time partitioning. 
--- a/widgets/BigQueryMultiTable-batchsink.json
+++ b/widgets/BigQueryMultiTable-batchsink.json
@@ -101,6 +101,19 @@
           }
         }
       ]
+    },
+    {
+      "label": "Auto Create",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Location",
+          "name": "location",
+          "widget-attributes": {
+            "default": "US"
+          }
+        }
+      ]
     }
   ],
   "outputs": [],
--- a/widgets/BigQueryTable-batchsink.json
+++ b/widgets/BigQueryTable-batchsink.json
@@ -132,6 +132,14 @@
       "label": "Auto Create",
       "properties": [
         {
+          "widget-type": "textbox",
+          "label": "Location",
+          "name": "location",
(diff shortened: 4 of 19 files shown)


widgets/GCSBucketCreate-action.json
```diff
@@ -56,6 +56,19 @@
           }
         }
       ]
+    },
+    {
+      "label": "Auto Create",
+      "properties": [
+        {
+          "widget-type": "textbox",
```
bdmogal, line 64: can it be a selector?

widgets/GCSBucketCreate-action.json
```diff
@@ -56,6 +56,19 @@
...
+          "label": "Location",
+          "name": "location",
+          "widget-attributes": {
+            "placeholder": "Location of the gcs buckets"
```
bdmogal: instead can we just show the default? If this can't be a selector right now?

docs/BigQueryMultiTable-batchsink.md
```diff
@@ -39,6 +39,8 @@ bucket will be created and then deleted after the run finishes.
 **Truncate Table:** Whether or not to truncate the table before writing to it.
 Should only be used with the Insert operation.
 
+**Location:** The location where the big query datasets will get created. This value is ignored for existing dataset.
```
albertshau: nit: for existing dataset -> if the dataset already exists

same for other places

widgets/GCSBucketCreate-action.json
yaojiefeng: This is very complicated since they can be single region or multi region, it is difficult to list all of them.

docs/BigQueryMultiTable-batchsink.md
albertshau: actually, this is also ignored if the temporary bucket already exists.

if the dataset or temporary bucket already exist.

widgets/GCSBucketCreate-action.json
yaojiefeng: You just pass null to the client in order to create on default location, it is different for different gcp products. I think it is better to not specify them so that we can leave it to the clients to determine what the default is. And in case the client changes in the future, it will not break the plugin

widgets/GCSBucketCreate-action.json
albertshau: selector is brittle, it means we have to update CDAP in order to accommodate a new location. Unless selector has a way to overwrite with a custom value, I think this should be a textbox

widgets/GCSBucketCreate-action.json
albertshau: +1 to defaulting

yaojiefeng: Changed to default

docs/BigQueryTable-batchsink.md
```diff
@@ -51,7 +51,8 @@ Should only be used with the Insert operation.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
-**Location:** The location where the big query dataset will get created. This value is ignored for existing dataset.
+**Location:** The location where the big query dataset will get created. This value is ignored
+if the dataset or temporary bucket already exist..
```
albertshau: double period

widgets/BigQueryMultiTable-batchsink.json
```diff
@@ -110,7 +110,7 @@
           "label": "Location",
           "name": "location",
           "widget-attributes": {
-            "placeholder": "Location of the big query datasets"
+            "placeholder": "US"
```
albertshau: not a placeholder, but an actual default

widgets/BigQueryTable-batchsink.json
```diff
@@ -136,7 +136,7 @@
           "label": "Location",
           "name": "location",
           "widget-attributes" : {
-            "placeholder": "Location of the big query dataset"
+            "placeholder": "US"
```
albertshau: is it significant that it's US here and us for gcs?

yaojiefeng: This is to be consistent with the gcp console. In gcs, we use `us` and in BQ we use `US`

albertshau (approved):

PR #167, opened 2019-10-16
PLUGIN-82: Fixed problem with duplicate during Upsert operation in BigQuery Sink plugin.

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/PLUGIN-82

In scope of this PR:

- Fixed problem with handling duplicates.
- Added additional validation for 'Table key' field for checking duplicate fields.
- Added validation for new field 'Ordered by'.
- Added possibility deleting temporary table.

### DIFF
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -51,6 +51,10 @@ Should only be used with the Insert operation.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
+**Dedupe By**: Column names and sort order used to choose which input record to update/upsert when there are 
+multiple input records with the same key. For example, if this is set to 'updated_time desc', then if there are 
+multiple input records with the same key, the one with the largest value for 'updated_time' will be applied.              
+
 **Create Partitioned Table**: Whether to create the BigQuery table with time partitioning. This value 
 is ignored if the table already exists.
 * When this is set to true, table will be created with time partitioning. 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -91,6 +91,8 @@ import javax.annotation.Nullable;
 public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<AvroKey<GenericRecord>, NullWritable> {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryOutputFormat.class);
 
+  private static final String SOURCE_DATA_QUERY = "(SELECT * FROM (SELECT row_number() OVER (PARTITION BY %s%s) " +
+    "as rowid, * FROM %s) where rowid = 1)";
   private static final String UPDATE_QUERY = "UPDATE %s T SET %s FROM %s S WHERE %s";
   private static final String UPSERT_QUERY = "MERGE %s T USING %s S ON %s WHEN MATCHED THEN UPDATE SET %s " +
     "WHEN NOT MATCHED THEN INSERT (%s) VALUES(%s)";
@@ -111,6 +113,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Avr
     private Operation operation;
     private TableReference temporaryTableReference;
     private List<String> tableKeyList;
+    private List<String> orderedByList;
     private List<String> tableFieldsList;
 
     private boolean allowSchemaRelaxation;
@@ -156,6 +159,9 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Avr
         .collect(Collectors.toList());
       String tableKey = conf.get(BigQueryConstants.CONFIG_TABLE_KEY, null);
       tableKeyList = Arrays.stream(tableKey != null ? tableKey.split(",") : new String[0]).map(String::trim)
+        .collect(Collectors.toList());
+      String dedupedBy = conf.get(BigQueryConstants.CONFIG_DEDUPE_BY, null);
+      orderedByList = Arrays.stream(dedupedBy != null ? dedupedBy.split(",") : new String[0])
         .collect(Collectors.toList());
       String tableFields = conf.get(BigQueryConstants.CONFIG_TABLE_FIELDS, null);
       tableFieldsList = Arrays.stream(tableFields != null ? tableFields.split(",") : new String[0])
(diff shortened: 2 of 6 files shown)


AntonGensitskiy:

docs/BigQueryTable-batchsink.md
```diff
@@ -51,6 +51,8 @@ Should only be used with the Insert operation.
 
 **Table Key**: List of fields that determines relation between tables during Update and Upsert operations.
 
+**Ordered By**: Column names and sort order that will be applied to source data during Upsert operation.
```
albertshau: This is a misleading name that doesn't really convey what this is used for. Let's change it to 'Dedupe By', with description:  

Column names and sort order used to choose which input record to update/upsert when there are multiple input records with the same key. For example, if this is set to 'updated_time desc', then if there are multiple input records with the same key, the one with the largest value for 'updated_time' will be applied.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -157,6 +160,9 @@ public void commitJob(JobContext jobContext) throws IOException {
       String tableKey = conf.get(BigQueryConstants.CONFIG_TABLE_KEY, null);
       tableKeyList = Arrays.stream(tableKey != null ? tableKey.split(",") : new String[0]).map(String::trim)
         .collect(Collectors.toList());
+      String orderedBy = conf.get(BigQueryConstants.CONFIG_ORDERED_BY, null);
+      orderedByList = Arrays.stream(orderedBy != null ? orderedBy.split(",") : new String[0]).map(String::trim)
+        .map(keyValue -> keyValue.replace(":", " ")).collect(Collectors.toList());
```
albertshau: the output format shouldn't need to replace any special characters. Whatever is configuring the outputformat should give well-formed values.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -444,6 +450,9 @@ private String generateQuery(TableReference tableRef) {
         case UPDATE:
           return String.format(UPDATE_QUERY, destinationTable, fieldsForUpdate, sourceTable, criteria);
         case UPSERT:
+          String orderedBy = orderedByList.isEmpty() ? "" : "ORDER BY " + String.join(", ", orderedByList);
+          sourceTable = String.format(SOURCE_DATA_QUERY, String.join(", ", tableKeyList), orderedBy,
```
albertshau: the same thing should be done for update if an order is specified. Otherwise I assume the existing merge query just applies it in whatever order it chooses?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -343,6 +356,53 @@ private void validateOperationProperties(@Nullable Schema schema, FailureCollect
...
+      .forEach(key -> collector.addFailure(
+        String.format("Table key field '%s' is duplicated.", key),
+        String.format("Remove duplicates of Table key field '%s'.", key))
+        .withConfigElement(NAME_TABLE_KEY, key)
+      );
+
+    if (Operation.UPSERT.equals(getOperation()) && getOrderedBy() != null) {
+      List<String> orderedByList = Arrays.stream(Objects.requireNonNull(getOrderedBy()).split(","))
+        .map(String::trim).collect(Collectors.toList());
+      LOG.debug("-> " + orderedByList);
```
albertshau: this should be removed

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -343,6 +356,53 @@ private void validateOperationProperties(@Nullable Schema schema, FailureCollect
...
+        .forEach(key -> collector.addFailure(
+          String.format("Ordered by field '%s' is duplicated.", key),
+          String.format("Remove duplicates of Ordered by field '%s'.", key))
+          .withConfigElement(NAME_ORDERED_BY, orderedByFieldValueMap.get(key))
+        );
+
+      orderedByList.stream()
+        .filter(v -> v.split(":").length != 2)
+        .forEach(v -> collector.addFailure(
+          "Missing sort order.",
```
albertshau: this seems like it should just be defaulted (to whatever order is default in bigquery). We wouldn't expect this when the pipeline is created from the UI, but it could happen when created through the REST API.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -343,6 +356,53 @@ private void validateOperationProperties(@Nullable Schema schema, FailureCollect
...
+        String.format("Remove duplicates of Table key field '%s'.", key))
+        .withConfigElement(NAME_TABLE_KEY, key)
+      );
+
+    if (Operation.UPSERT.equals(getOperation()) && getOrderedBy() != null) {
+      List<String> orderedByList = Arrays.stream(Objects.requireNonNull(getOrderedBy()).split(","))
+        .map(String::trim).collect(Collectors.toList());
+      LOG.debug("-> " + orderedByList);
+      orderedByList.stream()
+        .filter(v -> !fields.contains(v.split(":")[0]))
```
albertshau: can we use a space instead of ':'? That way it translates directly to what would be used in the query.

albertshau (commented):
mostly small comments. Please be sure to update the corresponding update/upsert integration tests in https://github.com/cdapio/cdap-integration-tests after the fix is merged

docs/BigQueryTable-batchsink.md
AntonGensitskiy: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: Updated. New logic applied to Update and Upsert operation.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
AntonGensitskiy: This validation was removed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java

AntonGensitskiy:

AntonGensitskiy:
Current integration update/upsert tests don't have duplicate values, that's why I add several new test for cover this functionality.

beetlejuice1000:
Could I please ask what the behaviour would be if there are duplicates in the destination table in BQ? Will this be merged to itself and effectively truncate the table? If not, how do we deal with duplicates in the destination table when merging with this fix?

albertshau:
This handles duplicates within a single pipeline run. If the pipeline is the only thing that has ever written to the table, then there will not be any duplicates. If something else is writing dupes, or if there are existing duplicates, this will not (and should not) try to de-duplicate things.

albertshau (approved):
lgtm, please squash commits and I'll merge.

AntonGensitskiy:
Commits are squashed.

PR #168, opened 2019-10-22
PLUGIN-83: Fixed problems with BigQuery Sink plugin.

AntonGensitskiy:
JIRA: https://issues.cask.co/browse/PLUGIN-83

In scope of this PR:
- Fixed issue with NPE during closing mAvroFileWriter.
- Fixed issue with processing data in case when gcsPaths is empty.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
@@ -109,6 +109,9 @@ public class AvroRecordWriter extends RecordWriter<AvroKey<GenericRecord>, NullW
   /** {@inheritDoc} */
   @Override
   public long sync() throws IOException {
-    return mAvroFileWriter.sync();
+    if (mAvroFileWriter != null) {
+      return mAvroFileWriter.sync();
+    }
+    return 0;
   }
 }
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
@@ -101,7 +101,9 @@ public class AvroRecordWriter extends RecordWriter<AvroKey<GenericRecord>, NullW
   /** {@inheritDoc} */
   @Override
   public void close(TaskAttemptContext context) throws IOException {
-    mAvroFileWriter.close();
+    if (mAvroFileWriter != null) {
+      mAvroFileWriter.close();
+    }
   }
 
   /** {@inheritDoc} */
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -196,84 +196,92 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Avr
                BigQueryStrings.toString(tableRef), gcsPaths.size(), gcsPaths.isEmpty() ? "(empty)" : gcsPaths.get(0),
                true);
 
-      // Create load conf with minimal requirements.
-      JobConfigurationLoad loadConfig = new JobConfigurationLoad();
-      loadConfig.setSchema(schema);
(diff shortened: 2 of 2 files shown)


AntonGensitskiy:

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -101,7 +101,9 @@ private void createFileWriter(Schema writerSchema) throws IOException {
   /** {@inheritDoc} */
   @Override
   public void close(TaskAttemptContext context) throws IOException {
-    mAvroFileWriter.close();
+    if (mAvroFileWriter != null) {
```
albertshau, line 104: a similar fix should be done in the sync() method

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -202,84 +202,92 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
-        if (partitionByField != null) {
-          timePartitioning.setField(partitionByField);
-        }
-        timePartitioning.setRequirePartitionFilter(requirePartitionFilter);
-        loadConfig.setTimePartitioning(timePartitioning);
-        if (!clusteringOrderList.isEmpty()) {
-          Clustering clustering = new Clustering();
-          clustering.setFields(clusteringOrderList);
-          loadConfig.setClustering(clustering);
+      if (!gcsPaths.isEmpty()) {
```
albertshau: it's hard to read if there is a bunch of code in an if block. Instead, do checks in their own if-block, and leave normal logic outside.

```
if (gcsPaths.isEmpty()) {
   // handle scenario
}

// normal logic
```

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -202,84 +202,92 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
-                                                        temporaryTableReference.getDatasetId(),
-                                                        temporaryTableReference.getTableId(), table).execute();
+        if (temporaryTableReference != null && bigQueryHelper.tableExists(temporaryTableReference)) {
+          long expirationMillis = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(1);
+          Table table = bigQueryHelper.getTable(temporaryTableReference).setExpirationTime(expirationMillis);
+          bigQueryHelper.getRawBigquery().tables().update(temporaryTableReference.getProjectId(),
+                                                          temporaryTableReference.getDatasetId(),
+                                                          temporaryTableReference.getTableId(), table).execute();
+        }
+      } else if (!bigQueryHelper.tableExists(tableRef)) {
```
albertshau: When does this happen? Can you add some comments?

albertshau (commented):
please rebase this against the relase/0.13 branch and merge it into release/0.13 instead of develop.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
AntonGensitskiy: I added null check for sync() method. However sync() method has return value, in case when v is null this method will be return 0. Is it acceptable solution?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
AntonGensitskiy: This block was added for case when gcsPaths list is empty and destination table not exist. In this case empty destination table will be created.

AntonGensitskiy:
@albertshau I addressed all your comments and rebase on branch release/0.13, could you, please, take a look.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
albertshau: yeah that's fine

albertshau (approved):

PR #176, opened 2020-01-06
Revert "Revert version bump"

mikkin:
This reverts commit abd7150852d758b62fc93cd8e125b73696be4746.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,8 +72,8 @@
     <avro.version>1.8.2</avro.version>
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
-    <cdap.version>6.1.0-SNAPSHOT</cdap.version>
-    <cdap.plugin.version>2.3.0-SNAPSHOT</cdap.plugin.version>
+    <cdap.version>6.1.1</cdap.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -610,8 +610,8 @@
         <version>1.1.0</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
```
MEseifan: Should we be bumping the lower bound of allowed versions for these plugins?

pom.xml
```diff
@@ -610,8 +610,8 @@
...
+            <parent>system:cdap-data-pipeline[6.1.1,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-streams[6.1.1,7.0.0-SNAPSHOT)</parent>
```
MEseifan: Should we be bumping the lower bound for allowed versions for these plugins?

MEseifan: Spoke with Yaojie, should keep this as `6.1.0-SNAPSHOT`

yaojiefeng (approved):

PR #185, opened 2020-03-13
[CDAP-16425] Fix macro support for output schema in BigQuery source plugin

rmstar:


rmstar:
@yaojiefeng I see the following error on shutdown when BQ source output schema is a macro:
```
Exception raised on destroy lifecycle method in class io.cdap.cdap.datapipeline.SmartWorkflow of the Workflow program of run program_run:default.5fe091ec-64c1-11ea-ba4b-000000351bec.-SNAPSHOT.workflow.DataPipelineWorkflow.62cbf20d-64c1-11ea-a58b-00000053f35f
--
java.lang.NullPointerException: null 	at io.cdap.cdap.etl.lineage.FieldLineageProcessor.validateAndConvert(FieldLineageProcessor.java:86) ~[na:na] 	at io.cdap.cdap.datapipeline.SmartWorkflow.destroy(SmartWorkflow.java:604) ~[na:na] 	at io.cdap.cdap.internal.app.runtime.AbstractContext.lambda$destroyProgram$2(AbstractContext.java:669) ~[na:na] 	at io.cdap.cdap.internal.app.runtime.AbstractContext$2.run(AbstractContext.java:555) ~[na:na] 	at io.cdap.cdap.data2.transaction.Transactions$CacheBasedTransactional.finishExecute(Transactions.java:224) ~[na:na] 	at io.cdap.cdap.data2.transaction.Transactions$CacheBasedTransactional.execute(Transactions.java:211) ~[na:na] 	at io.cdap.cdap.internal.app.runtime.AbstractContext.execute(AbstractContext.java:550) ~[na:na] 	at io.cdap.cdap.internal.app.runtime.AbstractContext.destroyProgram(AbstractContext.java:669) ~[na:na] 	at io.cdap.cdap.internal.app.runtime.workflow.WorkflowDriver.destroyWorkflow(WorkflowDriver.java:297) [na:na] 	at io.cdap.cdap.internal.app.runtime.workflow.WorkflowDriver.shutDown(WorkflowDriver.java:277) [na:na] 	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$1.run(AbstractExecutionThreadService.java:65) [com.google.guava.guava-13.0.1.jar:na] 	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_221]
```

I'm wondering if I we need to avoid [recording lineage](https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java#L172) in some cases?

yaojiefeng:
If the output schema is macro, then the stage spec does not have it, which causes this NPE, for the short term we can catch any exception in the destroy method, and log a message. For the long term, we should propagate the stage spec with the schema again if macro is enabled

rmstar:
Thanks Yaojie! Do you want me to open a bug for the CDAP change (catch exception in the destroy method)?

yaojiefeng (approved):

yaojiefeng:
I can do the CDAP change for field lineage, don't worry about it

PR #186, opened 2020-03-16
Feature release/instrument streaming

yaojiefeng:
Note the build will fail until https://github.com/cdapio/cdap/pull/11962 and https://github.com/cdapio/hydrator-plugins/pull/1038 are merged

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.13.3</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.13.3</version>
+  <version>0.13.4</version>
```
albertshau: better for this to be snapshot until we bump the dependencies to be non-snapshot

albertshau (approved):

PR #189, opened 2020-03-19
[CDAP-16068] Checking for schema emptiness in prepareRun rather than configurePipeline for BigQuery source plugin

vravish:


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -237,6 +240,12 @@ private void validateConfiguredSchema(Schema configuredSchema, FailureCollector
     com.google.cloud.bigquery.Schema bqSchema = getBQSchema(collector);
 
     FieldList fields = bqSchema.getFields();
+
+    if (configuredSchema == null) {
```
albertshau: This shouldn't result in an error. We do want to allow users to leave out a configured schema. In that case, the source should use whatever schema the table has.

For example, when the table is a macro, the schema may change from run to run as the table changes. In that case, the schema cannot be provided by the user.

What should happen is the plugin should error out in prepareRun if the bqSchema doesn't have any fields.

vravish: That is what this does. This method will not be called from configurePipeline if configuredSchema is null. See lines 107-112; in configurePipeline, we have a null check and return early if configuredSchema == null.

vravish: But it will be called from prepareRun and get to this failure case if the schema does not exist by the time the pipeline is run.

albertshau: no that's not what this does. This checks if the user has not given a schema and then errors. We only want to error if the BQ table doesn't have a schema.

albertshau: Basically, please check that you are able to deploy and run a pipeline where the configured schema is empty, and the schema in BQ is not.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -237,6 +240,12 @@ private void validateConfiguredSchema(Schema configuredSchema, FailureCollector
     com.google.cloud.bigquery.Schema bqSchema = getBQSchema(collector);
 
     FieldList fields = bqSchema.getFields();
+
+    if (fields.isEmpty()) {
```
albertshau: this is almost the same behavior as now, it will fail to deploy when there is a no-schema BQ table and a non-null configured schema.

This is what I suggest:
-  refactor this method and getSchema() to take the BQ schema as an argument
-  call getBQSchema() in both configurePipeline() and prepareRun()
-  in prepareRun(), if fields are empty, throw an exception

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -117,6 +117,13 @@ public void configurePipeline(PipelineConfigurer configurer) {
   public void prepareRun(BatchSourceContext context) throws Exception {
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
+
+    if (getBQSchema(collector).getFields().isEmpty()) {
+      collector.addFailure("No schema was specified in input BigQuery table.",
```
albertshau: better to include the table name:

BigQuery table '[dataset].[table]' does not have a schema. Please edit the table to add a schema.

albertshau (approved):
minor comment about the error message, otherwise lgtm

PR #194, opened 2020-03-25
Bump version to 0.14.2

ajainarayanan:


albertshau (approved):

albertshau (commented):
cdap-maven-plugin part should be updated to have 6.1.2 as the bottom version in the range

albertshau (approved):

PR #195, opened 2020-03-30
[CDAP-16481] Added BigQuery source and sink type conversions to docs

vravish:


docs/Bigtable-batchsink.md
```diff
@@ -40,3 +40,26 @@ Full list of allowed properties: https://cloud.google.com/bigtable/docs/hbase-cl
...
 When running on other clusters, the file must be present on every node in the cluster.
+
+Data Type Mappings from CDAP to BigQuery
+----------
+The following table lists out different CDAP data types, as well as the 
+corresponding BigQuery data type for each CDAP type, for updates and upserts.
+
+| CDAP type | BigQuery type |
+|-----------|---------------|
+| long      | int64         |
```
albertshau: should include all CDAP types, seems to be missing int, float, and union. I believe int -> int64, float -> float64, and union is not supported.

Also, were you able to verify that this information is correct? The stuff I put in the jira was off the top of my head.

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,24 @@ Only columns defined in schema will be included into output record.
...
+|---------------|-----------|
+| int64         | long      |
+| numeric       | numeric   |
+| float64       | double    |
+| boolean       | boolean   |
+| string        | string    |
+| bytes         | bytes     |
+| date          | date      |
+| datetime      | string    |
+| time          | time      |
```
albertshau: CDAP time and timestamp are in microseconds. What is the unit for BigQuery? If it is nanos or more granular, we should add a note that there will be some loss of granularity

docs/Bigtable-batchsink.md
vravish: How can I verify these types? Is there a master list somewhere? Or should I create a BQ source and see how the types are loaded into CDAP?

albertshau: this is the sink. You should verify that this information is correct by running pipelines

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,27 @@ Only columns defined in schema will be included into output record.
...
+| numeric       | decimal     |
+| float64       | double      |
+| bool          | boolean     |
+| string        | string      |
+| bytes         | bytes       |
+| date          | date        |
+| datetime      | string      |
+| time          | time        |
+| timestamp     | timestamp   |
+| array         | array       |
```
CuriousVini: Is `StandardSQLTypeName.ARRAY` type supported in source?

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,27 @@ Only columns defined in schema will be included into output record.
...
+
+Data Type Mappings from BigQuery to CDAP
+----------
+The following table lists out different BigQuery data types, as well as the 
+corresponding CDAP data type for each BigQuery type.
+
+| BigQuery type | CDAP type   |
+|---------------|-------------|
+| int64         | long        |
+| numeric       | decimal     |
```
CuriousVini: this should say what precision and scale.

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,27 @@ Only columns defined in schema will be included into output record.
...
+| bytes         | bytes       |
+| date          | date        |
+| datetime      | string      |
+| time          | time        |
+| timestamp     | timestamp   |
+| array         | array       |
+| struct        | record      |
+| geo           | unsupported |
+
+For time and timestamp, values are stored upto microsecond precision in both
```
CuriousVini: should we add this information in the type itself?

docs/Bigtable-batchsink.md
```diff
@@ -40,3 +40,29 @@ Full list of allowed properties: https://cloud.google.com/bigtable/docs/hbase-cl
...
+| string    | string        |
+| bytes     | bytes         |
+| date      | date          |
+| time      | time          |
+| timestamp | timestamp     |
+| array     | array         |
+| record    | struct        |
+| float     | float64       |
+| int       | int64         |
+| union     | unsupported   |
```
CuriousVini: map type is also not supported right? Does this include all the cdap types such as enum etc?

CuriousVini (commented):
few comments

docs/Bigtable-batchsink.md
```diff
@@ -49,16 +49,19 @@ corresponding BigQuery data type for each CDAP type, for updates and upserts.
...
-| boolean   | boolean       |
+| boolean   | bool          |
 | string    | string        |
 | bytes     | bytes         |
 | date      | date          |
 | time      | time          |
 | timestamp | timestamp     |
 | array     | array         |
 | record    | struct        |
+| float     | float64       |
```
albertshau: should put this above double, since they are similar

docs/Bigtable-batchsink.md
```diff
@@ -49,16 +49,19 @@ corresponding BigQuery data type for each CDAP type, for updates and upserts.
...
+| int       | int64         |
```
albertshau: should put this above long, since they are similar

docs/Bigtable-batchsource.md
```diff
@@ -61,18 +61,21 @@ Data Type Mappings from BigQuery to CDAP
...
```
albertshau: upto -> up to

albertshau (approved):
minor comments, otherwise lgtm

albertshau:
looks like Vinisha has some more comments, please address before merging

docs/Bigtable-batchsource.md
vravish: No; the only supported fields in source are BOOL, BYTES, DATE, DATETIME, FLOAT64, INT64, NUMERIC, STRING, STRUCT, TIME, TIMESTAMP

docs/Bigtable-batchsource.md
vravish: In the table entry you mean?

docs/Bigtable-batchsink.md
vravish: This includes all the types I was able to generate in CDAP, and I wasn't sure how to add map type.

albertshau:
@vravish this needed more review of changes for Vinisha's comments before merging

docs/Bigtable-batchsource.md
albertshau: This doesn't seem correct based on https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java#L325. Did you try it?

docs/Bigtable-batchsink.md
albertshau: Maps are a type in CDAP and should be mentioned. Not knowing how to add a type doesn't mean you can keep it undocumented.

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,24 @@ Only columns defined in schema will be included into output record.
...
+----------
+The following table lists out different BigQuery data types, as well as the 
+corresponding CDAP data type for each BigQuery type.
+
+| BigQuery type | CDAP type   | Precision / Range           |
+|---------------|-------------|-----------------------------|
+| array         | unsupported |                             |
+| bool          | boolean     |                             |
+| bytes         | bytes       |                             |
+| date          | date        | YYYY-MM-DD                  |
```
albertshau, line 69: don't need to specify a 'range' for date, this can mislead people into thinking its a string in that format.

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,24 @@ Only columns defined in schema will be included into output record.
...
+| datetime      | string      |                             |
+| float64       | double      | 8 bytes, double precision   |
```
albertshau, line 71: don't need to specify number of bytes for double and long

docs/Bigtable-batchsource.md
```diff
@@ -55,3 +55,24 @@ Only columns defined in schema will be included into output record.
 **onError**: Strategy used to handle errors during transformation of a text entry to record. Possible values are:
 - **Skip error** - Ignores erroneous records.
 - **Fail Pipeline** - Fails pipeline due to erroneous record.
+
+Data Type Mappings from BigQuery to CDAP
+----------
+The following table lists out different BigQuery data types, as well as the 
+corresponding CDAP data type for each BigQuery type.
+
+| BigQuery type | CDAP type   | Precision / Range           |
```
albertshau, line 64: this 3rd column is more confusing than helpful. You can just put the information in the CDAP type column. For example, timestamp (microseconds)

vravish:
#200 is the follow up PR

PR #196, opened 2020-04-01
support skip header

yaojiefeng:


chtyim (approved):

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -290,6 +296,11 @@ public boolean isCopyHeader() {
       return copyHeader != null && copyHeader;
     }
 
+    @Override
+    public boolean skipHeader() {
+      return skipHeader == null ? false : skipHeader;
```
albertshau, line 301: this will be null in the configure method if skipHeader is a macro. Are we only calling this from prepareRun() or are we calling it in configure too? If only in prepare, we're all good.

yaojiefeng: This isn't called at all at this moment, the value is set using reflections for format plugins

PR #204, opened 2020-04-07
Macro support for format field in GCSSource, GCSBatchSink

rmstar:
Depends on https://github.com/cdapio/hydrator-plugins/pull/1072


albertshau (approved):
lgtm, though be sure the other one is merged before merging this.

Will there also be a similar change for the S3 source?

rmstar:
> Will there also be a similar change for the S3 source?

I'll add that in another PR

PR #206, opened 2020-04-13
bump version to 0.14.3-SNAPSHOT

albertshau:


CuriousVini (commented):
should we update cdap version as well?

albertshau:
we don't need to bump cdap version unless there's some new api or something required by the plugin. This lets the plugin be compatible with as many CDAP versions as possible.

CuriousVini (approved):

PR #207, opened 2020-04-14
CDAP-15775 allow writing to an empty BQ table using sink

albertshau:
BigQuery tables can be created without any schema. When data is
loaded, the schema is set. Fixed the sink to allow writing to
these types of tables instead of failing during validation.


albertshau:
cherry-pick of #205

CuriousVini (approved):

PR #212, opened 2020-04-21
[CDAP-16630] Checking if source table definition instanceof StandardTableDefinition before casting

vravish:


chtyim (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -392,10 +392,12 @@ private void validatePartitionProperties(FailureCollector collector) {
     if (sourceTable == null) {
       return;
     }
-    TimePartitioning timePartitioning = ((StandardTableDefinition) (sourceTable).getDefinition())
-      .getTimePartitioning();
-    if (timePartitioning == null) {
-      return;
+    if (sourceTable.getDefinition() instanceof StandardTableDefinition) {
```
CuriousVini, line 395: can `sourceTable.getDefinition()` be instance of some other class? If so should we add a warn message saying that not supported or something similar?

vravish: The other class is autogenerated by BQ, and is only the definition if table's data comes from an external (non-BQ) source, e.g. Google Sheets.

CuriousVini: I see.

CuriousVini (approved):

albertshau:
@vravish be sure to cherrypick to release branches as well

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -392,10 +392,12 @@ private void validatePartitionProperties(FailureCollector collector) {
...
+      TimePartitioning timePartitioning = ((StandardTableDefinition) sourceTable.getDefinition())
+          .getTimePartitioning();
```
albertshau, line 397: indentation is off (should be 2 spaces). Please make sure your IDE has the right settings.

PR #215, opened 2020-04-22
[CDAP-16630] Checking if source table definition instanceof StandardTableDefinition before casting

vravish:


vravish:
Cherry pick of #212 and #213

CuriousVini (approved):

PR #216, opened 2020-04-25
[CDAP-16670] Fix OOM for bigquery sink

CuriousVini:
image

image

image

Tested witH 8MB default and Mapreduce engine and following worker config:
image

The pipeline reads 100 * 12000 records from 100 multiple tables from sql server and writes to bigquery using bigquery multi table.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -202,6 +202,9 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
                                  getConfig().isAllowSchemaRelaxation());
     baseConfiguration.setStrings(BigQueryConfiguration.OUTPUT_TABLE_WRITE_DISPOSITION_KEY,
                                  getConfig().getWriteDisposition().name());
+    // this setting is needed because gcs has default chunk size of 64MB. This is large default chunk size which can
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -202,6 +202,9 @@ private Configuration getBaseConfiguration(@Nullable String cmekKey) throws IOEx
                                  getConfig().isAllowSchemaRelaxation());
     baseConfiguration.setStrings(BigQueryConfiguration.OUTPUT_TABLE_WRITE_DISPOSITION_KEY,
                                  getConfig().getWriteDisposition().name());
+    // this setting is needed because gcs has default chunk size of 64MB. This is large default chunk size which can
+    // cause OOM issue if there are many tables being written. See this - CDAP-16670
+    baseConfiguration.set("fs.gs.outputstream.upload.chunk.size", "8388608");
```
chtyim: Can we make this configurable?

nitinmotgi: If we make it configurable then we to have best practice guide for that please.

chtyim: The default should be 8M, but making it configuration to give more flexibility for wider use cases

chtyim (approved):

PR #219, opened 2020-04-27
[CDAP-16664] Escape BigQuery table name

rmstar:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -457,7 +457,8 @@ private void updateTableSchema(TableReference tableRef) {
 
     private String generateQuery(TableReference tableRef) {
       String criteriaTemplate = "T.%s = S.%s";
-      String destinationTable = tableRef.getProjectId() + "." + tableRef.getDatasetId() + "." + tableRef.getTableId();
+      String destinationTable = "`" + tableRef.getProjectId() + "." + tableRef.getDatasetId() +
```
albertshau: This is hard to read, can we switch to using String.format() instead

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -466,9 +467,9 @@ private String generateQuery(TableReference tableRef) {
         .map(s -> String.format(criteriaTemplate, s, s)).collect(Collectors.joining(", "));
       String orderedBy = orderedByList.isEmpty() ? "" : " ORDER BY " + String.join(", ", orderedByList);
       String sourceTable = String.format(SOURCE_DATA_QUERY, String.join(", ", tableKeyList), orderedBy,
-                                         temporaryTableReference.getProjectId() + "." +
+                                         "`" + temporaryTableReference.getProjectId() + "." +
```
albertshau: similar comment here, a little hard to read. Would be better to make a separate tempTable variable and use String.format() to generate it.

albertshau (approved):

PR #220, opened 2020-04-27
Fixed warnings and added missing copyright

bdmogal:


albertshau (approved):

CuriousVini (approved):
could you please file jira to fail the build in case license is not provided.

PR #223, opened 2020-04-29
Don't ACK Pubsub messages when running preview

rmstar:


src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -31,12 +31,15 @@
 import org.apache.spark.storage.StorageLevel;
 import org.apache.spark.streaming.api.java.JavaDStream;
 import org.apache.spark.streaming.api.java.JavaReceiverInputDStream;
+import org.apache.spark.streaming.dstream.ReceiverInputDStream;
```
CuriousVini, line 34: Is validation working fine with this change?

rmstar: yes, verified that messages are not auto-acknowledged in preview mode, but are acked when pipeline is deployed.

CuriousVini (approved):
lgtm 👍

PR #227, opened 2020-05-14
[CDAP-16817] Fix bigquery source to emit bytes

CuriousVini:
Before:
image
image

After:
image

image

Bigquery to Bigquery:

image

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.gcp.bigquery.source;
 
+import io.cdap.cdap.api.common.Bytes;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.format.UnexpectedFormatException;
 import io.cdap.cdap.api.data.schema.Schema;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -103,4 +104,14 @@ protected Object convertField(Object field, Schema fieldSchema) throws IOExcepti
     }
     return super.convertField(field, fieldSchema);
   }
+
+  @Override
+  protected Object convertBytes(Object field) {
+    if (field instanceof ByteBuffer) {
+      ByteBuffer buf = (ByteBuffer) field;
+      return Arrays.copyOfRange(buf.array(), buf.arrayOffset() + buf.position(),
```
chtyim: Use `Bytes.toBytes(buf)` from cdap-api instead. It works for all types of ByteBuffer implementations.

chtyim (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -103,4 +104,9 @@ protected Object convertField(Object field, Schema fieldSchema) throws IOExcepti
     }
     return super.convertField(field, fieldSchema);
   }
+
+  @Override
+  protected Object convertBytes(Object field) {
```
nitinmotgi, line 109: Where is this method being called, dont' see this in PR.

CuriousVini: It gets called in Base class - `RecordConverter` which will get triggered from `super.convertField(field, fieldSchema)` method call in `transform` method of derived class -  `BigQueryAvroToStructuredTransformer`

PR #232, opened 2020-05-19
remove -SNAPSHOT version from release 0.15.0

elfenheart:


albertshau (commented):
lower bound in plugin parent should be bumped to 6.2.0

albertshau (approved):

PR #234, opened 2020-05-27
[CDAP-16120] Make allowSchemaRelaxation property optional

CuriousVini:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -66,6 +66,7 @@
   protected String gcsChunkSize;
 
   @Macro
+  @Nullable
   @Description("Whether to modify the BigQuery table schema if it differs from the input schema.")
   protected boolean allowSchemaRelaxation;
```
albertshau: does this work even if it's a boolean and not a Boolean? Will it get defaulted to false if no value is given?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -117,8 +118,8 @@ public String getBucket() {
     return bucket;
   }
 
-  public boolean isAllowSchemaRelaxation() {
-    return allowSchemaRelaxation;
+  public Boolean isAllowSchemaRelaxation() {
```
albertshau: should return boolean not Boolean, since it can't be null

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
CuriousVini: Modified it to be `Boolean` not sure why we used `boolean` before.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
CuriousVini: done

albertshau (approved):
lgtm

PR #239, opened 2020-06-08
CDAP-16815 - Emit metrics for insert, update and upsert from BQSink

greeshmaswaminathan:
Fix for CDAP-16815
- Emit metric from BQ sink for insert , update and upsert
- Ensure that count is correctly captured since BQ could emit large record counts

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -22,6 +22,10 @@ import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetInfo;
 import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.FieldList;
+import com.google.cloud.bigquery.Job;
+import com.google.cloud.bigquery.JobConfiguration;
+import com.google.cloud.bigquery.JobId;
+import com.google.cloud.bigquery.JobStatistics;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
@@ -75,12 +79,15 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
 
   private static final String gcsPathFormat = "gs://%s";
   private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s";
+  public static final String RECORDS_UPDATED_METRIC = "records.updated";
 
   // UUID for the run. Will be used as bucket name if bucket is not provided.
   // UUID is used since GCS bucket names must be globally unique.
   private final UUID uuid = UUID.randomUUID();
   protected Configuration baseConfiguration;
   private StructuredToAvroTransformer avroTransformer;
+  private final String jobId = UUID.randomUUID().toString();
+  private BigQuery bigQuery;
 
   /**
    * Executes main prepare run logic. Child classes cannot override this method,
@@ -98,10 +105,11 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
     Credentials credentials = serviceAccountFilePath == null ?
       null : GCPUtils.loadServiceAccountCredentials(serviceAccountFilePath);
     String project = config.getProject();
-    BigQuery bigQuery = GCPUtils.getBigQuery(project, credentials);
+    bigQuery = GCPUtils.getBigQuery(project, credentials);
     String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
     baseConfiguration = getBaseConfiguration(cmekKey);
     String bucket = configureBucket();
+    baseConfiguration.set(BigQueryConstants.CONFIG_JOB_ID, jobId);
     if (!context.isPreviewEnabled()) {
       createResources(bigQuery, GCPUtils.getStorage(project, credentials), config.getDataset(), bucket,
                       config.getLocation(), cmekKey);
@@ -112,6 +120,8 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
 
   @Override
   public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
+    recordMetric(succeeded, context);
+
     if (getConfig().getBucket() == null) {
       Path gcsPath = new Path(String.format(gcsPathFormat, uuid.toString()));
       try {
@@ -124,6 +134,47 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
         LOG.warn("Failed to delete bucket '{}': {}", gcsPath, e.getMessage());
       }
     }
+  }
+
+  private void recordMetric(boolean succeeded, BatchSinkContext context) {
+    if (!succeeded) {
+      return;
+    }
+    Job queryJob = bigQuery.getJob(getJobId());
+    if (queryJob == null) {
+      LOG.warn("Unable to find BigQuery job. No metric will be emitted for the number of affected rows.");
+      return;
+    }
+    long totalRows = getTotalRows(queryJob);
+    LOG.info("Job {} affected {} rows", queryJob.getJobId(), totalRows);
+    //work around since StageMetrics count() only takes int as of now
+    int cap = 10000; // so the loop will not cause significant delays
+    long count = totalRows / Integer.MAX_VALUE;
+    if (count > cap) {
+      LOG.warn("Total record count is too high! Metric for the number of affected rows may not be updated correctly");
+    }
+    count = count < cap ? count : cap;
+    for (int i = 0; i <= count && totalRows > 0; i++) {
+      int rowCount = totalRows < Integer.MAX_VALUE ? (int) totalRows : Integer.MAX_VALUE;
+      context.getMetrics().count(RECORDS_UPDATED_METRIC, rowCount);
+      totalRows -= rowCount;
+    }
+  }
+
+  private JobId getJobId() {
+    String location = bigQuery.getDataset(getConfig().getDataset()).getLocation();
+    return JobId.newBuilder().setLocation(location).setJob(jobId).build();
+  }
+
+  private long getTotalRows(Job queryJob) {
+    JobConfiguration.Type type = queryJob.getConfiguration().getType();
+    if (type == JobConfiguration.Type.LOAD) {
+      return ((JobStatistics.LoadStatistics) queryJob.getStatistics()).getOutputRows();
+    } else if (type == JobConfiguration.Type.QUERY) {
+      return ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getNumDmlAffectedRows();
+    }
+    LOG.warn("Unable to identify BigQuery job type. No metric will be emitted for the number of affected rows.");
+    return 0;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -112,6 +118,19 @@ public final void prepareRun(BatchSinkContext context) throws Exception {
 
   @Override
   public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
+    if (succeeded) {
+      Job queryJob = bigQuery.getJob(jobId);
```
albertshau: I believe we've had bugs in the past related to jobs in different locations. Be sure to test writing to datasets in non-default locations to make sure this works. You may have to use the method that takes a JobId and set the location on that JobId.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -112,6 +118,19 @@ public final void prepareRun(BatchSinkContext context) throws Exception {
...
+      if (queryJob == null) {
+        LOG.warn("Unable to find job, metrics cannot be collected");
+      } else {
```
albertshau: nit: in general, it's more readable to have less nesting and do these types of checks like:

```
if (!succeeded) {
  return;
}

Job queryJob = bigQuery.getJob(jobId);
if (queryJob == null) {
  LOG.warn(...);
  return;
}

// other logic here.
```

though you'll have to move this block to its own method or move it below the existing file delete logic.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
     }
   }
 
+  private void recordMetric(BatchSinkContext context, Job queryJob) throws Exception {
+    String operation = baseConfiguration.get(BigQueryConstants.CONFIG_OPERATION);
+    String metricName = String.format("records.%s", operation.toLowerCase());
```
albertshau: It may also make sense to just have a single metric like 'records.updated' rather than one per operation 'records.insert', 'records.update', 'record.upsert'. That way people only ever look for the same metric rather than changing it per pipeline.

@sreevatsanraman @bdmogal do you have any opinions here?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+    long totalRows = getTotalRows(queryJob);
```
albertshau, line 148: is this the only thing that can throw an exception? It would be cleaner to just have this return without emitting any metric rather than have it throw an exception for an unknown query type and make the caller handle an exception.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+
+  private long getTotalRows(Job queryJob) throws Exception {
+    final JobConfiguration.Type type = queryJob.getConfiguration().getType();
+    LOG.info(String.format("Job configuration type is %s" , type));
+    if (type == JobConfiguration.Type.LOAD) {
+      return ((JobStatistics.LoadStatistics) queryJob.getStatistics()).getOutputRows();
+    } else if (type == JobConfiguration.Type.QUERY) {
+      return ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getNumDmlAffectedRows();
+    }
+    throw new Exception("Unable to identify job type");
```
albertshau: avoid throwing generic Exception or RuntimeException in most situations. It makes it very difficult to do proper error handling in the caller, since they don't know what went wrong. 

In this particular situation, since we know it should never be any other type, we can either return 0 to tell the caller nothing needs to be emitted, or throw an IllegalStateException because this part of the code should never be reached.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+    count = count < cap ? count : cap;
+    for (int i = 0; i <= count && totalRows > 0; i++) {
+      int rowCount = totalRows < Integer.MAX_VALUE ? (int) totalRows : Integer.MAX_VALUE;
+      context.getMetrics().count(metricName, rowCount);
+      totalRows -= rowCount;
+    }
+  }
+
+  private long getTotalRows(Job queryJob) throws Exception {
+    final JobConfiguration.Type type = queryJob.getConfiguration().getType();
```
albertshau: doesn't need to be final

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+    LOG.info(String.format("Job configuration type is %s" , type));
```
albertshau: shouldn't log this, as it will always be dependent on the configured operation type.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
     }
   }
 
+  private void recordMetric(BatchSinkContext context, Job queryJob) throws Exception {
+    String operation = baseConfiguration.get(BigQueryConstants.CONFIG_OPERATION);
+    String metricName = String.format("records.%s", operation.toLowerCase());
+    long totalRows = getTotalRows(queryJob);
+    LOG.info(String.format("Total row count is %d" , totalRows));
+    //work around since StageMetrics count() only takes int as of now
+    final int cap = 10000; // so the loop will not cause significant delays
```
albertshau: doesn't need to be final

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +145,36 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
     }
   }
 
+  private void recordMetric(BatchSinkContext context, Job queryJob) throws Exception {
+    String operation = baseConfiguration.get(BigQueryConstants.CONFIG_OPERATION);
+    String metricName = String.format("records.%s", operation.toLowerCase());
+    long totalRows = getTotalRows(queryJob);
+    LOG.info(String.format("Total row count is %d" , totalRows));
```
albertshau: would be good to include the job id in this message. Also, logger can use '{}' as a placeholder instead of String.format().

`LOG.info("Job {} affected {} rows")`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -280,7 +282,9 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
         bigQueryHelper.getRawBigquery().datasets().get(tableRef.getProjectId(), tableRef.getDatasetId()).execute();
 
       JobReference jobReference =
-        bigQueryHelper.createJobReference(projectId, "direct-bigqueryhelper-import", dataset.getLocation());
+        new JobReference().setProjectId(projectId).
```
albertshau: style: we put the '.' on the next line:

```
new JobReference()
  .setProjectId()
```

albertshau (commented):
mostly small comments

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Thanks for pointing this out. Jobs were not being fetched for a different region. Fixed it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Updated the metric name to "records.updated"

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Fixed to log a warning and return 0.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Removed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -126,6 +136,47 @@ public final void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+  }
+
+  private long getTotalRows(Job queryJob) {
+    JobConfiguration.Type type = queryJob.getConfiguration().getType();
+    if (type == JobConfiguration.Type.LOAD) {
+      return ((JobStatistics.LoadStatistics) queryJob.getStatistics()).getOutputRows();
+    } else if (type == JobConfiguration.Type.QUERY) {
+      return ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getNumDmlAffectedRows();
+    }
+    LOG.warn("Unable to identify job type. Could not find affected rows");
```
albertshau: message should tell the user what the impact is. Something like:

Unable to identify BigQuery job type. No metric will be emitted for the number of affected rows.

albertshau (approved):
minor comment about an exception message, otherwise lgtm.

please fix, squash commits, then merge

PR #241, opened 2020-06-11
CDAP-16880 and CDAP-16879 - Fix BQ sink validations

greeshmaswaminathan:
- Fixes for CDAP-16880 and CDAP-16879
- If truncate table is set, do not validate for schema. This will work currently for partitioned tables since we are not allowing insertion into specific partitions
- Truncate table and update schema should not be set together

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -330,6 +330,13 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
       return;
     }
 
+    if (getConfig().isTruncateTableSet()) {
+      //no validation required for schema if truncate table is set.
+      // BQ will overwrite the schema for normal tables when write disposition is WRITE_TRUNCATE
+      //note - If write to single partition is supported in future, schema validation will be necessary
+      return;
+    }
+
     FieldList bqFields = bqSchema.getFields();
     List<Schema.Field> outputSchemaFields = Objects.requireNonNull(tableSchema.getFields());
 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -42,6 +42,7 @@ public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig 
   public static final String NAME_TRUNCATE_TABLE = "truncateTable";
   public static final String NAME_LOCATION = "location";
   private static final String NAME_GCS_CHUNK_SIZE = "gcsChunkSize";
+  private static final String NAME_UPDATE_SCHEMA = "allowSchemaRelaxation";
 
   @Name(NAME_DATASET)
   @Macro
@@ -65,6 +66,7 @@ public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig 
     "number of bytes. By default, 8388608 bytes (8MB) will be used as upload request chunk size.")
   protected String gcsChunkSize;
 
+  @Name(NAME_UPDATE_SCHEMA)
   @Macro
   @Nullable
   @Description("Whether to modify the BigQuery table schema if it differs from the input schema.")
@@ -74,7 +76,7 @@ public abstract class AbstractBigQuerySinkConfig extends GCPReferenceSinkConfig 
   @Macro
   @Nullable
   @Description("Whether or not to truncate the table before writing to it. "
-    + "Should only be used with the Insert operation.")
+    + "Should only be used with the Insert operation. This could overwrite the table schema")
   protected Boolean truncateTable;
 
   @Name(NAME_LOCATION)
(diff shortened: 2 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -386,6 +392,11 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
     collector.getOrThrowException();
   }
 
+  private Boolean isTruncateTableSet() {
+    AbstractBigQuerySinkConfig config = getConfig();
+    return (config == null || config.truncateTable == null) ? false : config.truncateTable;
```
albertshau: config will never be null, this method can be replaced with a call to config's isTruncateTableSet

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -330,6 +330,12 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
       return;
     }
 
+    if (isTruncateTableSet()) {
+      //no validation required for schema if truncate table is set.
```
albertshau, line 334: why isn't this needed? BQ will change the schema of the table change after a truncate? If so, add that information to the comment.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -138,5 +144,12 @@ public void validate(FailureCollector collector) {
     if (!containsMacro(NAME_DATASET)) {
       BigQueryUtil.validateDataset(dataset, NAME_DATASET, collector);
     }
+
+    //note - These options could be set together for write to single partition in the future
+    if (isTruncateTableSet() && isAllowSchemaRelaxation()) {
```
albertshau: I don't understand why these can't be used together. From the comment earlier, I assume that when the table is being truncated, the schema updates automatically.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Yes, truncate table sets writeDisposition as WRITE_TRUNCATE and this overwrites the schema. I will add details to comment

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
greeshmaswaminathan: If both are set, the BQ load fails with this message  - "Could not commit Job: Error occurred while importing data to BigQuery 'Schema update options should only be specified with WRITE_APPEND disposition, or with WRITE_TRUNCATE disposition on a table partition.' 
" .  https://b.corp.google.com/issues/155237532#comment7 gives some more insights as to why they are incompatible.

albertshau: that means we should not be setting both of those options for the BQ load, but it doesn't mean the user shouldn't be able to set these at the plugin level. From the user's perspective, the schema can be updated when the table is truncated, as it can be updated to anything. 

The fix should be that we don't set schema update options for the load job if the write disposition doesn't allow it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
greeshmaswaminathan: Changed to ensure schema update options are not set if write disposition is WRITE_TRUNCATE. User will not be given a validation error

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -255,7 +255,10 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
       } else {
         loadConfig.setDestinationTable(tableRef);
 
-        if (allowSchemaRelaxation) {
+        //Schema update options should only be specified with WRITE_APPEND disposition,
+        // or with WRITE_TRUNCATE disposition on a table partition - The logic below should change when we support
+        // insertion into single partition
+        if (allowSchemaRelaxation && !JobInfo.WriteDisposition.WRITE_TRUNCATE.toString().equals(writeDisposition)) {
```
albertshau: it's usually better to change the string into the Enum and compare enums instead of changing enum to string and comparing.

albertshau (approved):
minor comment, otherwise lgtm

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
greeshmaswaminathan: Done

PR #244, opened 2020-06-23
CDAP-17003  Remove unnecessary schema validations in initialize 

greeshmaswaminathan:
- Removed schema validations from BQ source plugin intialize method since the schema is already validated in prepareRun
- Removed call to find schema when macro is present and configuration does not have a schema. The transform will now use the schema from the GenericRecord .
- This avoids all calls to BigQuery.getTable() in initialize method.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
@@ -36,6 +36,16 @@ import javax.annotation.Nullable;
  * Create StructuredRecords from GenericRecords. Contains custom logic for BigQuery date and time types.
  */
 public class BigQueryAvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
+
+  private Schema genericRecordSchema;
+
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -36,6 +36,11 @@
  * Create StructuredRecords from GenericRecords. Contains custom logic for BigQuery date and time types.
  */
 public class BigQueryAvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
+
+  public StructuredRecord transform(GenericRecord genericRecord) throws IOException {
+    return transform(genericRecord, Schema.parseJson(genericRecord.getSchema().toString()));
```
albertshau: parsing schema is an expensive operation - we should avoid doing it for every single record.

Since we know the schema will always be the same for all records, we can have a Schema variable that gets set for the very first record so that we only parse once.

greeshmaswaminathan: Done.

albertshau (approved):

PR #257, opened 2020-07-16
[CDAP-16999] Add Validations and Clarify Error Messages for GCS Plugins

dli357:
See [CDAP-16999](https://issues.cask.co/browse/CDAP-16999).

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Cask Data, Inc.
+ * Copyright © 2019-2020 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -22,6 +22,7 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.CopyWriter;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
 import io.cdap.plugin.gcp.common.GCPUtils;
@@ -82,12 +83,28 @@ public class StorageClient {
   private void pairTraverse(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite,
                             Consumer<BlobPair> consumer) {
 
-    Bucket sourceBucket = storage.get(sourcePath.getBucket());
+    Bucket sourceBucket = null;
+    try {
+      sourceBucket = storage.get(sourcePath.getBucket());
+    } catch (StorageException e) {
+      // Add more descriptive error message
+      throw new RuntimeException(
+        String.format("Unable to access source bucket %s. ", sourcePath.getBucket())
+          + "Ensure you entered the correct bucket path.", e);
+    }
     if (sourceBucket == null) {
       throw new IllegalArgumentException(
         String.format("Source bucket '%s' does not exist.", sourcePath.getBucket()));
     }
-    Bucket destBucket = storage.get(destPath.getBucket());
+    Bucket destBucket = null;
+    try {
+      destBucket = storage.get(destPath.getBucket());
+    } catch (StorageException e) {
+      // Add more descriptive error message
+      throw new RuntimeException(
+        String.format("Unable to access destination bucket %s. ", destPath.getBucket())
+          + "Ensure you entered the correct bucket path.", e);
+    }
     if (destBucket == null) {
       throw new IllegalArgumentException(
         String.format("Destination bucket '%s' does not exist. Please create it first.", destPath.getBucket()));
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -192,11 +193,29 @@ public boolean failIfExists() {
 
     void validate(FailureCollector collector) {
       if (!containsMacro("paths")) {
+        Storage storage = null;
+        try {
+          Credentials credentials = getServiceAccountFilePath() == null ?
+            null : GCPUtils.loadServiceAccountCredentials(getServiceAccountFilePath());
+          storage = GCPUtils.getStorage(getProject(), credentials);
+        } catch (IOException e) {
+          collector.addFailure(e.getMessage(), "Ensure you entered the correct file path.")
```
CuriousVini: what kind of message is displayed ? Is `IOException` only thrown when path is not correct?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -192,11 +193,29 @@ public boolean failIfExists() {
...
-            GCSPath.from(path);
+            GCSPath gcsPath = GCSPath.from(path);
+            // Check if bucket exists and is accessible
+            if (storage != null) {
+              storage.get(gcsPath.getBucket());
+            }
           } catch (IllegalArgumentException e) {
             collector.addFailure(e.getMessage(), null).withConfigElement(NAME_PATHS, path);
+          } catch (StorageException e) {
+            collector.addFailure(e.getMessage(), "Ensure you entered the correct bucket path.")
```
CuriousVini: Is `StorageException` only thrown when bucket path is not correct?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -192,11 +193,29 @@ public boolean failIfExists() {
...
+            .withConfigProperty(NAME_SERVICE_ACCOUNT_FILE_PATH)
+            .withStacktrace(e.getStackTrace());
+        }
         for (String path : getPaths()) {
           try {
-            GCSPath.from(path);
+            GCSPath gcsPath = GCSPath.from(path);
+            // Check if bucket exists and is accessible
+            if (storage != null) {
+              storage.get(gcsPath.getBucket());
```
chtyim: It shouldn't be a validation error if the bucket doesn't exist during the pipeline deployment time. Also, what does the getProject() returns if it is auto-detect? Does it return the project where CDAP is running or the target Dataproc project?

dli357: I believe it should be where the getProject() is run since it just checks for the GOOGLE_CLOUD_PROJECT environment variable where it is running. I.e. CDAP during configuration, dataproc during runtime?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
dli357: In this case, it should be some kind of "File Not Found" error or, in cases where it is unable to open the file for some reason, it will display that. However, this is outdated since we are moving these checks to happen only at runtime.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
dli357: I believe that StorageException is simply an exception which encapsulates all possible API responses from the GCS API. However, this is outdated since we are moving these checks to happen only at runtime. In most cases, an exception happens due to lack of permissions to access a bucket (403). See https://cloud.google.com/storage/docs/json_api/v1/status-codes

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -82,12 +83,32 @@ public void move(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolea
...
-    Bucket sourceBucket = storage.get(sourcePath.getBucket());
+    Bucket sourceBucket = null;
+    try {
+      sourceBucket = storage.get(sourcePath.getBucket());
+    } catch (StorageException e) {
+      // Add more descriptive error message
+      RuntimeException re = new RuntimeException(
+        String.format("Unable to access source bucket %s. ", sourcePath.getBucket())
+          + "Ensure you entered the correct bucket path.");
+      re.initCause(e);
```
chtyim: Just do

```
throw new RuntimeException(String.format(....), e);
```

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
dli357: Although the permissions error is not the only message which can occur as a result of the StorageException, it is the most likely response assuming GCS doesn't throw a 500 or something. Another option is to parse the response error code and use that instead to determine when to throw a more detailed error message.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dli357: Fixed

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -192,6 +203,7 @@ public boolean failIfExists() {
 
     void validate(FailureCollector collector) {
       if (!containsMacro("paths")) {
+        Storage storage = null;
```
chtyim: Unused variable?

dli357: Fixed

chtyim (approved):

PR #264, opened 2020-07-28
BigQuery sink enhancements

ardian4:
Added: Support writing to a different project
Added: Support integer range partitioning

JIRA:
https://issues.cask.co/browse/PLUGIN-172

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.gcp.bigquery.sink;
 
 import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.RangePartitioning;
 import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TimePartitioning;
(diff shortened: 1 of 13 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -283,21 +345,60 @@ private void validatePartitionProperties(@Nullable Schema schema, FailureCollect
...
-      if (timePartitioning != null && timePartitioning.getField() != null
-        && !timePartitioning.getField().equals(partitionByField)) {
-        collector.addFailure(String.format("Destination table '%s' is partitioned by column '%s'.",
-                                           table.getTableId().getTable(),
-                                           timePartitioning.getField()),
-                             String.format("Set the partition field to '%s'.", timePartitioning.getField()))
-          .withConfigProperty(NAME_PARTITION_BY_FIELD);
+      RangePartitioning rangePartitioning = tableDefinition.getRangePartitioning();
+      if (timePartitioning == null && rangePartitioning == null && shouldCreatePartitionedTable()) {
+        LOG.warn(String.format(
```
rmstar, line 350: I'm curious why we just log an error instead of failing validation in this case? I see that we were following the same pattern in existing code (line 282). @albertshau ?

rmstar (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
albertshau: auto-create properties are a little funny because they only apply once. I don't remember the exact reason we decided this shouldn't be an error, but its good to follow the existing behavior.

In general, we should have failures for anything that will never be correct, like having a negative number when it's not allowed. We're more lax about things that can be fixed, like if a GCS path doesn't exist when we expect it to, as we don't want pipeline deployment to be dependent on having the environment set up. This category of potential problems fits better as a warning instead of a failure, though we don't currently have a way to surface these types of things anywhere except the logs.

CuriousVini (commented):
Please squash the commits

CuriousVini (approved):

PR #267, opened 2020-08-04
Emit records.updated metric from GCS sink.

greeshmaswaminathan:
Fix for https://issues.cask.co/browse/CDAP-17123
- Emit records.updated metric for GCS sink
- Depends on hydrator-plugins fix in AbstractFileSink.java (PR - https://github.com/cdapio/hydrator-plugins/pull/1157)
- Depends on version 2.5.0-SNAPSHOT of cdap plugin due to dependency on AbstractFileSink change
- RecordWriter keeps a running counter of the fields written
- The count for each output file as calculated by the RecordWriter is updated as metadata on the file itself at the end of each task commit.
- The total is calculated at the end of run and updated in the metric.
-Using Storage API to list blobs

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -28,8 +28,10 @@ import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
+import io.cdap.cdap.etl.api.StageMetrics;
 import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
+import io.cdap.cdap.etl.api.validation.ValidatingOutputFormat;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.format.FileFormat;
 import io.cdap.plugin.format.plugin.AbstractFileSink;
@@ -37,6 +39,9 @@ import io.cdap.plugin.format.plugin.FileSinkProperties;
 import io.cdap.plugin.gcp.common.GCPReferenceSinkConfig;
 import io.cdap.plugin.gcp.common.GCPUtils;
 import io.cdap.plugin.gcp.gcs.GCSPath;
+import io.cdap.plugin.gcp.gcs.StorageClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.text.SimpleDateFormat;
@@ -52,7 +57,14 @@ import javax.annotation.Nullable;
 @Name("GCS")
 @Description("Writes records to one or more files in a directory on Google Cloud Storage.")
 public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConfig> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(GCSBatchSink.class);
+  public static final String RECORD_COUNT = "recordcount";
+  private static final String RECORDS_UPDATED_METRIC = "records.updated";
+
   private final GCSBatchSinkConfig config;
+  private String outputPath;
+
 
   public GCSBatchSink(GCSBatchSinkConfig config) {
     super(config);
@@ -62,6 +74,18 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
+  }
+
+  @Override
+  public ValidatingOutputFormat getValidatingOutputFormat(PipelineConfigurer pipelineConfigurer) {
+    ValidatingOutputFormat delegate = super.getValidatingOutputFormat(pipelineConfigurer);
+    return new GCSOutputFormatProvider(delegate);
+  }
+
+  @Override
+  public ValidatingOutputFormat getOutputFormatForRun(BatchSinkContext context) throws InstantiationException {
+    ValidatingOutputFormat outputFormatForRun = super.getOutputFormatForRun(context);
+    return new GCSOutputFormatProvider(outputFormatForRun);
   }
 
   @Override
@@ -90,8 +114,66 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   }
 
   @Override
+  protected String getOutputDir(long logicalStartTime) {
+    this.outputPath = super.getOutputDir(logicalStartTime);
+    return this.outputPath;
+  }
+
+  @Override
   protected void recordLineage(LineageRecorder lineageRecorder, List<String> outputFields) {
     lineageRecorder.recordWrite("Write", "Wrote to Google Cloud Storage.", outputFields);
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, BatchSinkContext context) {
+    super.onRunFinish(succeeded, context);
+    emitMetrics(succeeded, context);
+  }
+
+  private void emitMetrics(boolean succeeded, BatchSinkContext context) {
+    if (!succeeded) {
+      return;
+    }
+
+    try {
+      StorageClient storageClient = StorageClient.create(config.getProject(), config.getServiceAccountFilePath());
+      storageClient.mapMetaDataForAllBlobs(outputPath, new MetricsEmitter(context.getMetrics())::emitMetrics);
+    } catch (Exception e) {
+      LOG.warn("Metrics for the number of affected rows in GCS Sink maybe incorrect.", e);
+    }
+  }
+
+  private static class MetricsEmitter {
+    private StageMetrics stageMetrics;
+
+    private MetricsEmitter(StageMetrics stageMetrics) {
+      this.stageMetrics = stageMetrics;
+    }
+
+    public void emitMetrics(Map<String, String> metaData) {
+      long totalRows = extractRecordCount(metaData);
+      if (totalRows == 0) {
+        return;
+      }
+
+      //work around since StageMetrics count() only takes int as of now
+      int cap = 10000; // so the loop will not cause significant delays
+      long count = totalRows / Integer.MAX_VALUE;
+      if (count > cap) {
+        LOG.warn("Total record count is too high! Metric for the number of affected rows may not be updated correctly");
+      }
+      count = count < cap ? count : cap;
+      for (int i = 0; i <= count && totalRows > 0; i++) {
+        int rowCount = totalRows < Integer.MAX_VALUE ? (int) totalRows : Integer.MAX_VALUE;
+        stageMetrics.count(RECORDS_UPDATED_METRIC, rowCount);
+        totalRows -= rowCount;
+      }
+    }
+
+    private long extractRecordCount(Map<String, String> metadata) {
+      String value = metadata.get(RECORD_COUNT);
+      return value == null ? 0L : Long.parseLong(value);
+    }
   }
 
   /**
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
@@ -1,0 +1,250 @@
+package io.cdap.plugin.gcp.gcs.sink;
+
+import com.google.cloud.storage.Blob;
+import com.google.common.annotations.VisibleForTesting;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.etl.api.validation.FormatContext;
+import io.cdap.cdap.etl.api.validation.ValidatingOutputFormat;
+import io.cdap.plugin.gcp.common.GCPUtils;
+import io.cdap.plugin.gcp.gcs.StorageClient;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.JobStatus;
+import org.apache.hadoop.mapreduce.OutputCommitter;
+import org.apache.hadoop.mapreduce.OutputFormat;
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+import org.apache.hadoop.util.ReflectionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * OutputFormatProvider for GCSSink
+ */
+public class GCSOutputFormatProvider implements ValidatingOutputFormat {
+
+  private static final Logger LOG = LoggerFactory.getLogger(GCSOutputFormatProvider.class);
+  private static final String DELEGATE_OUTPUTFORMAT_CLASSNAME = "gcssink.delegate.outputformat.classname";
+  private static final String OUTPUT_FOLDER = "gcssink.metric.output.folder";
+  public static final String RECORD_COUNT_FORMAT = "recordcount.%s";
+  private final ValidatingOutputFormat delegate;
+
+  public GCSOutputFormatProvider(ValidatingOutputFormat delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void validate(FormatContext context) {
+    delegate.validate(context);
+  }
+
+  @Override
+  public String getOutputFormatClassName() {
+    return GCSOutputFormat.class.getName();
+  }
+
+  @Override
+  public Map<String, String> getOutputFormatConfiguration() {
+    Map<String, String> outputFormatConfiguration = new HashMap<>(delegate.getOutputFormatConfiguration());
+    outputFormatConfiguration.put(DELEGATE_OUTPUTFORMAT_CLASSNAME, delegate.getOutputFormatClassName());
+    return outputFormatConfiguration;
+  }
+
+  /**
+   * OutputFormat for GCS Sink
+   */
+  public static class GCSOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+    private OutputFormat delegateFormat;
+
+    private OutputFormat getDelegateFormatInstance(Configuration configuration) throws IOException {
+      if (delegateFormat != null) {
+        return delegateFormat;
+      }
+
+      String delegateClassName = configuration.get(DELEGATE_OUTPUTFORMAT_CLASSNAME);
+      try {
+        delegateFormat = (OutputFormat) ReflectionUtils
+          .newInstance(configuration.getClassByName(delegateClassName), configuration);
+        return delegateFormat;
+      } catch (ClassNotFoundException e) {
+        throw new IOException(
+          String.format("Unable to instantiate output format for class %s", delegateClassName),
+          e);
+      }
+    }
+
+    @Override
+    public RecordWriter<NullWritable, StructuredRecord> getRecordWriter(TaskAttemptContext taskAttemptContext) throws
+      IOException, InterruptedException {
+      RecordWriter originalWriter = getDelegateFormatInstance(taskAttemptContext.getConfiguration())
+        .getRecordWriter(taskAttemptContext);
+      return new GCSRecordWriter(originalWriter);
+    }
+
+    @Override
+    public void checkOutputSpecs(JobContext jobContext) throws IOException, InterruptedException {
+      getDelegateFormatInstance(jobContext.getConfiguration()).checkOutputSpecs(jobContext);
+    }
+
+    @Override
+    public OutputCommitter getOutputCommitter(TaskAttemptContext taskAttemptContext) throws IOException,
+      InterruptedException {
+      OutputCommitter delegateCommitter = getDelegateFormatInstance(taskAttemptContext.getConfiguration())
+        .getOutputCommitter(taskAttemptContext);
+      return new GCSOutputCommitter(delegateCommitter);
+    }
+  }
+
+  /**
+   * OutputCommitter for GCS
+   */
+  public static class GCSOutputCommitter extends OutputCommitter {
+
+    private final OutputCommitter delegate;
+
+    public GCSOutputCommitter(OutputCommitter delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public void setupJob(JobContext jobContext) throws IOException {
+      delegate.setupJob(jobContext);
+    }
+
+    @Override
+    public void cleanupJob(JobContext jobContext) throws IOException {
+      delegate.cleanupJob(jobContext);
+    }
+
+    @Override
+    public void commitJob(JobContext jobContext) throws IOException {
+      delegate.commitJob(jobContext);
+    }
+
+    @Override
+    public void abortJob(JobContext jobContext, JobStatus.State state) throws IOException {
+      delegate.abortJob(jobContext, state);
+    }
+
+    @Override
+    public void setupTask(TaskAttemptContext taskAttemptContext) throws IOException {
+      delegate.setupTask(taskAttemptContext);
+    }
+
+    @Override
+    public boolean needsTaskCommit(TaskAttemptContext taskAttemptContext) throws IOException {
+      return delegate.needsTaskCommit(taskAttemptContext);
+    }
+
+    @Override
+    public void commitTask(TaskAttemptContext taskAttemptContext) throws IOException {
+      /*On commit task, there seems to be some inconsistency across different hadoop implementations regarding the path
+       where output file is stored. For some implementations it appears in the path returned by FileOutputCommitter
+       getCommittedTaskPath and for some it does not.Before commit, the files appear to be consistently present in path
+       returned by FileOutputCommitter getTaskAttemptPath. Hence, find the output file from taskAttemptPath and add
+       metadata before commit happens. After commit, file would have been moved out of the taskAttemptPath. */
+      try {
+        updateMetricMetaData(taskAttemptContext);
+      } catch (Exception exception) {
+        LOG.warn("Unable to record metric for task. Metric emitted for the number of affected rows may be incorrect.",
+                 exception);
+      }
+
+      delegate.commitTask(taskAttemptContext);
+    }
+
+    private void updateMetricMetaData(TaskAttemptContext taskAttemptContext) throws IOException {
+      if (!(delegate instanceof FileOutputCommitter)) {
+        return;
+      }
+
+      FileOutputCommitter fileOutputCommitter = (FileOutputCommitter) delegate;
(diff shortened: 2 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -64,6 +81,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  @Override
+  public ValidatingOutputFormat getValidatingOutputFormat(PipelineConfigurer pipelineConfigurer) {
+    ValidatingOutputFormat delegate = super.getValidatingOutputFormat(pipelineConfigurer);
+    return new GCSOutputFormatProvider(delegate);
+  }
+
+  @Override
+  public ValidatingOutputFormat getOutputFormatForRun(BatchSinkContext context) throws InstantiationException {
+    ValidatingOutputFormat outputFormatForRun = super.getOutputFormatForRun(context);
+    return  new GCSOutputFormatProvider(outputFormatForRun);
```
albertshau: extra space after 'return'

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
       GCPUtils.createBucket(storage, config.getBucket(), config.getLocation(), cmekKey);
     }
+
   }
 
   @Override
   protected Map<String, String> getFileSystemProperties(BatchSinkContext context) {
-    return GCPUtils.getFileSystemProperties(config, config.getPath(), new HashMap<>());
+    return GCPUtils
+      .getFileSystemProperties(config, config.getPath(), new HashMap<>());
```
albertshau: doesn't need to be a newline, we have a 120 char max line length

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+  }
+
+  @Override
+  protected String getOutputDir(long logicalStartTime) {
+    this.outputPath =  super.getOutputDir(logicalStartTime);
```
albertshau: extra space before super

albertshau: also, looks like this doesn't need to override

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+    recordMetric(succeeded, context);
+  }
+
+  private void recordMetric(boolean succeeded, BatchSinkContext context) {
+    if (!succeeded) {
+      return;
+    }
+
+    try {
+     long  totalRows = getTotalRows(context);
```
albertshau: extra space after long

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+      LOG.info("Affected {} rows", totalRows);
+      //work around since StageMetrics count() only takes int as of now
```
albertshau, line 159: instead of summing then emitting metric, you can emit metrics as you go through the file list. That would make it less likely to have to do this loop. Metrics are aggregated in-memory before being flushed so there isn't a problem making multiple count() calls.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+
+  private long getTotalRows(BatchSinkContext context) throws IOException {
+    GCSPath gcsPath = GCSPath.from(outputPath);
+    //add all the configurations necessary to create file system
+    Configuration configuration = new Configuration();
+    getFileSystemProperties(context).entrySet().stream()
+      .forEach(entry -> configuration.set(entry.getKey(), entry.getValue()));
+
+    try (FileSystem fileSystem = getFileSystem(outputPath, configuration)) {
+      if (!(fileSystem instanceof GoogleHadoopFileSystem)) {
```
albertshau: it may be simpler and also potentially more efficient to use the Storage client directly instead of the Hadoop FileSystem abstraction. For example, the metadata might come back in the list call directly, instead of requiring a list and then a get for each object.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,98 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+  private long getRecordCount(GoogleCloudStorageFileSystem gcsFs, URI uri) {
+    try {
+      GCSPath objPath = GCSPath.from(uri.toString());
+      Map<String, byte[]> metadata = gcsFs.getGcs()
+        .getItemInfo(new StorageResourceId(objPath.getBucket(), objPath.getName())).getMetadata();
+      byte[] value = metadata.getOrDefault(RECORD_COUNT, Long.valueOf(0).toString().getBytes());
+      return Long.parseLong(new String(value));
+    } catch (IOException e) {
+      //Required to handle checked exceptions in Stream API
+      throw new RuntimeException(e);
```
albertshau: if we're unable to get this information for some reason, it should result in a warning in the log, but it shouldn't cause the pipeline run to fail. I'm not sure if throwing an exception from onRunFinish() will cause it to fail or not, but it seems like something that should be handled by the plugin and not propagated up.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,274 @@
...
+  /**
+   * OutputFormat for GCS Sink
+   */
+  public static class GCSOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+    private OutputFormat delegateFormat;
+
+
+    private OutputFormat getDelegateFormatInstance(Configuration configuration) throws IOException {
+      String delegateClassName = configuration.get(DELEGATE_OUTPUTFORMAT_CLASSNAME);
+      if (delegateFormat == null) {
```
albertshau: nit: it's more readable to check the inverse, as it removes one layer of nesting:

```
if (delegateFormat != null) {
  return delegateFormat;
}
...
```

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        synchronized (this) {
```
albertshau: I don't think can be called concurrently, did you see this pattern from somewhere else?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  /**
+   * RecordWriter for GCSSink
+   */
+  public static class GCSRecordWriter extends RecordWriter<NullWritable, StructuredRecord> {
+
+    private RecordWriter originalWriter;
+    private AtomicLong recordCount;
```
albertshau: don't think this needs to be an AtomicLong, write() isn't called concurrently

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,274 @@
...
+      delegate.recoverTask(taskContext);
+    }
+  }
+
+  /**
+   * RecordWriter for GCSSink
+   */
+  public static class GCSRecordWriter extends RecordWriter<NullWritable, StructuredRecord> {
+
+    private RecordWriter originalWriter;
```
albertshau: nit: private final

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: There is a catch block in recordMetric() which will handle all exceptions and log it so pipeline run does not fail. Throwing the exception here will ensure wrong count is not recorded.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: This method is overridden to store the value of outputPath. Since the path is appended with a timeSuffix, the exact output path is not known otherwise.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: GCS does not really support a listing of files based on directory, it just lists all objects in the bucket. The GoogleCloudStorageFileSystem does some prefix matching and filtering to make the listing of files in directory happen. We will have to rewrite that logic if we use Storage API. Also, just found another method that returns full details including metadata in the GoogleCloudStorageFileSystem API - if it works , we can avoid the additional metadata calls.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: No, will fix to remove all concurrency constructs.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Fixed.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Fixed the space.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: done.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Removed.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,95 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+      }
+
+      GoogleCloudStorageFileSystem gcsFs = ((GoogleHadoopFileSystem) fileSystem).getGcsFs();
+      try {
+        //For each file, read metadata to get count and sum them up
+        List<FileInfo> fileInfos = gcsFs.listAllFileInfoForPrefix(gcsPath.getUri());
+        fileInfos.stream()
+          .forEach(fileInfo -> recordMetric(context.getMetrics(), extractRecordCount(fileInfo.getAttributes())));
+      } finally {
+        gcsFs.close();
```
chtyim: Is it safe to close since it actually is the underlying GcsFs of the FileSystem?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,95 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+      if (!(fileSystem instanceof GoogleHadoopFileSystem)) {
+        return;
+      }
+
+      GoogleCloudStorageFileSystem gcsFs = ((GoogleHadoopFileSystem) fileSystem).getGcsFs();
+      try {
+        //For each file, read metadata to get count and sum them up
+        List<FileInfo> fileInfos = gcsFs.listAllFileInfoForPrefix(gcsPath.getUri());
+        fileInfos.stream()
+          .forEach(fileInfo -> recordMetric(context.getMetrics(), extractRecordCount(fileInfo.getAttributes())));
```
chtyim: Use `map` to make it cleaner

```
fileInfos.stream()
  .map(FileInfo::getAttributes)
  .map(this::extractRecordCount)
  .forEach(count -> recordMetric(context.getMetrics(), count));
```

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,95 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+      } finally {
+        gcsFs.close();
+      }
+
+    }
+  }
+
+  private void recordMetric(StageMetrics metrics, long totalRows) {
```
chtyim: Call it emitMetrics is more appropriate

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,95 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+    count = count < cap ? count : cap;
+    for (int i = 0; i <= count && totalRows > 0; i++) {
+      int rowCount = totalRows < Integer.MAX_VALUE ? (int) totalRows : Integer.MAX_VALUE;
+      metrics.count(RECORDS_UPDATED_METRIC, rowCount);
+      totalRows -= rowCount;
+    }
+  }
+
+  private long extractRecordCount(Map<String, byte[]> metadata) {
+    byte[] value = metadata.getOrDefault(RECORD_COUNT, Long.valueOf(0).toString().getBytes());
```
chtyim: ```String.valueOf(0L).getBytes(StandardCharsets.UTF_8);```

Always provide a Charset when calling getBytes on String.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -82,18 +111,95 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+    return Long.parseLong(new String(value));
```
chtyim: Provide Charset

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
chtyim: Also, seems simpler if you do:

```
byte[] value = metadata.get(RECORD_COUNT);
return value == null ? 0L : Long.parseLong(new String(value, StandardCharsets.UTF_8));
```

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,268 @@
...
+    return outputFormatConfiguration;
+  }
+
+  /**
+   * OutputFormat for GCS Sink
+   */
+  public static class GCSOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+    private OutputFormat delegateFormat;
+
+
```
chtyim: Remove extra empty line

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,268 @@
...
+      try {
+        delegateFormat = (OutputFormat) ReflectionUtils
+          .newInstance(configuration.getClassByName(delegateClassName), configuration);
+      } catch (ClassNotFoundException e) {
+        throw new IOException(
+          String.format("Unable to instantiate output format for class %s", delegateClassName),
+          e);
+      }
+      return delegateFormat;
```
chtyim: Move this return inside the try

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,268 @@
...
+
+  /**
+   * OutputFormat for GCS Sink
+   */
+  public static class GCSOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+    private OutputFormat delegateFormat;
+
+
+    private OutputFormat getDelegateFormatInstance(Configuration configuration) throws IOException {
+      String delegateClassName = configuration.get(DELEGATE_OUTPUTFORMAT_CLASSNAME);
```
chtyim: Move this after the if.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    @Override
+    public OutputCommitter getOutputCommitter(TaskAttemptContext taskAttemptContext) throws IOException,
+      InterruptedException {
+      OutputCommitter delegateCommitter = getDelegateFormatInstance(taskAttemptContext.getConfiguration())
+        .getOutputCommitter(taskAttemptContext);
+      return new GCSOutputCommitter(delegateCommitter);
+    }
+
```
chtyim: Remove extra new line

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,268 @@
...
+        List<URI> files = gcsFs.listFileNames(gcsFs.getFileInfo(committedTaskPath.toUri()));
+        if (files == null || files.isEmpty()) {
+          return;
+        }
+        //update metadata on any file - typically only one will be there
+        GCSPath objPath = GCSPath.from(files.get(0).toString());
+        //read the count from configuration
+        String keyInConfig = String.format(RECORD_COUNT_FORMAT, taskAttemptContext.getTaskAttemptID());
+        Map<String, byte[]> metaData = new HashMap<>();
+        byte[] bytes = Long.valueOf(configuration.getLong(keyInConfig, 0)).toString().getBytes();
```
chtyim: Provide charset. Also simpler to use `String.valueOf(conf.getLong(keyInConfig, 0L))`

src/test/java/io/cdap/plugin/gcp/gcs/GCSOutputformatProviderTest.java
```diff
@@ -0,0 +1,115 @@
...
+    Mockito.when(gcsfsMock.listFileNames(Mockito.any())).thenReturn(files);
+    GoogleCloudStorage gcs = Mockito.mock(GoogleCloudStorage.class);
+    Mockito.when(gcsfsMock.getGcs()).thenReturn(gcs);
+
+    committer.commitTask(mockContext);
+    Mockito.verify(fileOutputCommitter, Mockito.times(1)).commitTask(mockContext);
+    Mockito.verify(gcs, Mockito.times(1)).updateItems(Mockito.any());
+
+  }
+
```
chtyim: Remove extra empty lines

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
albertshau: oh ok i see

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
albertshau: I see. Then I suggest propagating the exception as-is (declaring the method to throw IOException) to avoid wrapping unneccessarily

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
albertshau: there is a way to give a prefix as one of the options when using the list API, which is how "directory" listing is normally done. If this is what the GoogleCloudStorageFileSystem is doing then both approaches should be the same performance wise and it doesn't really matter. But if it is post filtering instead of passing in the option, then we should use the Storage API.

greeshmaswaminathan: Fixed to use Storage API

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Not applicable after latest changes

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Not applicable after changing code to use Storage API.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Not applicable after changing code to use Storage API.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Not applicable after changing code to use Storage API.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done

src/test/java/io/cdap/plugin/gcp/gcs/GCSOutputformatProviderTest.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    @Override
+    public boolean needsTaskCommit(TaskAttemptContext taskAttemptContext) throws IOException {
+      return delegate.needsTaskCommit(taskAttemptContext);
+    }
+
+    @Override
+    public void commitTask(TaskAttemptContext taskAttemptContext) throws IOException {
+      //Update the metric before committing the file.
```
albertshau: can you add some comments about what you found around why we need to update the metadata before the commit and not after

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -0,0 +1,246 @@
...
+      if (configuration == null || taskAttemptPath == null) {
+        return;
+      }
+
+      //read the count from configuration
+      String keyInConfig = String.format(RECORD_COUNT_FORMAT, taskAttemptContext.getTaskAttemptID());
+      Map<String, String> metaData = new HashMap<>();
+      metaData.put(GCSBatchSink.RECORD_COUNT, String.valueOf(configuration.getLong(keyInConfig, 0L)));
+      StorageClient storageClient = getStorageClient(configuration);
+      //update metadata on any file - typically only one will be there
```
albertshau: typically, or always? is the taskAttemptPath a file or a directory? If it's not always a single blob, wouldn't the metrics end up incorrect?

albertshau (commented):
one minor comment and one question

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: taskAttemptPath is a directory. The exact blob name depends on the format. Yes in this case the expectation is to find a unique file because each task attempt path contains the task Id in it and record writer for that task creates a unique file with task Id as well. I can change the comment to reflect that.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
greeshmaswaminathan: Done.

albertshau (approved):

PR #269, opened 2020-08-07
[PLUGIN-368] Disregard Some Validation Failures in Deployment

dli357:
See [PLUGIN-368](https://issues.cask.co/browse/PLUGIN-368) for more information.

Changes SpannerSource, Datastore source and sink, Bigtable source and sink, and BigQuery source and sink so that connection failure, resource access failure, and resource permissions failures do not inhibit deployment of a pipeline.

On another note, this highlights the need for the Validation API to be enhanced to include separate failure types, such as failures which should occur in runtime but not in deployment time or warnings.

EDIT: Seems like we need to consolidate our validation approach before we can fix these issues. Closing this PR for now.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Cask Data, Inc.
+ * Copyright © 2019-2020 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -16,6 +16,7 @@
 package io.cdap.plugin.gcp.bigquery.sink;
 
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.Table;
 import io.cdap.cdap.api.annotation.Description;
@@ -31,12 +32,19 @@ import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.StageConfigurer;
 import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
+import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
+import io.cdap.cdap.etl.api.validation.InvalidStageException;
+import io.cdap.plugin.gcp.bigquery.source.BigQuerySource;
+import io.cdap.plugin.gcp.bigquery.source.BigQuerySourceConfig;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
+import io.cdap.plugin.gcp.common.GCPConfig;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.mapred.AvroKey;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.List;
@@ -56,7 +64,7 @@ import java.util.stream.Collectors;
   + "BigQuery is Google's serverless, highly scalable, enterprise data warehouse. "
   + "Data is first written to a temporary location on Google Cloud Storage, then loaded into BigQuery from there.")
 public final class BigQuerySink extends AbstractBigQuerySink {
-
+  private static final Logger LOG = LoggerFactory.getLogger(BigQuerySink.class);
   public static final String NAME = "BigQueryTable";
 
   private final BigQuerySinkConfig config;
@@ -78,9 +86,18 @@ public final class BigQuerySink extends AbstractBigQuerySink {
     if (config.tryGetProject() == null || config.autoServiceAccountUnavailable()) {
       return;
     }
-    // validate schema with underlying table
-    Schema schema = configuredSchema == null ? inputSchema : configuredSchema;
-    validateConfiguredSchema(schema, collector);
+
+    // Try validations which may require connectivity
+    try {
+      // Validate partition properties
+      config.validatePartitionProperties(inputSchema, configuredSchema, collector);
+      // validate schema with underlying table
+      Schema schema = configuredSchema == null ? inputSchema : configuredSchema;
+      validateConfiguredSchema(schema, collector);
+    } catch (InvalidStageException e) {
+      // Allow deployment even if retrieving schema fails
+      LOG.warn("Unable to validate BigQuery schema with error:\n", e);
+    }
   }
 
   @Override
@@ -92,6 +109,18 @@ public final class BigQuerySink extends AbstractBigQuerySink {
   protected void prepareRunValidation(BatchSinkContext context) {
     FailureCollector collector = context.getFailureCollector();
     config.validate(context.getInputSchema(), config.getSchema(collector), collector);
+    try {
+      config.validatePartitionProperties(context.getInputSchema(), config.getSchema(collector), collector);
+    } catch (InvalidConfigPropertyException e) {
+      collector.addFailure(String.format("Unable to load credentials from %s.", config.getServiceAccountFilePath()),
+                           "Ensure the service account file is available on the local filesystem.")
+        .withConfigProperty(GCPConfig.NAME_SERVICE_ACCOUNT_FILE_PATH);
+
+    } catch (InvalidStageException e) {
+      collector.addFailure("Unable to get details about the BigQuery table: " + e.getMessage(), null)
+        .withConfigProperty(BigQuerySourceConfig.NAME_TABLE);
+    }
+    collector.getOrThrowException();
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -78,9 +84,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    // Try validations which may require connectivity
+    try {
+      // Validate partition properties
+      config.validatePartitionProperties(inputSchema, configuredSchema, collector);
+      // validate schema with underlying table
+      Schema schema = configuredSchema == null ? inputSchema : configuredSchema;
+      validateConfiguredSchema(schema, collector);
+    } catch (InvalidStageException e) {
+      // Allow deployment even if retrieving schema fails
+      LOG.warn("Unable to validate BigQuery schema with error:\n", e);
```
albertshau, line 97: shouldn't need to include a newline in the message, just the first part of the message should be fine:

Unable to validate BigQuery schema

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -92,6 +107,18 @@ protected BigQuerySinkConfig getConfig() {
   protected void prepareRunValidation(BatchSinkContext context) {
     FailureCollector collector = context.getFailureCollector();
     config.validate(context.getInputSchema(), config.getSchema(collector), collector);
+    try {
+      config.validatePartitionProperties(context.getInputSchema(), config.getSchema(collector), collector);
+    } catch (InvalidConfigPropertyException e) {
+      collector.addFailure(String.format("Unable to load credentials from %s.", config.getServiceAccountFilePath()),
```
albertshau, line 113: is this guaranteed to be a problem loading credentials? I think it's better to just let the exception propagate up if this is at prepare time. Whoever throws it should be using the right message.

the collector is mainly used to collector a bunch of problems for validation when the pipeline is being created, but using it adds a lot of complexity that isn't required if we're already at runtime.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -78,9 +84,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
       return;
     }
-    // validate schema with underlying table
-    Schema schema = configuredSchema == null ? inputSchema : configuredSchema;
-    validateConfiguredSchema(schema, collector);
+
+    // Try validations which may require connectivity
+    try {
+      // Validate partition properties
+      config.validatePartitionProperties(inputSchema, configuredSchema, collector);
```
albertshau, line 91: this wasn't being called before?

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -111,20 +111,27 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
   @Override
   public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
     FailureCollector collector = batchSourceContext.getFailureCollector();
     config.validate(collector);
-    Schema actualSchema = getSchema(collector);
+    Schema actualSchema = null;
+    try {
+      actualSchema = getSchema(collector);
+    } catch (IOException e) {
+      collector.addFailure("Unable to get Spanner Client: " + e.getMessage(), null)
```
albertshau, line 130: this message doesn't seem correct (existing problem), as the exception can happen for many reasons. Better to say something more generic, like 'Unable to fetch schema of Spanner table %s' and include the table id in the message.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
albertshau: same for other places, don't need to include 'with error:' in the message

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -92,6 +107,18 @@ protected BigQuerySinkConfig getConfig() {
   protected void prepareRunValidation(BatchSinkContext context) {
     FailureCollector collector = context.getFailureCollector();
     config.validate(context.getInputSchema(), config.getSchema(collector), collector);
+    try {
+      config.validatePartitionProperties(context.getInputSchema(), config.getSchema(collector), collector);
+    } catch (InvalidConfigPropertyException e) {
```
albertshau, line 112: i'm confused by this, don't see where validatePartitionProperties throws an exception

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -78,9 +84,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-    validateConfiguredSchema(schema, collector);
+
+    // Try validations which may require connectivity
+    try {
+      // Validate partition properties
+      config.validatePartitionProperties(inputSchema, configuredSchema, collector);
+      // validate schema with underlying table
+      Schema schema = configuredSchema == null ? inputSchema : configuredSchema;
+      validateConfiguredSchema(schema, collector);
+    } catch (InvalidStageException e) {
```
albertshau, line 95: this confuses me, did something get changed to throw this exception?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -92,6 +107,18 @@ protected BigQuerySinkConfig getConfig() {
   protected void prepareRunValidation(BatchSinkContext context) {
     FailureCollector collector = context.getFailureCollector();
     config.validate(context.getInputSchema(), config.getSchema(collector), collector);
+    try {
+      config.validatePartitionProperties(context.getInputSchema(), config.getSchema(collector), collector);
```
albertshau, line 111: this wasn't being called before, are we adding validation in this PR? The title indicates it's just about not failing deployment for certain types of validations.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -100,32 +103,53 @@ public void configurePipeline(PipelineConfigurer configurer) {
...
+        stageConfigurer.setOutputSchema(schema);
+        return;
+      }
 
-    if (configuredSchema == null) {
-      stageConfigurer.setOutputSchema(schema);
-      return;
+      validateConfiguredSchema(configuredSchema, collector);
+      stageConfigurer.setOutputSchema(configuredSchema);
+    } catch (IllegalArgumentException | InvalidStageException e) {
```
albertshau, line 117: the code is now really difficult to understand because you can't tell when an exception is thrown, when the collector is used, etc. without tracing through a bunch of code.

The decision about whether to log a warning and move on should be pushed down to the place where the exception originates, not bubbled all the way up here. There is no context here which makes it very easy to do the wrong thing.

I think it would be cleaner for these validate methods to always use the collector and never throw an exception. For validations that try to connect to an external system, those methods can take a boolean that tells them whether to warn and move on, or whether to add an error to the collector.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: Apologies for the confusion, but validatePartitionProperties was originally private method which was called in config.validate. Since it made checks which required connectivity, I separated it out from the original config.validate method and called it manually in all spots that call config.validate.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: So originally, there were two overloaded versions of the getBigQueryTable function in BigQueryUtils.java which are called in various places. One version takes a FailureCollector and one does not. Since calling FailureCollector causes an immediate failure, we needed to remove the overloaded method which calls FailureCollector and instead catch the error somewhere above the callstack. Thus, the FailureCollector overload was removed in favor of the other method, which instead throws InvalidConfigPropertyException and InvalidStageException.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: Fixed

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: Same as above. I separated out the validatePartitionProperties helper method which used to be private and called in config.validate. This way we are still running validation logic, but not failing immediately if connection problems arise.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: It makes calls to helper methods which eventually propagates down to BigQueryUtils.getBigQueryTable which throws these (InvalidConfigPropertyException, InvalidStageException) as runtime exceptions.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
dli357: In that situation it was, but it is probably worth refactoring the code to remove the complexity. Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
dli357: Understood, fixed

dli357:
Seems like we need to consolidate our validation approach before we can fix these issues. Closing this PR for now.

PR #279, opened 2020-08-12
Enable GCS Batch Sink to take file system properties

greeshmaswaminathan:
- Enable GCS Batch Sink to take additional file system properties similar to FileSink
- Calculate the `records.updated` metric correctly when `avro.mo.config.namedOutput` property is set
- Tested with the `avro.mo.config.namedOutput` property set. Metric is reflecting correctly.

### DIFF
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -47,5 +47,11 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 
+**Output File Prefix:** Prefix for the output file name.  
+If none is given, it will default to 'part', which means all data files written by the sink will look like 
+'part-r-00000', 'part-r-00001', etc.
+
+**File System Properties:** Additional properties to use with the OutputFormat.
+
 **Schema:** Schema of the data to write.
 The 'avro' and 'parquet' formats require a schema but other formats do not.
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -21,6 +21,9 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
@@ -45,6 +48,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.text.SimpleDateFormat;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -61,6 +65,8 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   private static final Logger LOG = LoggerFactory.getLogger(GCSBatchSink.class);
   public static final String RECORD_COUNT = "recordcount";
   private static final String RECORDS_UPDATED_METRIC = "records.updated";
+  public static final String AVRO_NAMED_OUTPUT = "avro.mo.config.namedOutput";
+  public static final String COMMON_NAMED_OUTPUT = "mapreduce.output.basename";
 
   private final GCSBatchSinkConfig config;
   private String outputPath;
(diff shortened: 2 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -137,12 +144,25 @@ private void emitMetrics(boolean succeeded, BatchSinkContext context) {
...
       StorageClient storageClient = StorageClient.create(config.getProject(), config.getServiceAccountFilePath());
-      storageClient.mapMetaDataForAllBlobs(outputPath, new MetricsEmitter(context.getMetrics())::emitMetrics);
+      storageClient.mapMetaDataForAllBlobs(getPrefixPath(), new MetricsEmitter(context.getMetrics())::emitMetrics);
     } catch (Exception e) {
       LOG.warn("Metrics for the number of affected rows in GCS Sink maybe incorrect.", e);
     }
   }
 
+  private String getPrefixPath() {
+    if (!config.getFileSystemProperties().containsKey(AVRO_NAMED_OUTPUT)) {
```
albertshau: this is the prefix property when the format is avro, but a different property is used for the other formats ("mapreduce.output.basename" from FileOutputFormat).

Also, I think it makes sense to add the file prefix as an explicit property in the advanced section along with the new FileSystemProperties. If the user sets a value, we can set it in the getFileSystemProperties() method. That will be a more user friendly way to do it, rather than forcing the user to know about these properties.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -137,12 +144,25 @@ private void emitMetrics(boolean succeeded, BatchSinkContext context) {
...
+      return outputPath;
+    }
+    //Following code is for handling a special case for saving files in same output directory with avro format.
+    //The specified file prefix from Filesystem Properties will be used to make the filename unique.
+    //The code is based on assumptions from the internal implementation of
+    //org.apache.avro.mapreduce.AvroOutputFormatBase and org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
+    return String
+      .format("%s/%s-", outputPath.endsWith("/") ? outputPath.substring(0, outputPath.length() - 1) :
```
albertshau: nit: it's a little difficult to read, can you create variables for the 2 values after the ?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Added the new property to take in file name prefix, this will have preference if the same properties are set using FileSystemProperties. Also considered the property for other formats.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -137,12 +152,41 @@ private void emitMetrics(boolean succeeded, BatchSinkContext context) {
...
+    }
+    //Following code is for handling a special case for saving files in same output directory.
+    //The specified file prefix from Filesystem Properties/Output file base name can be used to make filename unique.
+    //The code is based on assumptions from the internal implementation of
+    //org.apache.avro.mapreduce.AvroOutputFormatBase and org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
+    String outputPathPrefix = outputPath.endsWith("/") ? outputPath.substring(0, outputPath.length() - 1) : outputPath;
+    return String.format("%s/%s-", outputPathPrefix, filenameBase);
+  }
+
+  private String getFilenameBase() {
```
albertshau, line 175: annotate the method with `@Nullable`

docs/GCS-batchsink.md
```diff
@@ -47,5 +47,9 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 
+**Output File Prefix:** Prefix for the output file name.
```
albertshau: Prefix for the output file name. If none is given, it will default to 'part', which means all data files written by the sink will look like 'part-r-00000', 'part-r-00001', etc.

albertshau (approved):
minor comments, otherwise lgtm

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
greeshmaswaminathan: Done

docs/GCS-batchsink.md
greeshmaswaminathan: Done.

PR #281, opened 2020-08-13
Schema is marked as required

vjosadika:
Jira ticket: https://issues.cask.co/browse/PLUGIN-251

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
@@ -78,6 +78,10 @@ public class SpannerSourceConfig extends GCPReferenceSourceConfig {
     if (!containsMacro(NAME_SCHEMA) && schema != null) {
       SpannerUtil.validateSchema(schema, SUPPORTED_TYPES, collector);
     }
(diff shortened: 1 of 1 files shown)


rmstar (commented):
I don't think removing the Nullable annotation is sufficient here. You'll also need to add some validation to check that schema is not null.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
```diff
@@ -25,9 +25,9 @@
 import io.cdap.plugin.gcp.common.GCPReferenceSourceConfig;
 import io.cdap.plugin.gcp.spanner.common.SpannerUtil;
 
+import javax.annotation.Nullable;
```
rmstar: Why was this import reordered?

rmstar (commented):
lgtm, 1 minor comment.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
vjosadika: Fixed

rmstar:
Please squash commits.

CuriousVini (approved):

PR #287, opened 2020-08-17
[CDAP-16999] GCS Source Object Access Edge Case

dli357:
Fixes the case where GCS Source can access an object but not its bucket. Fix for #257


shifuxu0301 (approved):
LGTM

PR #290, opened 2020-08-18
PLUGIN-373 Fix bigquery sink upsert

CuriousVini:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -252,6 +252,23 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
           .setProjectId(tableRef.getProjectId())
           .setTableId(temporaryTableName);
         loadConfig.setDestinationTable(temporaryTableReference);
+      } else if (Operation.UPSERT.equals(operation)) {
```
shifuxu0301: 1. It seems like we should use condition "!Operation.INSERT.equals(operation)". I have take a look the definition for "update". 

Update - records that match on Table Key will be updated in the table. Records that do not match will be dropped.

2. There were some duplicated codes of creating temporaryTableName and temporaryTableReference. Maybe we can try to combine the previous condition and this one together as much as possible.

shifuxu0301: I think the correct behavior for update operation would be. 
If table does not exists:
   table will be created, but not records will be merged in. (no match)
else if table is empty:
    table will be kept as empty (no match)
else 
     matched records will be updated. drop no matched records.

CuriousVini: If the Update operation is provided and the table does not exist, it doesn't help to create an empty table as none of the records will match. Also I can not think of any practical usecase where user would want to create an empty table without updating any record in it.
In that case better I think behavior would be to throw an error to the user. 

Though the current behavior is completely wrong. It just dumps every incoming rows to the table. We should throw an error if table does not exist for update operation.

CuriousVini: There are 3 tests to perform for upsert operation:
1.) Destination Table does not exist. Expected behavior: Create the table
2.)Destination Table exists but is empty. Expected behavior: insert all the incoming rows
3.) Table exists with keys that matches the incoming records. While there are some incoming records that does not match existing keys. Expected behavior: Only matching records are updated, rest are inserted

3 test cases for update operation:
1.) Destination Table does not exist. Expected behavior: Fail the pipeline
2.)Destination Table exists but is empty. Expected behavior: Table remains empty as no matching records are found
3.) Table exists with keys that matches the incoming records. While there are some incoming records that does not match existing keys. Expected behavior: Only matching records are updated, rest are ignored

The bug exists for upsert and update operation for test case #1. Rest have correct existing behavior

bdmogal: For upsert, in 1. the expected behavior is to create the table, and insert all incoming rows, right?

For update, I'm a little wary of 1. Why should it fail the pipeline? Can't it be a no-op with a warning in the logs?

CuriousVini: 1. Correct. 
2. Had an offline discussion with @bdmogal . We have decided to add a warning message if the table does not exist. The reason is, failing pipeline would not be a good user experience in case many sinks are involved.

albertshau: +1 on not failing the pipeline.

shifuxu0301 (approved):
LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
shifuxu0301: Agreed.

PR #302, opened 2020-08-24
Fixed validation of Clustering order

vjosadika:
Jira ticket: https://issues.cask.co/browse/PLUGIN-196

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
@@ -37,7 +37,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -328,6 +328,13 @@ private void validateClusteringOrder(@Nullable Schema schema, FailureCollector c
     if (!shouldCreatePartitionedTable() || Strings.isNullOrEmpty(clusteringOrder) || schema == null) {
       return;
     }
+
+    if(!Strings.isNullOrEmpty(clusteringOrder) && Strings.isNullOrEmpty(partitionByField)){
+      collector.addFailure(String.format("Clustering order cannot be validated."),
```
rmstar, line 333: What happens if clusteringOrder is specified, and partitionByField is a macro? Can you test that and make sure it works as expected?

vjosadika: Fixed

rmstar (approved):

PR #308, opened 2020-09-17
feature/plugin-170-auth-update

ardian4:
Added: Ability to provide Service Account type as JSON

Jira Task: https://issues.cask.co/browse/PLUGIN-170

PR #313, opened 2020-09-17
Feature PLUGIN-170 GCS move copy sink multisink

ardian4:
Updated: Provide Service Account as JSON for GCS Move, GCS Copy, GCS Sink, GCS MultiSink plugins

Dependent on PR: #308

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -301,6 +301,16 @@ public class StorageClient {
     return new StorageClient(storage);
   }
 
+  public static StorageClient create(String project, @Nullable String serviceAccount,
+                                     Boolean isServiceAccountFilePath) throws IOException {
+    StorageOptions.Builder builder = StorageOptions.newBuilder().setProjectId(project);
+    if (serviceAccount != null) {
+      builder.setCredentials(GCPUtils.loadServiceAccountCredentials(serviceAccount, isServiceAccountFilePath));
+    }
+    Storage storage = builder.build().getService();
+    return new StorageClient(storage);
+  }
+
   /**
    * Represents a blob to be copied or moved.
    */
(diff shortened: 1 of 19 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
```diff
@@ -48,7 +48,15 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   public void run(ActionContext context) throws IOException {
     config.validate(context.getFailureCollector());
 
-    StorageClient storageClient = StorageClient.create(config.getProject(), config.getServiceAccountFilePath());
+    Boolean isServiceAccountFilePath = config.isServiceAccountFilePath();
+    if (isServiceAccountFilePath == null) {
+      context.getFailureCollector().addFailure("Service account type is undefined.",
```
rmstar, line 53: General comment applicable to these PRs: let's also add this validation in `configurePipeline`. We can do this as part of `config.validate`.

ardian4: @rmstar 

This check here is added if the value is @macro, because we already have set e default value for "serviceAccountType" in the Widget (the default value is filepath). However, we can remove the @nullable annotation for the property "serviceAccountType" in config in order to make sure all validation criterias are met. Does this still require validation in `config.validate(...)` to be added?

rmstar: I see. No need to add validation in `config.validate()`

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -301,6 +301,16 @@ public static StorageClient create(String project, @Nullable String serviceAccou
     return new StorageClient(storage);
   }
 
+  public static StorageClient create(String project, @Nullable String serviceAccount,
```
rmstar, line 304: Is `create(String project, @Nullable String serviceAccountPath)` (L295) still needed?

rmstar (commented):
Can you please merge #308 first, then create a new PR on top of it?

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
ardian4: @rmstar No, we have removed in new PR.

ardian4:
New PR: https://github.com/data-integrations/google-cloud/pull/356

PR #314, opened 2020-09-17
Feature/PLUGIN-170-gcs-argument-setter

ardian4:
Updated: Provide Service Account as JSON for GCS Argument Setter plugin

Dependent on PR: #308

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -28,9 +28,13 @@ import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import io.cdap.plugin.gcp.gcs.GCSPath;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.util.Base64;
+import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -49,11 +64,55 @@ public static ServiceAccountCredentials loadServiceAccountCredentials(String pat
...
+  }
+
+  public static String extractPrivateKey(ServiceAccountCredentials credentials) {
+    return String.format(PRIVATE_KEY_WRAP,
+                         Base64.getEncoder().encodeToString(credentials.getPrivateKey().getEncoded()));
+  }
+
+  public static Map<String, String> generateAuthProperties(String serviceAccount,
+                                                           boolean isServiceAccountFilePath,
+                                                           String... keyPrefix) throws IOException {
```
rmstar, line 84: Do we have a case where we need to support multiple key prefixes?

ardian4: Yes, here is a case: https://github.com/data-integrations/google-cloud/blob/48a86cecc1cdf7e45b95bfbe76b19662754b8f2a/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java#L155-L158

ardian4:
new PR: https://github.com/data-integrations/google-cloud/pull/371

PR #326, opened 2020-09-23
(PLUGIN-395) - Avoid underlying BigQuery schema changes when allowSchemaRelaxation is disabled in BigQuery sink plugins.

supriyapatil-google:
1. Moved BigQuery Table Schema conversion methods from BigQuerySource to BigQueryUtil so that they can be reused by BigQuerySink classes.
2. Updated AbstractBigQuerySink#validateSchema method to 

- validate schema changes when truncate table is true but allowSchemaRelaxation is false.
- ignore required fields when both allowSchemaRelaxation/truncate table are true.
- fixed null mode checking since mode is optional and uses NULLABLE by default.

3. Update BigQuerySinkTest to as per validation changes.
4. Updated BigQuerySink/MultiSink classes to always use table schema when allowSchemaRelaxation is set to false.
5. Set table schema update options if allowSchemaRelaxation is enabled.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -272,9 +272,9 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
    * @return
    */
   protected final Schema overrideOutputSchemaWithTableSchemaIfNeeded(
-      Schema configuredSchema,
-      String tableName,
-      FailureCollector collector) {
+    Schema configuredSchema,
+    String tableName,
+    FailureCollector collector) {
     AbstractBigQuerySinkConfig config = getConfig();
 
     if (!config.isAllowSchemaRelaxation()) {
@@ -315,7 +315,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
     com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
     if (bqSchema == null || bqSchema.getFields().isEmpty()) {
       // Table is created without schema, so no further validation is required.
-      LOG.info("Table [%s] doesn't have a valid schema. Using input schema for writing records.", config.getTable());
+      LOG.info("Table [%s] doesn't have a schema. Using input schema for writing records.", config.getTable());
       return null;
     }
     return BigQueryUtil.getTableSchema(bqSchema, collector);
@@ -345,25 +345,25 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
 
     List<String> missingBQFields = BigQueryUtil.getSchemaMinusBqFields(outputSchemaFields, bqFields);
 
-    if (allowSchemaRelaxation) {
+    if (allowSchemaRelaxation && !getConfig().isTruncateTableSet()) {
       // Required fields can be added only if truncate table option is set.
-      if (!getConfig().isTruncateTableSet()) {
-        List<String> nonNullableFields = missingBQFields.stream()
-            .map(tableSchema::getField)
-            .filter(Objects::nonNull)
-            .filter(field -> !field.getSchema().isNullable())
-            .map(Schema.Field::getName)
-            .collect(Collectors.toList());
+      List<String> nonNullableFields = missingBQFields.stream()
+        .map(tableSchema::getField)
+        .filter(Objects::nonNull)
+        .filter(field -> !field.getSchema().isNullable())
+        .map(Schema.Field::getName)
+        .collect(Collectors.toList());
 
-        for (String nonNullableField : nonNullableFields) {
-          collector.addFailure(
-              String.format("Required field '%s' does not exist in BigQuery table '%s.%s'.",
-                  nonNullableField, getConfig().getDataset(), tableName),
-              "Change the field to be nullable.")
-              .withInputSchemaField(nonNullableField).withOutputSchemaField(nonNullableField);
-        }
+      for (String nonNullableField : nonNullableFields) {
+        collector.addFailure(
+          String.format("Required field '%s' does not exist in BigQuery table '%s.%s'.",
+              nonNullableField, getConfig().getDataset(), tableName),
+          "Change the field to be nullable.")
+          .withInputSchemaField(nonNullableField).withOutputSchemaField(nonNullableField);
       }
-    } else {
+    }
+
+    if (!allowSchemaRelaxation) {
       // schema should not have fields that are not present in BigQuery table,
       for (String missingField : missingBQFields) {
         collector.addFailure(
@@ -393,9 +393,9 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
         // skip checking schema if field is missing in BigQuery
         if (!missingBQFields.contains(fieldName)) {
           ValidationFailure failure = BigQueryUtil.validateFieldSchemaMatches(
-              bqFields.get(field.getName()), field, getConfig().getDataset(), tableName,
-              AbstractBigQuerySinkConfig.SUPPORTED_TYPES, collector);
-          if (failure != null) {
+            bqFields.get(field.getName()), field, getConfig().getDataset(), tableName,
+            AbstractBigQuerySinkConfig.SUPPORTED_TYPES, collector);
+        if (failure != null) {
             failure.withInputSchemaField(fieldName).withOutputSchemaField(fieldName);
           }
         }
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -267,18 +267,20 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
    * Adjusts output format schema depending on allowSchemaRelaxation setting to avoid unexpected
    * schema changes to the underlying bigQuery table.
    * @param configuredSchema Schema configured for output format.
+   * @param bqSchema Optional BigQuery table schema to avoid duplicate calls.
    * @param tableName BigQuery table name for writing record.
    * @param collector Failure collector to report failures to the client.
    * @return
    */
   protected final Schema overrideOutputSchemaWithTableSchemaIfNeeded(
-    Schema configuredSchema,
     String tableName,
+    Schema configuredSchema,
+    @Nullable com.google.cloud.bigquery.Schema bqSchema,
     FailureCollector collector) {
     AbstractBigQuerySinkConfig config = getConfig();
 
     if (!config.isAllowSchemaRelaxation()) {
-      Schema tableSchema = getTableSchema(config.getTable(), collector);
+      Schema tableSchema = getTableSchema(tableName, bqSchema, collector);
       // We use GCS buckets to write AVRO files and import them in BigQuery.
       // Avro is a self describing format and BigQuery overwrites table schema with AVRO record
       // schema.
@@ -295,27 +297,36 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
   /**
    * Returns Bigtable schema in a CDAP schema format.
    * @param tableName BigQuery table name for writing record.
+   * @param bqSchema Optional BigQuery table schema to avoid duplicate calls.
    * @param collector Failure collector to report failures to the client.
    * @return
    */
   @Nullable
-  private Schema getTableSchema(String tableName, FailureCollector collector) {
-    AbstractBigQuerySinkConfig config = getConfig();
-    Table table = BigQueryUtil.getBigQueryTable(
+  private Schema getTableSchema(
+    String tableName,
+    @Nullable com.google.cloud.bigquery.Schema bqSchema,
+    FailureCollector collector) {
+    if (bqSchema == null) {
+      AbstractBigQuerySinkConfig config = getConfig();
+      Table table = BigQueryUtil.getBigQueryTable(
         config.getProject(),
-        config.getDataset(), tableName,
+        config.getDataset(),
+        tableName,
         config.getServiceAccountFilePath(),
         collector);
 
-    if (table == null) {
-      LOG.info("Table [%s] doesn't exist yet. Using input schema for writing records.", config.getTable());
-      return null;
+      if (table == null) {
+        LOG.info("Table [%s] doesn't exist yet. Using input schema for writing records.",
+          config.getTable());
+        return null;
+      }
+
+      bqSchema = table.getDefinition().getSchema();
     }
 
-    com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
     if (bqSchema == null || bqSchema.getFields().isEmpty()) {
       // Table is created without schema, so no further validation is required.
-      LOG.info("Table [%s] doesn't have a schema. Using input schema for writing records.", config.getTable());
+      LOG.info("Table [%s] doesn't have a schema. Using input schema for writing records.", tableName);
       return null;
     }
     return BigQueryUtil.getTableSchema(bqSchema, collector);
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -263,6 +263,64 @@ private String configureBucket() {
...
+  /**
+   * Adjusts output format schema depending on allowSchemaRelaxation setting to avoid unexpected
+   * schema changes to the underlying bigQuery table.
+   * @param configuredSchema Schema configured for output format.
+   * @param tableName BigQuery table name for writing record.
+   * @param collector Failure collector to report failures to the client.
+   * @return
+   */
+  protected final Schema overrideOutputSchemaWithTableSchemaIfNeeded(
+      Schema configuredSchema,
```
albertshau: style: indentation should be 2 spaces instead of 4

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -299,11 +299,11 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
       } else {
         loadConfig.setDestinationTable(tableRef);
 
-        //Schema update options should only be specified with WRITE_APPEND disposition,
+        // Schema update options should only be specified with WRITE_APPEND disposition,
         // or with WRITE_TRUNCATE disposition on a table partition - The logic below should change when we support
         // insertion into single partition
         if (allowSchemaRelaxation && !JobInfo.WriteDisposition.WRITE_TRUNCATE
-          .equals(JobInfo.WriteDisposition.valueOf(writeDisposition))) {
+            .equals(JobInfo.WriteDisposition.valueOf(writeDisposition))) {
```
albertshau: style: should not change the indentation here

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -92,7 +99,9 @@ protected void prepareRunInternal(BatchSinkContext context, BigQuery bigQuery, S
       }
 
       try {
-        Schema tableSchema = Schema.parseJson(argument.getValue());
+        Schema tableSchema = overrideOutputSchemaWithTableSchemaIfNeeded(
+            Schema.parseJson(argument.getValue()), tableName, collector);
```
albertshau: style: indentation should be 2 spaces instead of 4

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -29,14 +30,19 @@
...
 import org.apache.avro.mapred.AvroKey;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import javax.annotation.Nullable;
```
albertshau: added accidentally?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -47,6 +53,7 @@
   + "BigQuery is Google's serverless, highly scalable, enterprise data warehouse. "
   + "Data is first written to a temporary location on Google Cloud Storage, then loaded into BigQuery from there.")
 public class BigQueryMultiSink extends AbstractBigQuerySink {
+  private static final Logger LOG = LoggerFactory.getLogger(BigQueryMultiSink.class);
```
albertshau: seems like it's not used

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -50,6 +51,7 @@
 import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
```
albertshau: looks unused

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -108,11 +110,14 @@ protected void prepareRunValidation(BatchSinkContext context) {
   @Override
   protected void prepareRunInternal(BatchSinkContext context, BigQuery bigQuery, String bucket) throws IOException {
     FailureCollector collector = context.getFailureCollector();
+
     Schema configSchema = config.getSchema(collector);
-    Schema schema = configSchema == null ? context.getInputSchema() : configSchema;
-    configureTable(schema);
+    Schema outputSchema = overrideOutputSchemaWithTableSchemaIfNeeded(
+        configSchema == null ? context.getInputSchema() : configSchema, config.getTable(), collector);
```
albertshau: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -263,6 +263,64 @@ private String configureBucket() {
...
+
+    if (table == null) {
+      LOG.info("Table [%s] doesn't exist yet. Using input schema for writing records.", config.getTable());
+      return null;
+    }
+
+    com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
+    if (bqSchema == null || bqSchema.getFields().isEmpty()) {
+      // Table is created without schema, so no further validation is required.
+      LOG.info("Table [%s] doesn't have a valid schema. Using input schema for writing records.", config.getTable());
```
albertshau: nit: a valid schema -> a schema

putting 'valid' there makes it sound like the table has an invalid schema, when in actuality it just doesn't have one at all.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -282,32 +340,28 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
...
-      for (String nonNullableField : nonNullableFields) {
-        collector.addFailure(
-          String.format("Required field '%s' does not exist in BigQuery table '%s.%s'.",
-                        nonNullableField, getConfig().getDataset(), tableName),
-          "Change the field to be nullable.")
-          .withInputSchemaField(nonNullableField).withOutputSchemaField(nonNullableField);
+      // Required fields can be added only if truncate table option is set.
+      if (!getConfig().isTruncateTableSet()) {
+        List<String> nonNullableFields = missingBQFields.stream()
+            .map(tableSchema::getField)
```
albertshau: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -282,32 +340,28 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
...
-        .collect(Collectors.toList());
-
-      for (String nonNullableField : nonNullableFields) {
-        collector.addFailure(
-          String.format("Required field '%s' does not exist in BigQuery table '%s.%s'.",
-                        nonNullableField, getConfig().getDataset(), tableName),
-          "Change the field to be nullable.")
-          .withInputSchemaField(nonNullableField).withOutputSchemaField(nonNullableField);
+      // Required fields can be added only if truncate table option is set.
+      if (!getConfig().isTruncateTableSet()) {
```
albertshau: this can be combined with the outer if condition to remove one layer of nesting.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -322,26 +376,32 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
...
-          failure.withInputSchemaField(fieldName).withOutputSchemaField(fieldName);
+    // column type changes should be disallowed if either allowSchemaRelaxation or truncate table are not set.
+    if (!allowSchemaRelaxation || !getConfig().isTruncateTableSet()) {
+      // Match output schema field type with BigQuery column type
+      for (Schema.Field field : tableSchema.getFields()) {
+        String fieldName = field.getName();
+        // skip checking schema if field is missing in BigQuery
+        if (!missingBQFields.contains(fieldName)) {
+          ValidationFailure failure = BigQueryUtil.validateFieldSchemaMatches(
+              bqFields.get(field.getName()), field, getConfig().getDataset(), tableName,
```
albertshau: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
albertshau: There doesn't appear to be any schema validation happening here. This can lead to situations where the pipeline is configured incorrectly, trying to write to a table with a completely different schema. We would use the table's schema instead of the data's schema, which would likely lead to weird null pointer or class cast exceptions that would be very difficult for the user to understand.

Instead of having this method lookup the table schema, I think it would be better to look it up outside and validate that it is compatible before passing it in. This would also avoid an extra schema lookup in the normal BQ sink.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
supriyapatil-google: Schema validation added in prepareRunValidation(). Formatting fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
supriyapatil-google: Removed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
supriyapatil-google: Removed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
supriyapatil-google: Removed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Fixed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Fixed. My intellij setting was adding 4 spaces instead of 2.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -282,19 +340,13 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
       return;
     }
 
-    if (getConfig().isTruncateTableSet()) {
```
greeshmaswaminathan, line 285: Should we be worried about backward compatibility here? cc: @albertshau

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -123,4 +121,39 @@ public void transform(StructuredRecord input,
                         Emitter<KeyValue<AvroKey<GenericRecord>, NullWritable>> emitter) throws IOException {
     emitter.emit(new KeyValue<>(new AvroKey<>(toAvroRecord(input)), NullWritable.get()));
   }
+
+  /**
+   * Validates configured output schema against underlying BigQuery table schema.
+   * @param context Sink Context
+   * @param collector Failure collector to report failures to the client.
+   */
+  private void validateConfiguredSchema(BatchSinkContext context, FailureCollector collector) {
```
albertshau: this is a good amount of duplicate code with prepareRunInternal.  It's kind of arbitrary that this class has a prepareRunInternal and prepareRunValidate. I think it's better to move the schema validation part to prepareRunInternal rather than try to keep this logic in the validate method.

This also would remove a duplicate schema lookup by passing the schema into overrideOutputSchemaWithTableSchemaIfNeeded instead of looking it up both here and in the method.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: As per offline discussion, please call this out in the release notes.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -282,19 +340,13 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
...
-      return;
-    }
-
     FieldList bqFields = bqSchema.getFields();
     List<Schema.Field> outputSchemaFields = Objects.requireNonNull(tableSchema.getFields());
 
     List<String> missingBQFields = BigQueryUtil.getSchemaMinusBqFields(outputSchemaFields, bqFields);
 
-    if (allowSchemaRelaxation) {
+    if (allowSchemaRelaxation && !getConfig().isTruncateTableSet()) {
```
greeshmaswaminathan, line 362: Was this giving a runtime exception before?

greeshmaswaminathan: Nvm, the previous check was returning.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -322,26 +376,32 @@ protected void validateSchema(Table table, Schema tableSchema, boolean allowSche
...
-    // Match output schema field type with BigQuery column type
-    for (Schema.Field field : tableSchema.getFields()) {
-      String fieldName = field.getName();
-      // skip checking schema if field is missing in BigQuery
-      if (!missingBQFields.contains(fieldName)) {
-        ValidationFailure failure = BigQueryUtil.validateFieldSchemaMatches(
-          bqFields.get(field.getName()), field, getConfig().getDataset(), tableName,
-          AbstractBigQuerySinkConfig.SUPPORTED_TYPES, collector);
+    // column type changes should be disallowed if either allowSchemaRelaxation or truncate table are not set.
+    if (!allowSchemaRelaxation || !getConfig().isTruncateTableSet()) {
```
greeshmaswaminathan, line 403: So we will allow column type changes even if `update schema` option is not set? If so, please document the behavior.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
supriyapatil-google: I have moved code to prepareRunInternal and refactored code to avoid duplicate calls.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: This is just refactoring based on Albert's comment.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: column type changes are only allowed if both allowSchemaRelaxation and truncate table is set otherwise we will run into runtime exception with conflicts between existing data vs new data being added.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
supriyapatil-google: Will do.

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
greeshmaswaminathan: Makes sense, thanks.

greeshmaswaminathan (approved):
Please squash the commits before merge.

PR #327, opened 2020-09-24
feature/plugin-170 BigTable auth update

ailegion:
added: Use JSON key for service account for BigTable plugins

Dependent on PR: https://github.com/data-integrations/google-cloud/pull/308


ailegion:
old that was later replaced with https://github.com/data-integrations/google-cloud/pull/366

PR #328, opened 2020-09-24
PLUGIN-380 Filter Support

ailegion:
added: Filter support for Spanner source

Jira Ticket: https://issues.cask.co/browse/PLUGIN-380


widgets/Spanner-batchsource.json
```diff
@@ -64,6 +64,14 @@
               }
             ]
           }
+        },
+        {
+          "widget-type": "textarea",
+          "label": "Import Query",
```
CuriousVini, line 70: This should be added to spanner docs as well

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
```diff
@@ -67,6 +68,10 @@
   @Macro
   public String table;
 
+  @Description("The SELECT query to use to import data from the specified table")
+  @Macro
+  public String importQuery;
```
CuriousVini, line 74: If this is a required field it would break backwards compatibility. We should keep it optional and if user has not provided the import query we should be using existing behavior meaning - `select * from %s`

ailegion: updated PR

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -192,7 +195,8 @@ private void initializeConfig(Configuration configuration, String projectId) {
     setIfValueNotNull(configuration, SpannerConstants.SERVICE_ACCOUNT_FILE_PATH, config.getServiceAccountFilePath());
     setIfValueNotNull(configuration, SpannerConstants.INSTANCE_ID, config.instance);
     setIfValueNotNull(configuration, SpannerConstants.DATABASE, config.database);
-    setIfValueNotNull(configuration, SpannerConstants.QUERY, String.format("Select * from %s;", config.table));
+    setIfValueNotNull(configuration, SpannerConstants.QUERY, Strings.isNullOrEmpty(config.importQuery) ?
+      String.format("Select * from %s;", config.table) : config.importQuery);
```
CuriousVini, line 203: 1.) Is `;` needed at the end?
2.) `Select` should have `S` capitalized? Is the query case sensitive?
3.) What happens when table name is case sensitive? Is it supported?

CuriousVini (commented):
a few questions, code changes lgtm 👍

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ailegion: @CuriousVini this was how it was originally. As for the questions:
1. It's not mandatory
2. Base on https://cloud.google.com/spanner/docs/data-definition-language#naming_conventions schema object names are case-insensitive for queries.
3. same as 2

CuriousVini (approved):
lgtm. please squash the commits

ailegion:
> lgtm. please squash the commits

squashed

CuriousVini:
Please resolve the conflicts

ailegion:
> Please resolve the conflicts

resovled

CuriousVini (approved):

PR #331, opened 2020-09-24
(PLUGIN-395) - Avoid underlying BigQuery schema changes when allowSchemaRelaxation is disabled in BigQuery sink plugins.

MEseifan:
1. Moved BigQuery Table Schema conversion methods from BigQuerySource to BigQueryUtil so that they can be reused by BigQuerySink classes.
2. Updated AbstractBigQuerySink#validateSchema method to 

- validate schema changes when truncate table is true but allowSchemaRelaxation is false.
- ignore required fields when both allowSchemaRelaxation/truncate table are true.
- fixed null mode checking since mode is optional and uses NULLABLE by default.

3. Update BigQuerySinkTest to as per validation changes.
4. Updated BigQuerySink/MultiSink classes to always use table schema when allowSchemaRelaxation is set to false.
5. Set table schema update options if allowSchemaRelaxation is enabled.


supriyapatil-google (approved):
Thanks!

PR #334, opened 2020-09-25
fix: BQ Sink retain column descriptions when Truncate option is set

ardian4:
JiraTicket: https://issues.cask.co/browse/PLUGIN-387


CuriousVini (approved):

supriyapatil-google:
Do we see this problem after fixes for https://issues.cask.co/browse/PLUGIN-395?

We shouldn't be making any schema changes if "Update Table Schema" is false. Can you please add the check to make sure "Update Table Schema" is set to true here.

PR #335, opened 2020-09-25
Added: GCS Source plugin Min Split Size property

ardian4:
JiraTicket : https://issues.cask.co/browse/PLUGIN-352


src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -268,6 +275,10 @@ public long getMaxSplitSize() {
       return maxSplitSize;
     }
 
+    public Long getMinSplitSize() {
```
CuriousVini, line 279: Mark this as nullable.

CuriousVini (approved):
Fix and squash commits

PR #366, opened 2020-10-16
update/PLUGIN-170-BigTable-Auth-update

ailegion:
Added: Ability to provide Service Account type as JSON for BigTable Plugins

JIRA Task: https://issues.cask.co/browse/PLUGIN-170

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
@@ -72,12 +72,13 @@ public final class BigtableSinkConfig extends GCPReferenceSinkConfig {
   private final String bigtableOptions;
 
   public BigtableSinkConfig(String referenceName, String table, String instance, @Nullable String project,
-                            @Nullable String serviceFilePath, String keyAlias, String columnMappings,
-                            @Nullable String bigtableOptions) {
+                            @Nullable String serviceAccountType, @Nullable String serviceFilePath, String keyAlias,
+                            String columnMappings, @Nullable String bigtableOptions) {
     this.referenceName = referenceName;
     this.table = table;
     this.instance = instance;
     this.project = project;
+    this.serviceAccountType = serviceAccountType;
     this.serviceFilePath = serviceFilePath;
     this.keyAlias = keyAlias;
     this.columnMappings = columnMappings;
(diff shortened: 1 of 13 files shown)


src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
```diff
@@ -0,0 +1,38 @@
...
+
+package io.cdap.plugin.gcp.bigtable.common;
+
+import org.apache.hadoop.hbase.mapreduce.TableOutputFormat;
+import org.apache.hadoop.mapreduce.JobContext;
+
+import java.io.IOException;
+
+/**
+ * Table output format class with null credentials fix fon non-default service accounts
```
rmstar: Please add more details about how this class is used

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class TableOutputCustomFormat extends TableOutputFormat {
+  public TableOutputCustomFormat() {
+
+  }
+
+  @Override
+  public void checkOutputSpecs(JobContext context) throws IOException, InterruptedException {
+    // required for auth
```
rmstar: Please add more details. What configuration is being set here?

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
rmstar: Also, why is this only needed in the table ouput format but not the input format?

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
```diff
@@ -72,12 +72,13 @@
...
   public BigtableSinkConfig(String referenceName, String table, String instance, @Nullable String project,
-                            @Nullable String serviceFilePath, String keyAlias, String columnMappings,
-                            @Nullable String bigtableOptions) {
+                            @Nullable String serviceAccountType, @Nullable String serviceFilePath, String keyAlias,
+                            String columnMappings, @Nullable String bigtableOptions) {
     this.referenceName = referenceName;
     this.table = table;
     this.instance = instance;
     this.project = project;
+    this.serviceAccountType = serviceAccountType;
```
rmstar, line 81: where is serviceAccountType used?

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
ailegion: When using non-default credentials we faced null pointer exception issue  - this would happen only during sink operations. While researching the cause of the problem we found out that the reason behind it was ``checkOutputSpecs`` method which calls ``createConnection``, however at this stage the configuration is not set yet. In order to solve it we had to trigger ``setConf`` call before ``checkOutputSpecs`` call.

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
ailegion: updated

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSinkConfig.java
ailegion: > where is serviceAccountType used?

this was required by tests - in order to make existing tests work.

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
rmstar: I see, thanks for explaining!

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.hadoop.mapreduce.JobContext;
+
+import java.io.IOException;
+
+/**
+ * Table output format class - extends default {@link TableOutputFormat} in order to override checkOutputSpecs method
+ * to include configuration properties before calling `ConnectionFactory.createConnection`.
+ * Fixes null pointer exception during connection creation.
+ */
+public class TableOutputCustomFormat extends TableOutputFormat {
```
rmstar: Let's rename this to BigtableOutputFormat to be consistent with naming of other input, output formats in this repo.

rmstar (commented):
Please squash commits after addressing minor comment. Thanks!

src/main/java/io/cdap/plugin/gcp/bigtable/common/TableOutputCustomFormat.java
ailegion: updated

ailegion:
> Please squash commits after addressing minor comment. Thanks!

updated and squashed

rmstar (approved):

PR #367, opened 2020-10-16
Updated: Provide Service Account as JSON for Google Speech

ardian4:
added: Use JSON key for service account for Speech plugins

### DIFF
--- a/docs/SpeechToText-transform.md
+++ b/docs/SpeechToText-transform.md
@@ -42,3 +42,9 @@ in the array represents one part of the full audio data and will contain the tra
 
 **Transcription Text Field**: The field to store the transcription of the full audio data. It is generated using the
 transcription for each part with the highest confidence.
+
+**Service Account**  - service account key used for authorization
+* **File Path**: Path on the local file system of the service account key used for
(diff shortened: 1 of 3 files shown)


rmstar (commented):
Avoid code duplication by extending GCPConfig instead of PluginConfig, as you have done in other plugins.

ardian4:
> Avoid code duplication by extending GCPConfig instead of PluginConfig, as you have done in other plugins.

@rmstar 
The reason why it was extended from PluginConfig and not from GCPConfig is that GCPConfig has the property ``project``, which is not needed for this plugin.
 How we should proceed?  
 One option is to create the widget and hide it.
image

rmstar:
> One option is to create the widget and hide it.

@ardian4 yes, this might be a better option. I'd prefer to avoid code duplication.

ardian4:
> > One option is to create the widget and hide it.
> 
> @ardian4 yes, this might be a better option. I'd prefer to avoid code duplication.

@rmstar DONE

rmstar (approved):
Please squash commits

ardian4:
> Please squash commits

Done

PR #376, opened 2020-10-20
[PLUGIN-436]  Increment bytes read/written counter in Datastore sourc…

rmstar:
…e/sink.

* Use lower level proto library `com.google.datastore.v1` instead of `com.google.cloud.datastore` as this gives us access to protos sent/received on the network for bytes read/written accounting. Removed usage of `com.google.cloud.datastore` library to simplify the code.

* Bytes read/written are incremented in the record reader/writer. 

### Unit tests
Unit tests cover entity <-> StructuredRecord transformations

### Manual testing

* read, write from/to non-default namespace.
* test all supported key types in source and sink.
* set ancestor in source and sink.
* set filters in source.
* set number of splits in source.
* configure indexed properties in sink.
* set batch size in sink.
* test all supported property types.


src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
```diff
@@ -54,19 +65,60 @@ public DatastoreRecordWriter(TaskAttemptContext taskAttemptContext) {
...
+        AllocateIdsRequest.newBuilder()
+          .setProjectId(projectId)
+          .addKeys(entity.getKey())
+          .build();
+      AllocateIdsResponse response;
+      try {
+         response = datastore.allocateIds(request);
+      } catch (DatastoreException e) {
+        LOG.error("Failed to allocate id",  e);
+        return;
```
MEseifan: Are we sure these errors match the errors that were previously thrown by the other library? Both in terms of error code and error message

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -62,21 +45,33 @@
...
+    Datastore datastore = DatastoreUtil.getDatastoreV1(
       config.get(DatastoreSourceConstants.CONFIG_SERVICE_ACCOUNT),
       config.getBoolean(DatastoreSourceConstants.CONFIG_SERVICE_ACCOUNT_IS_FILE, true),
       config.get(DatastoreSourceConstants.CONFIG_PROJECT));
     LOG.trace("Executing query split: {}", query);
-    results = datastore.run(query);
+    RunQueryRequest request = RunQueryRequest.newBuilder()
+      .setQuery(query)
+      .setPartitionId(PartitionId.newBuilder()
+                        .setNamespaceId(config.get(DatastoreSourceConstants.CONFIG_NAMESPACE))
```
MEseifan, line 67: Maybe a small comment to explain the need for this partition since it wasnt in the old code

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -62,21 +45,33 @@
...
+                        .setProjectId(config.get(DatastoreSourceConstants.CONFIG_PROJECT)))
+      .build();
+    try {
+      QueryResultBatch batch = datastore.runQuery(request).getBatch();
+      taskAttemptContext.getCounter(FileInputFormatCounter.BYTES_READ).increment(batch.getSerializedSize());
+      results = batch.getEntityResultsList().iterator();
+    } catch (DatastoreException e) {
```
MEseifan, line 74: Same comment as above, same error code and message as before?

src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
```diff
@@ -54,19 +65,60 @@ public DatastoreRecordWriter(TaskAttemptContext taskAttemptContext) {
...
+      AllocateIdsRequest request =
+        AllocateIdsRequest.newBuilder()
+          .setProjectId(projectId)
+          .addKeys(entity.getKey())
+          .build();
+      AllocateIdsResponse response;
+      try {
+         response = datastore.allocateIds(request);
+      } catch (DatastoreException e) {
+        LOG.error("Failed to allocate id",  e);
```
rmstar: different from previous behavior

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
rmstar: done

src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
rmstar: Thanks for pointing this out. The old code [putWithDeferredIdAllocation](https://googleapis.dev/java/google-cloud-datastore/latest/com/google/cloud/datastore/DatastoreBatchWriter.html#putWithDeferredIdAllocation-com.google.cloud.datastore.FullEntity...-) throws com.google.cloud.datastore.DatastoreException. Fixed the code to throw an exception here and in other places where the old code threw an exception.

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
rmstar: Old code throws [com.google.cloud.datastore.DatastoreException](https://googleapis.dev/java/google-cloud-datastore/latest/com/google/cloud/datastore/Datastore.html#run-com.google.cloud.datastore.Query-com.google.cloud.datastore.ReadOption...-).

I throw an IOException here because RecordReader.initialize() throws [IOException, InterruptedException](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/mapreduce/RecordReader.html#initialize-org.apache.hadoop.mapreduce.InputSplit-org.apache.hadoop.mapreduce.TaskAttemptContext-).

MEseifan (approved):
LGTM. Please update the description to mention the manual and automated tests you did to confirm this refactor is non-breaking.

PR #380, opened 2020-10-22
[PLUGIN-440] Increment bytes read/written counter in Bigtable source/…

rmstar:
…sink


MEseifan (commented):
code LGTM, can we test this with spark UI before merging?

rmstar:
I have verified that bytes in/out show up in the spark UI.

MEseifan (approved):

PR #408, opened 2020-10-30
PLUGIN-461: fix issue with backward incompatibility

ardian4:
1. if property ``serviceAccountType`` is null and is not macro fallback to ``filePath``
2. Handle macro for serviceAccountType

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
@@ -89,18 +89,24 @@ public class GCPConfig extends PluginConfig {
     return serviceAccountJson;
   }
 
+  /**
+   * @return Service Account Type, defaults to filePath.
+   */
+  @Nullable
   public String getServiceAccountType() {
-    if (containsMacro(NAME_SERVICE_ACCOUNT_TYPE) || Strings.isNullOrEmpty(serviceAccountType)) {
+    if (containsMacro(NAME_SERVICE_ACCOUNT_TYPE)) {
       return null;
     }
-    return serviceAccountType;
+    return Strings.isNullOrEmpty(serviceAccountType) ? SERVICE_ACCOUNT_FILE_PATH : serviceAccountType;
   }
 
+  @Nullable
   public Boolean isServiceAccountJson() {
     String serviceAccountType = getServiceAccountType();
     return Strings.isNullOrEmpty(serviceAccountType) ? null : serviceAccountType.equals(SERVICE_ACCOUNT_JSON);
   }
 
+  @Nullable
   public Boolean isServiceAccountFilePath() {
     String serviceAccountType = getServiceAccountType();
     return Strings.isNullOrEmpty(serviceAccountType) ? null : serviceAccountType.equals(SERVICE_ACCOUNT_FILE_PATH);
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -101,14 +101,19 @@ public class GCPUtils {
                                                            String serviceAccountType,
                                                            String... keyPrefix) throws IOException {
     Map<String, String> properties = new HashMap<>();
+    if (serviceAccountType == null) {
+      return properties;
+    }
     String privateKeyData = null;
     properties.put(SERVICE_ACCOUNT_TYPE, serviceAccountType);
 
     boolean isServiceAccountFilePath = SERVICE_ACCOUNT_TYPE_FILE_PATH.equals(serviceAccountType);
 
     for (String prefix : keyPrefix) {
-      if (isServiceAccountFilePath && serviceAccount != null) {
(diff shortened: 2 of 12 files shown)


src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -90,10 +90,10 @@ public String getServiceAccountJson() {
   }
 
   public String getServiceAccountType() {
```
CuriousVini, line 96: If this returns null the method should be marked as nullable. Also make sure all the callers handle null values

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -107,8 +107,10 @@ public static String extractPrivateKey(ServiceAccountCredentials credentials) {
     boolean isServiceAccountFilePath = SERVICE_ACCOUNT_TYPE_FILE_PATH.equals(serviceAccountType);
 
     for (String prefix : keyPrefix) {
-      if (isServiceAccountFilePath && serviceAccount != null) {
```
CuriousVini, line 110: can it still be null if its a macro?

CuriousVini (commented):
- Can we add a test for this case?
- We observed similar behavior in GCS Source. Has it been fixed?
https://issues.cask.co/browse/PLUGIN-461?focusedCommentId=39830&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-39830

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
ardian4: `isServiceAccountFilePath` will be false when `serviceAccountType` is macro, but code is updated so when `serviceAccountType` is macro, it will not add any auth properties.

ardian4:
> * Can we add a test for this case?
> * We observed similar behavior in GCS Source. Has it been fixed?
>   https://issues.cask.co/browse/PLUGIN-461?focusedCommentId=39830&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-39830

- Tests added 
- Also issue with GCSSource was already fixed in previous commit 

https://github.com/data-integrations/google-cloud/blob/96509a39ecec83cea8e581d1041e85d3683acc0e/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L109-L115

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
CuriousVini: So basically below check is exactly same as this one. Then why do we need that change?

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -101,14 +101,19 @@ public static String extractPrivateKey(ServiceAccountCredentials credentials) {
                                                            String serviceAccountType,
                                                            String... keyPrefix) throws IOException {
     Map<String, String> properties = new HashMap<>();
+    if (serviceAccountType == null) {
```
CuriousVini, line 104: Why is this change added? why it was not needed before?

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -89,18 +89,21 @@ public String getServiceAccountJson() {
...
   }
 
+  @Nullable
   public String getServiceAccountType() {
-    if (containsMacro(NAME_SERVICE_ACCOUNT_TYPE) || Strings.isNullOrEmpty(serviceAccountType)) {
+    if (containsMacro(NAME_SERVICE_ACCOUNT_TYPE)) {
       return null;
     }
-    return serviceAccountType;
+    return Strings.isNullOrEmpty(serviceAccountType) ? SERVICE_ACCOUNT_FILE_PATH : serviceAccountType;
```
CuriousVini, line 100: Please add a comment in the javadoc explaining what is the default value that is interpreted

CuriousVini (commented):
Are these changes applicable to any of the other GCP plugins such as pubsub, speechtransform?

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
ardian4: This change is required for the case when serviceAccountType is filePath and serviceAccount is null - for autodetect scenario. If we leave as it was it would fallback back to serviceAccountType as it was a JSON key.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
ardian4: It was added to handle when serviceAccountType is macro, and this method is called in configuration stage.

ardian4:
> Are these changes applicable to any of the other GCP plugins such as pubsub, speechtransform?

Not all plugins use this method. pubsub and speechtransform are not using this method.

CuriousVini (approved):
lgtm. Please squash the commits

ardian4:
> lgtm. Please squash the commits

Squashed

CuriousVini (approved):

PR #421, opened 2020-11-13
Bug Fix: service account file path set to auto-detect

ardian4:
Fix issue when service account is set to filePath and the service account file path set to auto-detect.
Jira Task: https://issues.cask.co/browse/PLUGIN-481

Test scenario: 
1. Service account type filePath: with service account file provided, with service account set as auto-detect, with both as macro
2. Service account type JSON: with service account JSON provided, with service account JSON as macro, with both as macro

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java
@@ -106,12 +106,12 @@ public final class GCSArgumentSetter extends Action {
   }
 
   private static Storage getStorage(GCSArgumentSetterConfig config) throws IOException {
-    return StorageOptions.newBuilder()
-      .setProjectId(config.getProject())
-      .setCredentials(GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
-                                                             config.isServiceAccountFilePath()))
-      .build()
-      .getService();
+    String serviceAccount = config.getServiceAccount();
+    StorageOptions.Builder builder = StorageOptions.newBuilder().setProjectId(config.getProject());
+    if (serviceAccount != null) {
+      builder.setCredentials(GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath()));
+    }
+    return builder.build().getService();
   }
 
   public static String getContent(GCSArgumentSetterConfig config) throws IOException {
--- a/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetterConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetterConfig.java
@@ -56,7 +56,8 @@ public final class GCSArgumentSetterConfig extends GCPConfig {
       }
     }
 
-    if (isServiceAccountJson()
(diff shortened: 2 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java
```diff
@@ -106,12 +106,12 @@ public void run(ActionContext context) throws Exception {
...
-    return StorageOptions.newBuilder()
-      .setProjectId(config.getProject())
-      .setCredentials(GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
-                                                             config.isServiceAccountFilePath()))
-      .build()
-      .getService();
+    String serviceAccount = config.getServiceAccount();
+    StorageOptions.Builder builder = StorageOptions.newBuilder().setProjectId(config.getProject());
+    if (serviceAccount != null) {
+      builder.setCredentials(GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath()));
```
CuriousVini, line 112: What would be credentials when service account is null?

CuriousVini: Also `config.isServiceAccountFilePath()` can be null. Wouldnt `GCPUtils#loadServiceAccountCredentials` method throw a null pointer exception when NPE? Has this change been tested with that case?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetterConfig.java
```diff
@@ -56,7 +56,8 @@ public void validateProperties(FailureCollector collector) {
       }
     }
 
-    if (isServiceAccountJson()
+    Boolean isServiceAccountJson = isServiceAccountJson();
+    if (isServiceAccountJson != null && isServiceAccountJson
```
CuriousVini, line 60: This change is needed at Line#79 as well?

CuriousVini (commented):
Please add test scenarios for this fix in the PR description

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java
ardian4: when service account is null credentials are not set and the library loads default credentials as in other plugins. Regarding ``config.isServiceAccountFilePath()`` being null - this condition cannot be fullfilled due to check in line 111, if we have serviceAccount that means that serviceAccountType is not null.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetterConfig.java
ardian4: in order for isServiceAccountJson to be null serviceAccountType needs to be null - and this can only happen if it's set as macro in validation stage which we are checking with ``canConnect`` on line 75

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java
CuriousVini: I see

CuriousVini (approved):
lgtm

PR #422, opened 2020-11-13
[PLUGIN-474] Fixed BigQuery Plugins Not Deleting Temporary Files

dli357:
See [PLUGIN-474](https://issues.cask.co/browse/PLUGIN-474) for more information.

Currently, when a user specifies a temporary bucket for BigQuery plugins, the Avro temporary files left behind are not cleaned up. This PR solves this problem by making the following changes:
 - Change from using the deprecated [`fs.gs.system.bucket`](https://storage.googleapis.com/hadoop-conf/gcs-core-default.xml) to use [`fs.default.name`](https://hadoop.apache.org/docs/r1.2.1/core-default.html) so that we can specify the subdirectory of a GCS bucket as the Hadoop FS base path
 - Create a new temporary subdirectory in the GCS bucket, whether the created temporary bucket or the user provided bucket, using the run ID as the name, and set the Hadoop base path using this temporary subdirectory
 - Delete the temporary subdirectory after all runs

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
@@ -49,6 +49,7 @@ import io.cdap.plugin.gcp.common.GCPUtils;
 import org.apache.avro.generic.GenericData;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapreduce.Job;
@@ -147,7 +148,7 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
                             cmekKey);
     }
 
-    configuration.set("fs.gs.system.bucket", bucket);
+    configuration.set("fs.default.name", String.format("gs://%s/%s/", bucket, uuid));
     configuration.setBoolean("fs.gs.impl.disable.cache", true);
     configuration.setBoolean("fs.gs.metadata.cache.enable", false);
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -210,16 +211,21 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
...
+    }
     try {
-      if (config.getBucket() == null) {
-        FileSystem fs = gcsPath.getFileSystem(configuration);
-        if (fs.exists(gcsPath)) {
-          fs.delete(gcsPath, true);
-        }
+      FileSystem fs = gcsPath.getFileSystem(configuration);
+      if (fs.exists(gcsPath)) {
+        fs.delete(gcsPath, true);
```
CuriousVini, line 224: does this also delete temp bucket that was created?

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
```diff
@@ -120,7 +119,7 @@ private void testMetric(Job mockJob, long expectedCount, int invocations)
     BigQuerySink sink = getSinkToTest(mockJob);
     MockStageMetrics mockStageMetrics = Mockito.spy(new MockStageMetrics("test"));
     BatchSinkContext context = getContextWithMetrics(mockStageMetrics);
-    sink.onRunFinish(true, context);
+    sink.recordMetric(true, context);
```
CuriousVini, line 122: why is this change needed?

CuriousVini (commented):
2 minor questions.

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
dli357: Previously, onRunFinish would only execute the filesystem delete call when a temporary bucket was created (e.g. the temporary bucket name was left empty). However, since this change now will execute the delete call after all runs no matter whether a temporary bucket is provided, the unit test will hit that condition and result in a NullPointerException since the Hadoop baseConfiguration doesn't exist.

Since these unit tests only test the recordMetric method anyway, I figured it was cleaner to change it to directly testing that method instead of testing the entire onRunFinish method, since refactoring/mocking the entire Hadoop configuration is nontrivial.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
dli357: Nope, but that is already enabled via AbstractBigQuerySink.java:262. We only delete the bucket if the temporary bucket was created.

CuriousVini: ah yea - through `fs.gs.bucket.delete.enable`

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
CuriousVini: I see

CuriousVini (approved):
lgtm 👍

PR #425, opened 2020-11-17
[CDAP-17289] Implement Pub/Sub receiver with backpressure support

fernst:
Implemented Pub/Sub receiver using Google's Pub/Sub client library.

Implemented support for backpressure using Spark's Rate Controller mechanism.

This implementation also allows the end user to specify the number of receivers they want to run in parallel, allowing us to scale as long as resources ara available.


pom.xml
```diff
@@ -443,9 +443,16 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>${junit.version}</version>
+      <groupId>org.powermock</groupId>
```
fernst, line 458: This same dependencies are used for Wrangler https://github.com/data-integrations/wrangler/blob/a26fe155e4895b262e95852c6051f178764c134c/wrangler-core/pom.xml#L86

src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+
+import static org.mockito.Mockito.when;
+
+public class PubSubSubscriberTest {
+
+  @Test
```
fernst: I'm still working on writing these tests. I just want comments on the rest of the code so far.

src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiverMessagePullingTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(SubscriptionAdminClient.class)
+public class PubSubReceiverMessagePullingTest {
```
fernst, line 54: I'm still working on writing these tests. I just want comments on the rest of the code so far.

pom.xml
```diff
@@ -608,7 +611,10 @@
             <!--Only @Plugin classes in the export packages will be included as plugin-->
             <_exportcontents>
               io.cdap.plugin.gcp.*;
+              com.google.auth.oauth2.*;
               com.google.cloud.hadoop.*;
+              com.google.pubsub.v1.*;
+              com.google.protobuf.*;
```
chtyim: Why do we need to export so many things?? Generally it is not good. If it is for spark streaming serialization, we should be able to hide most of them by being more careful about what goes into class fields and marking them transient when appropriate.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Input Stream used to subscribe to a Pub/Sub topic and pull messages.
+ */
+public class PubSubInputDStream extends ReceiverInputDStream<ReceivedMessage> {
+
+  protected String project;
+  protected String topic;
```
chtyim: not a good idea to have protected non final fields.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
```diff
@@ -0,0 +1,58 @@
+package io.cdap.plugin.gcp.publisher.source;
```
chtyim, line 16: Missing copyright and license header.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+    this.project = project;
+    this.topic = topic;
+    this.subscription = subscription;
+    this.credentials = credentials;
+    this.autoAcknowledge = autoAcknowledge;
+  }
+
+  @Override
+  public void onStart() {
+    if (topic != null) {
```
chtyim, line 156: What does it mean when `topic == null`? Is it a valid use case?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+   * Receive new messages and store based on the Storage Level settings.
+   *
+   * @throws ApiException     when the Pub/Sub API throws a non-retryable exception.
+   * @throws RuntimeException when the GrpcSubscriberStub cannot be created.
+   */
+  public void receive() {
+    SubscriberStubSettings subscriberStubSettings = getSubscriberStubSettings();
+
+    if (LOG.isTraceEnabled()) {
+      LOG.trace("Receiver Started execution");
```
chtyim: Just log it as debug without the if check. An `if` check is only necessary if the logging itself requires heavy computation (e.g. stringifying a complex object).

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    try (SubscriberStub subscriber = getSubscriberStub(subscriberStubSettings)) {
+      String subscriptionName = ProjectSubscriptionName.format(project, subscription);
+      fetchMessagesWithRetry(subscriber, subscriptionName);
+    } catch (IOException | ApiException e) {
+      throw new RuntimeException("Failed to fetch new messages.", e);
```
chtyim: Give more context (e.g. on what topic, which subscriber... etc).

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+        .setMaxMessages(getMessageRate())
+        .setSubscription(subscriptionName)
+        .build();
+    PullResponse pullResponse = subscriber.pullCallable().call(pullRequest);
+
+    List<ReceivedMessage> receivedMessages = pullResponse.getReceivedMessagesList();
+    List<String> ackIds = receivedMessages.stream().map(ReceivedMessage::getAckId).collect(Collectors.toList());
+
+    //If there are no messages to process, continue.
+    if (receivedMessages.size() == 0) {
```
chtyim: This check should be before the `actIds` creation

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Yes, when the user supplies a Subscription name without a topic, we pull from the subscription (without attempting to create it).

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+    this.credentials = credentials;
+    this.autoAcknowledge = autoAcknowledge;
+  }
+
+  @Override
+  public void onStart() {
+    if (topic != null) {
+      createSubscription();
+    }
+    new Thread(this::receive).start();
```
chtyim: I don't see any looping in the `receive` method, so this Thread is short live that it only fetch messages once?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+   * The default rate is 1000 (matching Apache Bahir) if the receiver has not been able to calculate a rate.
+   *
+   * @return The current rate at which this receiver should fetch messages.
+   */
+  protected int getMessageRate() {
+    int messageRate = supervisor().getCurrentRateLimit() < (long) Integer.MAX_VALUE ?
+      (int) supervisor().getCurrentRateLimit() : 1000;
+
+    if (LOG.isTraceEnabled()) {
+      LOG.trace("Receiver rate is: " + messageRate);
```
chtyim: Use placeholder `{}` for logging instead of `+`.

pom.xml
fernst: @chtyim That's exactly what happened to me. Spark complains about missing classes if those 2 packages are not available.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,386 @@
...
+   */
+  public static class BackoffConfigBuilder implements Serializable {
+    public int initialBackoffMs = 100;
+    public int maximumBackoffMs = 10000;
+    public double backoffFactor = 2.0;
+
+    protected BackoffConfigBuilder() {
+    }
+
+    public static BackoffConfigBuilder getInstance() {
```
chtyim, line 551: Why need a static method? Can't the caller just `new` it?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
fernst: Oops, I'll fix that.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Good catch.

chtyim: Also, use `isEmpty()` instead of `size() == 0`

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: the `fetchMessagesWithRetry` method that is called from the receive method is the one where the while loop is implemented.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Good point. I have noticed that most of Google's libraries use the this pattern for builders (see https://github.com/data-integrations/google-cloud/pull/425/commits/cfdbfff52860de20f372963f2217ced95a559687#diff-db0dd3323d1daa18e728d38181a3b88dacb74f0dbfcb9867d534c9ed730c5a9cR190), so my thinking was "Let's keep it consistent".

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
```
albertshau: 2020

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,394 @@
...
+        throw new RuntimeException("Failed to create subscription.", ioe);
+      }
+
+    }
+
+    if (isStopped()) {
+      return;
+    }
+
+    throw new RuntimeException("Failed to create subscription after 5 attempts.");
```
albertshau: would be good to keep track of the last exception and add it as the cause for this exception.

albertshau: also would be good to include the topic name in the error message.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,394 @@
...
+      } catch (ApiException ae) {
+        //If the subscription already exists, ignore the error.
+        if (ae.getStatusCode().getCode().equals(StatusCode.Code.ALREADY_EXISTS)) {
+          return;
+        }
+
+        //This error is thrown is the Topic Name is not valid.
+        //Throw an Illegal Argument Exception so the pipeline fails.
+        if (ae.getStatusCode().getCode().equals(StatusCode.Code.NOT_FOUND)) {
+          throw new IllegalArgumentException("Failed to create subscription. Topic Name is invalid.", ae);
```
albertshau: nit: Name -> name.

Also, would be good to include the topic name in the error message.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
```diff
@@ -0,0 +1,184 @@
...
+      LineageRecorder recorder = new LineageRecorder(context, config.referenceName);
+      recorder.recordRead("Read", "Read from Pub/Sub",
+                          this.schema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList()));
+    }
+  }
+
+  @Override
+  public JavaDStream<T> getStream(StreamingContext streamingContext) throws Exception {
+    SparkConf sparkConf = streamingContext.getSparkStreamingContext().ssc().conf();
+    sparkConf.set("spark.streaming.receiver.writeAheadLog.enable", "true");
```
albertshau: I don't think this actually works, due to details around how CDAP and Spark are integrated. You can verify by looking at the environment section in the Spark UI to see if these are indeed set.

If it doesn't get set, I believe you have to set them in configurePipeline(), through context.setPipelineProperties(). (see https://github.com/data-integrations/salesforce-marketing/blob/develop/src/main/java/io/cdap/plugin/sfmc/MarketingCloudDataExtensionSink.java#L77)

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public String getSubscription() {
+      return subscription;
+    }
+
+    public String getTopic() {
+      return topic;
+    }
+
+    public int getNumberOfReceivers() {
```
albertshau: looks unused, I assume you'll handle this in a future PR?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
```diff
@@ -0,0 +1,184 @@
...
+  @Override
+  public void prepareRun(StreamingSourceContext context) throws Exception {
+    FailureCollector collector = context.getFailureCollector();
+    config.validate(collector);
+    // record dataset lineage
+    context.registerLineage(config.referenceName, this.schema);
+
+    if (this.schema.getFields() != null) {
+      LineageRecorder recorder = new LineageRecorder(context, config.referenceName);
+      recorder.recordRead("Read", "Read from Pub/Sub",
```
albertshau: would be good to include the topic name or subscription name in the description

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
albertshau: Also, this should be an advanced property for the plugin (defaulting to true is fine). Since it adds some overhead, if the user doesn't care about at least once guarantees, they should be able to turn it off.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Input Stream used to subscribe to a Pub/Sub topic and pull messages.
+ */
+public class PubSubInputDStream extends ReceiverInputDStream<ReceivedMessage> {
+
+  protected String project;
+  protected String topic;
+  protected String subscription;
+  protected ServiceAccountCredentials credentials;
```
albertshau: guessing this is the reason the oauth packages are being exported? Can we create it lazily instead and make this transient?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,394 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+/**
+ * Spark Receiver for Pub/Sub Messages.
+ * <p>
+ * If backpressure is enabled, the message ingestion rate for this receiver will be managed by Spark.
+ */
+public class PubSubReceiver extends Receiver<ReceivedMessage> {
```
albertshau: Is ReceivedMessage here the reason the pubsub and protobuf packages are being exported? It may seem a bit silly, but it would be better to create a separate class in this project that encapsulates the needed information from ReceivedMessage, in order to avoid the issues that would be introduced by exporting the classes.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
fernst: Yes, this is a placeholder for future implementations (allowing the users to specify the number of receivers they want to run in parallel).

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Bahir does something similar to this. 

The counter argument for this is "What happens if another user for this library needs fields X, Y and Z from the message, and we don't expose those fields in the separate class we implement?".

But your point is valid. Let me try copying the fields I need (and maybe some additional fields) into a separate class and see if I can get away with not needing to export the additional Google packages.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
fernst: @albertshau Thanks for that tip. Setting those config properties in the `getStream()` method did work during my brief testing, but I agree that it's not an "elegant" solution.

pom.xml
chtyim: Can you cleanup the fields serialization such that we can minimize the set of export packages?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,405 @@
...
+    List<ReceivedMessage> receivedMessages = pullResponse.getReceivedMessagesList();
+
+    //If there are no messages to process, continue.
+    if (receivedMessages.isEmpty()) {
+      return;
+    }
+
+    //Exit if the receiver is stopped before storing.
+    if (isStopped()) {
+      LOG.debug("Receiver stopped before store and ack.");
```
chtyim: Add more context in the log message (topic, subscriber, etc)

pom.xml
```diff
@@ -608,6 +611,7 @@
             <!--Only @Plugin classes in the export packages will be included as plugin-->
             <_exportcontents>
               io.cdap.plugin.gcp.*;
+              com.google.auth.oauth2.*;
```
chtyim: Possible to eliminate this as well?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Spark Receiver for Pub/Sub Messages.
+ * <p>
+ * If backpressure is enabled, the message ingestion rate for this receiver will be managed by Spark.
+ */
+public class PubSubMessage implements Serializable {
+  static final HashMap<String, String> EMPTY_MAP = new HashMap<>();
+
+  String messageId = null;
```
chtyim: No need to assign to null for fields.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,85 @@
...
+  String orderingKey = null;
+  String ackId;
+  byte[] data = null;
+  HashMap<String, String> attributes = null;
```
chtyim: Use `Map` instead of `HashMap`.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,85 @@
...
+import java.time.Instant;
+import java.util.HashMap;
+import javax.annotation.Nonnull;
+
+/**
+ * Spark Receiver for Pub/Sub Messages.
+ * <p>
+ * If backpressure is enabled, the message ingestion rate for this receiver will be managed by Spark.
+ */
+public class PubSubMessage implements Serializable {
```
chtyim: It is actually better to implements `Externalizable` and implements the readExternal / writeExternal methods

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,85 @@
...
+  static final HashMap<String, String> EMPTY_MAP = new HashMap<>();
+
+  String messageId = null;
+  String orderingKey = null;
+  String ackId;
+  byte[] data = null;
+  HashMap<String, String> attributes = null;
+  Instant publishTime = null;
+
+  public PubSubMessage(@Nonnull ReceivedMessage message) {
```
chtyim: No need to use `@Nonnull`. Only need to annotate if it is `@Nullable.`

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,422 @@
...
+  /**
+   * Sleep for a given number of milliseconds, calculate new backoff time and return.
+   *
+   * @param backoff the time in milliseconds to delay execution.
+   * @return the new backoff delay in milliseconds
+   */
+  protected int sleepAndIncreaseBackoff(int backoff) {
+    try {
+      if (!isStopped()) {
+        LOG.warn("Sleeping for {} ms.", backoff);
```
chtyim: Why this is a warn?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
chtyim: I think it is better just new.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
chtyim: These shouldn't be set in here. These properties are not plugin specific. They should be exposed as pipeline config / properties.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
```diff
@@ -0,0 +1,183 @@
...
+                          this.schema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList()));
+    }
+  }
+
+  @Override
+  public JavaDStream<T> getStream(StreamingContext streamingContext) throws Exception {
+    SparkConf sparkConf = streamingContext.getSparkStreamingContext().ssc().conf();
+    sparkConf.set("spark.streaming.receiver.writeAheadLog.enable", "true");
+    sparkConf.set("spark.streaming.driver.writeAheadLog.closeFileAfterWrite", "true");
+    sparkConf.set("spark.streaming.receiver.writeAheadLog.closeFileAfterWrite", "true");
```
chtyim: These two settings are checkpoint fs specific. E.g. if the checkpoint is on HDFS, you actually don't want these set to true.

pom.xml
fernst: I'll try to clean this up next.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
fernst: I've removed these seetings.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
fernst: You can see the implementation now. The user is able to specify how many receivers they'd like to use as a configuration parameter (Defaults to 1).

pom.xml
fernst: I'll try this next.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
fernst: I've removed these properties.

src/main/java/io/cdap/plugin/gcp/publisher/source/GCPCredentialsProvider.java
```diff
@@ -34,7 +34,7 @@
  * Hack to replace the one in Bahir, which is incompatible with the version of the pubsub library
  * used by this project.
  */
-public class GCPCredentialsProvider implements SparkGCPCredentials {
+public class  GCPCredentialsProvider implements SparkGCPCredentials {
```
chtyim: Remove extra whitespace

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
fernst: I have made this property transient and removed the need to export the oauth package.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I have solved this by creating my own implementation of a "PubSubMessage" class.

src/main/java/io/cdap/plugin/gcp/publisher/source/GCPCredentialsProvider.java
albertshau: remove extra space

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
albertshau: +1

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
albertshau: this is still a good idea, in case we update the code and it still needs to read the old serialized messages.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.pubsub.v1.ReceivedMessage;
+
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Spark Receiver for Pub/Sub Messages.
```
albertshau: would be good to add a line about why we're using this instead of the message proto class from the pubsub library.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+  private static final String CREATE_SUBSCRIPTION_RETRY_ERROR_MSG =
+    "Failed to create subscription '%s' after 5 attempts";
+  private static final String MISSING_TOPIC_ERROR_MSG =
+    "Failed to create subscription. Topic '%s' was not found in project '%s'.";
+  private static final String SUBSCRIBER_ERROR_MSG =
+    "Failed to create subscriber using subscription '%s' for project '%s'.";
+  private static final String FETCH_ERROR_MSG =
+    "Failed to fetch new messages using subscription '%s' for project '%s'.";
+
+  private PubSubSubscriberConfig config;
```
albertshau: private final for these

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+    this.project = config.getProject();
+    this.topic = TopicName.format(config.getProject(), config.getTopic());
+    this.subscription = ProjectSubscriptionName.format(config.getProject(), config.getSubscription());
+
+    try {
+      this.credentials = config.getServiceAccount() == null ?
+        null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
+                                                      config.isServiceAccountFilePath());
+    } catch (IOException e) {
+      LOG.error("Unable to get credentials.");
```
albertshau: does the stop method below log the message and stack trace? 

If so, we don't need to double log it here. If not, this message should also include the exception so that the root cause is included in the log.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+      }
+    }
+
+    LOG.info("Receiver completed execution");
+  }
+
+  /**
+   * Create a new subscription (if needed) for the supplied topic.
+   *
+   * @throws IllegalArgumentException when the specified Topic does not exists
```
albertshau: are these javadocs correct? the code looks like it stops whenever it hits a non-retryable exception or if it tries too many times.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+   * @return the new backoff delay in milliseconds
+   */
+  protected int sleepAndIncreaseBackoff(int backoff) {
+    try {
+      if (!isStopped()) {
+        LOG.debug("Backoff - Sleeping for {} ms.", backoff);
+        Thread.sleep(backoff);
+      }
+    } catch (InterruptedException e) {
+      LOG.warn("Interrupted Exception in PubSubReceiver.");
```
albertshau: seems more appropriate to propagate this exception up rather than swallow it here, so that the caller can abort what they're doing. For example, when this is called when creating the subscription, we'll want to stop retrying immediately rather than update the backoff.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+   * @throws RuntimeException         when the SubscriptionAdminClient cannot be created.
+   * @throws ApiException             when a non-retryable exception is thrown by the Pub/Sub client
+   */
+  protected void createSubscription() {
+    int backoff = backoffConfig.getInitialBackoffMs();
+    int attempts = 5;
+
+    ApiException lastApiException = null;
+
+    while (!isStopped() && attempts-- > 0) {
```
albertshau, line 231: There are a few places where retries are being done, it may be worth looking at https://jodah.net/failsafe/ to see if it would make things simpler.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+
+      try (SubscriptionAdminClient subscriptionAdminClient = getSubscriptionAdminClient()) {
+
+        int ackDeadline = 10; //10 seconds before resending the message.
```
albertshau: seems like this should be configurable as an advanced option, and/or defaulted to a higher number. If the batch size is sufficiently large, it seems plausible storing the messages may take longer than 10 seconds, resulting in dupes.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+  /**
+   * Build subscriber client and schedule fetch task to run at a 100 millisecond delay.
+   *
+   * @throws ApiException     when the Pub/Sub API throws a non-retryable exception.
+   * @throws RuntimeException when the GrpcSubscriberStub cannot be created.
+   */
+  public void scheduleFetch() {
+    try {
+      SubscriberStubSettings subscriberSettings = getSubscriberSettings();
+      subscriber = getSubscriber(subscriberSettings);
```
albertshau: I think it's cleaner to set all the variables in the onStart() method, unless there's a particular reason it's created here?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+
+      executor.scheduleWithFixedDelay(this::receiveMessages, 0, 100, TimeUnit.MILLISECONDS);
```
albertshau: would be better to have this outside the try-catch to make it clear that this isn't the thing that can throw IOException and stop the subscriber

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+        fetchAndAck();
+        return;
+      } catch (ApiException ae) {
+        if (ae.isRetryable()) {
+          backoff = sleepAndIncreaseBackoff(backoff);
+        } else {
+          //Restart the receiver if the exception is not retryable.
+          String message =
+            String.format(FETCH_ERROR_MSG, subscription, project);
+          restart(message, ae);
```
albertshau, line 338: what's the difference between restarting the receiver and retrying the message fetch?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,506 @@
...
+  /**
+   * Get the rate at which this receiver should pull messages and set this rate in the bucket we use for rate control.
+   * The default rate is Integer.MAX_VALUE if the receiver has not been able to calculate a rate.
+   */
+  protected void updateMessageRateAndFillBucket() {
+    int messageRate = (int) Math.min(supervisor().getCurrentRateLimit(), Integer.MAX_VALUE);
+
+    if (messageRate != previousFetchRate) {
+      previousFetchRate = messageRate;
+      LOG.debug("Receiver fetch rate is set to: {}", messageRate);
```
albertshau: better to limit how often this logs, as debug logs are still collected by default. Maybe max once a minute?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
albertshau: description should include subscription name

widgets/GoogleSubscriber-streamingsource.json
```diff
@@ -77,6 +77,20 @@
           "name": "serviceAccountJSON"
         }
       ]
+    },
+    {
+      "label": "Advanced",
+      "properties": [
+        {
+          "widget-type": "number",
+          "label": "Number of Receivers",
```
albertshau: 'Receiver' is a Spark term that most users probably won't understand. 'Reader' may be a little better.

albertshau: Can you also update the markdown doc to add information about this setting. Should mention how this relates to the number of executors required to run the pipeline.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: @albertshau What do you think about using `trace` as the log level?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: A retryable exception is something like "429 too many requests", which means the request can be safely retried at a later time (in this case, the exponential backoff logic starts increasing the delay for requests).

A non-retryable exception can be something related to credentials expired, subscription deleted, or an internal server error from Pub/Sub. In this case, the receiver gets restarted, credentials re-initialized, subscription recreated if needed, and then messages start pulling once again.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Good point, let me rearrange.

pom.xml
fernst: Removed.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Good point, I'll remove this log line.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: that's fine too

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: I see. Can you add a comment with this info?

widgets/GoogleSubscriber-streamingsource.json
fernst: @albertshau I will also modify the property name from `numberOfReceivers` to `numberOfReaders` for consistency.

fernst: I have changed the `numberOfReceivers` setting to `numberOfReaders` and updated the docs.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I have implemented this change.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I have added this comment.

docs/GoogleSubscriber-streamingsource.md
```diff
@@ -57,3 +57,6 @@ the creation of subscriber will be received.
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 * **JSON**: Contents of the service account JSON file.
+
+**Number of readers**: Number of Pub/Sub reader workers to run in parallel. Each reader requires one worker in 
+the cluster. The default numbgit ter of readers per Pub/Sub Streaming Source is 1.
```
albertshau: typos

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubInputDStream.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.spark.streaming.receiver.Receiver;
+
+import javax.annotation.Nullable;
+
+/**
+ * Input Stream used to subscribe to a Pub/Sub topic and pull messages.
+ */
+public class PubSubInputDStream extends ReceiverInputDStream<PubSubMessage> {
+
+  private PubSubSubscriberConfig config;
```
albertshau: private final

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    if (ackId != null) {
+      out.writeUTF(ackId);
+    }
+
+    out.writeObject(publishTime);
+
+    out.writeInt(attributes != null ? attributes.size() : -1);
+    if (attributes != null) {
+      for (Map.Entry<String, String> entry : attributes.entrySet()) {
+        out.writeObject(entry.getKey());
```
albertshau: should this be writeUTF?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    out.writeInt(messageId != null ? messageId.length() : -1);
+    if (messageId != null) {
+      out.writeUTF(messageId);
+    }
+
+    out.writeInt(orderingKey != null ? orderingKey.length() : -1);
```
albertshau: can this ever be null? If so, it should be annotated as `@Nullable` in the constructor

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    if (orderingKey != null) {
+      out.writeUTF(orderingKey);
+    }
+
+    out.writeInt(ackId != null ? ackId.length() : -1);
```
albertshau: similar question here, seems like it shouldn't ever be null?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    if (ackId != null) {
+      out.writeUTF(ackId);
+    }
+
+    out.writeObject(publishTime);
+
+    out.writeInt(attributes != null ? attributes.size() : -1);
```
albertshau, line 126: should make sure attributes is never null and an empty collection is used instead.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    return attributes;
+  }
+
+  public Instant getPublishTime() {
+    return publishTime;
+  }
+
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    out.writeInt(messageId != null ? messageId.length() : -1);
```
albertshau: is this needed? out.writeUTF(null) throws an exception?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: still have this question

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: still have this question, javadoc seems incorrect?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: still have this question

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,546 @@
...
+        //This error is thrown is the Topic does not exist.
+        // Call the stop method so the pipeline fails.
+        if (ae.getStatusCode().getCode().equals(StatusCode.Code.NOT_FOUND)) {
+          String message = String.format(MISSING_TOPIC_ERROR_MSG, topic, project);
+          stop(message, ae);
+          return;
+        }
+
+        //Retry if the exception is retriable.
+        if (ae.isRetryable()) {
```
albertshau: should this also be checking those retryable status codes from the top of the class?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: still have this question

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
albertshau: still have this concern

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
albertshau: still have this comment

albertshau (commented):
is this up to date? Still have several of the same comments/questions

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: Since the backing structure is a HashMap, null keys and null values are a possibility. `writeUTF` does not support null strings.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Ah, good catch. The initial implementation I had was throwing exceptions but I have rewritten it to use the Receiver state methods (`stop` and `restart`) instead.

I must have stashed these doc changes, let me add them to the PR.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Good point. I did not notice this in my testing, but we can always set this timeout to 60 seconds.

I can add a config option for this if you think this would be useful.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Hmm, good catch. I will use the same method call I used in the `fetchAndAck` method.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I can move this into the `onStart` method, but I think having this in a separate function makes it easier to visualize and write tests for.

pom.xml
chtyim: Don't remove the direct dependency on junit since you are still using it (e.g. `@Test`). Never rely on transitive dependencies.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: `stop` and `restart` do indeed log the exception, for example:

```
2021-01-14 17:07:47,694 - WARN  [receiver-supervisor-future-0:o.a.s.s.r.ReceiverSupervisorImpl@87] - Restarting receiver with delay 2000 ms: Failed to fetch new messages using subscription 'projects/fernando-test-project-291016/subscriptions/input-sub-1' for project 'fernando-test-project-291016'.
com.google.api.gax.rpc.DeadlineExceededException: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 24.999882589s. [remote_addr=pubsub.googleapis.com/64.233.181.95:443]
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:51) ~[na:na]
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72) ~[na:na]
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60) ~[na:na]
	at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:97) ~[na:na]
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:68) ~[na:na]
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1056) ~[na:na]
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30) ~[na:na]
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1138) ~[guava-13.0.1.jar:na]
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:958) ~[guava-13.0.1.jar:na]
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:748) ~[guava-13.0.1.jar:na]
	at io.grpc.stub.ClientCalls$GrpcFuture.setException(ClientCalls.java:563) ~[na:na]
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:533) ~[na:na]
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:617) ~[na:na]
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:70) ~[na:na]
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:803) ~[na:na]
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:782) ~[na:na]
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37) ~[na:na]
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123) ~[na:na]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_272]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_272]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180) ~[na:1.8.0_272]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293) ~[na:1.8.0_272]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_272]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_272]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_272]
Caused by: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 24.999882589s. [remote_addr=pubsub.googleapis.com/64.233.181.95:443]
	at io.grpc.Status.asRuntimeException(Status.java:533) ~[na:na]
	... 14 common frames omitted
```

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Wrapper class for Pub/Sub message contents.
+ *
+ * We use this class to mirror the contents of Google's com.google.pubsub.v1.ReceivedMessage without having to add
+ * an additional exported dependency for our workers.
+ */
+public class PubSubMessage implements Externalizable {
+
+  private static final long serialVersionUID = 1L;
```
chtyim: You don't need this if you implement Externalizable

src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubMessageTest.java
```diff
@@ -0,0 +1,124 @@
...
+    objectOutputStream.flush();
+    objectOutputStream.close();
+    outputStream.close();
+
+    byte[] serialized = outputStream.toByteArray();
+
+    ByteArrayInputStream inputStream = new ByteArrayInputStream(serialized);
+    ObjectInputStream objectInputStream = new ObjectInputStream(inputStream);
+
+    PubSubMessage deserialized = new PubSubMessage();
```
chtyim: You shouldn't call this and the next line directly. You should read the object from the ObjectInputStream and let Java to construct the object.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    this.messageId = messageId;
+    this.orderingKey = orderingKey;
+    this.ackId = ackId;
+    this.data = data;
+    this.attributes = attributes;
+    this.publishTime = publishTime;
+  }
+
+  public PubSubMessage(ReceivedMessage message) {
+    if (message == null) {
```
chtyim, line 62: Add `@Nullable` annotation to parameter that can be null.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+      return;
+    }
+
+    if (message.getMessage() != null) {
```
chtyim: There is too many calls to `message.getMessage()`, making it hard to read. Introduce a local variable.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+      this.messageId = message.getMessage().getMessageId();
+      this.orderingKey = message.getMessage().getOrderingKey();
+
+      if (message.getMessage().getData() != null) {
+        this.data = message.getMessage().getData().toByteArray();
```
chtyim: Don't call `toByteArray()` on `String` without providing an encoding.

chtyim: Also, seems like you should call the `decodeData` method instead to get the data as `byte[]`.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  public Instant getPublishTime() {
+    return publishTime;
+  }
+
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    out.writeInt(messageId != null ? messageId.length() : -1);
+    if (messageId != null) {
```
chtyim: An `if else` would be clearer:

```
if (messageId == null) {
  out.writeBoolean(false);
} else {
  out.writeBoolean(true);
  out.writeUTF(messageId);
}
```
}

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+    if (orderingKey != null) {
+      out.writeUTF(orderingKey);
+    }
+
+    out.writeInt(ackId != null ? ackId.length() : -1);
+    if (ackId != null) {
+      out.writeUTF(ackId);
+    }
+
+    out.writeObject(publishTime);
```
chtyim: Avoid using `writeObject` since it will rely on the object to implement Serializable / Externalizable. If it is the former case, compatibility is not guaranteed. In this particular case, write out a long and an int for the seconds and nanos instead.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
chtyim: Since you have many cases that need to write out a nullable String, introduce a private method for that purpose (write a boolean and then the non-null string as I commented above).

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,179 @@
...
+
+    out.writeInt(attributes != null ? attributes.size() : -1);
+    if (attributes != null) {
+      for (Map.Entry<String, String> entry : attributes.entrySet()) {
+        out.writeObject(entry.getKey());
+        out.writeObject(entry.getValue());
+      }
+    }
+
```
chtyim, line 133: Remove extra new line

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,546 @@
...
+    "Failed to create subscriber using subscription '%s' for project '%s'.";
+  private static final String FETCH_ERROR_MSG =
+    "Failed to fetch new messages using subscription '%s' for project '%s'.";
+
+  // Retryable status codes. These need to be handled in case Pub/Sub throws a StatusRuntimeException.
+  private static final int RESOURCE_EXHAUSTED = 429;
+  private static final int CANCELLED = 499;
+  private static final int INTERNAL = 500;
+  private static final int UNAVAILABLE = 503;
+  private static final int DEADLINE_EXCEEDED = 504;
```
chtyim: These are standard HTTP code. Use the HttpURLConnection.HTTP_* constants instead.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
chtyim: So you don't need to log here.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,546 @@
...
+
+    List<ReceivedMessage> receivedMessages = pullResponse.getReceivedMessagesList();
+
+    //If there are no messages to process, continue.
+    if (receivedMessages.isEmpty()) {
+      return;
+    }
+
+    //Decrement number of available messages in bucket.
+    bucket.updateAndGet(x -> x - receivedMessages.size());
```
chtyim, line 371: Seems like you can simply use `decrementAndGet(receivedMessages.size())`

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriber.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Base implementation of a Realtime source plugin to read from Google PubSub.
+ *
+ * @param <T> The type that the PubSubMessage will be mapped to by the mapping function.
+ */
+public abstract class PubSubSubscriber<T> extends StreamingSource<T> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(PubSubSubscriber.class);
+
+  protected PubSubSubscriberConfig config;
```
chtyim: final fields

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
```diff
@@ -0,0 +1,86 @@
...
+import org.slf4j.LoggerFactory;
+import scala.Serializable;
+import scala.reflect.ClassTag;
+
+import java.util.ArrayList;
+
+/**
+ * Utility class to create a JavaDStream of received messages.
+ */
+public class PubSubSubscriberUtil implements Serializable {
```
chtyim: Why a util class needs to be serializable?

chtyim: final and add a private constructor for util class.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I will, thanks for the advice @chtyim

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
fernst: Oops, good catch.

src/main/java/io/cdap/plugin/gcp/publisher/source/GCPCredentialsProvider.java
fernst: Removed.

src/test/java/io/cdap/plugin/gcp/publisher/source/PubSubMessageTest.java
fernst: I've implemented your change and now read from the ObjectInputStream directly.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Not all of them appear in the `HttpURLConnection` class.

Would it help to use Google's Status Codes? https://googleapis.dev/java/gax/1.44.0/com/google/api/gax/rpc/StatusCode.Code.html

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I don't think AtomicInteger implements this method in JDK8, at least https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html

pom.xml
fernst: I have added the jUnit dependency again.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I have added the `@Nullable` annotation

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I don't think it should, but since this comes from the Pub/Sub api, I prefer to guard against this case.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I would say that `null` represents "this was not present in original payload" vs empty which represents "No attributes in this particular message".

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Interesting. I do see the value, but I would like to keep this solution as-is for the time being. I can come back and refine after the initial PR has been merged.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I have added this annotation to the constructor.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: It does, writeUTF throws NullPointerException for a Null String.

fernst:
@albertshau @chtyim I have addressed your comments and questions. Let me know if you'd like additional changes.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,202 @@
...
+    PubsubMessage psMessage = message.getMessage();
+
+    if (psMessage != null) {
+      this.messageId = psMessage.getMessageId();
+      this.orderingKey = psMessage.getOrderingKey();
+
+      if (psMessage.getData() != null) {
+        this.data = psMessage.getData().toStringUtf8().getBytes(StandardCharsets.UTF_8);
```
chtyim: Why we turn the data to string utf8 and then turn it into bytes again? Shouldn't we call `getData().toByteArray()` to get the raw bytes directly?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
```diff
@@ -0,0 +1,202 @@
...
+      if (psMessage.getPublishTime() != null) {
+        this.publishTime = Instant.ofEpochSecond(psMessage.getPublishTime().getSeconds())
+          .plusNanos(psMessage.getPublishTime().getNanos());
+      }
+    }
+
+    this.ackId = message.getAckId();
+  }
+
+  public String getMessageId() {
```
chtyim, line 90: From the implementation, seems like all these fields can be null, meaning you'll need to add a `@Nullable` annotation to the getter methods.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,598 @@
...
+      }
+    }
+
+    LOG.info("Receiver completed execution");
+  }
+
+  /**
+   * Build credentials for our Pub/Sub client.
+   *
+   * @return boolean indicating if the credentials could be created (or declared null).
```
chtyim: This javadoc seems not accurate

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,598 @@
...
+   */
+  protected Credentials createCredentials() {
```
chtyim, line 200: The method should be annotated with `@Nullable`

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: That used to be the case, but then you pointed out what this meant we're assuming the underlying payload is already encoded as UTF-8.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I will add this annotation.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: Oops, let me review the javadoc.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I will add this annotation.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
chtyim: hum... I was looking at the wrong API. I thought it was a method on `String`. The `ByteString.toByteArray()` should be the right call.

fernst: But you're right. In theory, messages don't need to be strings, so this is making the assumption that the payload is a UTF-8 string. I'll change this.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I have added this annotation.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I have added these annotations.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: I have fixed the Javadoc.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubMessage.java
fernst: I have fixed this method.

docs/GoogleSubscriber-streamingsource.md
```diff
@@ -57,3 +57,6 @@ the creation of subscriber will be received.
 authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 * **JSON**: Contents of the service account JSON file.
+
+**Number of readers**: Number of Pub/Sub reader workers to run in parallel for this source. Each reader worker
```
albertshau: reader worker -> reader

albertshau (approved):

fernst:
@rmstar can you take a look at this PR? I had to include the schema changes into my changes for the new Pub/Sub Receiver implementation

docs/GoogleSubscriber-streamingsource.md
fernst: Updated.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,600 @@
...
+    add(UNAVAILABLE);
+    add(DEADLINE_EXCEEDED);
+  }});
+
+  private final PubSubSubscriberConfig config;
+  private final boolean autoAcknowledge;
+  private final BackoffConfig backoffConfig;
+  private int previousFetchRate = -1;
+
+  //Transient properties used by the receiver in the worker node.
```
rmstar, line 97: nit: space after //. Similar comment for other places in this file missing space.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,600 @@
...
+  protected void fetchAndAck() {
+    //Get the maximun number of messages to get. If this number is less or equal than 0, do not fetch.
+    int maxMessages = bucket.get();
+    if (maxMessages <= 0) {
+      return;
+    }
+
+    PullRequest pullRequest =
+      PullRequest.newBuilder()
+        .setMaxMessages(maxMessages)
```
rmstar, line 358: If a rate limit config is not specified will we fetch up to MAX_INT messages? If yes, we might want to limit the max number of messages fetched in a pull request to some reasonable number.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
```diff
@@ -0,0 +1,600 @@
...
+   * Implements exponential backoff strategy when a retryable exception is received.
+   * This method stops the receiver if a non retryable ApiException is thrown by the Google Cloud subscriber client.
+   */
+  protected void receiveMessages() {
+    int backoff = backoffConfig.getInitialBackoffMs();
+
+    //Try with backoff until stopped or the task succeeds.
+    while (!isStopped()) {
+      try {
+        fetchAndAck();
```
rmstar, line 322: if bucket.get() == 0, will we spin in a tight loop for up to a second until the bucket is refilled?

rmstar (commented):
CDAP sets `spark.executor.instances` based on the number of sources. I think the assumption is that each streaming source has 1 receiver. Does the new configuration to support multiple receivers break this assumption? 
CC @albertshau

fernst:
@rmstar @albertshau The `number of sources` calculation takes into account the number of required executors per receiver. In our case, this value is taken from the `Number of Readers` value our users specify.

https://github.com/cdapio/cdap/blob/6e073de9df71529b001e8d99ed10aa357b365616/cdap-app-templates/cdap-etl/cdap-data-streams/src/main/java/io/cdap/cdap/datastreams/DataStreamsSparkLauncher.java#L95

rmstar:
> @rmstar @albertshau The `number of sources` calculation takes into account the number of required executors per receiver. In our case, this value is taken from the `Number of Readers` value our users specify.
> 
> https://github.com/cdapio/cdap/blob/6e073de9df71529b001e8d99ed10aa357b365616/cdap-app-templates/cdap-etl/cdap-data-streams/src/main/java/io/cdap/cdap/datastreams/DataStreamsSparkLauncher.java#L95

I see, thanks @fernst

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: In practice, I saw this value ranging between ~3k and ~20k records per pull at most, depending on the size of messages.

fernst: This is merely a "suggestion/limit" for the Pub/Sub API.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubReceiver.java
fernst: This does not really "spin". If `bucket` has a value of 0, the `fetchAndAck` method will return before making any network calls and this method will sleep for 100 ms until the next iteration is invoked.

PR #429, opened 2020-12-01
[CDAP-17423] Use upsert when writing to datastore

rmstar:
Insert fails if a record with the same key already exists. Use upsert instead to handle this case.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
+++ b/src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
@@ -118,7 +118,7 @@ public class DatastoreRecordWriter extends RecordWriter<NullWritable, Entity> {
         .build();
       builder.addMutations(DatastoreHelper.makeInsert(fullEntity).build());
     } else {
-      builder.addMutations(DatastoreHelper.makeInsert(entity).build());
+      builder.addMutations(DatastoreHelper.makeUpsert(entity).build());
     }
     ++totalCount;
     ++numberOfRecordsInBatch;


src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
```diff
@@ -116,9 +116,9 @@ public void write(NullWritable key, Entity entity) throws IOException {
         .setKey(response.getKeysList().stream().collect(MoreCollectors.onlyElement()))
         .putAllProperties(entity.getPropertiesMap())
         .build();
-      builder.addMutations(DatastoreHelper.makeInsert(fullEntity).build());
+      builder.addMutations(DatastoreHelper.makeUpsert(fullEntity).build());
```
CuriousVini: why does this test fail now? Was there any backwards incompatible change merged to plugin?

rmstar: This is a regression in develop branch caused by #376 where I had migrated the plugin to use `com.google.datastore.v1` library instead of `com.google.cloud.datastore`.

CuriousVini: I see. Would this mean users who are using insert operation on plugin version that uses `com.google.cloud.datastore` api will see pipeline failures after upgrading to version that develop is using?

rmstar: yes, user's pipeline would fail if:
- their pipeline tries to write an entity that already exists in datastore.
- they upgrade to version that develop is using (doesn't have the fix in this in PR).

rmstar: The expected behavior in this case is to overwrite the entity that already exists. [integration test](https://github.com/cdapio/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/io/cdap/cdap/app/etl/gcp/GoogleCloudDatastoreTest.java#L214)

The old code uses  [put](https://github.com/data-integrations/google-cloud/pull/376/files#diff-9e0b36731c60e30d64ed56237bf2442cb1e23995435ccd5d84f17061776eab5dL69), which is an [upsert operation](https://googleapis.dev/java/google-cloud-datastore/latest/com/google/cloud/datastore/DatastoreBatchWriter.html#put-com.google.cloud.datastore.FullEntity-)

rmstar: hm actually only `put` is an upsert operation, [putWithDeferredIdAllocation](https://googleapis.dev/java/google-cloud-datastore/latest/com/google/cloud/datastore/DatastoreBatchWriter.html#putWithDeferredIdAllocation-com.google.cloud.datastore.FullEntity...-) is not. I'll update the PR to remove the unnecessary upsert.

MEseifan (approved):

src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreRecordWriter.java
CuriousVini: I see.

CuriousVini (approved):
lgtm 👍

PR #436, opened 2020-12-09
Fixed issue with output schema in PLUGIN-409

flakrimjusufi:
BigQuery sink should not require a set schema

JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-409

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -337,14 +337,14 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
     return BigQueryUtil.getTableSchema(bqSchema, collector);
   }
 
-  protected void validateInsertSchema(Table table, Schema tableSchema, FailureCollector collector) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -396,7 +396,7 @@ protected void validateSchema(
     Schema tableSchema,
```
CuriousVini: table schema is not marked as nullable

CuriousVini (commented):
could you please provide more information on how was this issue reproduced and tested. Also is it possible to add unit test for this?

flakrimjusufi:
> could you please provide more information on how was this issue reproduced and tested. Also is it possible to add unit test for this?

This is a simple step that was used to reproduce the error: 

**- Avro file in GCS Source -> Output schema empty**
**- BigQuery in sink -> Output schema empty** 

**Result: Null error occurred while configuring the stage.**

CuriousVini (approved):

CuriousVini:
Please squash the commits

bajram-adapt:
squashed

CuriousVini (approved):

PR #437, opened 2020-12-10
PLUGIN-318: Added the feature to support formats in PubSub Source/Sink

flakrimjusufi:
Support formats in PubSub Source/Sink

JIRA ticket: https://cdap.atlassian.net/browse/PLUGIN-318

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/publisher/PubSubOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/PubSubOutputFormat.java
@@ -25,7 +25,18 @@ import com.google.cloud.pubsub.v1.Publisher;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.ProjectTopicName;
 import com.google.pubsub.v1.PubsubMessage;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.format.StructuredRecordStringConverter;
+import io.cdap.cdap.format.io.StructuredRecordDatumWriter;
+import io.cdap.plugin.format.avro.StructuredToAvroTransformer;
+import io.cdap.plugin.gcp.bigtable.sink.BigtableSinkConfig;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import io.cdap.plugin.gcp.gcs.actions.GCSBucketCreate;
+import org.apache.avro.generic.GenericDatumWriter;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.BinaryEncoder;
+import org.apache.avro.io.DatumWriter;
+import org.apache.avro.io.EncoderFactory;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.io.Text;
@@ -38,6 +49,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.threeten.bp.Duration;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -48,7 +60,7 @@ import java.util.concurrent.atomic.AtomicReference;
 /**
  * OutputFormat to write to Pub/Sub topic.
  */
-public class PubSubOutputFormat extends OutputFormat<NullWritable, Text> {
+public class PubSubOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(PubSubOutputFormat.class);
 
   private static final String SERVICE_ACCOUNT = "service.account";
@@ -66,6 +78,8 @@ public class PubSubOutputFormat extends OutputFormat<NullWritable, Text> {
   public static void configure(Configuration configuration, GooglePublisher.Config config) {
 
     String serviceAccount = config.getServiceAccount();
+    String format = config.getFormat();
+    String delimiter = config.getDelimiter();
     if (serviceAccount != null) {
       configuration.set(SERVICE_ACCOUNT_TYPE, config.getServiceAccountType());
       configuration.set(SERVICE_ACCOUNT, config.getServiceAccount());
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -27,8 +28,10 @@
 import io.cdap.cdap.etl.api.streaming.StreamingContext;
 import io.cdap.cdap.etl.api.streaming.StreamingSource;
 import io.cdap.cdap.etl.api.streaming.StreamingSourceContext;
+import io.cdap.cdap.format.StructuredRecordStringConverter;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.gcp.common.GCPReferenceSourceConfig;
+import io.cdap.plugin.gcp.datastore.source.util.DatastoreSourceConstants;
```
rmstar: Don't add a dependency on other plugins. At some point, we want to split gcp plugins into multiple packages - see https://cdap.atlassian.net/browse/PLUGIN-478.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -75,14 +80,16 @@ public GoogleSubscriber(SubscriberConfig conf) {
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
-    pipelineConfigurer.getStageConfigurer().setOutputSchema(DEFAULT_SCHEMA);
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    config.validate(collector);
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
```
rmstar, line 94: Please test with schema set to macro.

If schema is a macro, getSchema() will return the default schema here, and you will set output  schema = default schema.

flakrimjusufi: Nice catch. You're right. 

Please check the updated PR, I made some changes in getSchema() and in validate() that handles this scenario.

rmstar (commented):
In this PR we are assuming that the message type is JSON if it's not bytes. Other plugins do not work like this. Instead, can we use format plugins to set message format? 
Also, can we make a similar change for Pub/Sub sink?
Screen Shot 2020-12-21 at 3 17 03 PM

bajram-adapt:
> In this PR we are assuming that the message type is JSON if it's not bytes. Other plugins do not work like this. Instead, can we use format plugins to set message format?
> Also, can we make a similar change for Pub/Sub sink?
> Screen Shot 2020-12-21 at 3 17 03 PM

Hi Prashant, for a requirements change of that nature we should make a new Jira item with the full requirements, and have Bhooshan pull the item into whatever priority he feels is necessary as the effort for requirement change is not calculated into our Dec timeline.

src/main/java/io/cdap/plugin/gcp/publisher/PubSubOutputFormat.java
```diff
@@ -166,6 +190,64 @@ public void onSuccess(String s) {
...
+          out.close();
+          serializedBytes = out.toByteArray();
+          data = ByteString.copyFrom(serializedBytes);
+          message = PubsubMessage.newBuilder().setData(data).build();
+          break;
+        }
+        case PubSubConstants.JSON:
+        case PubSubConstants.BLOB:
+        case PubSubConstants.TEXT: {
+          payload = StructuredRecordStringConverter.toJsonString(value);
```
rmstar, line 228: why are we converting blob and text to json string?

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -110,22 +126,80 @@ public void prepareRun(StreamingSourceContext context) throws Exception {
     ClassTag<SparkPubsubMessage> tag = scala.reflect.ClassTag$.MODULE$.apply(SparkPubsubMessage.class);
     JavaReceiverInputDStream<SparkPubsubMessage> pubSubMessages = new JavaReceiverInputDStream<>(stream, tag);
 
+    Schema customMessageSchema = getCustomMessageSchema();
+    final Schema outputSchema = config.getSchema();
+    final String format = config.format.toLowerCase();
```
rmstar: use getFormat() instead of accessing format directly.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -110,22 +126,80 @@ public void prepareRun(StreamingSourceContext context) throws Exception {
...
+        .setTimestamp("timestamp", getTimestamp(pubSubMessage.getPublishTime()))
+        .set("attributes", hashMap)
+        .build();
     });
   }
 
+  private StructuredRecord getStructuredRecord(Schema customMessageSchema, String format,
+                                               SparkPubsubMessage pubSubMessage) throws IOException {
+    StructuredRecord payload = null;
+    final String getData = new String(pubSubMessage.getData());
```
rmstar: Rename getData -> data

rmstar (commented):
Is this change backward compatible (message type is bytes if format is not specified)?

flakrimjusufi:
> Is this change backward compatible (message type is bytes if format is not specified)?

This change is backward compatible. If message type is bytes and the format is not specified, this will default to format TEXT which is a BYTE and it will work as expected. I can also add another check for message type as byte but I think this is not needed. Please correct me if my assumption is wrong.

src/main/java/io/cdap/plugin/gcp/publisher/PubSubOutputFormat.java
flakrimjusufi: PR updated.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
flakrimjusufi: PR updated.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
flakrimjusufi: PR updated.

rmstar (approved):
lgtm, please squash commits.

Also, please add an integration test. https://github.com/cdapio/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/io/cdap/cdap/app/etl/gcp/PubSubTest.java

flakrimjusufi:
> lgtm, please squash commits.
> 
> Also, please add an integration test. https://github.com/cdapio/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/io/cdap/cdap/app/etl/gcp/PubSubTest.java

Commits squashed.

rmstar (approved):
lgtm, please squash commits

bajram-adapt:
> lgtm, please squash commits

squashed

PR #438, opened 2020-12-11
fix: PLUGIN-481 Write nullable array as empty array

ardian4:
Regardless if array is nullable or not set mode to REPEATED, in case when array is null in BigQuery there will be empty array

Tests: 
   For BigQuery Sink:

1. Verify that with this dataset example pipeline fails:
```
       {"id":1,"names":["JKimiohn","Jane"]}
       {"id":2,"names":null} 
```
2. Do the code change, check in BigQuery table there are 2 rows and row with id 2 has empty value 
3. Verify also that nonnullable array is not broken  with this dataset example:
```
{"id":1,"names":["JKimiohn","Jane"]}
{"id":2,"names":["JKimiohn"]}

```
 For BigQueryMultiSink:

 1. Verify that with this dataset example pipeline fails
```
   {"id":1,"names":["JKimiohn","Jane"],"table":"table_a"}
   {"id":2,"names":null,"table":"table_b"} 
```
2. Do the code change, check in BigQuery if `table_b` has row with id 2 and names are empty 
3. Verify also that nonnullable array is not broken with this dataset example:
```
{"id":1,"names":["JKimiohn","Jane"],"table":"table_a"}
{"id":2,"names":["Jane"],"table":"table_b"}
```
   
  

JiraTicket: https://cdap.atlassian.net/browse/PLUGIN-403

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -527,9 +527,11 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
   }
 
   private Field.Mode getMode(Schema schema) {
-    if (schema.isNullable()) {
+    boolean isNullable = schema.isNullable();
+    Schema.Type nonNullableType = isNullable ? schema.getNonNullable().getType() : schema.getType();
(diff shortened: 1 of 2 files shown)


rmstar (approved):
Can you please list the tests that you ran to verify this fix? Thanks.

ardian4:
> Can you please list the tests that you ran to verify this fix? Thanks.

Done.

PR #444, opened 2020-12-15
PLUGIN-102 remove hardcoded format list

albertshau:
Removed hardcoded format list and switched to a dynamic list
based on the actual format plugins installed.


supriyapatil-google (approved):
Thanks!

PR #447, opened 2020-12-15
[CDAP-17473] Added conditions to not populate fileSystemProperties for path and projectId if either contains a macro

vravish:


src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -63,7 +63,8 @@ public String getProject() {
   @Nullable
   public String tryGetProject() {
     if (containsMacro(NAME_PROJECT) && Strings.isNullOrEmpty(project)) {
-      return null;
+      // macro substitution has not been performed yet, so return raw value for validation
+      return getRawProperties().getProperties().get(NAME_PROJECT);
```
albertshau: this is not the right fix, it will return an invalid value for the project. Whatever is trying to do validation with the project should make sure it's not a macro before doing the check.

CuriousVini: +1 to what Albert mentioned

src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
```diff
@@ -86,7 +87,7 @@ public int hashCode() {
    * @throws IllegalArgumentException if the path string is invalid
    */
   public static GCSPath from(String path) {
-    if (path.isEmpty()) {
+    if (Strings.isNullOrEmpty(path)) {
```
albertshau, line 90: why is this change needed? normally we assume the input is non-null unless it's annotated as nullable.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -140,9 +141,13 @@ public static String extractPrivateKey(ServiceAccountCredentials credentials) {
     }
     properties.put("fs.gs.impl", GoogleHadoopFileSystem.class.getName());
     properties.put("fs.AbstractFileSystem.gs.impl", GoogleHadoopFS.class.getName());
-    String projectId = config.getProject();
-    properties.put(FS_GS_PROJECT_ID, projectId);
-    properties.put("fs.gs.system.bucket", GCSPath.from(path).getBucket());
+    if (!config.containsMacro(GCPConfig.NAME_PROJECT)) {
```
albertshau, line 144: I don't understand how this fixes the issue in the bug. It mentions an error like "Could not detect Google Cloud project id from the environment." What is throwing that error?

vravish: That error happens during validation (https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java#L58), because during validation, [GCPUtils](https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L143) tries to populate the fileSystemProperties map with projectId and it tries to parse the gcsPath before either field is available. The solution, as in most cases, is to not validate either of those fields if they contain macros. And containsMacro returns false by pipeline runtime, so the map will be populated with actual values then.

src/main/java/io/cdap/plugin/gcp/gcs/GCSPath.java
vravish: If it is null, as it was before this bug fix for GCSSource, then the NullPointerException thrown is harder to debug than an IllegalArgumentException.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
albertshau: so that's the place that has the bug. It shouldn't be complaining about a missing project if the project is macro enabled, or the caller shouldn't be calling it if the project has a macro.

vravish: > so that's the place that has the bug

"that" meaning GCPUtils? If so, then this does fix it, since the caller (GCPUtils.getFileSystemProperties) does not call if there is a macro.

albertshau: these are used to configure the inputformat used to read data. It's not correct to just avoid putting these properties here, as whatever is expecting them will break. 

@CuriousVini it looks like there was a change in AbstractFileSource to call this, I think for schema detection? It should not blindly do schema detection if not all the required fields have values.

CuriousVini: yes that is correct. This is caused by following call
https://github.com/cdapio/hydrator-plugins/blob/develop/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java#L113

PR #456, opened 2020-12-17
Fixed: PLUGIN-499

nbesimi:
JIRA ticket: https://cdap.atlassian.net/browse/PLUGIN-499

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
@@ -24,6 +24,7 @@ import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.StandardTableDefinition;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableDefinition.Type;
+import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TimePartitioning;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import io.cdap.cdap.api.annotation.Description;
@@ -126,8 +127,7 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
     Schema configuredSchema = getOutputSchema(collector);
 
     String serviceAccount = config.getServiceAccount();
-    Credentials credentials = serviceAccount == null ?
-      null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
+    Credentials credentials = getCredentials(serviceAccount);
     BigQuery bigQuery = GCPUtils.getBigQuery(config.getDatasetProject(), credentials);
 
     uuid = UUID.randomUUID();
@@ -171,6 +171,9 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
     if (config.getViewMaterializationDataset() != null) {
       configuration.set(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_DATASET, config.getViewMaterializationDataset());
     }
+    String temporaryTableName = String.format("_%s_%s", config.getTable(),
+                                              UUID.randomUUID().toString().replaceAll("-", "_"));
+    configuration.set(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME, temporaryTableName);
 
     String temporaryGcsPath = String.format("gs://%s/%s/hadoop/input/%s", bucket, uuid, uuid);
     PartitionedBigQueryInputFormat.setTemporaryCloudStorageDirectory(configuration, temporaryGcsPath);
@@ -211,6 +214,24 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
 
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    deleteGcsTemporaryDirectory();
+    deleteBigQueryTemporaryTable();
+  }
+
+  private void deleteBigQueryTemporaryTable() {
+    String temporaryTable = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
+    try {
+      String serviceAccount = config.getServiceAccount();
+      Credentials credentials = getCredentials(serviceAccount);
+      BigQuery bigQuery = GCPUtils.getBigQuery(config.getDatasetProject(), credentials);
+      bigQuery.delete(TableId.of(config.getProject(), config.getDataset(), temporaryTable));
+      LOG.debug("Deleted temporary table '{}'", temporaryTable);
+    } catch (IOException e) {
+      LOG.error("Failed to load service account credentials: {}", e.getMessage(), e);
+    }
+  }
+
+  private void deleteGcsTemporaryDirectory() {
     org.apache.hadoop.fs.Path gcsPath = null;
     String bucket = config.getBucket();
     if (bucket == null) {
@@ -227,6 +248,12 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
     } catch (IOException e) {
       LOG.warn("Failed to delete temporary directory '{}': {}", gcsPath, e.getMessage());
     }
+  }
+
+  @Nullable
+  private Credentials getCredentials(@Nullable String serviceAccount) throws IOException {
+    return serviceAccount == null ?
+            null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
   }
 
   public Schema getSchema(FailureCollector collector) {
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
@@ -110,7 +110,7 @@ public class PartitionedBigQueryInputFormat extends AbstractBigQueryInputFormat<
       TableReference sourceTable = new TableReference().setDatasetId(datasetId).setProjectId(inputProjectId)
         .setTableId(tableName);
       String location = bigQueryHelper.getTable(sourceTable).getLocation();
-      String temporaryTableName = String.format("%s_%s", tableName, UUID.randomUUID().toString().replaceAll("-", "_"));
+      String temporaryTableName = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
       TableReference exportTableReference = createExportTableReference(type, inputProjectId, datasetId,
                                                                        temporaryTableName, configuration);
       runQuery(bigQueryHelper, inputProjectId, exportTableReference, query, location);
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
@@ -42,4 +42,5 @@ public interface BigQueryConstants {
   String CONFIG_PARTITION_INTEGER_RANGE_START = "cdap.bq.sink.partition.integer.range.start";
   String CONFIG_PARTITION_INTEGER_RANGE_END = "cdap.bq.sink.partition.integer.range.end";
   String CONFIG_PARTITION_INTEGER_RANGE_INTERVAL = "cdap.bq.sink.partition.integer.range.interval";
+  String CONFIG_TEMPORARY_TABLE_NAME = "cdap.bq.source.temporary.table.name";
 }


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -171,6 +172,9 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     if (config.getViewMaterializationDataset() != null) {
       configuration.set(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_DATASET, config.getViewMaterializationDataset());
     }
+    String temporaryTableName = String.format("%s_%s", config.getTable(),
```
bdmogal: minor nit: can you please prefix this by `_`? Tables that start with `_` are not displayed in the BQ Console by default.

nbesimi: changed

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
```diff
@@ -42,4 +42,5 @@
   String CONFIG_PARTITION_INTEGER_RANGE_START = "cdap.bq.sink.partition.integer.range.start";
   String CONFIG_PARTITION_INTEGER_RANGE_END = "cdap.bq.sink.partition.integer.range.end";
   String CONFIG_PARTITION_INTEGER_RANGE_INTERVAL = "cdap.bq.sink.partition.integer.range.interval";
+  String CONFIG_TEMPORARY_TABLE_NAME = "cdap.source.temporary.table.name";
```
CuriousVini: should follow same pattern of string as above

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +215,28 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
 
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    deleteGcsTemporaryDirectory();
+    deleteBigQueryTemporaryTable();
+  }
+
+  private void deleteBigQueryTemporaryTable() {
+    String temporaryTable = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
+    if (temporaryTable != null) {
```
CuriousVini: This will always be non null

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +215,28 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
...
+      String serviceAccount = config.getServiceAccount();
+      Credentials credentials = null;
+      try {
+        credentials = serviceAccount == null ?
+                null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
+      } catch (IOException e) {
+        LOG.warn("Failed to load service account credentials: {}", e.getMessage());
```
CuriousVini: if service account is failed to load is it safe to proceed to delete the table?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +215,28 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
...
+    deleteBigQueryTemporaryTable();
+  }
+
+  private void deleteBigQueryTemporaryTable() {
+    String temporaryTable = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
+    if (temporaryTable != null) {
+      String serviceAccount = config.getServiceAccount();
+      Credentials credentials = null;
+      try {
+        credentials = serviceAccount == null ?
```
CuriousVini: I see this logic at multiple places in this class. Lets create a method and reuse this logic

CuriousVini (changes requested):
1. Is this also affecting bq sink and bq multi sink? If so, we should fix it for those plugins too
2. What are the test cases that were performed for these change?

nbesimi:
> 1. Is this also affecting bq sink and bq multi sink? If so, we should fix it for those plugins too
> 2. What are the test cases that were performed for these change?

1. This PR is for ticket **PLUGIN-499** **BigQuery source does not delete the temporary tables created while exporting data to GCS**, so our focus was only for the BigQuery Source. 

2. The test cases that we performed for these changes:
  - Use BigQuery source without filter and table without partition: No temporary table gets created.
  - Use BigQuery source with filter: Temporary table gets created and gets deleted at the end.
  - Use BigQuery source for time partitioned table: Temporary table gets created and gets deleted at the end.
  - Use BigQuery source for VIEW: Temporary table gets created and gets deleted at the end.
  - Use BigQuery source for MATERIALIZED_VIEW: Temporary table gets created and gets deleted at the end.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java
nbesimi: Changed to: `String CONFIG_TEMPORARY_TABLE_NAME = "cdap.bq.source.temporary.table.name";`

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: @CuriousVini This will be null only in the case where we use the BigQuerySource for a Table without **filter** and table without **partition** configured. Therefore we had to add this check. Test case #1 from the comment below.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: Good catch. If the service account fails we are not proceeding delete table.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: Refactored. A new method got created:

```
private Credentials getCredentials(final String serviceAccount) throws IOException {
    return serviceAccount == null ?
            null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
  }
```

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
CuriousVini: 1. So this is an actual error. In that case, move the code below catch into try block. After that, there is no need to add return.
2. The logger level should be "ERROR"
3. In the log we should include exception 
LOG.error("Failed to load service account credentials: {}", e.getMessage(), e);

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
CuriousVini: We always set it in line 174 right?

nbesimi: @CuriousVini in line 174 only the table name gets created. In the class `PartitionedBigQueryInputFormat` line 114 the temporary table gets created if `query != null` (this happens when we query a Table with filter or **partition** configured.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: Updated

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
CuriousVini: so this check is for the string right? Meaning it will never be null since we always create it. Wondering why we need this check then? 
```
if (temporaryTable != null)

**Comment by CuriousVini** [comment]
For 1, it applies to bigquery sink as well. Please file a separate jira for it:
https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java#L365

File: src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java [hunk 6]
```diff
@@ -110,7 +110,7 @@ private void processQuery(JobContext context) throws IOException, InterruptedExc
       TableReference sourceTable = new TableReference().setDatasetId(datasetId).setProjectId(inputProjectId)
         .setTableId(tableName);
       String location = bigQueryHelper.getTable(sourceTable).getLocation();
-      String temporaryTableName = String.format("%s_%s", tableName, UUID.randomUUID().toString().replaceAll("-", "_"));
+      String temporaryTableName = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
```
CuriousVini, line 113: Also why are we not expecting `temporaryTableName` to be null here?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: @CuriousVini we apologize for the confusion, yes you are right. We do not need this check here. It does not make sense.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -229,6 +250,11 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     }
   }
 
+  private Credentials getCredentials(final String serviceAccount) throws IOException {
```
CuriousVini: Mark method and parameter as nullable

CuriousVini: no need to make this final

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +214,24 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
...
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    deleteGcsTemporaryDirectory();
+    deleteBigQueryTemporaryTable();
+  }
+
+  private void deleteBigQueryTemporaryTable() {
+    String temporaryTable = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
+    String serviceAccount = config.getServiceAccount();
+    try {
+      Credentials credentials = getCredentials(serviceAccount);
```
CuriousVini, line 225: handle null credentials here

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -126,8 +127,7 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     Schema configuredSchema = getOutputSchema(collector);
 
     String serviceAccount = config.getServiceAccount();
-    Credentials credentials = serviceAccount == null ?
-      null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
+    Credentials credentials = getCredentials(serviceAccount);
```
CuriousVini, line 130: handle null credentials

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +214,24 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
 
   @Override
   public void onRunFinish(boolean succeeded, BatchSourceContext context) {
+    deleteGcsTemporaryDirectory();
+    deleteBigQueryTemporaryTable();
+  }
+
+  private void deleteBigQueryTemporaryTable() {
+    String temporaryTable = configuration.get(BigQueryConstants.CONFIG_TEMPORARY_TABLE_NAME);
```
CuriousVini, line 222: nit: move this closer to where its used

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -211,6 +214,24 @@ public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<
...
+    String serviceAccount = config.getServiceAccount();
```
CuriousVini: this can be in try as well

CuriousVini (commented):
minor comments, fix and squash

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
nbesimi: null credentials can also be as a result of auto-detect.

In `getCredentials` method we are already handling null for `serviceAccount`:

```
@Nullable
  private Credentials getCredentials(@Nullable String serviceAccount) throws IOException {
    return serviceAccount == null ?
            null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
  }
```

And credentials null are also handled in `GCPUtils`, `getBigQuery`:

```
    if (credentials != null) {
      bigqueryBuilder.setCredentials(credentials);
    }
    return bigqueryBuilder.build().getService();
  }
```

rmstar (approved):

rmstar:
>> Is this also affecting bq sink and bq multi sink? If so, we should fix it for those plugins too
>
>  This PR is for ticket PLUGIN-499 BigQuery source does not delete the temporary tables created while exporting data to GCS, so our focus was only for the BigQuery Source.

Can you please check if this issue is also applicable to BQ sink and multi sink? If yes, we can open JIRAs to track fixing the sinks.

nbesimi:
> > > Is this also affecting bq sink and bq multi sink? If so, we should fix it for those plugins too
> > 
> > 
> > This PR is for ticket PLUGIN-499 BigQuery source does not delete the temporary tables created while exporting data to GCS, so our focus was only for the BigQuery Source.
> 
> Can you please check if this issue is also applicable to BQ sink and multi sink? If yes, we can open JIRAs to track fixing the sinks.

This is the JIRA to track fixing for sink and multi sink: https://cdap.atlassian.net/browse/PLUGIN-525

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
CuriousVini: I see

CuriousVini (approved):
lgtm

PR #457, opened 2020-12-17
updated/PLUGIN-296

ailegion:
added: Output schema to match fields from import query

Jira Ticket: https://cdap.atlassian.net/browse/PLUGIN-296

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
@@ -52,6 +52,7 @@ import io.cdap.plugin.gcp.common.Schemas;
 import io.cdap.plugin.gcp.spanner.SpannerArrayConstants;
 import io.cdap.plugin.gcp.spanner.SpannerConstants;
 import io.cdap.plugin.gcp.spanner.common.SpannerUtil;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
 import org.slf4j.Logger;
@@ -62,8 +63,11 @@ import java.io.IOException;
 import java.io.ObjectOutputStream;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -82,6 +86,8 @@ public class SpannerSource extends BatchSource<NullWritable, ResultSet, Structur
   private static final Statement.Builder SCHEMA_STATEMENT_BUILDER = Statement.newBuilder(
     String.format("SELECT  t.column_name,t.spanner_type, t.is_nullable FROM information_schema.columns AS t WHERE " +
                     "  t.table_catalog = ''  AND  t.table_schema = '' AND t.table_name = @%s", TABLE_NAME));
+  private static final String LIMIT = "limit";
+
   public static final String NAME = "Spanner";
   private final SpannerSourceConfig config;
   private Schema schema;
@@ -243,26 +249,55 @@ public class SpannerSource extends BatchSource<NullWritable, ResultSet, Structur
                                                          projectId)) {
       DatabaseClient databaseClient =
         spanner.getDatabaseClient(DatabaseId.of(projectId, config.instance, config.database));
-      Statement getTableSchemaStatement = SCHEMA_STATEMENT_BUILDER.bind(TABLE_NAME).to(config.table).build();
-      try (ResultSet resultSet = databaseClient.singleUse().executeQuery(getTableSchemaStatement)) {
-        List<Schema.Field> schemaFields = new ArrayList<>();
-        while (resultSet.next()) {
-          String columnName = resultSet.getString("column_name");
-          String spannerType = resultSet.getString("spanner_type");
-          String nullable = resultSet.getString("is_nullable");
-          boolean isNullable = "YES".equals(nullable);
-          Schema typeSchema = parseSchemaFromSpannerTypeString(columnName, spannerType, collector);
-          if (typeSchema == null) {
-            // this means there were failures added to failure collector. Continue to collect more failures
-            continue;
+      if (Strings.isNullOrEmpty(config.importQuery)) {
+        Statement getTableSchemaStatement = SCHEMA_STATEMENT_BUILDER.bind(TABLE_NAME).to(config.table).build();
+        try (ResultSet resultSet = databaseClient.singleUse().executeQuery(getTableSchemaStatement)) {
+          List<Schema.Field> schemaFields = new ArrayList<>();
+          while (resultSet.next()) {
+            String columnName = resultSet.getString("column_name");
+            String spannerType = resultSet.getString("spanner_type");
+            String nullable = resultSet.getString("is_nullable");
+            boolean isNullable = "YES".equals(nullable);
+            Schema typeSchema = parseSchemaFromSpannerTypeString(columnName, spannerType, collector);
+            if (typeSchema == null) {
+              // this means there were failures added to failure collector. Continue to collect more failures
+              continue;
+            }
+            Schema fieldSchema = isNullable ? Schema.nullableOf(typeSchema) : typeSchema;
+            schemaFields.add(Schema.Field.of(columnName, fieldSchema));
           }
-          Schema fieldSchema = isNullable ? Schema.nullableOf(typeSchema) : typeSchema;
-          schemaFields.add(Schema.Field.of(columnName, fieldSchema));
+          if (schemaFields.isEmpty() && !collector.getValidationFailures().isEmpty()) {
+            collector.getOrThrowException();
+          }
+          return Schema.recordOf("outputSchema", schemaFields);
         }
-        if (schemaFields.isEmpty() && !collector.getValidationFailures().isEmpty()) {
-          collector.getOrThrowException();
+      } else {
+        final Map<String, Boolean> nullableFields = getFieldsNullability(databaseClient);
+        Statement importQueryStatement = getStatementForOneRow(config.importQuery);
+        List<Schema.Field> schemaFields = new ArrayList<>();
+        try (ResultSet resultSet = databaseClient.singleUse().executeQuery(importQueryStatement)) {
+          while (resultSet.next()) {
+            final List<Type.StructField> structFields = resultSet.getCurrentRowAsStruct().getType().getStructFields();
+            for (Type.StructField structField : structFields) {
+              final Type fieldSpannerType = structField.getType();
+              final String columnName = structField.getName();
+              // there are cases when column name is not in metadata table such as "Select FirstName as name",
+              // so fallback is nullable
+              final boolean isNullable = nullableFields.getOrDefault(columnName, true);
+              final Schema typeSchema = parseSchemaFromSpannerType(fieldSpannerType, columnName, collector);
+              if (typeSchema == null) {
+                // this means there were failures added to failure collector. Continue to collect more failures
+                continue;
+              }
+              Schema fieldSchema = isNullable ? Schema.nullableOf(typeSchema) : typeSchema;
+              schemaFields.add(Schema.Field.of(columnName, fieldSchema));
+            }
+          }
+          if (schemaFields.isEmpty() && !collector.getValidationFailures().isEmpty()) {
+            collector.getOrThrowException();
+          }
+          return Schema.recordOf("outputSchema", schemaFields);
         }
-        return Schema.recordOf("outputSchema", schemaFields);
       }
     } catch (IOException e) {
       collector.addFailure("Unable to get Spanner Client: " + e.getMessage(), null)
@@ -272,6 +307,22 @@ public class SpannerSource extends BatchSource<NullWritable, ResultSet, Structur
       throw collector.getOrThrowException();
     }
 
+  }
+
+  private Statement getStatementForOneRow(String importQuery) {
+    String query;
+    // Matches any String containing the word 'limit' followed by a number
+    // ex: SELECT NAME FROM TABLE LIMIT 15
+    String regex = "^(?:[^;']|(?:'[^']+'))+ LIMIT +\\d+(.*)";
+    Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE | Pattern.CASE_INSENSITIVE);
+    if (pattern.matcher(importQuery).matches()) {
+      int index = StringUtils.lastIndexOf(importQuery, LIMIT);
+      String substringToReplace = importQuery.substring(index);
+      query = importQuery.replace(substringToReplace, "limit 1");
+    } else {
+      query = String.format("%s limit 1", importQuery);
+    }
+    return Statement.newBuilder(query).build();
   }
 
   @Nullable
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -78,10 +82,20 @@
 public class SpannerSource extends BatchSource<NullWritable, ResultSet, StructuredRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(SpannerSource.class);
   private static final String TABLE_NAME = "TableName";
+  private static final String TABLE_COLUMNS = "TableColumns";
   // listing table's schema documented at https://cloud.google.com/spanner/docs/information-schema
   private static final Statement.Builder SCHEMA_STATEMENT_BUILDER = Statement.newBuilder(
     String.format("SELECT  t.column_name,t.spanner_type, t.is_nullable FROM information_schema.columns AS t WHERE " +
                     "  t.table_catalog = ''  AND  t.table_schema = '' AND t.table_name = @%s", TABLE_NAME));
+  private static final Statement.Builder SCHEMA_STATEMENT_BUILDER_WITH_COLUMNS = Statement.newBuilder(
+    String.format("SELECT  t.column_name,t.spanner_type, t.is_nullable FROM information_schema.columns AS t WHERE " +
```
CuriousVini: if prefix of this string is same as above string, we can just concat 2 strings

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +255,40 @@ private Schema getSchema(FailureCollector collector) {
                                                          projectId)) {
       DatabaseClient databaseClient =
         spanner.getDatabaseClient(DatabaseId.of(projectId, config.instance, config.database));
-      Statement getTableSchemaStatement = SCHEMA_STATEMENT_BUILDER.bind(TABLE_NAME).to(config.table).build();
+      Map<String, String> columnNameMap = new HashMap<>();
+      // get columns from import query when query does not contain the '*' or 'case'
```
CuriousVini: why are we making this assumption? also why are we checking for 'case'?

CuriousVini:
Schema should be populated in accordance with import query

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ailegion: updated

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ailegion: Regarding ``*`` it indicates that user selects all column of the table.
Regarding ``case`` user can type select statement like in example:
``SELECT A, B,
  CASE A
    WHEN 90 THEN 'red'
    WHEN 50 THEN 'blue'
    ELSE 'green'
  END
  AS result
FROM Numbers``
in these cases we fallback to getting full schema of the table

ardian4:
> Schema should be populated in accordance with import query

update the PR to populated the schema accordance with import query

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
```diff
@@ -84,10 +84,6 @@ public void validate(FailureCollector collector) {
     if (!containsMacro(NAME_SCHEMA) && schema != null) {
       SpannerUtil.validateSchema(schema, SUPPORTED_TYPES, collector);
     }
-    if (!containsMacro(NAME_SCHEMA) && schema == null) {
```
CuriousVini: this will undo the changes for https://cdap.atlassian.net/browse/PLUGIN-251
Correct fix is to do something similar to bigquery source:
https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java#L194

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +252,41 @@ private Schema getSchema(FailureCollector collector) {
...
+      if (config.importQuery != null && !config.importQuery.contains("*") &&
```
CuriousVini: check for null and empty query

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
CuriousVini: What is `case` used for?

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +252,41 @@ private Schema getSchema(FailureCollector collector) {
...
+            if (column.toLowerCase().contains(" as ")) {
+              String[] columnNameAndAlias = column.split(COLUMN_ALIAS_SPLIT_PATTERN);
+              columnNameMap.put(columnNameAndAlias[0].trim(), columnNameAndAlias[1].trim());
+            } else {
+              columnNameMap.put(column.trim(), column.trim());
+            }
+          }
+        }
+      }
+      Statement getTableSchemaStatement = null;
```
CuriousVini: there is no need to initialize this as null

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +252,41 @@ private Schema getSchema(FailureCollector collector) {
...
+      if (columnNameMap.isEmpty()) {
+        getTableSchemaStatement = SCHEMA_STATEMENT_BUILDER.bind(TABLE_NAME).to(config.table).build();
+      } else {
+        getTableSchemaStatement = SCHEMA_STATEMENT_BUILDER.append(SCHEMA_STATEMENT_COLUMN_FILTER)
+          .bind(TABLE_NAME).to(config.table).bind(TABLE_COLUMNS)
+          .toStringArray(columnNameMap.keySet()).build();
```
CuriousVini: what happens when column map is empty? Has that case been tested?

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +252,41 @@ private Schema getSchema(FailureCollector collector) {
...
+      }
       try (ResultSet resultSet = databaseClient.singleUse().executeQuery(getTableSchemaStatement)) {
         List<Schema.Field> schemaFields = new ArrayList<>();
         while (resultSet.next()) {
           String columnName = resultSet.getString("column_name");
+          // remap column name to alias
+          if (!columnNameMap.isEmpty() && columnNameMap.containsKey(columnName)) {
```
CuriousVini: why is `!columnNameMap.isEmpty()` check needed? Why not just do `columnNameMap.containsKey(columnName)`?

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -241,11 +252,41 @@ private Schema getSchema(FailureCollector collector) {
...
+        !config.importQuery.toLowerCase().contains(" case ")) {
+        Matcher extractedColumnName = TABLE_COLUMNS_PATTERN.matcher(config.importQuery);
+        if (extractedColumnName.find() && extractedColumnName.groupCount() == 1) {
+          String selectedColumns = extractedColumnName.group(1);
+          String[] columns = selectedColumns.split(",");
+          for (String column : columns) {
+            // check for column aliases
+            if (column.toLowerCase().contains(" as ")) {
+              String[] columnNameAndAlias = column.split(COLUMN_ALIAS_SPLIT_PATTERN);
+              columnNameMap.put(columnNameAndAlias[0].trim(), columnNameAndAlias[1].trim());
```
CuriousVini: should check for the size of `columnNameAndAlias`

CuriousVini (commented):
Could you please provide list of select statements this fix has been tested with?

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSourceConfig.java
ardian4: updated

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ardian4: In the line 279 if the column map is empty it will retrieve all the columns.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ardian4: updated

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
ardian4: If user types a query like: 
```
SELECT A, B, CASE A WHEN 90 THEN 'red' WHEN 50 THEN 'blue' ELSE 'green' END AS result FROM Numbers
```
in these cases we fallback to getting full schema of the table

ardian4:
> Could you please provide list of select statements this fix has been tested with?

Output schema null and  for import query:

1. empty import query
2. Select * from testtable
3. Select FirstName from testtable
4. Select FirstName as name from testtable
5. Select FirstName, Identifier from testtable
6. Select FirstName, Identifier as Id from testtable
7. Select FirstName, Identifier as from testtable
8. Select FirstName, as from testtable

CuriousVini:
> > Could you please provide list of select statements this fix has been tested with?
> 
> Output schema null and for import query:
> 
> 1. empty import query
> 2. Select * from testtable
> 3. Select FirstName from testtable
> 4. Select FirstName as name from testtable
> 5. Select FirstName, Identifier from testtable
> 6. Select FirstName, Identifier as Id from testtable
> 7. Select FirstName, Identifier as from testtable
> 8. Select FirstName, as from testtable

What is the behavior in case of following queries (in case of invalid queries)?
```
Select FirstName, Identifier as from testtable
Select FirstName, as from testtable

**Comment by CuriousVini** [comment]
also please test with queries like:
```
SELECT A, B, CASE A WHEN 90 THEN 'red' WHEN 50 THEN 'blue' ELSE 'green' END AS result FROM Numbers

CuriousVini:
This should also have an integration test

ardian4:
> Select FirstName, as from testtable

Screen Shot 2020-12-30 at 3 02 35 PM

> Select FirstName, Identifier as from testtable
Screen Shot 2020-12-30 at 3 01 53 PM

ardian4:
> also please test with queries like:
> 
> ```
> SELECT A, B, CASE A WHEN 90 THEN 'red' WHEN 50 THEN 'blue' ELSE 'green' END AS result FROM Numbers
> ```

When import query contain  ```case```  it will get all the fields 

Screen Shot 2020-12-30 at 3 07 56 PM

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
rmstar: hm this approach is not ideal. You can construct queries where we unnecessarily fallback to getting full table schema.
For example, if the query is `SELECT COUNT(*) FROM table`, we'll get the full schema. 

Instead of parsing the query, can we execute the query with  `LIMIT 1` and get the schema from the query result?

ardian4: Executing the query with `LIMIT 1` may be a better approach but it will require additional effort and time to  replace this approach and test the new approach.  So our question is whether we should continue on this PR with the new approach?

However, the new approach may not be ideal when aggregation functions are used along GROUP BY as part of the importQuery. For instance cases like:

```
SELECT COUNT(CustomerID), Country
FROM Customers
GROUP BY Country
```

ardian4:
> This should also have an integration test

https://github.com/cdapio/cdap-integration-tests/pull/1084

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
rmstar: Let's implement this correctly so we won't need to rework the fix later.
Ideally, we would get the correct schema for all queries.

ardian4: @rmstar  With the suggested implementation we face the following issue:
Currently with Spanner cloud library we can get column name and colum type. We are missing the information whether column is nullable or not.
Spanner has this information in 'ResultSetMetadata' class -however this class is onlyavailable in 'com.google.spanner' package (https://googleapis.dev/java/google-cloud-spanner/3.2.1/com/google/spanner/v1/ResultSetMetadata.html) and not in 'com.google.cloud.spanner' package ('https://googleapis.dev/java/google-cloud-spanner/latest/com/google/cloud/spanner/ResultSet.html')

rmstar: What's the effort involved to use `com.google.spanner` library to get column metadata, including whether it's nullable or not?

If that's not feasible, can we use `com.google.cloud.spanner` to get the column names, and then use the existing approach in this PR (query information_schema) to check if the column is nullable?

ardian4: We will not be able to use the alternative 1 `com.google.spanner` library.

Yes we can use the `com.google.cloud.spanner` to get the column names, and then use the existing approach (information_schema) to check if the column is nullable. However, there are cases where we can't get the nullable information like:

1. Querying using aliases
2. Aggregate Functions


For these cases should we default to nullable column?

rmstar: Yes, if you can't get nullable information, then default to nullable column.

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -272,6 +309,18 @@ private Schema getSchema(FailureCollector collector) {
 
   }
 
+  private Statement getStatementForOneRow(String importQuery) {
+    String query;
+    if (StringUtils.containsIgnoreCase(importQuery, LIMIT)) {
+      int total = StringUtils.lastIndexOf(importQuery, LIMIT);
+      String substringToReplace = importQuery.substring(total);
+      query = importQuery.replace(substringToReplace, "limit 1");
```
rmstar, line 321: This can break in some corner cases where the table name contains "limit" substring. For example something like `SELECT <columns> from limited`.
Can you make sure we only replace "limit < number >" with "limit 1".

ardian4: Done

src/main/java/io/cdap/plugin/gcp/spanner/source/SpannerSource.java
```diff
@@ -272,6 +309,20 @@ private Schema getSchema(FailureCollector collector) {
 
   }
 
+  private Statement getStatementForOneRow(String importQuery) {
+    String query;
+    String regex = "^(?:[^;']|(?:'[^']+'))+ LIMIT +\\d+(.*)";
```
rmstar, line 316: Please add a comment that explains what this regex matches.

rmstar (approved):
lgtm, please address minor comment and squash commits.

ardian4:
> squash

Done

PR #462, opened 2020-12-21
Plugin-498: Added content-type to GCS sink

flakrimjusufi:
Add content-type to GCS sink

JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-498

### DIFF
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -45,6 +45,23 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 
 **Location:** The location where the gcs bucket will get created. This value is ignored if the bucket already exists.
 
+**Content Type:** The Content Type entity is used to indicate the media type of the resource.
+Defaults to 'application/octet-stream'. The following table shows valid content types for each format.
+
+| Format type   | Content type                                                                               |
+|---------------|--------------------------------------------------------------------------------------------|
+| avro          | application/avro, application/octet-stream                                                 |
+| csv           | text/csv, application/csv, text/plain, application/octet-stream                            |
+| delimited     | text/csv, application/csv, text/tab-separated-values, text/plain, application/octet-stream |
+| json          | application/json, text/plain, application/octet-stream                                     |
+| orc           | application/octet-stream                                                                   |
+| parquet       | application/octet-stream                                                                   |
+| tsv           | text/tab-separated-values, text/plain, application/octet-stream                            |
+
+**Custom Content Type:** The Custom Content Type is used when the value of Content-Type is set to other.
+User can provide specific Content-Type, different from the options in the dropdown.
+More information about the Content-Type can be found at https://cloud.google.com/storage/docs/metadata
+
 **Service Account**  - service account key used for authorization
 
 * **File Path**: Path on the local file system of the service account key used for
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -67,6 +67,7 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   private static final String RECORDS_UPDATED_METRIC = "records.updated";
   public static final String AVRO_NAMED_OUTPUT = "avro.mo.config.namedOutput";
   public static final String COMMON_NAMED_OUTPUT = "mapreduce.output.basename";
+  public static final String CONTENT_TYPE = "io.cdap.gcs.batch.sink.content.type";
 
   private final GCSBatchSinkConfig config;
   private String outputPath;
@@ -125,6 +126,7 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
   @Override
   protected Map<String, String> getFileSystemProperties(BatchSinkContext context) {
     Map<String, String> properties = GCPUtils.getFileSystemProperties(config, config.getPath(), new HashMap<>());
+    properties.put(GCSBatchSink.CONTENT_TYPE, config.getContentType());
     properties.putAll(config.getFileSystemProperties());
     String outputFileBaseName = config.getOutputFileNameBase();
     if (outputFileBaseName == null || outputFileBaseName.isEmpty()) {
@@ -242,6 +244,23 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
     private static final String NAME_LOCATION = "location";
     private static final String NAME_FS_PROPERTIES = "fileSystemProperties";
     private static final String NAME_FILE_NAME_BASE = "outputFileNameBase";
+    private static final String NAME_CONTENT_TYPE = "contentType";
+    private static final String NAME_CUSTOM_CONTENT_TYPE = "customContentType";
+    private static final String DEFAULT_CONTENT_TYPE = "application/octet-stream";
+    private static final String CONTENT_TYPE_OTHER = "other";
+    private static final String CONTENT_TYPE_APPLICATION_JSON = "application/json";
+    private static final String CONTENT_TYPE_APPLICATION_AVRO = "application/avro";
+    private static final String CONTENT_TYPE_APPLICATION_CSV = "application/csv";
+    private static final String CONTENT_TYPE_TEXT_PLAIN = "text/plain";
+    private static final String CONTENT_TYPE_TEXT_CSV = "text/csv";
+    private static final String CONTENT_TYPE_TEXT_TSV = "text/tab-separated-values";
+    private static final String FORMAT_AVRO = "avro";
+    private static final String FORMAT_CSV = "csv";
+    private static final String FORMAT_JSON = "json";
+    private static final String FORMAT_TSV = "tsv";
+    private static final String FORMAT_DELIMITED = "delimited";
+    private static final String FORMAT_ORC = "orc";
+    private static final String FORMAT_PARQUET = "parquet";
 
     private static final String SCHEME = "gs://";
     @Name(NAME_PATH)
@@ -279,6 +298,18 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
     @Description("The location where the gcs bucket will get created. " +
                    "This value is ignored if the bucket already exists")
     protected String location;
+
+    @Macro
+    @Description("The Content Type property is used to indicate the media type of the resource." +
+      "Defaults to 'application/octet-stream'.")
+    @Nullable
+    protected String contentType;
+
+    @Macro
+    @Description("The Custom Content Type is used when the value of Content-Type is set to other." +
+      "User can provide specific Content-Type, different from the options in the dropdown.")
+    @Nullable
+    protected String customContentType;
 
     @Name(NAME_FS_PROPERTIES)
     @Macro
@@ -326,10 +357,19 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
         }
       }
 
+      if (!containsMacro(NAME_CONTENT_TYPE) && !containsMacro(NAME_CUSTOM_CONTENT_TYPE)
+        && !Strings.isNullOrEmpty(contentType) && !contentType.equalsIgnoreCase(CONTENT_TYPE_OTHER)
+        && !containsMacro(NAME_FORMAT)) {
+        if (!contentType.equalsIgnoreCase(DEFAULT_CONTENT_TYPE)) {
+          validateContentType(collector);
+        }
+      }
+
       try {
         getSchema();
       } catch (IllegalArgumentException e) {
-        collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_SCHEMA).withStacktrace(e.getStackTrace());
+        collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_SCHEMA)
+          .withStacktrace(e.getStackTrace());
       }
 
       try {
@@ -337,6 +377,69 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
       } catch (IllegalArgumentException e) {
         collector.addFailure("File system properties must be a valid json.", null)
           .withConfigProperty(NAME_FS_PROPERTIES).withStacktrace(e.getStackTrace());
+      }
+    }
+
+    //This method validates the specified content type for the used format.
+    public void validateContentType(FailureCollector failureCollector) {
+      switch (format) {
+        case FORMAT_AVRO:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_AVRO)) {
+            failureCollector.addFailure(String.format("Valid content types for avro are %s, %s.",
+                                                      CONTENT_TYPE_APPLICATION_AVRO, DEFAULT_CONTENT_TYPE), null)
+              .withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_JSON:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_JSON)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for json are %s, %s, %s.", CONTENT_TYPE_APPLICATION_JSON,
+              CONTENT_TYPE_TEXT_PLAIN, DEFAULT_CONTENT_TYPE), null
+            ).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_CSV:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for csv are %s, %s, %s, %s.", CONTENT_TYPE_APPLICATION_CSV,
+              CONTENT_TYPE_TEXT_PLAIN, CONTENT_TYPE_TEXT_CSV, DEFAULT_CONTENT_TYPE), null
+            ).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_DELIMITED:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_TSV)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for delimited are %s, %s, %s, %s, %s.", CONTENT_TYPE_TEXT_PLAIN,
+              CONTENT_TYPE_TEXT_CSV, CONTENT_TYPE_APPLICATION_CSV, CONTENT_TYPE_TEXT_TSV, DEFAULT_CONTENT_TYPE), null
+            ).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_PARQUET:
+          if (!contentType.equalsIgnoreCase(DEFAULT_CONTENT_TYPE)) {
+            failureCollector.addFailure(String.format("Valid content type for parquet is %s.", DEFAULT_CONTENT_TYPE),
+                                        null).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_ORC:
+          if (!contentType.equalsIgnoreCase(DEFAULT_CONTENT_TYPE)) {
+            failureCollector.addFailure(String.format("Valid content type for orc is %s.", DEFAULT_CONTENT_TYPE),
+                                        null).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_TSV:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_TSV)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for tsv are %s, %s, %s.", CONTENT_TYPE_TEXT_TSV, CONTENT_TYPE_TEXT_PLAIN,
+              DEFAULT_CONTENT_TYPE), null).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
       }
     }
 
@@ -381,6 +484,30 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
     @Nullable
     public String getLocation() {
       return location;
+    }
+
+    /*  This method gets the value of content type. Valid content types for each format are:
+     *
+     *  avro -> application/avro, application/octet-stream
+     *  json -> application/json, text/plain, application/octet-stream
+     *  csv -> application/csv, text/csv, text/plain, application/octet-stream
+     *  delimited -> application/csv, text/csv, text/plain, text/tsv, application/octet-stream
+     *  orc -> application/octet-stream
+     *  parquet -> application/octet-stream
+     *  tsv -> text/tab-separated-values, application/octet-stream
+     */
+    @Nullable
+    public String getContentType() {
+      if (!Strings.isNullOrEmpty(contentType)) {
+        if (contentType.equals(CONTENT_TYPE_OTHER)) {
+          if (Strings.isNullOrEmpty(customContentType)) {
+            return DEFAULT_CONTENT_TYPE;
+          }
+          return customContentType;
+        }
+        return contentType;
+      }
+      return DEFAULT_CONTENT_TYPE;
     }
 
     public Map<String, String> getFileSystemProperties() {
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
@@ -95,7 +95,6 @@ public class GCSMultiBatchSink extends BatchSink<StructuredRecord, NullWritable,
     collector.getOrThrowException();
 
     Map<String, String> baseProperties = GCPUtils.getFileSystemProperties(config, config.getPath(), new HashMap<>());
-
     Map<String, String> argumentCopy = new HashMap<>(context.getArguments().asMap());
 
     String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
@@ -137,7 +136,7 @@ public class GCSMultiBatchSink extends BatchSink<StructuredRecord, NullWritable,
       outputProperties.putAll(RecordFilterOutputFormat.configure(validatingOutputFormat.getOutputFormatClassName(),
                                                                  config.splitField, name, schema));
       outputProperties.put(FileOutputFormat.OUTDIR, config.getOutputDir(context.getLogicalStartTime(), name));
-
+      outputProperties.put(GCSBatchSink.CONTENT_TYPE, config.getContentType());
       context.addOutput(Output.of(
         config.getReferenceName() + "_" + name,
         new SinkOutputFormatProvider(RecordFilterOutputFormat.class.getName(), outputProperties)));
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright © 2015-2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.gcs.sink;
+
+import com.google.cloud.storage.Blob;
+import com.google.common.annotations.VisibleForTesting;
+import io.cdap.plugin.gcp.common.GCPUtils;
+import io.cdap.plugin.gcp.gcs.StorageClient;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.JobStatus;
+import org.apache.hadoop.mapreduce.OutputCommitter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * OutputCommitter for GCS
+ */
+public class GCSOutputCommitter extends OutputCommitter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(GCSOutputFormatProvider.class);
+  public static final String RECORD_COUNT_FORMAT = "recordcount.%s";
+
+  private final OutputCommitter delegate;
+
+  public GCSOutputCommitter(OutputCommitter delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void setupJob(JobContext jobContext) throws IOException {
+    delegate.setupJob(jobContext);
+  }
+
+  @Override
+  public void cleanupJob(JobContext jobContext) throws IOException {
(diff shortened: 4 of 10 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -316,7 +330,15 @@ public void validate(FailureCollector collector) {
             .withConfigProperty(NAME_SUFFIX).withStacktrace(e.getStackTrace());
         }
       }
-
+      if (contentType == null) {
+        contentType = DEFAULT_CONTENT_TYPE;
```
chtyim: Shouldn't modify field in the validate method. Instead, introduce "getter" method do the defaulting logic in there. In the validate method, call the getter method for validation if needed.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -67,6 +67,7 @@
   private static final String RECORDS_UPDATED_METRIC = "records.updated";
   public static final String AVRO_NAMED_OUTPUT = "avro.mo.config.namedOutput";
   public static final String COMMON_NAMED_OUTPUT = "mapreduce.output.basename";
+  public static final String CONTENT_TYPE = "Content-Type";
```
chtyim: Prefix the FS configuration key. Also, use "." notation. E.g. "io.cdap.gcs.batch.sink.content.type".

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -125,6 +126,7 @@ public void prepareRun(BatchSinkContext context) throws Exception {
   @Override
   protected Map<String, String> getFileSystemProperties(BatchSinkContext context) {
     Map<String, String> properties = GCPUtils.getFileSystemProperties(config, config.getPath(), new HashMap<>());
+    properties.put(GCSBatchSink.CONTENT_TYPE, config.contentType);
```
chtyim: Use getter method instead of accessing the field directly.

CuriousVini:
@flakrimjusufi If multiple GCS sinks are being used in a single pipeline with different content type. What is the behavior? Any conflicts for content type?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -280,6 +284,16 @@ private long extractRecordCount(Map<String, String> metadata) {
                    "This value is ignored if the bucket already exists")
     protected String location;
 
+    @Macro
+    @Description("The Content Type property is used to indicate the media type of the resource.")
+    protected String contentType;
```
CuriousVini, line 306: having required types in plugins is a backwards incompatible change. Please make it nullable

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -280,6 +284,16 @@ private long extractRecordCount(Map<String, String> metadata) {
...
+
+    @Macro
+    @Nullable
+    @Description("The Custom Content Type is used when the value of Content-Type is set to other." +
+      "User can provide specific Content-Type, different from the options in the dropdown.")
+    protected String customContentType;
```
CuriousVini, line 312: same here

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -316,7 +330,15 @@ public void validate(FailureCollector collector) {
             .withConfigProperty(NAME_SUFFIX).withStacktrace(e.getStackTrace());
         }
       }
-
+      if (contentType == null) {
```
CuriousVini: check for null and empty

docs/GCS-batchsink.md
```diff
@@ -44,6 +44,11 @@ If the format is a macro, only the pre-packaged formats can be used.
 The delimiter will be ignored if the format is anything other than 'delimited'.
 
 **Location:** The location where the gcs bucket will get created. This value is ignored if the bucket already exists.
+                           
+**Content Type:** The Content-Type entity is used to indicate the media type of the resource.
```
CuriousVini: should mention what is the default value

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -316,7 +330,15 @@ public void validate(FailureCollector collector) {
...
+        contentType = DEFAULT_CONTENT_TYPE;
+      }
+      if (contentType.equals(CONTENT_TYPE_OTHER)) {
+        if (customContentType == null) {
+          contentType = DEFAULT_CONTENT_TYPE;
```
CuriousVini: there is no need to have assignment here as well. can do:
```
if (contentType.equals(CONTENT_TYPE_OTHER) && !Strings.isNullOrEmpty(customContentType)) {
 contentType = customContentType;
}

File: docs/GCS-batchsink.md [hunk 9]
```diff
@@ -44,6 +44,11 @@ If the format is a macro, only the pre-packaged formats can be used.
...
+
+**Custom Content Type:** The Custom Content Type is used when the value of Content-Type is set to other.
+User can provide specific Content-Type, different from the options in the dropdown.
```
CuriousVini, line 62: can add a link to list of other content types:
https://cloud.google.com/storage/docs/metadata

CuriousVini (changes requested):
Please make sure changes have been tested for cases:
1. Macros
2. Backwards compatibility
3. Also please add unit tests to these and modify integration tests to support this fix

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: Please check the updated PR.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: This was changed to `public static final String CONTENT_TYPE = "io.cdap.gcs.batch.sink.content.type";` in the updated PR.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: Please check the updated PR.

flakrimjusufi:
> Please make sure changes have been tested for cases:
> 
> 1. Macros
> 2. Backwards compatibility
> 3. Also please add unit tests to these and modify integration tests to support this fix

1. We tested with **Macro** and everything was okay. The pipeline was able to run successfully. 
2. We tested the pipelines with older versions of plugin and the pipelines were able to run successfully. We also did some further tests in **CDAP 6.1.4** and **CDAP 6.2.3** and everything worked fine. 
3. Please check the updated PR.

flakrimjusufi:
> @flakrimjusufi If multiple GCS sinks are being used in a single pipeline with different content type. What is the behavior? Any conflicts for content type?

We did some tests with **multiple GCS sink with different content types** and the behavior was as expected. There were no conflicts for content type. The pipelines were able to run successfully.

CuriousVini:
> > Please make sure changes have been tested for cases:
> > 
> > 1. Macros
> > 2. Backwards compatibility
> > 3. Also please add unit tests to these and modify integration tests to support this fix
> 
> 1. We tested with **Macro** and everything was okay. The pipeline was able to run successfully.
> 2. We tested the pipelines with older versions of plugin and the pipelines were able to run successfully. We also did some further tests in **CDAP 6.1.4** and **CDAP 6.2.3** and everything worked fine.
> 3. Please check the updated PR.

3. Could you please add a link to updated integration test PR here?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -383,6 +399,20 @@ public String getLocation() {
...
+      if (!Strings.isNullOrEmpty(contentType)) {
+        if (contentType.equals(CONTENT_TYPE_OTHER)) {
+          if (Strings.isNullOrEmpty(customContentType)) {
+            return DEFAULT_CONTENT_TYPE;
+          }
+          return customContentType;
+        }
+        return contentType;
+      }
+      return DEFAULT_CONTENT_TYPE;
```
CuriousVini, line 510: Do we use default content type if the field is a macro?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -383,6 +399,20 @@ public String getLocation() {
       return location;
     }
 
+    @Nullable
+    public String getContentType() {
+      if (!Strings.isNullOrEmpty(contentType)) {
+        if (contentType.equals(CONTENT_TYPE_OTHER)) {
+          if (Strings.isNullOrEmpty(customContentType)) {
```
CuriousVini, line 503: This means if `customContentType` is a macro, we use default content type?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -383,6 +399,20 @@ public String getLocation() {
       return location;
     }
 
+    @Nullable
+    public String getContentType() {
```
CuriousVini, line 500: what happens if someone uses content type as `text/plain`, and format as `avro`? Do we have any validations that would prevent that error configuration during deployment

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: If the field is macro, the value that was specified by the user is being used.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: Same in here as well, if the value of customContentType is a macro, the value that was specified by the user is being used.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: When someone uses content type as text/plain and format as avro, the pipeline runs successfully since we don't have any validation that would prevent any error during the deployment.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
```diff
@@ -184,7 +184,8 @@ private void updateMetricMetaData(TaskAttemptContext taskAttemptContext) throws
         LOG.info("Could not find a file in path %s to apply count metadata.", taskAttemptPath.toString());
         return;
       }
-      storageClient.setMetaData(blob, metaData);
+      blob.toBuilder().setContentType(configuration.get(GCSBatchSink.CONTENT_TYPE)).setMetadata(metaData).build()
+        .update();
```
CuriousVini: what is the difference between `storageClient.setMetadata()` and this call?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
CuriousVini, line 500: Please add a javadoc to it

CuriousVini: That is contradicting behavior right? If we are adding a new content type field, we should make sure such validation happens. Also we should have unit tests for those as well

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
CuriousVini: If this is a macro, during config time we will provide wrong value `DEFAULT_CONTENT_TYPE`. We should return null if there is a macro during config time.

bajram-adapt:
Integration tests: https://github.com/cdapio/cdap-integration-tests/pull/1083

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated with necessary validations and unit tests.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java
flakrimjusufi: So the difference between these methods was that `storageClient.setMetadata()` was not giving the content type but the default value (application/octet-stream).

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR updated.

flakrimjusufi:
> > > Please make sure changes have been tested for cases:
> > > 
> > > 1. Macros
> > > 2. Backwards compatibility
> > > 3. Also please add unit tests to these and modify integration tests to support this fix
> > 
> > 
> > 
> > 1. We tested with **Macro** and everything was okay. The pipeline was able to run successfully.
> > 2. We tested the pipelines with older versions of plugin and the pipelines were able to run successfully. We also did some further tests in **CDAP 6.1.4** and **CDAP 6.2.3** and everything worked fine.
> > 3. Please check the updated PR.
> 
> 1. Could you please add a link to updated integration test PR here?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +387,73 @@ public void validate(FailureCollector collector) {
...
+            failureCollector.addFailure(String.format(
+              "Valid content types for delimited are %s, %s, %s, %s", CONTENT_TYPE_TEXT_PLAIN,
+              CONTENT_TYPE_TEXT_CSV, CONTENT_TYPE_APPLICATION_CSV, CONTENT_TYPE_TEXT_TSV), null
+            ).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_PARQUET:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)) {
+            failureCollector.addFailure(String.format("Valid content type for parquet is %s",
+                                                      CONTENT_TYPE_TEXT_PLAIN),
```
rmstar: Is text/plain the right content type for parquet and orc formats?

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
```diff
@@ -132,4 +143,127 @@ public void close(TaskAttemptContext context) throws IOException, InterruptedExc
       delegate.close(context);
     }
   }
+
+  /**
+   * OutputCommitter for GCS
+   */
+  public static class GCSOutputCommitter extends OutputCommitter {
```
rmstar: Why is this needed for setting content type?

flakrimjusufi: We need to set up the content type when the job is finished. This is the same as GCS Sink plugin. However, if there's another way to do it, please let me know.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: I changed `text/plain `to `application/octet-stream` for parquet and orc formats. This is also the default content type.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -326,10 +357,26 @@ public void validate(FailureCollector collector) {
         }
       }
 
+      if (containsMacro(NAME_CONTENT_TYPE)) {
```
CuriousVini: no need for this check, by default if its a macro or not provided at configure time, it will be null

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -326,10 +357,26 @@ public void validate(FailureCollector collector) {
...
+        contentType = null;
+      }
+
+      if (containsMacro(NAME_CUSTOM_CONTENT_TYPE)) {
```
CuriousVini: same here

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +387,73 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    public void validateContentType(FailureCollector failureCollector) {
```
CuriousVini, line 384: please include valid types that are supported per format type. Also it would be good to add this information in plugin doc.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +387,73 @@ public void validate(FailureCollector collector) {
...
+              .withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_JSON:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_JSON)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for json are %s, %s", CONTENT_TYPE_APPLICATION_JSON,
+              CONTENT_TYPE_TEXT_PLAIN),
+                                        null
```
CuriousVini: nit: move remaining to above line

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +387,73 @@ public void validate(FailureCollector collector) {
...
+          break;
+        case FORMAT_CSV:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_CSV)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for csv are %s, %s, %s", CONTENT_TYPE_APPLICATION_CSV,
+              CONTENT_TYPE_TEXT_PLAIN,
+              CONTENT_TYPE_TEXT_CSV),
+                                        null
```
CuriousVini: nit: move remaining to above line

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +387,73 @@ public void validate(FailureCollector collector) {
...
+            failureCollector.addFailure(String.format("Valid content type for orc is %s", DEFAULT_CONTENT_TYPE),
+                                        null).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_TSV:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_TSV)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for tsv are %s, %s", CONTENT_TYPE_TEXT_TSV, CONTENT_TYPE_TEXT_PLAIN),
+                                        null
```
CuriousVini: nit: move remaining to above line

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
CuriousVini: Is `GCSOutputCommitter` exactly same as 
https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputFormatProvider.java#L108?

If so why not make it a class file and use the same class at both the places?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
```diff
@@ -137,7 +136,7 @@ public void prepareRun(BatchSinkContext context) throws IOException, Instantiati
       outputProperties.putAll(RecordFilterOutputFormat.configure(validatingOutputFormat.getOutputFormatClassName(),
                                                                  config.splitField, name, schema));
       outputProperties.put(FileOutputFormat.OUTDIR, config.getOutputDir(context.getLogicalStartTime(), name));
-
+      outputProperties.put(GCSBatchSink.CONTENT_TYPE, config.getContentType());
```
CuriousVini, line 139: does multi sink need same validation for format and content type?

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
flakrimjusufi: Yes. We need the same validation for format and content type in GCSMultiBatchSink as well. This is already in place.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
bajram-adapt: Format Types supported are in validation method already

bajram-adapt: Info is in plugin doc already

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
```diff
@@ -132,4 +136,8 @@ public void close(TaskAttemptContext context) throws IOException, InterruptedExc
       delegate.close(context);
     }
   }
+
+  /**
+   * OutputCommitter for GCS
```
CuriousVini: unintentional?

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
```diff
@@ -35,6 +37,7 @@
  * An OutputFormat that filters records before sending them to a delegate
  */
 public class RecordFilterOutputFormat extends OutputFormat<NullWritable, StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(GCSOutputFormatProvider.class);
```
CuriousVini: where is this logger being used?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
```diff
@@ -0,0 +1,149 @@
+package io.cdap.plugin.gcp.gcs.sink;
```
CuriousVini, line 17: missing license

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
CuriousVini: I was referring to javadoc for this public method.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +379,70 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    public void validateContentType(FailureCollector failureCollector) {
+      switch (format) {
```
CuriousVini, line 385: format can be null if its a macro. I believe null with switch case will not work.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +379,70 @@ public void validate(FailureCollector collector) {
...
+        case FORMAT_AVRO:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_APPLICATION_AVRO)) {
+            failureCollector.addFailure(String.format("Valid content type for avro is %s",
```
CuriousVini: Please add period to the end of all the validation error messages

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +379,70 @@ public void validate(FailureCollector collector) {
...
+                                                      CONTENT_TYPE_APPLICATION_AVRO), null)
```
CuriousVini: we also support `application/octet-stream` for all the types. Should we add that to the message as well?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -340,6 +379,70 @@ public void validate(FailureCollector collector) {
...
+          if (!contentType.equalsIgnoreCase(DEFAULT_CONTENT_TYPE)) {
+            failureCollector.addFailure(String.format("Valid content type for orc is %s", DEFAULT_CONTENT_TYPE),
+                                        null).withConfigProperty(NAME_CONTENT_TYPE);
+          }
+          break;
+        case FORMAT_TSV:
+          if (!contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_PLAIN)
+            && !contentType.equalsIgnoreCase(CONTENT_TYPE_TEXT_TSV)) {
+            failureCollector.addFailure(String.format(
+              "Valid content types for tsv are %s, %s", CONTENT_TYPE_TEXT_TSV, CONTENT_TYPE_TEXT_PLAIN), null
```
CuriousVini: is `text/plain` supported for this? If so, we should add that to plugin documentation.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java
CuriousVini: I see.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
```diff
@@ -0,0 +1,149 @@
...
+
+    //read the count from configuration
+    String keyInConfig = String.format(RECORD_COUNT_FORMAT, taskAttemptContext.getTaskAttemptID());
+    Map<String, String> metaData = new HashMap<>();
+    metaData.put(GCSBatchSink.RECORD_COUNT, String.valueOf(configuration.getLong(keyInConfig, 0L)));
+    StorageClient storageClient = getStorageClient(configuration);
+    //update metadata on the output file present in the directory for this task
+    Blob blob = storageClient.pickABlob(taskAttemptPath.toString());
+    if (blob == null) {
+      LOG.info("Could not find a file in path %s to apply count metadata.", taskAttemptPath.toString());
```
CuriousVini: `"Could not find a file in path {} to apply count metadata."`

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
flakrimjusufi: Unintentional.

src/main/java/io/cdap/plugin/gcp/gcs/sink/RecordFilterOutputFormat.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSOutputCommitter.java
flakrimjusufi: PR Updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: We were able to handle null exceptions when macro is null in **GCSBatchSink** but in **GCSMultiBatchSink** it seems like there's a scenario when format is required even after validation and when the format is macro (null) this always fails with null pointer exception. 

Please check the line 75 in this file: 
**src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSMultiBatchSink.java.**

```
  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
    config.validate(collector);

    FileFormat format = config.getFormat();
    // add schema as a macro since we don't know it until runtime
    PluginProperties formatProperties = PluginProperties.builder()
      .addAll(config.getProperties().getProperties())
      .add("schema", String.format("${%s}", SCHEMA_MACRO)).build();
```

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkTest.java
```diff
@@ -60,4 +60,25 @@ public void testInvalidFSProperties() throws NoSuchFieldException {
     config.validate(collector);
     Assert.assertEquals(1, collector.getValidationFailures().size());
   }
+
+  @Test
+  public void testContentType() throws Exception {
```
CuriousVini: Please add negative tests as well. 
What is the behavior when incompatible content type and format is provided.

Also please include unit tests for different content types.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -383,6 +486,21 @@ public String getLocation() {
       return location;
     }
 
+    @Nullable
+    @Description("Gets the value of content type")
```
CuriousVini: this should be a javadoc not annotation. The javadoc should include mapping of format -> content type

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
CuriousVini: That indicates a bug. Can you file a jira for it?

CuriousVini (commented):
a few comments. Please squash the commits after fixing. 

Also please file a jira for GCS Multi sink plugin

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkTest.java
flakrimjusufi: PR updated.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
flakrimjusufi: PR updated.

flakrimjusufi:
> a few comments. Please squash the commits after fixing.
> 
> Also please file a jira for GCS Multi sink plugin

The PR is updated according to the latest review. Commits are also squashed. 

Here you have the JIRA tikcket for GCS Multi Sink plugin:
https://cdap.atlassian.net/browse/PLUGIN-553

CuriousVini (approved):

PR #464, opened 2020-12-23
[CDAP-17248] Added File Encoding support for GCS Source Plugin

fernst:
This PR Depends on cdapio/hydrator-plugins#1255

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
@@ -30,8 +30,10 @@ import io.cdap.cdap.etl.api.batch.BatchSource;
 import io.cdap.cdap.etl.api.batch.BatchSourceContext;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.format.FileFormat;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -320,6 +337,11 @@ public boolean skipHeader() {
       return skipHeader == null ? false : skipHeader;
     }
 
+    @Nullable
+    public String getFileEncoding() {
```
yaojiefeng, line 341: Just curious here, from https://github.com/cdapio/hydrator-plugins/pull/1255/files, I saw this method is only used in the validation, how does the input format knows which encoding is or am I missing anything

yaojiefeng (commented):
One comment

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
fernst: @yaojiefeng The final piece is this PR in `core-plugins`. It's not part of the original PR as these are 2 modules with cross dependencies and I can't really build both of them on the same PR https://github.com/cdapio/hydrator-plugins/pull/1297

yaojiefeng (approved):

PR #466, opened 2020-12-28
Fixed issue with macros on the Key Type property for the Datastore plugin

fernst:
I noticed the Datastore plugin attempts to connect when macros are present on the Key Type and Key Alias fields. I fixed this behavior.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreSourceConfig.java
@@ -561,12 +561,14 @@ public class DatastoreSourceConfig extends GCPReferenceSourceConfig {
   boolean shouldConnect() {
     return !containsMacro(DatastoreSourceConstants.PROPERTY_SCHEMA) &&
       !containsMacro(NAME_SERVICE_ACCOUNT_TYPE) &&
-      !(containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_FILE_PATH) ||
-        containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_JSON)) &&
+      !containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_FILE_PATH) &&
+      !containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_JSON) &&
       !containsMacro(DatastoreSourceConfig.NAME_PROJECT) &&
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreSourceConfig.java
```diff
@@ -564,6 +564,8 @@ boolean shouldConnect() {
       !(containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_FILE_PATH) ||
         containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_JSON)) &&
       !containsMacro(DatastoreSourceConfig.NAME_PROJECT) &&
+      !(containsMacro(DatastoreSourceConstants.PROPERTY_KEY_TYPE) ||
```
albertshau: unless I'm misunderstanding the doc, shouldn't both of these required to be non-macro instead of just one?

fernst: @albertshau 

I think a customer might have a scenario where they set their **Key Type** to either `none` or `key literal` as needed. Given that the **Key Alias** is ignored when **Key Type** is set to none, this is a valid scenario.

The other case is where a customer always sets their **Key Type** as `key literal` but wants to use a Macro for the **Key Alias**.

albertshau: I got confused by the parantheses. This is logically equivalent to `!containsMacro(key) && !containsMacro(key_alias)`, which would be a simpler way to express it.

Same is true for the service account properties that are checked above.

fernst: I wanted to keep it "grouped" in the same way the NAME_SERVICE_ACCOUNT_FILE_PATH and NAME_SERVICE_ACCOUNT_JSON are. I can transform `!(a || b)` into `!a && !b` if you prefer.

albertshau: yes, can you change them both. We shouldn't have a mix of them

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreSourceConfig.java
fernst: Done.

PR #468, opened 2020-12-28
[PLUGIN-518] Datastore namespace doc fix

CuriousVini:
1. Fix datastore doc 
2. Fix bug where namespace with value `default` was throwing error. The fix is to use empty string when namespace value is `default`
3. Fix bug when datastore kind is a macro, null error is thrown

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreSinkConfig.java
@@ -29,6 +29,7 @@ import io.cdap.plugin.gcp.datastore.sink.util.DatastoreSinkConstants;
 import io.cdap.plugin.gcp.datastore.sink.util.IndexStrategy;
 import io.cdap.plugin.gcp.datastore.sink.util.SinkKeyType;
 import io.cdap.plugin.gcp.datastore.source.DatastoreSourceConfig;
+import io.cdap.plugin.gcp.datastore.source.util.DatastoreSourceConstants;
 import io.cdap.plugin.gcp.datastore.util.DatastorePropertyUtil;
 import io.cdap.plugin.gcp.datastore.util.DatastoreUtil;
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/datastore/sink/DatastoreSinkConfig.java
```diff
@@ -403,7 +404,9 @@ public boolean shouldConnect() {
     return !(containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_FILE_PATH) ||
       containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_JSON)) &&
       !containsMacro(NAME_SERVICE_ACCOUNT_TYPE) &&
-      !containsMacro(DatastoreSourceConfig.NAME_PROJECT) &&
+      !containsMacro(DatastoreSourceConfig.NAME_PROJECT) && !containsMacro(DatastoreSourceConstants.PROPERTY_KIND) &&
```
fernst: nit: move the second condition to a new line.

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreSourceConfig.java
```diff
@@ -563,8 +563,9 @@ boolean shouldConnect() {
       !containsMacro(NAME_SERVICE_ACCOUNT_TYPE) &&
       !(containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_FILE_PATH) ||
         containsMacro(DatastoreSourceConfig.NAME_SERVICE_ACCOUNT_JSON)) &&
-      !containsMacro(DatastoreSourceConfig.NAME_PROJECT) &&
-      tryGetProject() != null &&
+      !containsMacro(DatastoreSourceConfig.NAME_PROJECT) && !containsMacro(DatastoreSourceConstants.PROPERTY_KIND) &&
```
fernst: Same nit as before.

fernst (approved):
LGTM

PR #473, opened 2020-12-29
PLUGIN-516 do not enforce output schema for spanner sink

yaojiefeng:
Do not enforce a schema for spanner sink unless the table does not exist

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
@@ -99,7 +99,8 @@ public final class SpannerSink extends BatchSink<StructuredRecord, NullWritable,
     // throw a validation exception if any failures were added to the collector.
     collector.getOrThrowException();
 
-    Schema schema = config.getSchema(collector);
+    Schema configuredSchema = config.getSchema(collector);
+    Schema schema = configuredSchema == null ? context.getInputSchema() : configuredSchema;
     if (!context.isPreviewEnabled()) {
       try (Spanner spanner = SpannerUtil.getSpannerService(config.getServiceAccount(),
                                                            config.isServiceAccountFilePath(), config.getProject())) {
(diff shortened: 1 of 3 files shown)


tivv (commented):
Please add some unit tests

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
```diff
@@ -133,14 +134,19 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
-                                       Schema schema) throws ExecutionException, InterruptedException {
+                                       @Nullable Schema schema) throws ExecutionException, InterruptedException {
     boolean tableExists = isTablePresent(dbClient);
     if (!tableExists) {
       if (Strings.isNullOrEmpty(config.getKeys())) {
         throw new IllegalArgumentException(String.format("Spanner table %s does not exist. To create it from the " +
                                                            "pipeline, primary keys must be provided",
                                                          config.getTable()));
       }
+      if (schema == null) {
```
CuriousVini: this check can be moved to line 103

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
```diff
@@ -209,9 +215,8 @@ public void initialize(BatchRuntimeContext context) throws Exception {
     super.initialize(context);
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
-    Schema schema = config.getSchema(collector);
```
CuriousVini, line 212: what is the current behavior if user only selects a,b,c columns to be written from record of a,b,c,d? would this code change modify current behavior?

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
yaojiefeng: This check can't be at line 103 since it only needs to be checked if table not exist, if table already exist, we do not need to enforce the schema field

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
yaojiefeng: Good catch, fixed

CuriousVini (approved):
lgtm 👍🏼

PR #537, opened 2021-02-09
Support datetime datatype in BQ source, sink and multisink. 

greeshmaswaminathan:
Support datetime datatype in BQ source, sink and multisink
- BQ source will allow mapping from BQ datetime to CDAP datetime
- BQ source will also allow mapping string to CDAP datetime (Runtime failure if string is not ISO 8601 format), mapping from BQ datetime to CDAP string 
- BigQuery sink will allow mapping CDAP datetime to BQ datetime
- BQ sink will also allow mapping from cdap datetime to BQ string , cdap string to BQ datetime (Error on loading if not valid ISO 8601 string)
- BigQuery Sink does not support mapping of any AVRO type to datetime. (https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro#logical_types). To work around this issue, anytime when a datetime field is present in output schema, JSON format for file will be used to load data in to BQ.

Tested the following scenarios
- BQ source date time to cdap date time
- BQ source date time to cdap string
- BQ source string to cdap datetime 
- CDAP datetime to BQ sink datetime
- CDAP datetime to BQ sink string
- CDAP string to BQ sink datetime
- BQ sink with both datetime and bytes fields to test JSON support for bytes
- BQ Multisink with filter where one source schema has datetime  and one does not.(This will use AVRO for one source and JSON for the other)

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryAvroConverter.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryAvroConverter.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.gcp.bigquery.sink;
+
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.common.RecordConverter;
+import io.cdap.plugin.format.avro.StructuredToAvroTransformer;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.mapred.AvroKey;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * BigQueryAvroConverter converts a {@link StructuredRecord} to {@link AvroKey<GenericRecord>}
+ */
+public class BigQueryAvroConverter extends RecordConverter<StructuredRecord, AvroKey<GenericRecord>> {
+
+  private final StructuredToAvroTransformer structuredToAvroTransformer;
+
+  BigQueryAvroConverter() {
+    structuredToAvroTransformer = new StructuredToAvroTransformer(null);
+  }
+
+  @Override
+  public AvroKey<GenericRecord> transform(StructuredRecord record, @Nullable Schema schema) throws IOException {
+    return new AvroKey<>(structuredToAvroTransformer.transform(record, schema == null ? record.getSchema() : schema));
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryJsonConverter.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryJsonConverter.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.gcp.bigquery.sink;
+
+import com.google.gson.JsonObject;
+import com.google.gson.internal.bind.JsonTreeWriter;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.common.RecordConverter;
+
+import java.io.IOException;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * BigQueryJsonConverter converts a {@link StructuredRecord} to {@link JsonObject}
+ */
+public class BigQueryJsonConverter extends RecordConverter<StructuredRecord, JsonObject> {
+
+  @Override
+  public JsonObject transform(StructuredRecord input, @Nullable Schema schema) throws IOException {
+    try (JsonTreeWriter writer = new JsonTreeWriter()) {
+      writer.beginObject();
+      for (Schema.Field recordField : Objects.requireNonNull(input.getSchema().getFields())) {
+        // From all the fields in input record, write only those fields that are present in output schema
+        if (schema != null && schema.getField(recordField.getName()) == null) {
+          continue;
+        }
+        BigQueryRecordToJson.write(writer, recordField.getName(), input.get(recordField.getName()),
+                                   recordField.getSchema());
+      }
+      writer.endObject();
+      return writer.get().getAsJsonObject();
+    }
+  }
+}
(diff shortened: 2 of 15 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryAvroConverter.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private final StructuredToAvroTransformer structuredToAvroTransformer;
+
+  BigQueryAvroConverter() {
+    structuredToAvroTransformer = new StructuredToAvroTransformer(null);
+  }
+
+  @Override
+  public AvroKey<GenericRecord> transform(StructuredRecord record, Schema schema) throws IOException {
+    return new AvroKey<>(structuredToAvroTransformer.transform(record, schema));
```
fernst: I would label the `schema` attribute as Nullable since Avro supports a nullable schema now.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryJsonConverter.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.util.Objects;
+
+/**
+ * BigQueryJsonConverter converts a {@link StructuredRecord} to {@link JsonObject}
+ */
+public class BigQueryJsonConverter extends RecordConverter<StructuredRecord, JsonObject> {
+
+  @Override
+  public JsonObject transform(StructuredRecord input, Schema schema) throws IOException {
```
fernst: `schema` could also be nullable here, and we just write all fields in the Strucured record.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
```diff
@@ -160,7 +165,11 @@ private static void writeSimpleTypes(JsonWriter writer, String name, boolean isA
       case STRING:
         writer.value(object.toString());
         break;
-      // TODO CDAP-15342 write byte type to json writer
+      case BYTES:
+        if (object instanceof byte[]) {
```
albertshau, line 171: byte fields are also sometimes ByteBuffer, should add that instanceof check as well.

If it's neither, should throw an exception rather than silently drop the data.

albertshau (commented):
does this PR add support in the source? I see source in the title but everything here looks related to the sink. (Fine to add source in a separate PR, just trying to understand).

Also, are there any doc changes needed?

greeshmaswaminathan:
> does this PR add support in the source? I see source in the title but everything here looks related to the sink. (Fine to add source in a separate PR, just trying to understand).
> 
> Also, are there any doc changes needed?

@albertshau , this file has the change for source - src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java

For doc, we have not changed anything visible other than the additional data type which should show up for the schema(once UI changes are in).

greeshmaswaminathan:
> > does this PR add support in the source? I see source in the title but everything here looks related to the sink. (Fine to add source in a separate PR, just trying to understand).
> > Also, are there any doc changes needed?
> 
> @albertshau , this file has the change for source - src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
> 
> For doc, we have not changed anything visible other than the additional data type which should show up for the schema(once UI changes are in).

I see there is some doc for supported types etc, will update that.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryAvroConverter.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryJsonConverter.java
greeshmaswaminathan: Done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
```diff
@@ -168,6 +169,11 @@ private static void writeSimpleTypes(JsonWriter writer, String name, boolean isA
       case BYTES:
         if (object instanceof byte[]) {
           writer.value(Base64.getEncoder().encodeToString((byte[]) object));
+        } else if (object instanceof ByteBuffer) {
+          writer.value(Base64.getEncoder().encodeToString(((ByteBuffer) object).array()));
```
albertshau: array() doesn't always work properly, use io.cdap.cdap.api.common.Bytes.toBytes() instead

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
```diff
@@ -25,8 +25,12 @@
 import io.cdap.plugin.gcp.bigquery.sink.BigQueryRecordToJson;
 import org.junit.Assert;
 import org.junit.Test;
+import py4j.Base64;
```
albertshau: this is a weird one, did you mean to use java.util.Base64?

albertshau:
From a compatibility perspective, what happens if somebody is using the BQ source to read from a table with DateTime, but has the schema for that field set to a string? Need to verify that we allow this behavior and don't fail.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java
greeshmaswaminathan: Done

src/test/java/io/cdap/plugin/gcp/bigquery/BigQueryRecordToJsonTest.java
greeshmaswaminathan: Yes, fixed it.

greeshmaswaminathan:
> From a compatibility perspective, what happens if somebody is using the BQ source to read from a table with DateTime, but has the schema for that field set to a string? Need to verify that we allow this behavior and don't fail.

Supporting datetime <-> string between CDAP and BQ.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -86,8 +91,15 @@ protected Object convertField(Object field, Schema fieldSchema) throws IOExcepti
           case TIMESTAMP_MICROS:
             return field;
           case DATETIME:
-            //BQ datetime should already be in the expected ISO 8601 format for LocalDateTime
-            return LocalDateTime.parse(field.toString());
+            try {
+              return LocalDateTime.parse(field.toString());
+            } catch (DateTimeParseException exception) {
+              throw new UnexpectedFormatException(
+                String.format("Field of type '%s' with value '%s' is not in ISO-8601 format.",
```
albertshau: would be good to include the field name in the error message.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -86,8 +91,15 @@ protected Object convertField(Object field, Schema fieldSchema) throws IOExcepti
           case TIMESTAMP_MICROS:
             return field;
           case DATETIME:
-            //BQ datetime should already be in the expected ISO 8601 format for LocalDateTime
-            return LocalDateTime.parse(field.toString());
+            try {
+              return LocalDateTime.parse(field.toString());
```
albertshau: think it's a little better to parse for validation, but still return the string. That way we don't need the instanceof check above.

albertshau (approved):
minor comments. 

I don't see any change to validation logic, so I'm assuming this means there aren't currently any validation checks that the BQ type must match exactly the CDAP type.

greeshmaswaminathan:
> minor comments.
> 
> I don't see any change to validation logic, so I'm assuming this means there aren't currently any validation checks that the BQ type must match exactly the CDAP type.

Validation checks for datatype match between BQ and CDAP uses the type maps in src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java which will allow mapping between string and datetime.

PR #538, opened 2021-02-10
PLUGIN-191: Sanitized output name when special characters are present

flakrimjusufi:
BigQuery Multisink fails on table names with special characters
JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-191

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.io.NullWritable;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * This plugin allows users to write {@link StructuredRecord} entries to multiple Google Big Query tables.
@@ -50,7 +51,7 @@ import java.util.Map;
   + "Data is first written to a temporary location on Google Cloud Storage, then loaded into BigQuery from there.")
 public class BigQueryMultiSink extends AbstractBigQuerySink {
   private static final String TABLE_PREFIX = "multisink.";
-
+  private static final String OUTPUT_PATTERN = "[A-Za-z0-9_-]+";
   private final BigQueryMultiSinkConfig config;
 
   public BigQueryMultiSink(BigQueryMultiSinkConfig config) {
@@ -112,12 +113,25 @@ public class BigQueryMultiSink extends AbstractBigQuerySink {
         }
 
         String outputName = String.format("%s-%s", config.getReferenceName(), tableName);
+        outputName = sanitizeOutputName(outputName);
         initOutput(context, bigQuery, outputName, tableName, tableSchema, bucket, context.getFailureCollector());
       } catch (IOException e) {
         collector.addFailure("Invalid schema: " + e.getMessage(), null);
       }
     }
     collector.getOrThrowException();
+  }
+
+  /**
+   * This method sanitizes outputName when there is an un-allowed special character in dataset
+   * For example: If we have a dataset with outputName testtable$2020, this method will sanitize it to testtable_2020
+   */
+  private String sanitizeOutputName(String outputName) {
+    // Output name before regex: testtable$2020
+    final Pattern compilePattern = Pattern.compile(OUTPUT_PATTERN);
+    final boolean validatePattern = compilePattern.matcher(outputName).matches();
+    // Output name after regex: testtable_2020
+    return (!validatePattern) ? outputName.replaceAll("[^\\p{Alpha}\\p{Digit}-]+", "_") : outputName;
   }
 
   @Override


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -120,6 +122,14 @@ protected void prepareRunInternal(BatchSinkContext context, BigQuery bigQuery, S
     collector.getOrThrowException();
   }
 
+  private String sanitizeOutputName(String outputName) {
```
ailegion, line 129: missing javadoc

flakrimjusufi: PR updated.

ailegion: looks ok

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -50,7 +51,7 @@
   + "Data is first written to a temporary location on Google Cloud Storage, then loaded into BigQuery from there.")
 public class BigQueryMultiSink extends AbstractBigQuerySink {
   private static final String TABLE_PREFIX = "multisink.";
-
+  private static final String OUTPUT_PATTERN = "[A-Za-z0-9_-]+";
```
rmstar, line 54: From [this regex](https://github.com/cdapio/cdap/blob/2de0ef365a77ef549d071e52c89e73a0b0652b93/cdap-proto/src/main/java/io/cdap/cdap/proto/id/EntityId.java#L67), it seems like  '.' and '$' are allowed for dataset ids.
Can you please try to reproduce the issue and figure out the exact cause of the failure?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -120,6 +122,18 @@ protected void prepareRunInternal(BatchSinkContext context, BigQuery bigQuery, S
     collector.getOrThrowException();
   }
 
+  /**
+   * This method sanitizes outputName when there is an un-allowed special character in dataset
+   * For example: If we have a dataset with outputName testtable$2020, this method will sanitize it to testtable_2020
+   */
+  private String sanitizeOutputName(String outputName) {
+    //Output name before regex: testtable$2020
```
rmstar: nit: add space after //

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -120,6 +122,18 @@ protected void prepareRunInternal(BatchSinkContext context, BigQuery bigQuery, S
...
+    final Pattern compilePattern = Pattern.compile(OUTPUT_PATTERN);
+    final boolean validatePattern = compilePattern.matcher(outputName).matches();
+    //Output name after regex: testtable_2020
+    return (!validatePattern) ? outputName.replaceAll("[^\\p{Alpha}\\p{Digit}]+", "_") : outputName;
```
rmstar: '-' is an allowed character. Seems it will get replaced by '_' here?

rmstar (commented):
Can you please add a test for `sanitizeOutputName`?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
flakrimjusufi: First of all, are we talking about CDAP Datasets?

Because according to the description in JIRA:
`The BigQuery multi sink fails when tables have characters that aren't allowed by CDAP's dataset system (ex: $).`

We used this documentation as reference for CDAP Dataset Names: https://cdap.atlassian.net/wiki/spaces/DOCS/pages/477593783/Naming+Conventions

In here, the output name is generated with combination of table name, but the table name is validated before creating outputName, so if there is any issue will validation, the pipeline with fail before creating the outputName. 

Please check this line of code: https://github.com/data-integrations/google-cloud/blob/799447e7ef8b8ff83f72fa149f18e06e0381a804/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java#L115

The only difference between CDAP Dataset system and BigQuery naming convention is that BQ table name can contain $ (in case when table is partition by time, for example **client$2020**).

When BQ table exists and is partitioned by time (ex. partition by year), then BQMultisink can write data to this specific table **client$2020**

When we tested with scenario, the pipeline was able to run successfully, but we were unable to view the Lineage Records in CDAP because of an unexpected error that says: 

`Document.querySelector: '#target_ns-PLUGIN_191_ds-mlt-client$2020_fd-surname' is not a valid selector` 

I think this was the reason why special characters were asked to be sanitized. 

Please correct me if my assumption is wrong.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
flakrimjusufi: Added space after //.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
flakrimjusufi: You're right about '-'. I added in the regex in order to prevent replacement with '_'.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
rmstar: I see, thanks for the explanation!

rmstar (approved):
lgtm. please squash commits

flakrimjusufi:
> lgtm. please squash commits

Commits squashed.

PR #541, opened 2021-02-11
PLUGIN-572 - BigQuery Sink fails when the user specifies arbitrary GCS Upload Request Chunk Size

ailegion:
BigQuery Sink fails when the user specifies arbitrary GCS Upload Request Chunk Size
JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-572

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
@@ -588,15 +588,16 @@ public final class BigQueryUtil {
    * @param collector failure collector
    */
   public static void validateGCSChunkSize(String chunkSize, String chunkSizePropertyName, FailureCollector collector) {
-    String errorMessage1 = String.format("Value must be a multiple of %s", MediaHttpUploader.MINIMUM_CHUNK_SIZE);
-    String errorMessage2 = "Error parsing the provided value";
     if (!Strings.isNullOrEmpty(chunkSize)) {
       try {
         if (Integer.parseInt(chunkSize) % MediaHttpUploader.MINIMUM_CHUNK_SIZE != 0) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -579,6 +580,27 @@ public static void validateTable(String table, String tablePropertyName, Failure
...
+  public static void validateGCSChunkSize(String chunkSize, String chunkSizePropertyName, FailureCollector collector) {
+    String errorMessage1 = String.format("Value must be a multiple of %s", MediaHttpUploader.MINIMUM_CHUNK_SIZE);
+    String errorMessage2 = "Error parsing the provided value";
+    if (!Strings.isNullOrEmpty(chunkSize)) {
+      try {
+        if (Integer.parseInt(chunkSize) % MediaHttpUploader.MINIMUM_CHUNK_SIZE != 0) {
+          collector.addFailure(errorMessage1, null).withConfigProperty(chunkSizePropertyName);
+        }
+      } catch (NumberFormatException e) {
+        collector.addFailure(errorMessage2, null).withConfigProperty(chunkSizePropertyName);
```
ardian4: You can use e.getMessage() as error message and add corrective action.

ailegion: fixed

CuriousVini (commented):
add a test case where actual value is correct. Usually its best practice to add happy case and multiple failure cases covering code path.

ailegion:
> add a test case where actual value is correct. Usually its best practice to add happy case and multiple failure cases covering code path.

added more tests

CuriousVini (approved):
please squash the commits

PR #545, opened 2021-02-12
[PLUGIN-470] Cherry Pick into 0.15

fernst:


CuriousVini (approved):

pom.xml
```diff
@@ -73,7 +73,7 @@
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
     <cdap.version>6.2.0</cdap.version>
-    <cdap.plugin.version>2.4.2</cdap.plugin.version>
+    <cdap.plugin.version>2.4.4-SNAPSHOT</cdap.plugin.version>
```

fernst: As mentioned in the other PR, some of the changes in my other PR for hydrator plugins are needed here.

The snapshot dependency also signals the need to release the core plugins before this dependency can be tagged with a new patch release.

PR #546, opened 2021-02-12
[PLUGIN-470] Cherry Pick into 0.14

fernst:
…ommitter.

Added some improvements based on Terence's suggestions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
     <cdap.version>6.1.2</cdap.version>
-    <cdap.plugin.version>2.3.7</cdap.plugin.version>
+    <cdap.plugin.version>2.3.8-SNAPSHOT</cdap.plugin.version>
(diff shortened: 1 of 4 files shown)


CuriousVini (approved):

pom.xml
```diff
@@ -73,7 +73,7 @@
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
     <cdap.version>6.1.2</cdap.version>
-    <cdap.plugin.version>2.3.7</cdap.plugin.version>
+    <cdap.plugin.version>2.3.8-SNAPSHOT</cdap.plugin.version>
```

fernst: Versión 2.3.8 contains changes to the Avro formatter which are needed for the GCS MultiSink to work.

CuriousVini: Does that mean if a customer uses this jar and deploys on instance with 2.3.7 hydrator jar, this change would still work? If not, that would make this backwards incompatible change.

CuriousVini: Had an offline discussion. This is not because of an api change but just a behavior change. so not backwards incompatible change
https://github.com/cdapio/hydrator-plugins/pull/1323/files

PR #547, opened 2021-02-12
[PLUGIN-470] Cherry pick into 0.16

fernst:


pom.xml
```diff
@@ -73,7 +73,7 @@
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
     <cdap.version>6.3.0</cdap.version>
-    <cdap.plugin.version>2.5.0</cdap.plugin.version>
+    <cdap.plugin.version>2.5.1-SNAPSHOT</cdap.plugin.version>
```

CuriousVini (approved):
lgtm

pom.xml
fernst: Discussed offline that this is related to the changes introduced in https://github.com/cdapio/hydrator-plugins/pull/1324

PR #549, opened 2021-02-15
Support string to datetime mapping and vice versa in datastore, spanner

greeshmaswaminathan:
Support string to datetime mapping and vice versa in datastore, spanner
- Fixed a bug in BQ support to override the right overloaded convertField method in BigQueryAvroToStructuredTransformer . The calling method will add the field name in the final exception thrown
- Datastore and spanner source will allow string to be mapped to CDAP datetime and validate for correct format 
- Datastore and spanner sink will allow cdap datetime to be written as string
- Tested GCS source and sink. Datetime is supported with changes in https://github.com/cdapio/hydrator-plugins/pull/1327

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
@@ -61,12 +61,11 @@ public class BigQueryAvroToStructuredTransformer extends RecordConverter<Generic
 
   @Override
   @Nullable
-  protected Object convertField(Object field, Schema.Field schemaField) throws IOException {
+  protected Object convertField(Object field, Schema fieldSchema) throws IOException {
     if (field == null) {
       return null;
     }
 
-    Schema fieldSchema = schemaField.getSchema();
     // Union schema expected to be nullable schema. Underlying non-nullable type should always be a supported type
     fieldSchema = fieldSchema.isNullable() ? fieldSchema.getNonNullable() : fieldSchema;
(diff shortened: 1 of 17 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
```diff
@@ -92,8 +91,7 @@ protected Object convertField(Object field, Schema.Field schemaField) throws IOE
               LocalDateTime.parse(field.toString());
             } catch (DateTimeParseException exception) {
               throw new UnexpectedFormatException(
-                String.format("Field '%s' of type '%s' with value '%s' is not in ISO-8601 format.",
-                              schemaField.getName(),
+                String.format("Datetime field with value '%s' is not in ISO-8601 format.",
```
albertshau, line 94: This means the field name is no longer in the error message? Or does this exception get caught somewhere else and rethrown with the field name added?

albertshau (approved):
lgtm if the field name ends up in the error message, otherwise please fix the error message.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryAvroToStructuredTransformer.java
greeshmaswaminathan: Gets caught in https://github.com/cdapio/hydrator-plugins/blob/develop/hydrator-common/src/main/java/io/cdap/plugin/common/RecordConverter.java#L99 and field name gets added there.

PR #553, opened 2021-02-15
PLUGIN 517- fix issue with validation of partition field property

ardian4:
Fixed: Validate partition field property when partition type is integer:
    
    1. Throw validation issue in configuration stage when partition field is empty and partition type is Integer and destination table does not exist
    2. Throw validation issue in configuration stage when partition field is empty and partition type is Integer and destination table exist
    3. Throw validation issue in prepare run stage when partition field or partition type is macro and partition field is empty

Jira Ticket: https://cdap.atlassian.net/browse/PLUGIN-517

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
@@ -389,7 +389,7 @@ public final class BigQuerySinkConfig extends AbstractBigQuerySinkConfig {
       validateColumnForPartition(partitionByField, schema, collector);
       return;
     }
-    if (!shouldCreatePartitionedTable()) {
+    if (shouldCreatePartitionedTable()) {
       validateColumnForPartition(partitionByField, schema, collector);
     }
   }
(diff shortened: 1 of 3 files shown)


flakrimjusufi (commented):
Looks good to me.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -404,7 +404,16 @@ private void validateRangePartitionTableWithInputConfiguration(Table table, Rang
 
   private void validateColumnForPartition(@Nullable String columnName, @Nullable Schema schema,
                                           FailureCollector collector) {
-    if (columnName == null || schema == null) {
+    if (containsMacro(NAME_PARTITION_BY_FIELD) || containsMacro(NAME_PARTITIONING_TYPE) || schema == null) {
+      return;
+    }
+    PartitionType partitioningType = getPartitioningType();
+    if (Strings.isNullOrEmpty(columnName)) {
+      if (partitioningType == PartitionType.INTEGER) {
```
rmstar, line 441: Is this check not require if partitioningType == PartitionType.TIME?

rmstar: Also, can we add a unit test for this in BigQuerySinkTest?

ardian4: BigQuery has 3 types of partitions: 

1. **Ingestion time:** Tables are partitioned based on the data's ingestion (load) time or arrival time.
2. **Date/timestamp/datetime:** Tables are partitioned based on a TIMESTAMP, DATE, or DATETIME column.
3. **Integer range:** Tables are partitioned based on an integer column.

https://cloud.google.com/bigquery/docs/partitioned-tables

In Sink  for Time Partition if there is no field specified, the table will partition by ingestion time, so partition columnName can be empty.


We have added a unit test where partition type is TIME and the partition field is provided

rmstar: I see, thanks for explaining!

rmstar (approved):
lgtm, please squash commits

PR #555, opened 2021-02-17
PLUGIN-557: BigQuerySink fix for upsert operation in existing table but in different project 

flakrimjusufi:
BigQuery sink pipeline fails doing upsert operation to existing table in different project. The table is partitioned by ingestion-time (day). Partition field is left empty.

JIRA: https://cdap.atlassian.net/browse/PLUGIN-557

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
@@ -231,7 +231,7 @@ public final class BigQuerySink extends AbstractBigQuerySink {
    */
   private void configureTable(Schema schema) {
     AbstractBigQuerySinkConfig config = getConfig();
-    Table table = BigQueryUtil.getBigQueryTable(config.getProject(), config.getDataset(),
+    Table table = BigQueryUtil.getBigQueryTable(config.getDatasetProject(), config.getDataset(),
                                                 config.getTable(),
                                                 config.getServiceAccount(),
                                                 config.isServiceAccountFilePath());
@@ -253,7 +253,7 @@ public final class BigQuerySink extends AbstractBigQuerySink {
     }
 
     String tableName = config.getTable();
-    Table table = BigQueryUtil.getBigQueryTable(config.getProject(), config.getDataset(), tableName,
+    Table table = BigQueryUtil.getBigQueryTable(config.getDatasetProject(), config.getDataset(), tableName,
                                                 config.getServiceAccount(), config.isServiceAccountFilePath(),
                                                 collector);
 


ardian4:
lgtm

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -231,7 +231,7 @@ private void configureBigQuerySink() {
    */
   private void configureTable(Schema schema) {
     AbstractBigQuerySinkConfig config = getConfig();
-    Table table = BigQueryUtil.getBigQueryTable(config.getProject(), config.getDataset(),
+    Table table = BigQueryUtil.getBigQueryTable(config.getDatasetProject(), config.getDataset(),
```
CuriousVini, line 234: what if `config.getProject()` is provided? In that case we will rely on `config.getDatasetProject()`?

CuriousVini (commented):
Is there any unit test or integration test that can be added for this bug?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
flakrimjusufi: If `config.getProject()` and `config.getDatasetProject()` are provided, then `datasetProject` will get priority. 

If `datasetProject` is not defined, then `config.getProject()` will be set up in configuration. 
That's how it is defined in **AbstractBigQuerySinkConfig**
https://github.com/data-integrations/google-cloud/blob/86f4b91d9bf2916f48ba9fd12aacacecc8e805e5/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java#L114-L120

flakrimjusufi:
> Is there any unit test or integration test that can be added for this bug?

We created this JIRA Ticket https://cdap.atlassian.net/browse/PLUGIN-606 in order to add integration test in the future because currently it would be hard to add without having another project for integration test.

CuriousVini (approved):

CuriousVini:
Has this fix been tested with dataset id being a macro?

flakrimjusufi:
> Is there any unit test or integration test that can be added for this bug?



> Has this fix been tested with dataset id being a macro?

Yes, this fix has been tested with dataset id being a macro. 
There's no exception or anything like that. 
If datasetId is a macro, the validation will not proceed. 

This is the block of code that prevents any validation exception if any of the fields contains macro: 

https://github.com/data-integrations/google-cloud/blob/86f4b91d9bf2916f48ba9fd12aacacecc8e805e5/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java#L251-L253

And this is the code of shouldConnect(): 

https://github.com/data-integrations/google-cloud/blob/86f4b91d9bf2916f48ba9fd12aacacecc8e805e5/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java#L611-L619

CuriousVini (approved):

PR #573, opened 2021-02-24
[PLUGIN-470] Implemented DelegatingOutputFormat for schemaless BigQuery multisink

fernst:
I have added the option to use the BigQuery Multisink without having to set schemas as pipeline arguments. This ties to the Database MultiSource when used with custom SQL statements.

Users now have a configuration option in the BigQuery MultiSink which allows the  BigQuery Multisink to run without having the record(s) schema set in the pipeline. However, this means that we will not try to validate the BigQuery schema before writing to output.

### DIFF
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -67,6 +67,9 @@ In this case, those fields in the BigQuery schema will be modified to become nul
                          
 Incompatible schema changes will result in pipeline failure.
 
+**Allow flexible schemas in Output**: When enabled, this sink will write out records with arbitrary schemas. 
+Records may not have a well defined schema depending on the source.
+
 **Service Account**  - service account key used for authorization
 
 * **File Path**: Path on the local file system of the service account key used for
--- a/docs/GCSMultiFiles-batchsink.md
+++ b/docs/GCSMultiFiles-batchsink.md
@@ -67,6 +67,10 @@ When running on other clusters, the file must be present on every node in the cl
 The 'avro' supports 'snappy' and 'deflate'. The parquet supports 'snappy' and 'gzip'. 
 Other formats does not support compression.
 
+**Allow flexible schemas in Output**: When enabled, this sink will write out records with arbitrary schemas. 
+Records may not have a well defined schema depending on the source.
+When enabled, the format must be one of 'avro', 'json', 'csv', 'tsv', 'delimited'.
+
 **Split Field:** The name of the field that will be used to determine which directory to write to. 
 Defaults to 'tablename'.
 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sink;
+
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
+import com.google.cloud.hadoop.io.bigquery.output.BigQueryOutputConfiguration;
+import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
+import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
+import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Utility class for the BigQuery DelegatingMultiSink.
+ *
+ * The logic in this class has been extracted from the {@link AbstractBigQuerySink} in order to make this functionality
+ * available to other classes in this package.
+ */
+final class BigQuerySinkUtils {
+
+  private static final String gcsPathFormat = "gs://%s/%s";
+  private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s";
+  private static final String DATETIME = "DATETIME";
+
+  public static void configureBigQueryOutput(Configuration configuration,
+                                             String projectName,
+                                             String datasetName,
+                                             String tableName,
+                                             String temporaryGCSPath,
+                                             List<BigQueryTableFieldSchema> fields) throws IOException {
+    BigQueryTableSchema outputTableSchema = new BigQueryTableSchema();
+    if (!fields.isEmpty()) {
+      outputTableSchema.setFields(fields);
+    }
+
+    BigQueryFileFormat fileFormat = getFileFormat(fields);
+    BigQueryOutputConfiguration.configure(
+      configuration,
+      String.format("%s:%s.%s", projectName, datasetName, tableName),
+      outputTableSchema,
+      temporaryGCSPath,
+      fileFormat,
+      getOutputFormat(fileFormat));
+  }
+
+  public static void configureBigQueryOutputForMultiSink(JobContext context,
+                                                         String projectName,
+                                                         String datasetName,
+                                                         String bucketName,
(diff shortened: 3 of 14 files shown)


fernst:
I will update the docs for this and the GCS MultiSink in a later PR.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -0,0 +1,197 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Utility class for the BigQuery DelegatingMultiSink.
+ */
+public class BigQuerySinkUtils {
```
albertshau: I'm assuming this is just code that's been moved. If there's been change in logic, please point it out.

docs/BigQueryMultiTable-batchsink.md
```diff
@@ -67,6 +67,10 @@ In this case, those fields in the BigQuery schema will be modified to become nul
                          
 Incompatible schema changes will result in pipeline failure.
 
+**Allow flexible schemas in Output**: When enabled, this sink will write out records with arbitrary schemas. 
+Records may not have a well defined schema depending on the source.
+When enabled, the BigQuery table schema will not be validated or updated.
```
albertshau: I see that update schema is hidden when this is set, but I don't see any changes in the logic around this. For example, if somebody deploys a pipeline through REST when both enabled. Is it handled somewhere?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
```diff
@@ -0,0 +1,176 @@
...
+  private String datasetName;
+  private String bucketName;
+  private String bucketPathUniqueId;
+
+  public DelegatingMultiSinkOutputCommitter() {
+    committerMap = new HashMap<>();
+    schemaMap = new HashMap<>();
+  }
+
+  public void setProjectName(String projectName) {
```
albertshau: should avoid setters whenever possible, as it kind of exposes the internals and makes it easier to unintentionally couple logic from classes and create weird bugs. It's better to create the committer lazily in the OutputFormat, unless there's some reason that doesn't work?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
supriyapatil-google: keep this class final at the package level.

supriyapatil-google: Can we make sure we keep the history here that this was copied from AbstractBigQuerySink class?

docs/GCSMultiFiles-batchsink.md
```diff
@@ -67,6 +67,10 @@ When running on other clusters, the file must be present on every node in the cl
 The 'avro' supports 'snappy' and 'deflate'. The parquet supports 'snappy' and 'gzip'. 
 Other formats does not support compression.
 
+**Allow flexible schemas in Output**: When enabled, this sink will write out records with arbitrary schemas.
```
supriyapatil-google, line 70: Why is this file being updated?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Output Committer which creates and delegates operations to other Bigquery Output Committer instances.
+ *
+ * Delegated instances are supplied along with a schema, which is used to configure the commit operation.
+ */
+public class DelegatingMultiSinkOutputCommitter extends OutputCommitter {
```
supriyapatil-google, line 34: why are we changing the way the current sink plugin works? I would rather keep the same processing logic for both of them. Why do you need a separate impl here? You will still have an input stream with an Avro record format.

docs/BigQueryMultiTable-batchsink.md
supriyapatil-google: +1

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: Yes, this is code that has been moved. I will add additional comments to highlights this.

docs/BigQueryMultiTable-batchsink.md
fernst: By default, the `allowFlexibleSchemas` option takes precedence. When `allowFlexibleSchemas` is enabled, the schema relaxation (updating the BQ schema) doesn't ever take place.

Should I return an error message if both options are enabled at the same time?

docs/GCSMultiFiles-batchsink.md
fernst: The GCS MultiSink PR was merged earlier but at the time the docs were not updated.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
fernst: By default, the "current" way the plugin works remains the same. The schema gets validated and updated (if the user specifies the Update Schema option) and the records are stored in GCS and then copied to BQ.

The "new" solution will only be used when the users specify the "Allow Flexible Schemas" option. This is used when the record schema is now known at pipeline execution time (Right now, this is only possible when using the Multi Table Source with custom SQL statements: https://github.com/data-integrations/multi-table-plugins/blob/develop/docs/MultiTableDatabase-batchsource.md#custom-sql-statements

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
fernst: Good point, I don't know why I didn't think of this. Let me make sure that it works and I'll update this logic.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: I will add this as a comment.

fernst: I have added a comment highlighting the original source of this code.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
fernst: Updated. I tested it and it works as expected.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: Updated.

docs/BigQueryMultiTable-batchsink.md
albertshau: I see that it doesn't get validated, but does BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION still get set in the OutputFormat and used by the delegate? 

It would be good to try this scenario out (existing table with a different schema) to see what happens.

fernst: I have honestly not tried that. Let me run some tests and see how it goes.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkRecordWriter.java
```diff
@@ -0,0 +1,114 @@
...
+      delegate = delegateMap.get(tableName);
+    } else {
+      delegate = getRecordWriterDelegate(tableName, key.getSchema());
+    }
+
+    delegate.write(key, value);
+  }
+
+  @Override
+  public void close(TaskAttemptContext context) throws IOException, InterruptedException {
```
fernst, line 79: @albertshau I realized that, for some reason, the Context instance that is supplied to this method does not include all the configuration options that I expect.

If I commit the job with the original instance that was used to create the committer and record writer, the OutputCommitter has everything it needs to commit and update schema.

I wonder if you have any cues as to why the instance of the TaskAttemptContext supplied here would be missing some fields.

fernst:
@albertshau @supriyapatil-google I was able to validate that the sink works just fine with the **Update Table Schema** option enabled or disabled, so I have added this option back to the UI..

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkRecordWriter.java
albertshau: hm that's odd, I'm not sure why that would be the case. Worth adding a comment though, in case somebody wonders why it's being done the way it is.

docs/BigQueryMultiTable-batchsink.md
albertshau: should this line be removed now?

albertshau (approved):
minor comments about doc and comment

docs/BigQueryMultiTable-batchsink.md
fernst: Yes, I will update the docs.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkRecordWriter.java
fernst: I'll add a comment.

docs/BigQueryMultiTable-batchsink.md
fernst: Updated.

PR #574, opened 2021-02-24
Support datetime for Bigtable source and sink

greeshmaswaminathan:
- Datetime support for Bigtable source and sink
- Fixed other plugin tests to test with formatted datetime string to avoid flakiness 
- Tested Bigtable source , sink with valid and invalid datetime values


fernst (approved):
LGTM

PR #587, opened 2021-03-03
Fixed NPE when Path in GCS Source is a Macro

fernst:
Bumped to next snapshot version.

### DIFF
--- a/widgets/GCSFile-batchsource.json
+++ b/widgets/GCSFile-batchsource.json
@@ -40,11 +40,21 @@
           }
         },
         {
-          "widget-type": "plugin-list",
+          "widget-type": "select",
           "label": "Format",
           "name": "format",
           "widget-attributes": {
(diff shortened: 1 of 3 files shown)


widgets/GCSFile-batchsource.json
```diff
@@ -40,11 +40,21 @@
...
+              "avro",
+              "blob",
+              "csv",
+              "delimited",
+              "json",
+              "parquet",
+              "text",
+              "tsv"
+            ],
+            "default": "text"
```
greeshmaswaminathan, line 57: Nit: Do we need a default value? In 6.3 , it forces user to choose a value.

greeshmaswaminathan (approved):
Minor comment, otherwise LG.

widgets/GCSFile-batchsource.json
fernst: This is for 6.1

PR #590, opened 2021-03-09
CDAP-17746: Field validation logic in BigQuery sink pipelines causes a NullPointerException

jasir99:
API returns null for BigQuery Field Mode. Which by defaults means it's NULLABLE.

Reference: https://github.com/googleapis/google-cloud-java/issues/3509
Jira: https://cdap.atlassian.net/browse/CDAP-17746


rmstar (commented):
Can you please add a test?

rmstar:
Also, it would be good to have integration test coverage for this - https://github.com/cdapio/cdap-integration-tests/blob/develop/integration-test-remote/src/test/java/io/cdap/cdap/app/etl/gcp/GoogleBigQueryTest.java

jasir99:
> Can you please add a test?

This requires Schema from BigQuery - which means making an API call to BigQuery, and I didn't find any similar approach in the current unit tests. Do you have any suggestions?

rmstar:
Ok, I see that you have added an integration test. That's good for now.
For unit test you could consider using a mock.

rmstar (approved):

PR #610, opened 2021-04-08
[PLUGIN-376] Fixed: GCS copy & move plugins - Bucket and Object permission check discrepancy

flakrimjusufi:
GCS Plugins Bucket and Object Permissions Check Discrepancy
JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-376
Test cases: https://docs.google.com/spreadsheets/d/10CSYkzomHJ3wNiqlqnK7U2y4YxV7PXo3sNhhjx858qI/edit?usp=sharing

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -115,9 +115,66 @@ public class StorageClient {
    * @param recursive whether to copy objects in all subdirectories
    * @param overwrite whether to overwrite existing objects
    * @throws IllegalArgumentException if overwrite is false and copying would overwrite an existing object
+   * @return copied blob object from google bucket storage or null if we are dealing with a directory/bucket
    */
-  public void copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+  public Blob copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+    // If sourcePath is not a bucket and sourcePath doesn't end with "/"
+    // For example: gs://plugin_376/csvexample.csv
+    // It means we are dealing with a single object which can be copied to the destination
+    if (!sourcePath.isBucket() && !sourcePath.getName().endsWith("/")) {
+      return copyObjectToDestination(sourcePath, destPath, overwrite);
+    }
     pairTraverse(sourcePath, destPath, recursive, overwrite, BlobPair::copy);
+    return null;
+  }
+
+  /**
+   * Copy a single object from the source path to the destination path.
+   *
+   * @param sourcePath the path to copy object from
+   * @param destPath   the path to copy object to
+   * @param overwrite  whether to overwrite existing object
+   * @throws RuntimeException if overwrite is false and copying would overwrite an existing object
+   * @return copied blob object from google bucket storage
+   */
+  private Blob copyObjectToDestination(GCSPath sourcePath, GCSPath destPath, boolean overwrite) {
+    Blob sourceBlob;
+    try {
+      sourceBlob = storage.get(sourcePath.getBucket(), sourcePath.getName());
+      if (sourceBlob == null) {
+        throw new RuntimeException(
+          String.format("Object %s. not found", sourcePath.getUri())
+            + "Ensure you entered the correct path.");
+      }
+    } catch (StorageException e) {
+      throw new RuntimeException(
+        String.format("Unable to access source bucket %s. ", sourcePath.getBucket())
+          + "Ensure you entered the correct bucket path.", e);
+    }
+
+    boolean destinationObjectExists = false;
+    String destinationPath = String.format("%s%s", destPath.getName(), sourcePath.getName());
+    if (storage.get(BlobId.of(destPath.getBucket(), destinationPath)) != null) {
+      destinationObjectExists = true;
+    }
+
+    final BlobId destination = BlobId.of(destPath.getBucket(), append(destPath.getName(), sourcePath.getName()));
+    if (!overwrite && destinationObjectExists) {
+      throw new IllegalArgumentException(String.format("%s already exists.", toPath(destination)));
+    }
+
+    LOG.debug("Copying {} to {}.", toPath(sourceBlob.getBlobId()), toPath(destination));
+    Blob copied;
+    try {
+      CopyWriter copyWriter = sourceBlob.copyTo(destination);
+      copied = copyWriter.getResult();
+      LOG.debug("Successfully copied {} to {}.", toPath(sourceBlob.getBlobId()), toPath(destination));
+    } catch (StorageException e) {
+      throw new RuntimeException(
+        String.format("Unable to access source bucket %s. ", destPath.getBucket())
+          + "Ensure you entered the correct bucket path.", e);
+    }
+    return copied;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -115,9 +115,67 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
    * @param overwrite whether to overwrite existing objects
    * @throws IllegalArgumentException if overwrite is false and copying would overwrite an existing object
+   * @return copied blob object from google bucket storage or null if we are dealing with a directory/bucket
    */
-  public void copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+  public Blob copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+    //If sourcePath is not a bucket and sourcePath doesn't end with "/"
+    //For example: gs://plugin_376/csvexample.csv
+    //It means we are dealing with a single object which can be copied to the destination
+    if (!sourcePath.isBucket() && !sourcePath.getName().endsWith("/")) {
```
ailegion, line 124: Is there no other approach on how to check difference between file and directory?

flakrimjusufi: Well, apparently, for our case, this is the only way. 

If for example, we want to copy a _directory_ from source path to destination path and the source path is in this way --> **gs://cirus-test-bucket/output/csv**

We will have a **null** in return which means the directory cannot be found.
That's because google cloud storage can find directories only when we have a '/' character at the end of path, like in this way: **gs://cirus-test-bucket/output/csv/**

We've tried with Blob's _isDirectory()_ method but according to java doc: 
https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/BlobInfo.html#isDirectory--

**public boolean isDirectory()**
`Returns true if the current blob represents a directory. This can only happen if the blob is returned by Storage.list(String, Storage.BlobListOption...) when the Storage.BlobListOption.currentDirectory() option is used. When this is the case only getBlobId() and getSize() are set for the current blob: BlobId.getName() ends with the '/' character, BlobId.getGeneration() returns null and getSize() is 0.`

https://github.com/googleapis/java-storage/blob/4e5ee0398e700baf4f88224f66309e426f9532d7/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java#L931-L941

ailegion: I see. lgtm.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -115,9 +115,67 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    * @param recursive whether to copy objects in all subdirectories
    * @param overwrite whether to overwrite existing objects
    * @throws IllegalArgumentException if overwrite is false and copying would overwrite an existing object
+   * @return copied blob object from google bucket storage or null if we are dealing with a directory/bucket
    */
-  public void copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+  public Blob copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+    //If sourcePath is not a bucket and sourcePath doesn't end with "/"
```
rmstar: nit: add space after //

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -130,9 +188,68 @@ public void copy(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolea
    * @param recursive whether to move objects in all subdirectories
    * @param overwrite whether to overwrite existing objects
    * @throws IllegalArgumentException if overwrite is false and moving would overwrite an existing object
+   * @return moved blob object from google bucket storage or null if we are dealing with a directory/bucket
    */
-  public void move(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+  public Blob move(GCSPath sourcePath, GCSPath destPath, boolean recursive, boolean overwrite) {
+    //If sourcePath is not a bucket and sourcePath doesn't end with "/"
```
rmstar: nit: add space after //

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -115,9 +115,67 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
+    Blob sourceBlob;
+    try {
+      sourceBlob = storage.get(sourcePath.getBucket(), sourcePath.getName());
+      if (sourceBlob == null) {
+        throw new RuntimeException(
+          String.format("Object %s. not found", sourcePath.getUri())
+            + "Ensure you entered the correct path.");
+      }
+    } catch (StorageException e) {
+      throw new RuntimeException(
```
rmstar, line 150: Why are we wrapping StoroageException in RuntimeException? javadoc for this method indicates that it throws StorageException

rmstar (commented):
Can you please list the tests that you ran to verify this fix?

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
flakrimjusufi: It's something related with this ticket: https://cdap.atlassian.net/browse/PLUGIN-202

Something similar was also done in _pairTraverse_ method: https://github.com/data-integrations/google-cloud/blob/69c37be020c1d33d6ee8ecf8d856161945530a34/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java#L276-L283 ...and the reason for this was to catch some edge cases and to present more meaningful messages to the user.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
flakrimjusufi: > Can you please list the tests that you ran to verify this fix?

Here you have a spreadsheet with all the test cases we've done to verify this fix: 

https://docs.google.com/spreadsheets/d/10CSYkzomHJ3wNiqlqnK7U2y4YxV7PXo3sNhhjx858qI/edit?usp=sharing

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
rmstar: I see, then let's just fix the javadoc (@throws) as we are not throwing StorageException.

rmstar (approved):
Thanks for adding the test cases! Please make sure you have tested the following case mentioned in the Jira description: try to read/write object that is public but the bucket is not accessible.

lgtm, please address minor comment and squash commits.

flakrimjusufi:
> Thanks for adding the test cases! Please make sure you have tested the following case mentioned in the Jira description: try to read/write object that is public but the bucket is not accessible.
> 
> lgtm, please address minor comment and squash commits.

We've tested the mentioned case (read/write) object that is public but the bucket is not accessible and everything is working as expected. 

Commits squashed.

PR #612, opened 2021-04-08
PLUGIN-654:Pub/Sub source fails when subscription is a macro.

jasir99:
Pub/Sub source fails when subscription is a macro.

JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-654


flakrimjusufi (commented):
Looks good to me.

rmstar (approved):

PR #629, opened 2021-04-22

ardian4:
Cherrypick commit https://github.com/data-integrations/google-cloud/commit/78cc2bcd359e1251f95bdba2ba38c6497e204d7e which already fixed issue in release/0.17

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -34,6 +34,7 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.common.base.Strings;
+import com.google.gson.Gson;
 import io.cdap.cdap.api.data.batch.Output;
 import io.cdap.cdap.api.data.batch.OutputFormatProvider;
 import io.cdap.cdap.api.data.format.StructuredRecord;
@@ -76,6 +77,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
   private static final String gcsPathFormat = "gs://%s/%s";
   private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s";
   public static final String RECORDS_UPDATED_METRIC = "records.updated";
+  private static final Gson GSON = new Gson();
 
   // UUID for the run. Will be used as bucket name if bucket is not provided.
   // UUID is used since GCS bucket names must be globally unique.
@@ -135,14 +137,15 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, A
 
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
-    avroTransformer = new StructuredToAvroTransformer(null);
+    Schema schema = GSON.fromJson(context.getArguments().get(BigQueryConstants.CDAP_BQ_SINK_TABLE), Schema.class);
+    avroTransformer = new StructuredToAvroTransformer(schema);
   }
 
   /**
    * Transform the given {@link StructuredRecord} into avro {@link GenericRecord} with the same schema.
    */
   protected final GenericRecord toAvroRecord(StructuredRecord record) throws IOException {
-    return avroTransformer.transform(record, record.getSchema());
+    return avroTransformer.transform(record);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -106,6 +106,22 @@
           "LIKE", "NOT LIKE", "BETWEEN", "NOT BETWEEN", "IN", "NOT IN", "IS NULL", "IS NOT NULL",
           "IS TRUE", "IS NOT TRUE", "IS FALSE", "IS NOT FALSE");
 
+  @Override
+  public RecordWriter<StructuredRecord, NullWritable> getRecordWriter(TaskAttemptContext taskAttemptContext)
```
greeshmaswaminathan: The classes BigQueryRecordWriter and BigQueryAvroConverter were introduced as an abstraction layer to switch between JSON and AVRO output format at runtime. Since prior to 6.4 we only use AVRO , these are unnecessary here. Use StructuredToAvroTransformer with the getOutputSchema to write directly with the existing RecordWriter.

greeshmaswaminathan: This PR will also introduce the problem in this bug - https://cdap.atlassian.net/browse/PLUGIN-678. Please also cherry pick the fix for that as part of this PR.

ardian4: We followed your advice. Also the fix for PLUGIN-678 is included.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -187,12 +186,7 @@ public String getOutputFormatClassName() {
 
       @Override
       public Map<String, String> getOutputFormatConfiguration() {
-        Map<String, String> configToMap = BigQueryUtil.configToMap(configuration);
```
greeshmaswaminathan, line 190: This change seems like was introduced by a cherrypick -  https://github.com/data-integrations/google-cloud/pull/645/files. I don't think it is required, but please do test that PLUGIN-655 works.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -135,14 +137,15 @@ public void onRunFinish(boolean succeeded, BatchSinkContext context) {
 
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
-    avroTransformer = new StructuredToAvroTransformer(null);
+    Schema schema = GSON.fromJson(context.getArguments().get(BigQueryConstants.CDAP_BQ_SINK_TABLE), Schema.class);
```
greeshmaswaminathan, line 140: The value could be null. Check if a null check is required or GSON handles it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -170,6 +173,7 @@ protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, Str
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
     recordLineage(context, outputName, tableSchema, fieldNames);
+    context.getArguments().set(BigQueryConstants.CDAP_BQ_SINK_TABLE, GSON.toJson(tableSchema));
```
greeshmaswaminathan, line 176: Does this show up in the runtime arguments list for a pipeline run ?

jetroni: It doesn't show up in the runtime arguments list.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
jetroni: GSON automatically handles it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
jetroni: We tested it and it works.

greeshmaswaminathan (approved):

rmstar (approved):

PR #630, opened 2021-04-22

ardian4:
Cherrypick commit 78cc2bc which already fixed issue in release/0.17

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,7 @@
     <httpclient.version>4.5.6</httpclient.version>
     <jackson.core.version>2.8.11.1</jackson.core.version>
     <junit.version>4.12</junit.version>
+    <powermock.version>2.0.2</powermock.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -451,6 +452,18 @@
       <artifactId>junit</artifactId>
       <version>${junit.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
```
rmstar, line 456: why was this dependency added?

jetroni: In release/0.15 we would get an error while building stating that this dependency was missing. We think that it might be missing from this PR: https://github.com/data-integrations/google-cloud/pull/627

Should we create a new PR for this build fix?

rmstar: I see, thanks.

rmstar (approved):

PR #631, opened 2021-04-22

ardian4:
Cherrypick commit 78cc2bc which already fixed issue in release/0.17


rmstar (commented):
This looks a bit different compared to #629. Can you please clarify what the differences are?

ardian4:
> This looks a bit different compared to #629. Can you please clarify what the differences are?

In development branch commit for `DATETIME` also fixed this issue but `DATETIME` is not supported below CDAP 6.4 and we tried to patch only relevant part and mantain same flow. After that we had suggestion https://github.com/data-integrations/google-cloud/pull/629#discussion_r623441687 to simplify solution.

rmstar (approved):

PR #633, opened 2021-04-23
PLUGIN-678 - Do not override with table Schema in BQ Sink 

greeshmaswaminathan:
- Do not override configured schema in BQ sink with target BQ table schema
- BQ INT/FLOAT gets mapped to CDAP long/double which might conflict with the configured schema where it could be int/float
- For the fix for PLUGIN-395, re-ordering the configured schema to match the target table schema
- Field type validation is done separately, so not checking while re-ordering
- This might re-surface some validation errors (like configured as nullable but target is required)

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -25,10 +25,7 @@ import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
-import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
-import com.google.cloud.hadoop.io.bigquery.output.BigQueryOutputConfiguration;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
-import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
@@ -55,6 +52,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -296,7 +294,7 @@ protected final Schema overrideOutputSchemaWithTableSchemaIfNeeded(
   private Schema getTableSchema(
     String tableName,
     @Nullable com.google.cloud.bigquery.Schema bqSchema,
-    FailureCollector collector) {
+    FailureCollector collector, Schema configuredSchema) {
```
albertshau, line 297: annotate configuredSchema as `@Nullable`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -321,7 +319,28 @@ private Schema getTableSchema(
       LOG.info("Table [%s] doesn't have a schema. Using input schema for writing records.", tableName);
       return null;
     }
-    return BigQueryUtil.getTableSchema(bqSchema, collector);
+    return getOrderedSchema(bqSchema, configuredSchema);
+  }
+
+  private Schema getOrderedSchema(com.google.cloud.bigquery.Schema bqSchema, Schema configuredSchema) {
```
albertshau, line 325: annotate configuredSchema as `@Nullable`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -321,7 +319,28 @@ private Schema getTableSchema(
       LOG.info("Table [%s] doesn't have a schema. Using input schema for writing records.", tableName);
       return null;
     }
-    return BigQueryUtil.getTableSchema(bqSchema, collector);
+    return getOrderedSchema(bqSchema, configuredSchema);
```
albertshau, line 322: I think it's better for this call to happen outside of this method, in the overrideOutputSchemaWithTableSchemaIfNeeded method, and keep getTableSchema() the same so that it only fetches the table schema. Otherwise the contract is a bit confusing.

albertshau (commented):
please be sure to mention in the bug release notes the impact of this change.

It fixes some bugs that could potentially cause existing pipelines to fail to deploy when upgraded. Should mention what should be done to fix it in these situations.

albertshau (approved):

greeshmaswaminathan:
Merged another PR for the fix - https://github.com/data-integrations/google-cloud/pull/636

PR #635, opened 2021-04-24
Bump version to 0.16.2

rmstar:


pom.xml
```diff
@@ -458,6 +459,18 @@
       <artifactId>junit</artifactId>
       <version>${junit.version}</version>
     </dependency>
+    <dependency>
```
rmstar, line 462: This dependency was added in https://github.com/data-integrations/google-cloud/commit/9452b29bf925a9a74e707910d089d3e9b14169f2. Powermock is used in BigQueryExecuteConfigTest.

CuriousVini (approved):

PR #636, opened 2021-04-26
PLUGIN-678 - Override configured schema only for load job config 

greeshmaswaminathan:
- Configured schema was overridden for BQSink in prepareRun for preventing unexpected schema changes in target table (https://cdap.atlassian.net/browse/PLUGIN-395)
- With date time changes (https://github.com/data-integrations/google-cloud/commit/9bde7d253ec42f43919170a97ea8b470040d9dfc), the configured schema was used to create the AVRO schema and this resulted in the bug https://cdap.atlassian.net/browse/PLUGIN-678 . This was done because configured schema could be a subset of the data schema and is consistent with the BigQueryJsonConverter behavior.
- Moved the schema overriding to OutputCommitter only for the load job setting so that the configured schema remains the same and target table schema is not changed
- Tested for existing/new tables with INTEGER, Float, Date, Time, Timestamp, String , Datetime
- Tested for upsert on existing tables
- Tested pipeline with no schema (GCS avro to BQ sink)
- Tested that schema change happens when `update schema` is set
- Tested BQ multi table sink with Integer

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -25,10 +25,7 @@ import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
-import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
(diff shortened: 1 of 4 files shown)


tivv (commented):
Could you add a test?

albertshau (approved):
lgtm. Please also make sure to add some integration tests in the other project.

greeshmaswaminathan:
> Could you add a test?

I will add some integration tests.

greeshmaswaminathan:
> lgtm. Please also make sure to add some integration tests in the other project.

Yes, working on it.

PR #637, opened 2021-04-27
PLUGIN-678 - Override configured schema only for load job config 

greeshmaswaminathan:
Cherry pick for https://github.com/data-integrations/google-cloud/pull/636


albertshau (approved):
lgtm. Do we need to cherry-pick into any other branch?

greeshmaswaminathan:
> lgtm. Do we need to cherry-pick into any other branch?
Not immediately required I think, because regression is only for 6.4. Will keep a watch on issues reported for BQ in older versions.

PR #640, opened 2021-04-27
Fix issue with auto-detect on project for BQ Argument Setter

fernst:


rmstar (approved):

rmstar:
It would be better to first merge #641, then cherrypick to all relevant release branches.

greeshmaswaminathan (approved):

PR #649, opened 2021-05-07
Restructured some classes to make code easier to use from the BQ Pushdown implementation.

fernst:
Moved some common properties that will be used by the Source, Sink and SQL engine into the `BigQueryBaseConfig.java` class.

Moved some methods from the `AbstractBigQuerySink` into the `BigQuerySinkUtils` class in order to make them reusable in the SQL Engine implementation, since the BQ Source and BQ Sink share a lot of functionality with the BQ Push/Pull operations.


pom.xml
```diff
@@ -72,7 +72,7 @@
     <avro.version>1.8.2</avro.version>
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
-    <cdap.version>6.4.0</cdap.version>
+    <cdap.version>6.5.0-SNAPSHOT</cdap.version>
     <cdap.plugin.version>2.6.0</cdap.plugin.version>
```
tivv: I highly recommend changing this to 2.7.0-SNAPSHOT as well. There were some coordinated changes related to Spark 1 dropped

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -48,6 +58,107 @@
...
+  public static Configuration getOutputConfiguration(Configuration baseConfiguration,
+                                                     BigQueryBaseConfig config,
+                                                     String bucket,
+                                                     String pathPrefix,
+                                                     String tableName,
+                                                     List<BigQueryTableFieldSchema> fields) throws IOException {
+    Configuration configuration = new Configuration(baseConfiguration);
+    String temporaryGcsPath = getTemporaryGcsPath(bucket, pathPrefix, tableName);
+
+    BigQuerySinkUtils.configureBigQueryOutput(configuration,
```
tivv: So you need to qualify it here? It's same class

tivv: Also I believe you can make configureBigQueryOutput private now (good as it takes mutable argument) and make naming consistent. We have `configureBigQueryOutput`, `configureBigQueryOutputForMultiSink`. The third method should have consistent naming

src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
```diff
@@ -0,0 +1,106 @@
...
+  @Nullable
+  public String getBucket() {
+    if (bucket != null) {
+      bucket = bucket.trim();
+      if (bucket.isEmpty()) {
+        return null;
+      }
+      // remove the gs:// scheme from the bucket name
+      if (bucket.startsWith(SCHEME)) {
+        bucket = bucket.substring(SCHEME.length());
```
tivv, line 100: It would be really great to add a validation that it's not the whole URI. It's pretty confusing right now

pom.xml
fernst: I will bump this as well.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: Good catch on both.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -48,12 +58,113 @@
...
-                                             List<BigQueryTableFieldSchema> fields) throws IOException {
+  /**
+   * Creates Hadoop configuration for the given table and its fields.
+   *
+   * @param bucket bucket name
+   * @param tableName table name
+   * @param fields list of Big Query fields
+   * @return Hadoop configuration
+   */
+  public static Configuration getOutputConfiguration(Configuration baseConfiguration,
```
tivv: Please make this method name follow same naming (configure*Output)

fernst: I'd argue that they are different because this method builds a new configuration instance (by copying the supplied one and adding to it), but the other one operates on the supplied configuration by modifying the underlying config instance.

fernst: Did additional refactor/cleanup. Take a look now.

tivv (approved):

PR #655, opened 2021-05-10
Handle nullable arrays in BigQuery Sink

VictorD-Veolia:
BigQuery fields can't be both NULLABLE and REQUIRED. However, it is still possible to insert null arrays BigQuery, which are then converted into empty arrays.

There is then no reason to deny users from using schema with nullable arrays in the BigQuery Sink.


VictorD-Veolia:
Hi @bdmogal. I updated the unit test for the behavior fixed in this PR and created an issue in JIRA : https://cdap.atlassian.net/browse/PLUGIN-795

itsankit-google:
Merged as part of https://github.com/data-integrations/google-cloud/pull/1001.

PR #656, opened 2021-05-10
Handle nullable arrays of records in BigQuery Sink 

VictorD-Veolia:
Fix to handle properly nullable array of records and arrays of nullable records in BigQuery Sink

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -29,6 +29,7 @@ import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
+import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -210,27 +211,60 @@ public static String getTemporaryGcsPath(String bucket, String pathPrefix, Strin
...
+   * Return true if field is of type array and false else
+   * @param field the field
+   * @return isArray
+   */
+  private static boolean fieldIsArray(Schema.Field field) {
+    Schema fieldSchema = field.getSchema();
+
+    if (fieldSchema.getType().equals(Schema.Type.ARRAY)) {
+      return true;
+    } else if (fieldSchema.getType().equals(Schema.Type.UNION)) {
```
chtyim: No need to have `else` after an `if` block that returns.

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtilsTest.java
```diff
@@ -0,0 +1,124 @@
...
+import io.cdap.cdap.api.data.schema.Schema;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.List;
+
+public class BigQuerySinkUtilsTest {
+
+    @Test
+    public void testGenerateTableFieldSchema() {
```
chtyim: Please fix the indentation. We use 2 spaces as indentation.

PR #657, opened 2021-05-10
Additional refactoring for SQL Engine integration

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -50,10 +50,10 @@
  * The logic in this class has been extracted from the {@link AbstractBigQuerySink} in order to make this functionality
  * available to other classes in this package.
  */
-final class BigQuerySinkUtils {
+public final class BigQuerySinkUtils {
 
-  private static final String gcsPathFormat = "gs://%s/%s";
-  private static final String temporaryBucketFormat = gcsPathFormat + "/input/%s-%s";
+  public static final String GS_PATH_FORMAT = "gs://%s/%s";
+  private static final String temporaryBucketFormat = GS_PATH_FORMAT + "/input/%s-%s";
```
tivv: Could you please change this name to uppercase as well.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -130,27 +147,42 @@ private static void createBucket(Storage storage, String bucket, @Nullable Strin
     }
   }
 
+  /**
+   * Updates {@link Configuration} with bucket details.
+   * Uses provided bucket, otherwise generates random.
```
tivv: It's actually not plain random, it's equal to runid.

tivv (commented):
Could you also add some unit tests for this?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: Good point. let me fix this.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: Done.

tivv (commented):
Please add a unit test for the utility method introduced.

fernst:
@tivv Added a test for the `configureBucket` method.

tivv (approved):

PR #658, opened 2021-05-12
Updated BigQuery source in order to make code reusable for BQ Pushdown.

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public String getInputFormatClassName() {
+    return PartitionedBigQueryInputFormat.class.getName();
+  }
+
+  @Override
+  public Map<String, String> getInputFormatConfiguration() {
+    Map<String, String> config = new HashMap<>();
```
tivv: Can be done as
`StreamSupport.stream(configuration.spliterator(), false).collect(Collectors.toMap(Entry::getKey, Entry::getValue))`
Wondering why do we need to copy the whole configuration. In any case, it would be great to move to constructor.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -0,0 +1,208 @@
...
+                            cmekKey);
+    }
+
+    return bucket;
+  }
+
+  /**
+   * Updates {@link Configuration} with bucket details.
+   */
+  public static void configureBucket(Configuration configuration, String bucket, String runId) {
```
tivv: It looks like it's not only bucket, but a full path inside the bucket.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -0,0 +1,208 @@
...
+    configuration.set("fs.default.name", String.format("gs://%s/%s/", bucket, runId));
```
tivv: GCS_PATH_FORMAT?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -82,7 +71,7 @@
   private Configuration configuration;
   private final BigQueryAvroToStructuredTransformer transformer = new BigQueryAvroToStructuredTransformer();
   // UUID for the run. Will be used as bucket name if bucket is not provided.
-  private UUID uuid;
+  private String runId;
```
tivv: We are already using a run id term (ProgramRunId) and it looks like this guy is something else. Please use a different name

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
tivv: Also, do we log this value anywhere? It makes sense to log it early to ensure we have it in case of any issues (e.g. permissions)

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
tivv: I am wondering if we should switch from making Configuration that would have all the defaults, then adding few items and changing it to a map that potentially clashes with the different defaults on the other side.

Instead we should make a map and then create a Configuration only when needed (when calling org.apache.hadoop.fs.Path#getFileSystem)

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -0,0 +1,208 @@
...
+    configuration.setBoolean("fs.gs.metadata.cache.enable", false);
+  }
+
+  /**
+   * Sets up service account credentials into supplied Hadoop configuration.
+   *
+   * @param configuration Hadoop Configuration instance.
+   * @param config Big Query configuration.
+   */
+  public static void configureServiceAccount(Configuration configuration, BigQueryBaseConfig config) {
```
tivv, line 98: Where is this one used?

tivv (commented):
Generally I don't feel good about having this number of small methods. When using it's very easy to screw up. Also it's hard to follow which methods are called and what is set. Can we have a smaller number of bigger methods?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
fernst: This is an identifier for a given execution. Each source has it's own "runId" that is used to create a new path to store objects related to this source.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
fernst: Oops, will add it back.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
fernst: You're right, this should be the static variable.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
tivv: Well, it's not a key for execution, you can't retrieve an execution using it. It would be no problems if we did not have already io.cdap.cdap.common.app.RunIds widely used, but now those two will be confused and if someone try to retrieve a run using this id he/she would fail.

fernst: I see your point now.

fernst: I have renamed this variable to something that is less confusing.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
yaojiefeng: You can use `SourceInputFormatProvider` to just return the non-default configs, there is logic to compute the non-default here - https://github.com/cdapio/hydrator-plugins/blob/develop/hydrator-common/src/main/java/io/cdap/plugin/common/SourceInputFormatProvider.java#L40. It is in hydrator-common, so you should be able to directly use it here.

tivv:
Generally I don't feel good about having this number of small methods. When using it's very easy to screw up. Also it's hard to follow which methods are called and what is set. Can we have a smaller number of bigger methods?

fernst:
@tivv from the perspective of the SQL engine, I only need these 3 methods:

```
    // Configure GCS Bucket to use
    String gcsPath = BigQuerySQLEngineUtils.getGCSPath(bucket, runId, table);
    BigQuerySourceUtils.configureBucket(configuration, gcsPath);

    // Set up temporary table name. This will be used if the source table is a view
    BigQuerySourceUtils.configureTemporaryTableName(configuration, table);

    // Configure BigQuery input format.
    BigQuerySourceUtils.configureBigQueryInput(configuration, project, bqDataset, table, gcsPath);
```

That's the reason I build all of these small methods in the utility class, it allows me to bring only the small bits of functionality I need to make the SQL engine implementation work.

fernst:
@tivv condensed some methods into one.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -82,7 +71,7 @@
   private Configuration configuration;
   private final BigQueryAvroToStructuredTransformer transformer = new BigQueryAvroToStructuredTransformer();
   // UUID for the run. Will be used as bucket name if bucket is not provided.
-  private UUID uuid;
+  private String bucketPath;
```
tivv, line 74: nit: bucketName, it's not a full path

tivv (approved):

PR #659, opened 2021-05-20
Initial implementation for BigQuery SQL engine.

fernst:
At this stage, the push/pull stages have been tested, and the Join on Keys operation is implemented.

Join on conditions will be implemented in a later PR.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatProvider.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatProvider.java
@@ -35,6 +35,13 @@ public class BigQueryOutputFormatProvider implements OutputFormatProvider {
   protected Schema tableSchema;
 
   /**
+   * This constructor is only used when Spark serializes this class.
+   */
+  protected BigQueryOutputFormatProvider() {
+    // no-op
+  }
+
+  /**
    * @param configuration Hadoop Configuration object, must be set up by using {@link BigQuerySinkUtils#configureOutput}
    * @param tableSchema CDAP Schema for the output table.
    */
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
@@ -31,6 +31,13 @@ public class BigQueryInputFormatProvider implements InputFormatProvider {
 
   protected Map<String, String> inputFormatConfiguration;
 
+  /**
+   * This constructor is only used when Spark serializes this class.
+   */
+  protected BigQueryInputFormatProvider() {
+    // no-op
+  }
+
   public BigQueryInputFormatProvider(Configuration configuration) {
     this.inputFormatConfiguration = StreamSupport
       .stream(configuration.spliterator(), false)
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sqlengine;
+
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.Job;
+import com.google.cloud.bigquery.JobConfiguration;
+import com.google.cloud.bigquery.JobId;
+import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.Table;
+import com.google.cloud.bigquery.TableId;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.engine.sql.SQLEngineException;
+import io.cdap.cdap.etl.api.engine.sql.dataset.SQLDataset;
+import io.cdap.cdap.etl.api.engine.sql.request.SQLJoinRequest;
+import io.cdap.cdap.etl.api.join.JoinCondition;
+import io.cdap.cdap.etl.api.join.JoinDefinition;
+import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQuerySQLBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * SQL Dataset that represents the result of a Join operation that is executed in BigQuery.
+ */
+public class BigQueryJoinDataset implements SQLDataset, BigQuerySQLDataset {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BigQueryJoinDataset.class);
+
+  private final String datasetName;
+  private final JoinDefinition joinDefinition;
+  private final BigQuery bigQuery;
+  private final String location;
+  private final String project;
+  private final String bqDataset;
+  private final String bqTable;
+  private final String jobId;
+  private final BigQuerySQLBuilder helper;
+  private Long numRows;
+
+  private BigQueryJoinDataset(String datasetName,
+                              JoinDefinition joinDefinition,
+                              Map<String, String> stageToTableNameMap,
+                              BigQuery bigQuery,
+                              String location,
+                              String project,
+                              String bqDataset,
+                              String bqTable,
+                              String jobId) {
+    this.datasetName = datasetName;
+    this.joinDefinition = joinDefinition;
+    this.bigQuery = bigQuery;
+    this.location = location;
+    this.project = project;
+    this.bqDataset = bqDataset;
+    this.bqTable = bqTable;
+    this.jobId = jobId;
+    this.helper = new BigQuerySQLBuilder(this.project, this.bqDataset, stageToTableNameMap);
+  }
+
+  public static BigQueryJoinDataset getInstance(SQLJoinRequest joinRequest,
+                                                Map<String, String> bqTableNamesMap,
+                                                BigQuery bigQuery,
+                                                String location,
+                                                String project,
+                                                String dataset,
+                                                String runId) {
+
+    // Get new Job ID for this push operation
+    String jobId = BigQuerySQLEngineUtils.getIdentifier();
+
+    // Build new table name for this dataset
+    String table = BigQuerySQLEngineUtils.getNewTableName(runId);
+
+    BigQueryJoinDataset instance = new BigQueryJoinDataset(joinRequest.getDatasetName(),
+                                                           joinRequest.getJoinDefinition(),
+                                                           bqTableNamesMap,
+                                                           bigQuery,
+                                                           location,
+                                                           project,
+                                                           dataset,
+                                                           table,
+                                                           jobId);
+    instance.executeJoin();
+    return instance;
+  }
+
+  private void executeJoin() {
+    TableId destinationTable = TableId.of(project, bqDataset, bqTable);
+
+    String query = getQuery(joinDefinition);
+    LOG.info("Creating table `{}` using job: {} with SQL statement: {}", bqTable, jobId, query);
+
+    // Run BigQuery job with generated SQL statement, store results in a new table, and set priority to BATCH
+    // TODO: Make priority configurable
+    QueryJobConfiguration queryConfig =
+      QueryJobConfiguration.newBuilder(query)
+        .setDestinationTable(destinationTable)
+        .setPriority(QueryJobConfiguration.Priority.BATCH)
+        .build();
+
+    // Create a job ID so that we can safely retry.
+    JobId bqJobId = JobId.newBuilder().setJob(jobId).setLocation(location).setProject(project).build();
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(bqJobId).build());
+
+    // Wait for the query to complete.
+    try {
+      queryJob = queryJob.waitFor();
+    } catch (InterruptedException ie) {
+      throw new SQLEngineException("Interrupted exception when executing Join operation", ie);
+    }
+
+    // Check for errors
+    if (queryJob == null) {
+      throw new SQLEngineException("BigQuery job not found: " + jobId);
+    } else if (queryJob.getStatus().getError() != null) {
+      throw new SQLEngineException(String.format(
+        "Error executing BigQuery Job: '%s' in Project '%s', Dataset '%s', Location'%s' : %s",
+        jobId, project, bqDataset, location, queryJob.getStatus().getError().toString()));
+    }
+
+    LOG.info("Created BigQuery table `{}` using Job: {}", bqTable, jobId);
+  }
+
+  protected String getQuery(JoinDefinition joinDefinition) {
+    if (joinDefinition.getCondition().getOp() == JoinCondition.Op.KEY_EQUALITY) {
+      return helper.getFieldEqualityQuery(joinDefinition);
+    } else {
+      //TODO: implement
+      throw new SQLEngineException("Advanced joins not currently supported");
+    }
+  }
+
+  @Override
+  public String getDatasetName() {
+    return datasetName;
+  }
+
+  @Override
+  public Schema getSchema() {
+    return joinDefinition.getOutputSchema();
+  }
+
+  @Override
+  public long getNumRows() {
+    // Get the number of rows from BQ if not known at this time.
+    if (numRows == null) {
+      numRows = BigQuerySQLEngineUtils.getNumRows(bigQuery, project, bqDataset, bqTable);
+    }
+
+    return numRows;
+  }
+
+  @Override
+  public String getBigQueryTableName() {
+    return bqTable;
+  }
+
+  @Override
+  @Nullable
+  public String getGCSPath() {
+    return null;
+  }
+
+  @Override
+  public String getJobId() {
+    return jobId;
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sqlengine;
+
+import com.google.cloud.bigquery.BigQuery;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.dataset.lib.KeyValue;
+import io.cdap.cdap.etl.api.Transform;
+import io.cdap.cdap.etl.api.engine.sql.dataset.SQLPullDataset;
+import io.cdap.cdap.etl.api.engine.sql.request.SQLPullRequest;
+import io.cdap.plugin.gcp.bigquery.source.BigQueryInputFormatProvider;
+import io.cdap.plugin.gcp.bigquery.source.BigQuerySourceUtils;
+import io.cdap.plugin.gcp.bigquery.sqlengine.transform.PullTransform;
+import org.apache.avro.generic.GenericData;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.LongWritable;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * SQL Pull Dataset implementation for BigQuery backed datasets.
+ */
+public class BigQueryPullDataset extends BigQueryInputFormatProvider
+  implements SQLPullDataset<StructuredRecord, LongWritable, GenericData.Record>, BigQuerySQLDataset {
+
+  private final BigQuery bigQuery;
+  private final String datasetName;
+  private final Schema schema;
+  private final String project;
+  private final String bqDataset;
+  private final String bqTable;
+  private final String gcsPath;
+  private Long numRows;
+
+  private BigQueryPullDataset(Configuration configuration,
+                              String datasetName,
+                              Schema schema,
+                              BigQuery bigQuery,
+                              String project,
+                              String bqDataset,
+                              String bqTable,
+                              String gcsPath) {
+    super(configuration);
+    this.datasetName = datasetName;
+    this.schema = schema;
+    this.bigQuery = bigQuery;
+    this.project = project;
+    this.bqDataset = bqDataset;
+    this.bqTable = bqTable;
+    this.gcsPath = gcsPath;
+  }
+
+  public static BigQueryPullDataset getInstance(SQLPullRequest pullRequest,
+                                                Configuration baseConfiguration,
+                                                BigQuery bigQuery,
+                                                String project,
+                                                String bqDataset,
+                                                String bqTable,
+                                                String bucket,
+                                                String runId) throws IOException {
+
+    // Clone configuration object
+    Configuration configuration = new Configuration(baseConfiguration);
+
+    // Configure BigQuery input format.
+    String gcsPath = BigQuerySQLEngineUtils.getGCSPath(bucket, runId, bqTable);
+    BigQuerySourceUtils.configureBigQueryInput(configuration, project, bqDataset, bqTable, gcsPath);
+
+    return new BigQueryPullDataset(configuration,
+                                   pullRequest.getDatasetName(),
+                                   pullRequest.getDatasetSchema(),
+                                   bigQuery,
+                                   project,
+                                   bqDataset,
+                                   bqTable,
+                                   gcsPath);
+  }
+
+  @Override
+  public Transform<KeyValue<LongWritable, GenericData.Record>, StructuredRecord> fromKeyValue() {
+    return new PullTransform(schema);
+  }
+
+  @Override
+  public String getDatasetName() {
+    return datasetName;
+  }
+
+  @Override
+  public Schema getSchema() {
+    return schema;
+  }
+
+  @Override
+  public long getNumRows() {
+    // Get the number of rows from BQ if not known at this time.
+    if (numRows == null) {
+      numRows = BigQuerySQLEngineUtils.getNumRows(bigQuery, project, bqDataset, bqTable);
+    }
+
+    return numRows;
+  }
+
+  @Override
+  public String getBigQueryTableName() {
+    return bqTable;
+  }
+
+  @Override
+  @Nullable
+  public String getJobId() {
+    return null;
+  }
+
+  @Override
+  public String getGCSPath() {
+    return gcsPath;
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.sqlengine;
+
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.dataset.lib.KeyValue;
+import io.cdap.cdap.etl.api.Transform;
+import io.cdap.cdap.etl.api.engine.sql.dataset.SQLPushDataset;
+import io.cdap.cdap.etl.api.engine.sql.request.SQLPushRequest;
+import io.cdap.plugin.gcp.bigquery.sink.BigQueryOutputFormatProvider;
+import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
+import io.cdap.plugin.gcp.bigquery.sink.Operation;
+import io.cdap.plugin.gcp.bigquery.sqlengine.transform.PushTransform;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.NullWritable;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * SQL Push Dataset implementation for BigQuery backed datasets.
+ */
+public class BigQueryPushDataset extends BigQueryOutputFormatProvider
+  implements SQLPushDataset<StructuredRecord, StructuredRecord, NullWritable>, BigQuerySQLDataset {
+
+  private final String datasetName;
+  private final BigQuery bigQuery;
+  private final String project;
+  private final String bqDataset;
+  private final String bqTable;
+  private final String gcsPath;
+  private final String jobId;
+  private Long numRows;
+
+  private BigQueryPushDataset(String datasetName,
+                              Schema tableSchema,
+                              Configuration configuration,
+                              BigQuery bigQuery,
+                              String project,
+                              String bqDataset,
+                              String bqTable,
+                              String jobId,
+                              String gcsPath) {
+    super(configuration, tableSchema);
+    this.datasetName = datasetName;
+    this.bigQuery = bigQuery;
+    this.project = project;
+    this.bqDataset = bqDataset;
+    this.bqTable = bqTable;
+    this.jobId = jobId;
+    this.gcsPath = gcsPath;
+  }
+
+  protected static BigQueryPushDataset getInstance(SQLPushRequest pushRequest,
+                                                   Configuration baseConfiguration,
+                                                   BigQuery bigQuery,
+                                                   String project,
+                                                   String dataset,
+                                                   String bucket,
+                                                   String runId) throws IOException {
+    // Get new Job ID for this push operation
+    String jobId = BigQuerySQLEngineUtils.getIdentifier();
+
+    // Build new table name for this dataset
+    String table = BigQuerySQLEngineUtils.getNewTableName(runId);
+
+    // Clone configuration object
+    Configuration configuration = new Configuration(baseConfiguration);
+
+    // Set required configuration properties.
+    configuration.set(BigQueryConstants.CONFIG_JOB_ID, jobId);
+    configuration.set(BigQueryConstants.CONFIG_OPERATION, Operation.INSERT.name());
+
+    // Configure output.
+    String gcsPath = BigQuerySQLEngineUtils.getGCSPath(bucket, runId, table);
+    List<BigQueryTableFieldSchema> fields =
+      BigQuerySinkUtils.getBigQueryTableFieldsFromSchema(pushRequest.getDatasetSchema());
+    BigQuerySinkUtils.configureBucket(configuration, bucket, runId);
+    BigQuerySinkUtils.configureOutput(configuration, project, dataset, table, gcsPath, fields);
+
+    //Build new Instance
+    return new BigQueryPushDataset(pushRequest.getDatasetName(),
+                                   pushRequest.getDatasetSchema(),
+                                   configuration,
+                                   bigQuery,
+                                   project,
+                                   dataset,
+                                   table,
+                                   jobId,
+                                   gcsPath);
+  }
+
+  @Override
+  public Transform<StructuredRecord, KeyValue<StructuredRecord, NullWritable>> toKeyValue() {
+    return new PushTransform();
+  }
+
+  @Override
+  public String getDatasetName() {
+    return datasetName;
+  }
+
+  @Override
+  public Schema getSchema() {
+    return tableSchema;
+  }
+
+  @Override
+  public long getNumRows() {
+    // Get the number of rows from BQ if not known at this time.
+    if (numRows == null) {
+      numRows = BigQuerySQLEngineUtils.getNumRows(bigQuery, project, bqDataset, bqTable);
+    }
+
+    return numRows;
+  }
+
+  @Override
+  public String getBigQueryTableName() {
+    return bqTable;
+  }
+
+  @Override
+  public String getGCSPath() {
+    return gcsPath;
+  }
+
+  @Override
+  public String getJobId() {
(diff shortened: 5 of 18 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+    // Handle Strings that are stored as a Byte Buffer.
+    if (fieldSchema.getType() == Schema.Type.STRING && field instanceof ByteBuffer) {
+      return StandardCharsets.UTF_8.decode((ByteBuffer) field).toString();
+    }
+
+    return super.convertField(field, fieldSchema);
+  }
+
+  @VisibleForTesting
+  protected Integer mapInteger(Object field) throws SQLEngineException {
```
lahwran: will mapInteger ever receive an Integer as input?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      return result;
+    } else {
+      throw new SQLEngineException(String.format(INTEGER_CONVERSION_ERROR, field));
+    }
+  }
+
+  @VisibleForTesting
+  protected Float mapFloat(Object field) {
+    if (field instanceof Double) {
```
lahwran, line 92: Same question here - will mapFloat ever receive a Float as input?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
fernst: No, the reason I had to implement this function is because BigQuery stores integers as INT64 (which get mapped to Long when reading form the Avro input).

Same for Floats, as they are stored in Float64 (which get read as Double from Avro)

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatProvider.java
```diff
@@ -34,6 +34,13 @@
   protected Configuration configuration;
   protected Schema tableSchema;
 
+  /**
+   * This constructor is only used when Spark serializes this class.
+   */
+  public BigQueryOutputFormatProvider() {
```
tivv: Does it need to be public?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
```diff
@@ -31,6 +31,13 @@
 
   protected Map<String, String> inputFormatConfiguration;
 
+  /**
+   * This constructor is only used when Spark serializes this class.
+   */
+  public BigQueryInputFormatProvider() {
```
tivv: Does it need to be public?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,182 @@
...
+  private final JoinDefinition joinDefinition;
+  private final BigQuery bigQuery;
+  private final String project;
+  private final String bqDataset;
+  private final String bqTable;
+  private final String jobId;
+  private final BigQuerySQLBuilder helper;
+  private Long numRows;
+
+  public BigQueryJoinDataset(String datasetName,
```
tivv: Is one always supposed to use getInstance? If yes, make this private.
Also consider moving executeJoin to getInstance, it's better to do heavylifting outside of constructor.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,154 @@
...
+  private String datasetName;
+  private Schema schema;
+  private String project;
+  private String bqDataset;
+  private String bqTable;
+  private String gcsPath;
+  private Long numRows;
+  private final SQLEngineAvroToStructuredTransformer transformer = new SQLEngineAvroToStructuredTransformer();
+
+  public BigQueryPullDataset() {
```
tivv: Does it need to be public?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,154 @@
...
+    super();
```
tivv: You usually don't need to explicilty call super without parameters

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,154 @@
...
+                                   bigQuery,
+                                   project,
+                                   bqDataset,
+                                   bqTable,
+                                   gcsPath);
+  }
+
+  @Override
+  public Transform<KeyValue<LongWritable, GenericData.Record>, StructuredRecord> fromKeyValue() {
+    return new SerializableTransform<KeyValue<LongWritable, GenericData.Record>, StructuredRecord>() {
```
tivv: In which case this guy is used?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
```diff
@@ -0,0 +1,166 @@
...
+  private String datasetName;
+  private transient BigQuery bigQuery;
+  private String project;
+  private String bqDataset;
+  private String bqTable;
+  private String gcsPath;
+  private String jobId;
+  private Long numRows;
+
+  public BigQueryPushDataset() {
```
tivv: Does it need to be public?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
```diff
@@ -0,0 +1,166 @@
...
+    super();
```
tivv: You usually don't need explicit super call without parameters

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @Override
+  public Schema getSchema() {
+    return tableSchema;
+  }
+
+  @Override
+  public long getNumRows() {
+    // Get the number of rows from BQ if not known at this time.
+    if (numRows == null) {
```
tivv, line 129: This code is duplicated at least three times. Does it make sense to make a utility method that provides a lazy supplier or function that calculates number of rows?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+      .map(this::buildSelectedField)
+      .collect(Collectors.joining(" , "));
+  }
+
+  protected String buildSelectedField(JoinField joinField) {
+    StringBuilder builder = new StringBuilder();
+
+    builder.append(getBQTableNameForStage(joinField.getStageName()));
+    builder.append(".");
+    builder.append(joinField.getFieldName());
```
tivv: Do we need any quoting? Would it work with spaces / special chars in field names?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+
+    if (joinField.getAlias() != null) {
+      builder.append(" AS ");
+      builder.append(joinField.getAlias());
```
tivv: Also a quoting question

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,284 @@
...
+    storage = GCPUtils.getStorage(project, credentials);
+
+    String cmekKey = context.getRuntimeArguments().get(GCPUtils.CMEK_KEY);
+    configuration = BigQueryUtil.getBigQueryConfig(config.getServiceAccount(), config.getProject(),
+                                                   cmekKey, config.getServiceAccountType());
+    BigQuerySinkUtils.createResources(bigQuery, storage, dataset, bucket, config.getLocation(), cmekKey);
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, RuntimeContext context) {
```
tivv: Why do you need to override just to call super?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  @Override
+  public boolean exists(String datasetName) throws SQLEngineException {
+    return datasets.containsKey(datasetName);
+  }
+
+  @Override
+  public boolean canJoin(SQLJoinDefinition sqlJoinDefinition) {
+    //TODO: implement
```
tivv: Please check at least for advanced joins that is not supported right now

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,284 @@
...
+      ));
+  }
+
+  /**
+   * Stops the running job for the supplied dataset
+   *
+   * @param datasetName the name of the dataset (representing the stage name in CDAP)
+   * @param bqDataset the BigQuery Dataset Instance
+   */
+  protected void cancelJob(String datasetName, BigQuerySQLDataset bqDataset) {
```
tivv: So, should you call it stage name? Also consider passing it less, here it's used only to log

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+  }
+
+  /**
+   * Deletes the BigQuery table for the supplied dataset
+   *
+   * @param datasetName the name of the dataset (representing the stage name in CDAP)
+   * @param bqDataset the BigQuery Dataset Instance
+   */
+  protected void deleteTable(String datasetName, BigQuerySQLDataset bqDataset) {
```
tivv: So, should you call it stage name? Also consider passing it less, here it's used only to log

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineUtils.java
```diff
@@ -0,0 +1,42 @@
...
+  public static String getGCSPath(String bucket, String runId, String tableId) {
+    return String.format(GCS_PATH_FORMAT, bucket, runId, tableId);
+  }
+
+  public static String getIdentifier() {
+    return UUID.randomUUID().toString().replace("-", "");
+  }
+
+  public static String getNewTableName(String runId) {
+    return String.format(BQ_TABLE_NAME_FORMAT, runId, BigQuerySQLEngineUtils.getIdentifier());
```
tivv: You don't need qualifier `BigQuerySQLEngineUtils` here

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+      Double value = (Double) field;
+      Float result = value.floatValue();
+      if ((double) result != value) {
```
tivv: This is pretty fragile.

tivv:
```groovy
double d = 12345/10000.0
float f = (float)d
double d2 = f
if (d2 != d) {
    println d + " != " + f
}
println Double.doubleToLongBits(d)
println Double.doubleToLongBits(d2)
```
Output:
```
1.2345 != 1.2345
4608238512912635789
4608238513140269056
```

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -643,4 +651,23 @@ private static void match(String text, String propertyName, String pattern,
...
+  public static void deleteTemporaryDirectory(Configuration configuration, String dir) {
+    Path path = new Path(dir);
+    try {
+      FileSystem fs = path.getFileSystem(configuration);
+      if (fs.exists(path)) {
+        fs.delete(path, true);
+        LOG.debug("Deleted temporary directory '{}'", path);
+      }
+    } catch (IOException e) {
+      LOG.warn("Failed to delete temporary directory '{}': {}", path, e.getMessage());
```
tivv: Please log an error. If it happens in pipeline we start collecting buckets and we want to know about it sooner than later

src/test/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLBuilderTest.java
```diff
@@ -0,0 +1,225 @@
...
+  @Test
+  public void testBuildJoinOperationLeftJoin() {
+    Map<String, JoinKey> stringJoinKeyMap = new HashMap<>();
+    configureJoinOperationTest(stringJoinKeyMap);
+
+    JoinStage leftStage = JoinStage.builder("leftStage", null).setRequired(true).build();
+    JoinStage rightStage = JoinStage.builder("rightStage", null).setRequired(false).build();
+
+    helper.addJoinOnKeyOperation(leftStage, rightStage, stringJoinKeyMap);
+    Assert.assertEquals("LEFT OUTER JOIN `rightTable` ON some join condition",
```
tivv: So, why do we quote table names only in some cases?

tivv (commented):
Also please ensure proper coverage (80%+) on all added code

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
tivv: Note: this is groovy code, you can run it in Tools -> Groovy Console in Intellij.

fernst: Hmm, good point. I had skipped over these float/double arithmetic quirks.

Do you have any suggestions on how to validate? The other strategy I see is to make sure the double falls "in range" with a `-Float.MAX_VALUE >= double && Float.MAX_VALUE <= double`

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: Will do.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatProvider.java
fernst: Not really, let me fix.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryInputFormatProvider.java
fernst: Not really, let me fix.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: The answer to your question is "yes", so I will make this private.

I will also move the "execution" to the `getInstance` method.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
tivv: Could you clarify what are you trying to validate here? For floating point there is also loss of precision when you convert it to float. 
FYI: Float.MIN_VALUE is 2^-149 (positive)

src/test/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLBuilderTest.java
fernst: So each tests is working on a different part of the code. When fully executed, the result will be a fully quoted table name.

For the purposes of some of my test, I just want to ensure that the output string from a given method has the right spacing/separation when building SQL.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
fernst: BigQuery stores all floats as Float64, This means Float fields in CDAP get stored as Float64 in BQ.

Since records in BQ are stored as Float64, they get mapped to Double when read from Avro.

My goal here is to ensure that this Double "fits" into a Float.

tivv: In what cases it won't fit and what should happen? Where does this joiner output schema come from? Can user set joiner schema output be float with double as input and what would happen for BQ pushdown enabled vs disabled?

fernst: I don't expect this to be an issue with joins. However, in the future, when aggregations are supported, we need to account for the posibility of something like an "overflow" (where, for example `SUM(some_int_column)` results in an integer overflow).

I changed the logic to ensure that the Float is "close enough" to the double value (within 1e-15) in order to assume the conversion from double to float is "valid".

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
fernst: These transform operations are serialized and supplied to the Spark workers when reading/writing into BQ.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: I renamed this to Stage name.

I agree that I'm only passing this for logging purposes, but I think it's important to let the end user know which stage(s) caused issues.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,180 @@
...
+  private final BigQuery bigQuery;
+  private final String project;
+  private final String bqDataset;
+  private final String bqTable;
+  private final String jobId;
+  private final BigQuerySQLBuilder helper;
+  private Long numRows;
+
+  private BigQueryJoinDataset(String datasetName,
+                             JoinDefinition joinDefinition,
```
albertshau: indentation is off

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,180 @@
...
+    TableId destinationTable = TableId.of(project, bqDataset, bqTable);
+
+    String query = getQuery(joinDefinition);
+    LOG.info("Creating table `{}` using job: {} with SQL statement: {}", bqTable, jobId, query);
+
+    // Run BigQuery job with generated SQL statement, store results in a new table, and set priority to BATCH
+    QueryJobConfiguration queryConfig =
+      QueryJobConfiguration.newBuilder(query)
+        .setDestinationTable(destinationTable)
+        .setPriority(QueryJobConfiguration.Priority.BATCH)
```
albertshau, line 116: this should probably be configurable (can happen in a later PR)

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,180 @@
...
+        .build();
+
+    // Create a job ID so that we can safely retry.
+    JobId bqJobId = JobId.of(jobId);
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(bqJobId).build());
```
albertshau, line 121: believe location needs to be set here, otherwise it assumes US location

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,180 @@
...
+    // Wait for the query to complete.
+    try {
+      queryJob = queryJob.waitFor();
+    } catch (InterruptedException ie) {
+      throw new SQLEngineException("Interrupted exception when executing Join operation", ie);
+    }
+
+    // Check for errors
+    if (queryJob == null) {
+      throw new SQLEngineException("BigQuery job not found: " + jobId);
```
albertshau, line 132: when does this happen?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -0,0 +1,180 @@
...
+    } else if (queryJob.getStatus().getError() != null) {
+      // You can also look at queryJob.getStatus().getExecutionErrors() for all
+      // errors, not just the latest one.
+      throw new SQLEngineException("Error executing BigQuery Job: "
```
albertshau: should also include all info (jobid, project?, location?) needed to allow the user to find the actual BQ job to take a closer look.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * SQL Pull Dataset implementation for BigQuery backed datasets.
+ */
+public class BigQueryPullDataset extends BigQueryInputFormatProvider
+  implements SQLPullDataset<StructuredRecord, LongWritable, GenericData.Record>, BigQuerySQLDataset {
+
+  // Note that these properties are not final in order to support Spark's serialization.
+  private transient BigQuery bigQuery;
+  private String datasetName;
```
albertshau: do you get an error if they are final? Normally they should be set with reflection when deserializing

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,155 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * SQL Pull Dataset implementation for BigQuery backed datasets.
+ */
+public class BigQueryPullDataset extends BigQueryInputFormatProvider
+  implements SQLPullDataset<StructuredRecord, LongWritable, GenericData.Record>, BigQuerySQLDataset {
+
+  // Note that these properties are not final in order to support Spark's serialization.
+  private transient BigQuery bigQuery;
```
albertshau: this will be null then when deserialized?

I'm also confused about why this is getting serialized at all, could it be avoided by tweaking the app code?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,330 @@
...
+                                        project,
+                                        dataset,
+                                        bucket,
+                                        runId);
+
+      LOG.info("Executing Push operation for dataset {} stored in table {}",
+               sqlPushRequest.getDatasetName(),
+               pushDataset.getBigQueryTableName());
+
+      datasets.put(sqlPushRequest.getDatasetName(), pushDataset);
```
albertshau, line 134: it would be better if the engine didn't have to keep track of state like this (it means every implementation would have to do something similar).

It seems like you don't actually need to keep track though, unless I'm missing something? For pulling and cleanup, the source of truth is always if the table exists in BQ.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * Validate that this schema is supported in BigQuery.
+   *
+   * We don't support CDAP types ENUM, MAP and UNION in BigQuery.
+   *
+   * @param schema input schema
+   * @return boolean determining if this schema is supported in BQ.
+   */
+  protected boolean supportsSchema(Schema schema) {
```
albertshau: nit: annotate schema as nullable

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,330 @@
...
+    try {
+      String tableName = bqDataset.getBigQueryTableName();
+      TableId tableId = TableId.of(project, dataset, tableName);
+
+      if (!bigQuery.delete(tableId)) {
+        LOG.error("Unable to delete BigQuery table '{}' for stage '{}'",  tableName, stageName);
+      }
+    } catch (BigQueryException bqe) {
+      // Log exception and continue execution.
+      LOG.error("Exception when deleting BigQuery table", bqe);
```
albertshau: doesn't seem right to swallow this exception. The cleanup() method declares that it can throw a SQLEngineException, but these methods log errors and move on. 

Would be better to propagate this exception up and have the cleanup() method catch and log/handle appropriately.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineUtils.java
```diff
@@ -0,0 +1,58 @@
...
+  public static Long getNumRows(BigQuery bigQuery, String project, String dataset, String table) {
+    TableId tableId = TableId.of(project, dataset, table);
+    Table bgTable = bigQuery.getTable(tableId);
+
+    if (bgTable == null) {
+      throw new SQLEngineException(String.format("Table '%s' could not be found on dataset '%s' and project `%s`",
+                                                 table, dataset, project));
+    }
+
+    return bgTable.getNumRows().longValue();
```
albertshau, line 56: I'm assuming this is an efficient operation and not a scan over the entire table?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
```diff
@@ -0,0 +1,105 @@
...
+ * Create StructuredRecords from GenericRecords when pulling records from BigQuery.
+ *
+ * The following transformations are supported:
+ * BigQuery INT64 -> Integer (with Overflow check)
+ * BigQuery FLOAT64 -> Float (with Overflow check)
+ *
+ * Otherwise, field conversion gets delegated to the {@link BigQueryAvroToStructuredTransformer}
+ */
+public class SQLEngineAvroToStructuredTransformer extends BigQueryAvroToStructuredTransformer
+  implements Serializable {
```
albertshau, line 40: why does this need to be serializable?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -643,4 +651,23 @@ private static void match(String text, String propertyName, String pattern,
...
+   */
+  public static void deleteTemporaryDirectory(Configuration configuration, String dir) {
+    Path path = new Path(dir);
+    try {
+      FileSystem fs = path.getFileSystem(configuration);
+      if (fs.exists(path)) {
+        fs.delete(path, true);
+        LOG.debug("Deleted temporary directory '{}'", path);
+      }
+    } catch (IOException e) {
```
albertshau: better for the method to throw this exception and let the caller handle it, as they can give a more descriptive error message and may even decide not to log at error level.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: Oops, will fix.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: Will add a TODO to address this later.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/SQLEngineAvroToStructuredTransformer.java
fernst: This transformer gets sent from the driver into the workers to execute the read from GCS after an export. If not serializable, Spark will fail.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineUtils.java
fernst: This makes a network call to the BigQuery API to get the number of rows. Since this response is instant instead of taking longer for longer tables, this is probably equivalent to querying the INFORMATION_SCHEMA for the table.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: good catch, will adjust.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: Only if/when the job is deleted via the CLI or BQ API

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineUtils.java
fernst: This is an RPC call that queries the BQ tables API. This is probably querying the informaiton schema.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: Will add better logging here.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: Tweaked this.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: I have re-factored the error handling when executing the cleanup step, PTAL.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -0,0 +1,287 @@
...
+   * @param index the index to generate
+   * @return the table alias for the current index.
+   */
+  protected String buildTableAlias(int index) {
+
+    StringBuilder alias = new StringBuilder();
+
+    // Build all preceding characters in the alias, if needed.
+    if ((index / 26) > 0) {
+      alias.append(buildTableAlias((index / 26) - 1));
```
tivv: Can you make it a for loop and not recursion? You are recursively making stringbuilders and converting them to string and appending to other stringbuilders

fernst: Iterating means I have to "prepend" into a String, which is a N^2 operation as I have to copy the string.

I can re-factor this to not use the StringBuilder and build the string in-place.

fernst: Re-factored in order to avoid using StringBuilder in the recursive call.

tivv (commented):
Generally looks good for me.
Next things are unclear:
 * Do you need to quote field names?
 * Why do you override BigQuerySQLEngine.onRunFinish

fernst:
@tivv Answering your questions:

1) Field names in CDAP are compatible with BQ, at least from the UI perspective. I will add quotes for field names in a later PR as I think this would make the solution sturdier.
2) I don't currently do anything with this, but in the build we are providing for one of our customers for testing, we will add additional logging/output information here.

tivv (approved):
LGTM. Please also address comments from Albert

albertshau (commented):
BigQuerySQLEngine still has some unaddressed comments about some methods that don't need to exist (since they just call super), and a question about whether it really needs to keep track of datasets itself.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -0,0 +1,155 @@
...
+  private String datasetName;
+  private Schema schema;
+  private String project;
+  private String bqDataset;
+  private String bqTable;
+  private String gcsPath;
+  private Long numRows;
+  private final SQLEngineAvroToStructuredTransformer transformer = new SQLEngineAvroToStructuredTransformer();
```
albertshau: since this is used within the anonymous function returned by fromKeyValue(), Spark needs to serialize this entire class. It would be better to pull that Transform into its own class and move this variable to that class. That way BigQueryPullDataset doesn't need to be serializable, and this becomes more straightforward to understand.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
albertshau: See comment about the SQLEngineAvroToStructuredTransformer below.

Its confusing to make this transient and everything else non-final, as it suggests part of this class is used in the Spark executors and part is used on the driver.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * SQL Push Dataset implementation for BigQuery backed datasets.
+ */
+public class BigQueryPushDataset extends BigQueryOutputFormatProvider
+  implements SQLPushDataset<StructuredRecord, StructuredRecord, NullWritable>, BigQuerySQLDataset {
+
+  // Note that these properties are not final in order to support Spark's serialization.
```
albertshau: Does this class need to be serializable? I don't see anything here that would looks like Spark would need to serialize this.

albertshau (commented):
some other questions around serialization

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
fernst: I'll tweak the code so hopefully we won't need to serialize the whole class.

However, since this class extends from `BigQueryInputFormatProvider`, the methods to get the input format and configuration might also need to be serialized. Let me test and see how the class gets serialized.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
fernst: I'll try this and submit a change after testing.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
fernst: The `toKeyValue` has to be sent to the worker in order to write records to output. I'll try to adjust this so we can skip serializing the whole class.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: No particular reason at this time. I'll remove it and add it later if needed.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: When executing the cleanup method, we use this dataset to identify several things:

1) Which table in BQ represents this dataset.
2) Which temporary path in GCS is used as temporary storage for this dataset (if any). This is used for push/pull datasets.
3) Cancel the job that is currently being executed for this dataset if we get the order to shut-down the executor.

An alternative here is to re-write this to just delete the entire GCS path and all tables that start with the `runId` prefix on run finish.

Let me know your thoughts on this.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
fernst: Yep, that worked.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPullDataset.java
```diff
@@ -44,22 +40,14 @@
   implements SQLPullDataset<StructuredRecord, LongWritable, GenericData.Record>, BigQuerySQLDataset {
 
   // Note that these properties are not final in order to support Spark's serialization.
```
albertshau: can remove this comment now

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
albertshau: can remove this comment now

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/transform/PullTransform.java
```diff
@@ -0,0 +1,47 @@
...
+    this.schema = schema;
+  }
+
+  @Override
+  public void transform(KeyValue<LongWritable, GenericData.Record> input, Emitter<StructuredRecord> emitter)
+    throws Exception {
+    StructuredRecord transformed = transformer.transform(input.getValue(), schema);
+    emitter.emit(transformed);
+  }
+};
```
albertshau: remove extra semicolon

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -0,0 +1,351 @@
...
+    // Initialize BQ and GCS clients.
+    bigQuery = GCPUtils.getBigQuery(project, credentials);
+    storage = GCPUtils.getStorage(project, credentials);
+
+    String cmekKey = context.getRuntimeArguments().get(GCPUtils.CMEK_KEY);
+    configuration = BigQueryUtil.getBigQueryConfig(config.getServiceAccount(), config.getProject(),
+                                                   cmekKey, config.getServiceAccountType());
+    BigQuerySinkUtils.createResources(bigQuery, storage, dataset, bucket, config.getLocation(), cmekKey);
+  }
+  @Override
```
albertshau, line 117: nit: add newline between methods

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
albertshau: hm I see, let's keep this for now.

Is there a TTL being set on tables that are created? We also need to consider situations where the job is killed/dies without cleanup being run. Can do this in a follow-up PR if not there.

albertshau (approved):
lgtm, with a question about what happens if cleanup can't run because the job is killed or dies.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: We can add table TTL on table creation (using the table output format provider). I will add this in a later PR.

PR #661, opened 2021-05-25
CDAP-17907 migrate BigQuery Connector

seanzhougoogle:


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -35,21 +40,35 @@
 /**
  * Holds configuration required for configuring {@link BigQuerySource}.
  */
-public final class BigQuerySourceConfig extends BigQueryBaseConfig {
+public final class BigQuerySourceConfig extends PluginConfig {
```
fernst, line 43: Why this change? The reason BigQuery source config extends from BigQuery base config is so the same configuration class can be referenced in sources, sinks and BQ Pushdown.

fernst (dismissed):

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
seanzhougoogle: This is the major purpose for the migration, We need to extract connection config into BigQueryConnector config. The first step is to migrate BigQuery source, will migrate BigQueryTarget and probably BQ Pushdown later. I will send you some related design doc offline :)

seanzhougoogle: `BigQueryBaseConfig` itself is fine, but it inherits from `GCPConfig` which we need to decouple from because it contains connection properties.

Actually in a final state, we can have `BigQueryConnectorConfig` in `BigQueryBaseConfig` as a sub config and make `BigQueryBaseConfig` not inherit from `GCPConfig`. But for now we are only migrating BigQuerySource since `BigQueryBaseConfig` is also inherited by BigQuerySink , so in the first step , I have to decouple it.

Will need to do a refactoring when all BigQuery Source and Sink are migrated.

fernst: This will cause a conflict with the BQ Pushdown changes in https://github.com/data-integrations/google-cloud/pull/659/files#diff-1e91c596d76d44c6146342154a5f84c569e2d639ef56eb4020251d4d6fd0a538

seanzhougoogle: no worries, you can merge first, and I will resolve the conflicts. Only one method in the BigQuerySourceUtils is impacted. That method actually just need the bucket name, no need to pass in the whole config.

widgets/BigQueryTable-batchsource.json
```diff
@@ -4,6 +4,35 @@
...
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "false"
+          }
+        },
+        {
+          "widget-type": "textbox",
```
elfenheart: the widget-type for this should be `connection-select`

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    Table result = bigQuery.getTable(TableId.of(dataset, table));
+    if (result == null) {
+      throw new IllegalArgumentException(String.format("Cannot find such table: %s.%s.", dataset, table));
+    }
+    return BrowseDetail.builder()
+      .addEntity(BrowseEntity.builder(table, "/" + dataset + "/" + table, ENTITY_TYPE_TABLE).canSample(true).build())
+      .build();
+  }
+
+  private String[] parsePath(String originalPath) {
```
albertshau: It would be cleaner to have a BigQueryPath class that contains the parsing and validation logic (may also need to validate the character set?). This would remove some of the duplicate code in this class and also make it easily unit testable.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    if (table.isEmpty()) {
+      throw new IllegalArgumentException(String.format("Path %s is invalid. Table name is empty.", path));
+    }
+    return getTable(bigQuery, dataset, table);
+  }
+
+  private BrowseDetail getTable(BigQuery bigQuery, String dataset, String table) {
+    Table result = bigQuery.getTable(TableId.of(dataset, table));
+    if (result == null) {
+      throw new IllegalArgumentException(String.format("Cannot find such table: %s.%s.", dataset, table));
```
albertshau: nit: remove 'such' 

Cannot find table %s.%s

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    String[] parts = path.split("/");
+    // path should contains at most two part : dataset and table
+    if (parts.length > 2) {
+      throw new IllegalArgumentException(
+        String.format("Path %s is invalid. Path should at most contain two parts.", originalPath));
+    }
+    return parts;
+  }
+
+  private BrowseDetail listTables(BigQuery bigQuery, String dataset, Integer limit) {
```
albertshau: annotate limit as nullable

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    int countLimit = limit == null ? Integer.MAX_VALUE : limit;
+    int count = 0;
+    BrowseDetail.Builder browseDetailBuilder = BrowseDetail.builder();
+    DatasetId datasetId = DatasetId.of(dataset);
+    Page<Table> tablePage = bigQuery.listTables(datasetId);
+    String parentPath = "/" + dataset + "/";
+    for (Table table : tablePage.iterateAll()) {
+      String name = table.getTableId().getTable();
+      browseDetailBuilder
```
albertshau: does Table have a type? I seem to remember the listing can return tables, views, and models. Models cannot be sampled (maybe they shouldn't even be listed)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  private BigQuery getBigQuery() throws IOException {
+    GoogleCredentials credentials = null;
+    //validate service account
+    if (config.isServiceAccountJson() || config.getServiceAccountFilePath() != null) {
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+    }
+    ;
```
albertshau: remove

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    String query = String.format("SELECT * FROM `%s.%s.%s` LIMIT %d", config.getDatasetProject(), dataset, table,
+      limit);
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
+    JobId jobId = JobId.of(UUID.randomUUID().toString());
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
+    // Wait for the job to finish
+    try {
+      queryJob = queryJob.waitFor();
+    } catch (InterruptedException e) {
+      throw new IOException("Query job interrupted.", e);
```
albertshau: should include the job id in the error message

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    // check for errors
+    if (queryJob == null) {
+      throw new IOException("Job no longer exists.");
```
albertshau: include the job id in the error message

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    return GCPUtils.getBigQuery(config.getDatasetProject(), credentials);
+  }
+
+  private List<StructuredRecord> getData(BigQuery bigQuery, String dataset, String table, int limit)
+    throws IOException {
+    String query = String.format("SELECT * FROM `%s.%s.%s` LIMIT %d", config.getDatasetProject(), dataset, table,
+      limit);
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
+    JobId jobId = JobId.of(UUID.randomUUID().toString());
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
```
albertshau, line 220: I believe this means the job is always in the default location (US). Have you tested on tables in different locations?

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,370 @@
...
+    } catch (InterruptedException e) {
+      throw new IOException("Query results interrupted.", e);
+    }
+    List<StructuredRecord> samples = new ArrayList<>();
+
+    com.google.cloud.bigquery.Schema schema = result.getSchema();
+    Schema cdapSchema = BigQueryUtil.getTableSchema(schema, null);
+
+    FieldList fields = schema.getFields();
+    for (FieldValueList fieldValues : result.iterateAll()) {
```
albertshau: would be good to pull out this logic into a separate class that is only responsible for converting FieldValueList into a StructuredRecord. Would also make it easier to unit test.

albertshau (commented):
Please also add some unit tests. Some logic can be pulled out into different classes to make things more testable, and we should also be able to test directly against BigQuery if service account credentials are given, similar to what we do in the BigQuery delta target plugin.

widgets/BigQueryTable-batchsource.json
seanzhougoogle: done:)

seanzhougoogle:
> Please also add some unit tests. Some logic can be pulled out into different classes to make things more testable, and we should also be able to test directly against BigQuery if service account credentials are given, similar to what we do in the BigQuery delta target plugin.

Thanks for the suggestion. I was just migrating the BigQueryHandler from Wrangler to BigQueryConnector and didn't notice the code structure and testability. Done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done:)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: table doesn't have a type here. this is returned from `bigquery.lisTables` and there is another method for list models : `bigquery.listModels`

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: good point. It's the original logic from Wrangler.
And I just tested , it can query tables in europe.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
+/*
+ *
```
yaojiefeng: nit - extra line

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+    String table = path.getTable();
+    if (table == null) {
+      throw new IllegalArgumentException("Path should contain both dataset and table name.");
+    }
+    return getTableData(getBigQuery(), dataset, table, sampleRequest.getLimit());
+  }
+
+  @Override
+  public void test(FailureCollector failureCollector) throws ValidationException {
+    // Should not happen
```
yaojiefeng: You can remove this if condition, it should fail much earlier when it instantiates this plugin if config is not found

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -0,0 +1,101 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * BigQuery Path that parses the path in the request of connection service
+ */
+public class BigQueryPath {
+  private String dataset;
+  private String table;
+  private static final int NAME_MAX_LENGTH = 1024;
```
yaojiefeng, line 35: Do you know where does this max length come from?

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -0,0 +1,101 @@
...
+    parsePath(path);
+  }
+
+  private void parsePath(String path) {
+    if (path == null) {
+      throw new IllegalArgumentException("Path should not be null.");
+    }
+
+    //remove heading "/" if exists
+    if (path.startsWith("/")) {
```
yaojiefeng, line 48: Can use `"path.replaceAll("^/+", "")` to remove leading "/" and `path.replaceAll("/+$", "")` to remove trailing "/"

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -0,0 +1,101 @@
...
+    // both "" and "/" are taken as root path
+    if (path.isEmpty()) {
+      return;
+    }
+
+    String[] parts = path.split("/", -1);
+
+    // path should contain at most two part : dataset and table
+    // if path ends with / , it's allowed and the last part will be empty string
+    if (parts.length == 3 && !parts[2].isEmpty() || parts.length > 3) {
```
yaojiefeng: You don't need this length == 3 check if you trim the "/" using the code above, the length can also be > 3, if the path is like "//a//b//"

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -0,0 +1,101 @@
...
+      throw new IllegalArgumentException("Path should at most contain two parts.");
+    }
+
+    dataset = parts[0];
+    validateName("Dataset" , dataset);
+
+    // if path ends with "/" , e.g. "dataset/" we allow it
+    if (parts.length == 2 && !path.endsWith("/") || parts.length > 2) {
```
yaojiefeng: Don't need this if you trim the "/"

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
yaojiefeng: `listTables` does not return the actual type where I feel it should. Here I am pretty sure view is included in the table list, can you open a JIRA to investigate if this listing needs to be improved to identify table, view, and model?

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+    if (result == null) {
+      throw new IllegalArgumentException(String.format("Cannot find table: %s.%s.", dataset, table));
+    }
+    return BrowseDetail.builder()
+      .addEntity(BrowseEntity.builder(table, "/" + dataset + "/" + table, ENTITY_TYPE_TABLE).canSample(true).build())
+      .setTotalCount(1).build();
+  }
+
+  private BrowseDetail listTables(BigQuery bigQuery, String dataset, @Nullable Integer limit) {
+    int countLimit = limit == null ? Integer.MAX_VALUE : limit;
```
yaojiefeng: limit == null || limit <=0

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+      }
+      throw e;
+    }
+    String parentPath = "/" + dataset + "/";
+    for (Table table : tablePage.iterateAll()) {
+      String name = table.getTableId().getTable();
+      browseDetailBuilder
+        .addEntity(BrowseEntity.builder(name, parentPath + name, ENTITY_TYPE_TABLE).canSample(true).build());
+      count++;
+      if (count == countLimit) {
```
yaojiefeng: Should not break here, the count should include the total count of the tables, you should stop adding the table to the browse detail if count hits the limit, but you need to loop till the end to get the exact total count.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+    int countLimit = limit == null ? Integer.MAX_VALUE : limit;
+    int count = 0;
+    BrowseDetail.Builder browseDetailBuilder = BrowseDetail.builder();
+    for (Dataset dataset : datasetPage.iterateAll()) {
+      String name = dataset.getDatasetId().getDataset();
+      browseDetailBuilder
+        .addEntity(BrowseEntity.builder(name, "/" + name, ENTITY_TYPE_DATASET).canBrowse(true).build());
+      count++;
+      if (count == countLimit) {
+        break;
```
yaojiefeng: same here, should not break

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+    }
+    return GCPUtils.getBigQuery(config.getDatasetProject(), credentials);
+  }
+
+  private List<StructuredRecord> getTableData(BigQuery bigQuery, String dataset, String table, int limit)
+    throws IOException {
+    String query =
+      String.format("SELECT * FROM `%s.%s.%s` LIMIT %d", config.getDatasetProject(), dataset, table, limit);
```
yaojiefeng, line 216: We have received multiple support tickets on this since starting a query will start a job on bigquery which requires extra permission. Can you check if `listTableData` is easier to use?

yaojiefeng: On a second thought, this `listTableData` might need more time to investigate on what permission and which location it will run, let's keep this and open a JIRA to track if we can get rid of the extra permission.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorConfig.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @Name(NAME_SERVICE_ACCOUNT_JSON)
+  @Description("Content of the service account file.")
+  @Macro
+  @Nullable
+  protected String serviceAccountJson;
+
+
+  public String getDatasetProject() {
+    if (datasetProject == null || AUTO_DETECT.equalsIgnoreCase(datasetProject)) {
```
yaojiefeng, line 79: don't need the null check here, `AUTO_DETECT.equals` should evaluate to false if this is null

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorConfig.java
```diff
@@ -0,0 +1,174 @@
...
+import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
+import io.cdap.cdap.api.plugin.PluginConfig;
+
+import javax.annotation.Nullable;
+
+/**
+ * Plugin config for BiqueryConnector
+ */
+public class BigQueryConnectorConfig extends PluginConfig {
```
yaojiefeng, line 31: Does this class mostly come from `GCPConfig`? Can you open a JIRA as an improvement to track if we can get rid of the duplicate code?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -153,6 +248,8 @@ public Type getSourceTableType() {
   }
 
 
+
```
yaojiefeng, line 251: nit - remove these two lines

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
yaojiefeng: This seems to add back a lot of configs that got refactored to a base config. Can you rebase against latest develop to see if we can do better on the grouping of the config?

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+
+  @Override
+  public ConnectorSpec generateSpec(ConnectorSpecRequest connectorSpecRequest) {
+    BigQueryPath path = new BigQueryPath(connectorSpecRequest.getPath());
+    String dataset = path.getDataset();
+    String table = path.getTable();
+    if (table == null) {
+      throw new IllegalArgumentException("Path should contain both dataset and table name.");
```
yaojiefeng: I don't think you need to fail here, it is fairly possible to generate spec on a partial path.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+      result = queryJob.getQueryResults();
+    } catch (InterruptedException e) {
+      throw new IOException("Query results interrupted.", e);
+    }
+    return BigQueryDataParser.parse(result);
+  }
+
+
+  @Override
+  public ConnectorSpec generateSpec(ConnectorSpecRequest connectorSpecRequest) {
```
yaojiefeng: Please do a `mvn clean install` on this branch in this pr https://github.com/cdapio/cdap/pull/13430, we added some changes to generateSpec, now it will need to return schema and related plugins

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,250 @@
...
+    BigQueryPath path = new BigQueryPath(connectorSpecRequest.getPath());
+    String dataset = path.getDataset();
+    String table = path.getTable();
+    if (table == null) {
+      throw new IllegalArgumentException("Path should contain both dataset and table name.");
+    }
+    return ConnectorSpec.builder().addProperty(BigQuerySourceConfig.NAME_DATASET, dataset)
```
yaojiefeng: here you need to check if a table is a view, the source will need this property like here: https://github.com/data-integrations/wrangler/commit/c18cf5e8f40994e4875125f3477202fceedc4295#diff-036a61e7dab7d973c262043ecb842472c2f61b90f8bfbcc15f4f9402417f7920R289

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Map;
+
+import static io.cdap.plugin.gcp.bigquery.connector.BigQueryConnectorConfig.SERVICE_ACCOUNT_JSON;
```
yaojiefeng: don't use static import

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -0,0 +1,199 @@
...
+ * service.account.file-- the path to the service account key file
+ */
+public class BigQueryConnectorTest {
+  private static String serviceAccountKey;
+  private static String project;
+  private static String datasetProject;
+  private static String dataset;
+  private static String table;
+  private static String serviceAccountFilePath;
+
```
yaojiefeng, line 60: nit - remove extra line

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * BigQuery Connector integration test. This test will only be run when below property is provided:
+ * project.id -- the name of the project where temporary table or staging bucket may be created. It will default to
+ * active google project if you have google cloud client installed.
+ * dataset.project -- optional, the name of the project where the dataset is
+ * dataset.name -- the name of the dataset
+ * table.name -- the name of the table
+ * service.account.file-- the path to the service account key file
+ */
+public class BigQueryConnectorTest {
```
yaojiefeng, line 53: This class will need some environment property to be passed in, is this meant for integration tests or on what condition can a build run these tests?

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+public class BigQueryPathTest {
+
```
yaojiefeng, line 24: remove extra line

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
```diff
@@ -0,0 +1,108 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.connector;
+
+import joptsimple.internal.Strings;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
```
yaojiefeng: Don't use static import, can just do `Assert.assertEquals`

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    //table exceeds max length
+    assertThrows("Table is invalid, it should contain at most 1024 characters.", IllegalArgumentException.class,
+      () -> new BigQueryPath("/b/" + Strings.repeat('a', 1025)));
+
+    //table contains invalid character
+    assertThrows("Dataset is invalid, it should contain only letters, numbers, and underscores.",
+      IllegalArgumentException.class,
+      () -> new BigQueryPath("/a/%"));
+
```
yaojiefeng: nit - remove extra lines

widgets/BigQueryTable-batchsource.json
```diff
@@ -4,6 +4,35 @@
   },
   "display-name": "BigQuery",
   "configuration-groups": [
+    {
+      "label": "Connection",
+      "properties": [
```
yaojiefeng, line 9: it is good to put all the connection related configs under this label, so user know exactly what fields get hidden if they turn on connection

src/test/java/io/cdap/plugin/gcp/bigquery/util/SimpleFailureCollector.java
```diff
@@ -0,0 +1,47 @@
...
+
+import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.cdap.etl.api.validation.ValidationException;
+import io.cdap.cdap.etl.api.validation.ValidationFailure;
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class SimpleFailureCollector implements FailureCollector {
```
yaojiefeng: I don't think this is needed, can you remove it?

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
yaojiefeng: This applies to all copyright files in this pr

yaojiefeng (commented):
Have some comments, please reinstall cdap based on https://github.com/cdapio/cdap/pull/13430, we have changed the `ConnectorSpec` to return more things.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
seanzhougoogle: the latest commit is based on the latest develop branch. As I mentioned, BigQueryBaseConfig is fine. The only issue is it inherits from GCPConfig. We can still inherit from BigQueryBaseConfig in the future and put BigQueryConnectorConfig in it and make it not inherit from GCPConfig.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: Ah, good catch :)

done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/test/java/io/cdap/plugin/gcp/bigquery/util/SimpleFailureCollector.java
seanzhougoogle: good catch. Done :)

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
seanzhougoogle: done :)

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
seanzhougoogle: sorry, may I know the rationale ? I have multiple assertEquals , using static import can make the code leaner here. Let's chat offline :)

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: actually I don't want to remove all leading or tailing "/"

I want to take "//" as invalid one.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: same as previous comment, I want to take "//a/b//" as invalid path

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: It's the requirement of BigQuery. If you try to create a dataset/table with name exceeds 1024, you will get an error telling you the maximum length is 1024.

widgets/BigQueryTable-batchsource.json
seanzhougoogle: Good point :) done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done : CDAP-18028

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: looks like it's available via table.getDefinition().getType()

let's discuss of line to check what exactly should be returned for browse dataset.

seanzhougoogle: I've tested that listTables won't get models in the result

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
seanzhougoogle: discussed offline, we will leave the code as it is for now. once we migrate all BQ related source/sink to connection , we can do what I suggested above.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: as discussed offline, will put type information in BrowseEntity.type

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: discussed offline, updated java doc to reflect it.

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPathTest.java
seanzhougoogle: discussed offline, it's a coding convention of CDAP. Done :)

seanzhougoogle:
> Have some comments, please reinstall cdap based on [cdapio/cdap#13430](https://github.com/cdapio/cdap/pull/13430), we have changed the `ConnectorSpec` to return more things.

done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -62,8 +69,10 @@
   "tables. BigQuery is Google's serverless, highly scalable, enterprise data warehouse.")
 public final class BigQueryConnector implements DirectConnector {
   public static final String NAME = "BigQuery";
-  public static final String ENTITY_TYPE_DATASET = "dataset";
-  public static final String ENTITY_TYPE_TABLE = "table";
+  public static final String ENTITY_TYPE_DATASET = "DATASET";
+  public static final String ENTITY_TYPE_TABLE = TableDefinition.Type.TABLE.name();
+  public static final String ENTITY_TYPE_VIEW = TableDefinition.Type.VIEW.name();
```
yaojiefeng: I don't see where this is used. I remember we plan to add this type to the `BrowseEntityProperty`?

yaojiefeng: I see this type is already added in the `getTableDetail` method, we can just remove these

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -237,14 +245,25 @@ private BigQuery getBigQuery() throws IOException {
...
-    if (table == null) {
+    String datasetName = path.getDataset();
+    String tableName = path.getTable();
+    if (tableName == null) {
       throw new IllegalArgumentException("Path should contain both dataset and table name.");
     }
-    return ConnectorSpec.builder().addProperty(BigQuerySourceConfig.NAME_DATASET, dataset)
-      .addProperty(BigQuerySourceConfig.NAME_TABLE, table).build();
+
+    Map<String, String> properties = new HashMap<>();
```
yaojiefeng: Can you also add the properties of useConnection flag and connection here? This is to ensure UI have all the properties needed when going to the source, you can find the example here: https://github.com/cdapio/hydrator-plugins/compare/file-connector#diff-52ec8fe889fbb4fe210de39c3f35ffbe41f27ba518b80a9d0a60f4d43906d554R139-R140

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
yaojiefeng: This is not addressed yet

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
yaojiefeng: You can do `path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path` before you split so you don't need this complex condition

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * BigQuery Connector Plugin
+ */
+@Plugin(type = Connector.PLUGIN_TYPE)
+@Name(BigQueryConnector.NAME)
```
yaojiefeng, line 68: add a category `@Category("Google Cloud Platform")`, so UI can group based on these categories

yaojiefeng (commented):
Some minor comments

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: Do we have somewhere define all the catetory values ? So that I can reference the value instead of hardcoding the value here which is error prone.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: Here I'd like to expose all the possible `type` values because the `type` in BrowseEntityProperty is a string.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: added the hardcoded category value for now.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: The major reason I do this here without first removing the ending slash is because I want to parse and validate the dataset in the same way even for some special case. I added more comments to make it clear.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
yaojiefeng: If they are not used just remove them since this is dependent on the client, if bigquery decides to add more types in the future, it won't be reflected here, I think `definition.getType()` is enough in `getTableDetail`

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
yaojiefeng: for "//", the `path.isEmpty()` at line 53 should already return, so you can still get rid of this complex condition.

yaojiefeng (approved):
Just 2 minor comments, please fix and merge.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: Good catch. I was confused. done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: makes sense.  done :)

PR #664, opened 2021-06-01
Remove application/csv. text/csv is the accepted standard by IANA

omarismail94:
Remove `application/csv` from supported types as it is not a Content-Types from the ones officially recognized: https://www.iana.org/assignments/media-types/media-types.xhtml


omarismail94:
R: @albertshau

albertshau:
This is a backwards incompatible change. Is there any harm in leaving application/csv?

PR #665, opened 2021-06-01
Implemented support for Advanced joins in BigQuery Pushdown

fernst:
Added further tests and code improvements.

Refactored SQL Join builder to take Join Definition as input and expose simple method to get join SQL.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
@@ -18,17 +18,14 @@ package io.cdap.plugin.gcp.bigquery.sqlengine;
 
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.Job;
-import com.google.cloud.bigquery.JobConfiguration;
 import com.google.cloud.bigquery.JobId;
 import com.google.cloud.bigquery.JobInfo;
 import com.google.cloud.bigquery.QueryJobConfiguration;
-import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.engine.sql.SQLEngineException;
 import io.cdap.cdap.etl.api.engine.sql.dataset.SQLDataset;
 import io.cdap.cdap.etl.api.engine.sql.request.SQLJoinRequest;
-import io.cdap.cdap.etl.api.join.JoinCondition;
 import io.cdap.cdap.etl.api.join.JoinDefinition;
 import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQuerySQLBuilder;
 import org.slf4j.Logger;
@@ -52,7 +49,7 @@ public class BigQueryJoinDataset implements SQLDataset, BigQuerySQLDataset {
   private final String bqDataset;
   private final String bqTable;
   private final String jobId;
-  private final BigQuerySQLBuilder helper;
+  private final BigQuerySQLBuilder queryBuilder;
   private Long numRows;
 
   private BigQueryJoinDataset(String datasetName,
@@ -72,7 +69,7 @@ public class BigQueryJoinDataset implements SQLDataset, BigQuerySQLDataset {
     this.bqDataset = bqDataset;
     this.bqTable = bqTable;
     this.jobId = jobId;
-    this.helper = new BigQuerySQLBuilder(this.project, this.bqDataset, stageToTableNameMap);
+    this.queryBuilder = new BigQuerySQLBuilder(this.joinDefinition, this.project, this.bqDataset, stageToTableNameMap);
   }
 
   public static BigQueryJoinDataset getInstance(SQLJoinRequest joinRequest,
@@ -105,7 +102,7 @@ public class BigQueryJoinDataset implements SQLDataset, BigQuerySQLDataset {
   private void executeJoin() {
     TableId destinationTable = TableId.of(project, bqDataset, bqTable);
 
-    String query = getQuery(joinDefinition);
+    String query = queryBuilder.getQuery();
     LOG.info("Creating table `{}` using job: {} with SQL statement: {}", bqTable, jobId, query);
 
     // Run BigQuery job with generated SQL statement, store results in a new table, and set priority to BATCH
@@ -137,15 +134,6 @@ public class BigQueryJoinDataset implements SQLDataset, BigQuerySQLDataset {
     }
 
     LOG.info("Created BigQuery table `{}` using Job: {}", bqTable, jobId);
-  }
-
-  protected String getQuery(JoinDefinition joinDefinition) {
-    if (joinDefinition.getCondition().getOp() == JoinCondition.Op.KEY_EQUALITY) {
-      return helper.getFieldEqualityQuery(joinDefinition);
-    } else {
-      //TODO: implement
-      throw new SQLEngineException("Advanced joins not currently supported");
-    }
   }
 
   @Override
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
@@ -22,13 +22,13 @@ import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.storage.Storage;
+import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.RuntimeContext;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
-import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.engine.sql.BatchSQLEngine;
 import io.cdap.cdap.etl.api.engine.sql.SQLEngineException;
 import io.cdap.cdap.etl.api.engine.sql.dataset.SQLDataset;
@@ -39,6 +39,7 @@ import io.cdap.cdap.etl.api.engine.sql.request.SQLJoinRequest;
 import io.cdap.cdap.etl.api.engine.sql.request.SQLPullRequest;
 import io.cdap.cdap.etl.api.engine.sql.request.SQLPushRequest;
 import io.cdap.cdap.etl.api.join.JoinCondition;
+import io.cdap.cdap.etl.api.join.JoinDefinition;
 import io.cdap.cdap.etl.api.join.JoinStage;
 import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
@@ -51,7 +52,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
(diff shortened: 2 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -171,20 +173,33 @@ public boolean exists(String datasetName) throws SQLEngineException {
...
   @Override
   public boolean canJoin(SQLJoinDefinition sqlJoinDefinition) {
-    // TODO: remove when expression joins are supported.
-    if (sqlJoinDefinition.getJoinDefinition().getCondition().getOp() != JoinCondition.Op.KEY_EQUALITY) {
-      return false;
-    }
 
-    //Ensure none of the schemas contains unsupported types.
-    for (JoinStage inputStage : sqlJoinDefinition.getJoinDefinition().getStages()) {
+    List<String> issues = new LinkedList<>();
```
tivv: How is this issues list used? I see you add data to it, but not retrieve. Also please use ArrayList. LinkedList makes sense only in very edge cases. In most cases ArrayList is faster and use less memory.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -72,7 +69,7 @@ private BigQueryJoinDataset(String datasetName,
     this.bqDataset = bqDataset;
     this.bqTable = bqTable;
     this.jobId = jobId;
-    this.helper = new BigQuerySQLBuilder(this.project, this.bqDataset, stageToTableNameMap);
+    this.helper = new BigQuerySQLBuilder(this.joinDefinition, this.project, this.bqDataset, stageToTableNameMap);
```
tivv: please name this `queryBuilder` or similar. It's better to have specific names

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -171,20 +173,33 @@ public boolean exists(String datasetName) throws SQLEngineException {
...
+
+    JoinDefinition joinDefinition = sqlJoinDefinition.getJoinDefinition();
+
+    // Ensure none of the input schemas contains unsupported types or invalid stage names.
+    for (JoinStage inputStage : joinDefinition.getStages()) {
       if (!supportsSchema(inputStage.getSchema())) {
-        return false;
+        issues.add(String.format("Input schema from stage '%s' has one of more unsupported field types.",
```
albertshau: of -> or

albertshau: it would be more useful if the message contained which field is unsupported

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -171,20 +173,33 @@ public boolean exists(String datasetName) throws SQLEngineException {
...
+                                 inputStage.getStageName()));
       }
     }
 
-    if (!supportsSchema(sqlJoinDefinition.getJoinDefinition().getOutputSchema())) {
-      return false;
+    // Ensure the output schema doesn't contain unsupported types
+    if (!supportsSchema(joinDefinition.getOutputSchema())) {
+      issues.add("Output Schema has one of more unsupported field types.");
```
albertshau: similar comment here, message should include which field(s) is unsupported

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -349,4 +365,18 @@ protected boolean supportsSchema(@Nullable Schema schema) {
...
 
+  /**
+   * Ensure the Stage name is valid for execution in BQ pushdown.
+   *
+   * Due to differences in character escaping rules in Spark and BigQuery, identifiers that are accepted in Spark
+   * might not be valid in BigQuery. Due to this limitation, we don't support stage names or aliases containing
+   * backslash \ or backtick ` characters at this time.
+   *
+   * @param identifier stage name or alias to validate
+   * @return wether this stage name is valid for BQ Pushdown.
```
albertshau: typo: whether

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -24,33 +24,90 @@
...
+ * <p>
+ * Many methods in this class have proteced visibility for the purposes of testing.
  */
 public class BigQuerySQLBuilder {
+
+  /**
+   * Allows us to find a table name in the user supplied advanced SQL Join expression.
+   * Explanation of groups:
+   * 1. Matches on 3 possibilities:
+   *   a. The beggining of the string,
```
albertshau: typo: beginning

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -24,33 +24,90 @@
...
+   *   b. A non-word character (like a parenthesys or symbol)
```
albertshau: typo: parantheses

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -24,33 +24,90 @@
...
+   * <p>
+   * This can match expressions such as:
+   * table.column
+   * `table`.column
+   * table .column
+   * `table` .column
+   * table    .  column
+   * `table`    .  column
+   */
+  private static final String TABLE_NAME_PATTERN = "(^|\\W|^\\.\\s*)(`?)(%s)(`?)(\\s*)(\\.)";
```
tivv: `^\\.\\s*` does not mean "not a dot". It means start of line followed by dot. "^" would mean not only inside square brackets.
Please add separate unit tests to ensure the pattern does what's expected

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -106,28 +189,127 @@ protected void appendFieldEqualityClause(JoinDefinition joinDefinition) {
...
+
+    String replacement1 = String.format(TABLE_NAME_PATTERN, Pattern.quote(leftIdentifier));
+    String replacement2 = String.format(TABLE_NAME_PATTERN, Pattern.quote(rightIdentifier));
+    String replacement3 = String.format(TABLE_NAME_PATTERN, Pattern.quote(temp));
+
+
+    // Replace the left idenfier with the left alias, and the right identifier with the right alias.
+    // Note that Left is transformer to the temporary placeholder before flipping to the right value.
+    String replaced = expression
+      .replaceAll(replacement1, "$1" + Matcher.quoteReplacement(temp) + "$5$6")
```
tivv: nit: consider using named groups, it would be a bit more readable. Also you can join groups 5 and 6.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
```diff
@@ -106,28 +189,127 @@ protected void appendFieldEqualityClause(JoinDefinition joinDefinition) {
...
+   * Executes replace operation to replace table names in the supplied string for BigQuery-safe table aliases.
+   *
+   * @param expression       expression to manipulate
+   * @param leftIdentifier   Left Stage name
+   * @param leftReplacement  Left Stage replacement alias
+   * @param rightIdentifier  Right Stage Name
+   * @param rightReplacement Right Stage replacement alias
+   */
+  @VisibleForTesting
+  protected void appendOnExpressionClause(String expression,
```
albertshau: this seems overly complex, it's hard to follow what is happening, examples would help. In what situations do we need to do this replacement vs just directly using the expression that the user provided? Is it when an invalid table name is being used?

It seems more straightforward and understandable to always use the user provided input as-is and fail if it happens to be invalid (assuming the query gets logged)

albertshau: this is also the behavior in the Spark implementation, it just uses the user given input as-is.

fernst: My idea with the re-write was to ensure no conditions can cause a syntax error in BQ (as Spark and BQ have slightly different syntax rules).

I agree that, for simplicity, it would be easier to just paste the input as-is, but I wouldn't like for the user to run into a condition where preview works but then BQ fails.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: Will fix.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: Will fix.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Will fix.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
fernst: Good point, let me rename this.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: Oof, good catch.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Oh, good catch, I must have deleted the block where I print out a warning. Let me add it again.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Take a look now. If issues are found for this stage, I will print one message that mentions the stage and then all error messages concatenated.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: But, to your point, SparkSQL will be as picky as BQ is will something like a non-ascii dataset/table name..

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: Will add further regex tests

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: After some further testing, I realized your point is pretty valid. Since I'm already checking that stage names and stage aliases do not contain the escape characters for spark ( ` ) and BigQuery ( \ ), supplying the raw SQL without modifications should not cause any problems in BigQuery that were not present in Spark.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/builder/BigQuerySQLBuilder.java
fernst: I have removed the regex based on Albert's comments.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: If you look now, you'll see the issues are used to build a Logger Warning message, which includes the stage name.

fernst: I will add logic to list all invalid fields.

albertshau (approved):

tivv (approved):

PR #667, opened 2021-06-02
Restuctured code for easier maintainability and legibility.

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * Utility Class for the BigQuery SQL Engine implementation.
+ */
+public class BigQuerySQLEngineUtils {
```
tivv, line 36: If this class has only static methods, add final and a private constructor to ensure it's not instantiated.

tivv: Also as soon as this is public utility class, add javadocs to methods / constants

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public static final String GCS_PATH_FORMAT = BigQuerySinkUtils.GS_PATH_FORMAT + "/%s";
+  public static final String BQ_TABLE_NAME_FORMAT = "%s_%s";
+
+  public static String getGCSPath(String bucket, String runId, String tableId) {
+    return String.format(GCS_PATH_FORMAT, bucket, runId, tableId);
+  }
+
+  public static String getIdentifier() {
```
tivv: If it's table name identifier, make method name less generic. `getIdentifier` can mean pretty much anything.

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
fernst: It can also be a runID, that's why I used `identifier`.

PR #670, opened 2021-06-05
fix dataset project settings

seanzhougoogle:
the project set in the bigquery client should be the default project to list dataset and run BQ jobs

so the service account specified should have "BigQuery Data Viewer" role on the "dataset project" and "BigQuery Job User" role on the "project"

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
@@ -83,12 +83,13 @@ public final class BigQueryConnector implements DirectConnector {
   @Override
   public List<StructuredRecord> sample(ConnectorContext context, SampleRequest sampleRequest) throws IOException {
     BigQueryPath path = new BigQueryPath(sampleRequest.getPath());
-    String dataset = path.getDataset();
     String table = path.getTable();
     if (table == null) {
       throw new IllegalArgumentException("Path should contain both dataset and table name.");
     }
-    return getTableData(getBigQuery(), dataset, table, sampleRequest.getLimit());
+    String dataset = path.getDataset();
+    return getTableData(getBigQuery(config.getProject()), config.getDatasetProject(), dataset, table,
+      sampleRequest.getLimit());
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -177,7 +188,8 @@ private void testBrowse(BigQueryConnector connector) throws IOException {
     Assert.assertEquals(1, detail.getTotalCount());
     Assert.assertEquals(1, detail.getEntities().size());
     for (BrowseEntity entity : detail.getEntities()) {
-      Assert.assertEquals("TABLE", entity.getType());
+      System.out.println(entity.getType() + " : " + entity.getName());
```
yaojiefeng, line 201: nit - remove this

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -167,7 +177,8 @@ private void testBrowse(BigQueryConnector connector) throws IOException {
     Assert.assertTrue(detail.getTotalCount() > 0);
     Assert.assertTrue(detail.getEntities().size() > 0);
     for (BrowseEntity entity : detail.getEntities()) {
-      Assert.assertEquals("TABLE", entity.getType());
+      System.out.println(entity.getType() + " : " + entity.getName());
```
yaojiefeng, line 189: nit - remove this

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -157,6 +166,7 @@ private void testBrowse(BigQueryConnector connector) throws IOException {
     Assert.assertTrue(detail.getTotalCount() > 0);
     Assert.assertTrue(detail.getEntities().size() > 0);
     for (BrowseEntity entity : detail.getEntities()) {
+      System.out.println(entity.getType() + " : " + entity.getName());
```
yaojiefeng, line 177: nit - remove this

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -129,39 +130,39 @@ public void test(FailureCollector failureCollector) throws ValidationException {
   public BrowseDetail browse(BrowseRequest browseRequest) throws IOException {
     BigQueryPath path = new BigQueryPath(browseRequest.getPath());
 
-    BigQuery bigQuery = getBigQuery();
-    if (path.isRoot()) {
+    String dataset = path.getDataset();
+    if (dataset == null) {
       // browse project to list all datasets
-      return listDatasets(bigQuery, browseRequest.getLimit());
+      return listDatasets(getBigQuery(config.getDatasetProject()), browseRequest.getLimit());
```
yaojiefeng, line 138: A little confused here, there are couple places `getBigQuery` is called, sometimes it uses project, sometimes it uses dataset project, what is the difference and how do we decide which project to use?

yaojiefeng (commented):
Some comments

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: as commented in getBigQuery method, it's where BQ list dataset and where the BQ job is run. 

So to list dataset, we should use `dataset project`. List tables already have dataset set in the tableId, so for other call we just use `project`

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
seanzhougoogle: done

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
seanzhougoogle: done

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
seanzhougoogle: also add a javadoc for that method to make it more clear

yaojiefeng (approved):

PR #674, opened 2021-06-09
Improved logging and fixed issue with nullable fields for BQ Pushdown

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
```diff
@@ -87,6 +92,7 @@ public static Long getNumRows(BigQuery bigQuery, String project, String dataset,
                                                  table, dataset, project));
     }
 
+    LOG.info("Table {} contains {} records", table, bgTable.getNumRows().longValue());
```
tivv: Please surround with `if (LOG.isInfoEnabled())` `getNumRows` is definitely not free

tivv (approved):
LGTM with minor comment

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -218,18 +224,24 @@ public SQLDataset join(SQLJoinRequest sqlJoinRequest) throws SQLEngineException
...
     datasets.put(sqlJoinRequest.getDatasetName(), joinDataset);
 
     return joinDataset;
   }
 
   @Override
   public void cleanup(String datasetName) throws SQLEngineException {
+
     BigQuerySQLDataset bqDataset = datasets.get(datasetName);
+    //if (bqDataset == null || config.skipDelete()) {
```
albertshau: unintended change?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
```diff
@@ -79,6 +83,7 @@ public static String getNewTableName(String runId) {
    * @return number of rows for this table.
    */
   public static Long getNumRows(BigQuery bigQuery, String project, String dataset, String table) {
+    LOG.info("Getting number of records stored in table {}", table);
```
albertshau: this seems more like a debug message to me

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
albertshau: should set bgTable.getNumRows().longValue() in a variable and refer to it here in the log statement before returning it below.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Yes! This is something I plan to introduce later as a way to add a few debugging behaviors to the plugin.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/util/BigQuerySQLEngineUtils.java
fernst: The expensive operation here is the `Table bgTable = bigQuery.getTable(tableId)`. In that response, the number of rows is returned.

I will make this a local variable as I'm using this value in 2 places.

PR #676, opened 2021-06-13
CDAP-17913 migrate GCS connector

yaojiefeng:
depends on https://github.com/cdapio/hydrator-plugins/pull/1415


docs/GCSFile-batchsource.md
```diff
@@ -25,6 +25,11 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this source for lineage, annotating metadata, etc.
 
+**Use Connection** A boolean indicates whether to use a connection, if a connection is used, 
+the credentials does not need to be provided.
+
+**Connection** The connection name, should have the macro function ${conn:(connection-name)} to provide.
```
nitinmotgi: Is there a drop-down for connection?

yaojiefeng: Yes, UI will show a page to let user choose the connection

docs/GCS-connector.md
```diff
@@ -0,0 +1,35 @@
...
+This plugin can be used to browse and sample from Google Cloud Storage.
+
+Cloud Storage allows world-wide storage and retrieval of any amount of data at any time.
+You can use Cloud Storage for a range of scenarios including serving website content,
+storing data for archival and disaster recovery,
+or distributing large data objects to users via direct download.
+
+Credentials
+-----------
+If the plugin is run on a Google Cloud Dataproc cluster, the service account key does not need to be
```
albertshau: This section should be modified, as connectors don't run on Dataproc.

docs/GCS-connector.md
```diff
@@ -0,0 +1,35 @@
...
+provided and can be set to 'auto-detect'.
+Credentials will be automatically read from the cluster environment.
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
+The service account key can be found on the Dashboard in the Cloud Platform Console.
+Make sure the account key has permission to access BigQuery and Google Cloud Storage.
```
albertshau: doesn't need access to BigQuery

docs/GCS-connector.md
albertshau: I think we can shorten and move this section to the Service Account part of the properties. Something like:

When running on Google Cloud Platform, the service account key does not need to be provided, as it can automatically be read from the environment. In other environments, the service account key must be provided.

docs/GCS-connector.md
```diff
@@ -0,0 +1,35 @@
...
+storing data for archival and disaster recovery,
+or distributing large data objects to users via direct download.
+
+Credentials
+-----------
+If the plugin is run on a Google Cloud Dataproc cluster, the service account key does not need to be
+provided and can be set to 'auto-detect'.
+Credentials will be automatically read from the cluster environment.
+
+If the plugin is not run on a Dataproc cluster, the path to a service account key must be provided.
```
albertshau: path doesn't need to be provided if the contents are given.

docs/GCSFile-batchsource.md
```diff
@@ -25,6 +25,11 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this source for lineage, annotating metadata, etc.
 
+**Use Connection** A boolean indicates whether to use a connection, if a connection is used,
```
albertshau: Don't need to mention the data type.

Whether to use a connection.

docs/GCSFile-batchsource.md
albertshau: We don't include 'The' or 'A' in the first sentence of these descriptions.

Name of the connection to use. Project and service account information will be provided by the connection.

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
```diff
@@ -0,0 +1,234 @@
...
+    if (!failureCollector.getValidationFailures().isEmpty()) {
+      return;
+    }
+
+    Credentials credentials = null;
+    try {
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+    } catch (IOException e) {
+      failureCollector.addFailure(String.format("Service account provided is not valid: %s.", e.getMessage()),
```
albertshau, line 102: would be good to add the stack trace to the failure here and below, as the exception message may not be enough to figure out what is wrong.

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
```diff
@@ -0,0 +1,234 @@
...
+    // entity for the path itself will also get returned in the result since this is a prefix search.
+    BrowseEntity entityForPath = null;
+    for (Blob blob : blobs.iterateAll()) {
+      String blobName = blob.getName();
+
+      if (count >= limit) {
+        if (!blobName.equals(pathBlobName)) {
+          count++;
+        }
+        continue;
```
albertshau: why not break here?

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
```diff
@@ -0,0 +1,234 @@
...
+                                    Storage.BlobListOption.prefix(pathBlobName));
+    int count = 0;
+    BrowseDetail.Builder builder = BrowseDetail.builder();
+    // entity for the path itself will also get returned in the result since this is a prefix search.
+    BrowseEntity entityForPath = null;
+    for (Blob blob : blobs.iterateAll()) {
+      String blobName = blob.getName();
+
+      if (count >= limit) {
+        if (!blobName.equals(pathBlobName)) {
```
albertshau: why does this need to be checked?

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
yaojiefeng: Because we still need to count to get the totalCount

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
yaojiefeng: so since this is prefix search, the query path itself will get returned in response, this is ensure it is not counted to the totalCount

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
yaojiefeng: Discussed offline, we will not continue counting for the number to avoid the request hang for large number of files, so changed to break here

albertshau (approved):

PR #677, opened 2021-06-14
Added logic to support configurable BigQuery job priorities for BQ Join jobs.

fernst:
Added logic to validate configuration for the SQL Engine Config.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineConfig.java
@@ -16,9 +16,11 @@
 
 package io.cdap.plugin.gcp.bigquery.sqlengine;
 
+import com.google.cloud.bigquery.QueryJobConfiguration;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineConfig.java
```diff
@@ -65,4 +82,21 @@ public Boolean shouldRetainTables() {
...
+
+  /**
+   * Validates configuration properties
+   */
+  public void validate() {
+    // Ensure value for the job priority configuration property is valid
+    if (jobPriority != null && !containsMacro(NAME_JOB_PRIORITY)
+      && !PRIORITY_BATCH.equalsIgnoreCase(jobPriority)
+      && !PRIORITY_INTERACTIVE.equalsIgnoreCase(jobPriority)) {
+      throw new SQLEngineException("Property 'jobPriority' must be 'batch' or 'interactive'");
```
fernst, line 99: Unfortunately, the Failure Collector is not available in the context of the SQL Engine execution, so I have to throw an exception here.

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineConfig.java
albertshau: would be good to open a jira to add this capability in some later release. I imagine it would be useful to have a 'validate' button for the sql engine.

PR #684, opened 2021-06-25
Added logic to relax BQ schema even when the output contains no records

fernst:
...as long as the user has specified the  option


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -278,6 +278,14 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
           table.setTableReference(tableRef);
```
fernst, line 281: @albertshau notice how we do create an empty table with the right schema as long as the table doesn't exist. This applies the "Schema relaxation" operation even when no records are going to be written.

albertshau (approved):

PR #688, opened 2021-06-29
fix gcs connector credential npe

yaojiefeng:


seanzhougoogle (dismissed):

src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java
```diff
@@ -97,7 +97,8 @@ public void test(ConnectorContext context) throws ValidationException {
     Credentials credentials = null;
     try {
       credentials =
-        GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+        config.getServiceAccount() == null ? null :
+          GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
```
seanzhougoogle, line 101: what will happen when you get the GCS client with null credentials?

yaojiefeng: it won't set the credentials if it is null, on data fusion it is mostly null, meaning we let the client auto detect the credentials to use https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L173

seanzhougoogle: sg :)

seanzhougoogle (approved):

PR #692, opened 2021-07-03
Add bypass_cmek_check requirements to plugins that do not have CMEK s…

hsiaoching:
…upport


greeshmaswaminathan (approved):
Are these the only plugins that need CMEK?

hsiaoching:
> Are these the only plugins that need CMEK?

These are the GCP plugins that do not support CMEK yet, thus require bypass_cmek_check to be enabled on cdap instance otherwise user would not be allowed to use them.

PR #697, opened 2021-07-09
Updated logic to run BQ Join jobs in the same location as the dataset.

fernst:
Discovered bug where we were using the supplied dataset location even when the dataset was previously created in another region. This fixes this issue.


tivv (commented):
There is an explicit "Location" configuration in BQ pushdown UI settings. Should we use that one if specified?

fernst:
@tivv The Location is only used when creating a new dataset. If the dataset already exists, we disregard that configuration option completely (which is also true for the BQ Sink).

fernst:
Actually, the bug was caused because of this. If the dataset was created in, let's say, EU, and I set the location as US, the code was trying to run the join job in US instead of EU (where the data actually lives).

tivv (approved):
nit: please add a unit test

PR #700, opened 2021-07-13
Added support for Snappy compression in Avro records for BigQuery Pushdown jobs

fernst:
Note that a table is uploaded using the Json format (when the schema contains Datetime fields), no compression will be used.

Most of the logic implemented in this classes is a straight copy from the GCP Hadoop connector, with some logic removed/added as needed.


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineAvroInputFormat.java
```diff
@@ -0,0 +1,133 @@
...
+ * InputFormat which extends {@link AvroBigQueryInputFormat} and uses
+ * {@link BigQuerySQLEngineUnshardedExportToCloudStorage} as the export logic, which allows us to use Snappy to
+ * compress the output from BigQuery jobs.
+ */
+public class BigQuerySQLEngineAvroInputFormat extends AvroBigQueryInputFormat {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BigQuerySQLEngineAvroInputFormat.class);
+
+  /**
+   * Copy of {@link AvroBigQueryInputFormat#getSplits(JobContext)} with a custom Export logic.
```
albertshau, line 52: can you add comments for the line(s) that are different than the copy?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineInputFormat.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public List<InputSplit> getSplits(JobContext context) throws IOException, InterruptedException {
+    processQuery(context);
+
+    return delegateInputFormat.getSplits(context);
+  }
+}
+
+
```
albertshau: nit: remove extra newlines

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineInputFormatProvider.java
```diff
@@ -0,0 +1,42 @@
...
+
+package io.cdap.plugin.gcp.bigquery.sqlengine.input;
+
+import io.cdap.plugin.gcp.bigquery.source.BigQueryInputFormatProvider;
+import org.apache.hadoop.conf.Configuration;
+
+/**
+ * Input format provider for BigQuery SQL Engine Pull operations.
+ */
+public class BigQuerySQLEngineInputFormatProvider extends BigQueryInputFormatProvider {
```
albertshau, line 27: this is to prevent conflicting exports? if so, would be good to add a comment.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineUnshardedExportToCloudStorage.java
```diff
@@ -0,0 +1,93 @@
...
+        projectId, "exporttocloudstorage", tableToExport.getLocation());
+
+    Job job = new Job();
+    job.setConfiguration(config);
+    job.setJobReference(jobReference);
+
+    // Insert and run job.
+    try {
+      Job response = bigQueryHelper.insertJobOrFetchDuplicate(projectId, job);
+      LOG.debug("Got response '{}'", response);
```
albertshau, line 87: what ends up getting logged here? Would be better to be more specific with the message than 'response'.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/output/BigQuerySQLEngineOutputFormat.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+/**
+ * An Output Format which extends {@link BigQueryOutputFormat}.
+ *
+ * This implementation overrides the implementation of
+ * {@link ForwardingBigQueryFileOutputFormat#checkOutputSpecs(JobContext)} in order to skip a check that prevents
+ * compression from being used.
```
albertshau, line 39: do we know why this was prevented to begin with? Seems odd that it was explicitly checking and preventing compression.

albertshau:
Is snappy packaged as a dependency in the plugin jar or does it assume it's available in the cluster environment?

It may be good to have a way to shut off compression just in case, perhaps by reading some runtime argument. That way, if somebody happens to have problems with it, they can set a system preference and disable it for all their pipelines.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineAvroInputFormat.java
fernst: I will add these comments.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineInputFormatProvider.java
fernst: No, the reason is that this Input Format Provider overrides `getInputFormatClassName` from `BigQueryInputFormatProvider` in order to return the SQL engine version of the input format: `BigQuerySQLEngineInputFormat`

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineUnshardedExportToCloudStorage.java
fernst: This is a straight copy from the underlying implementation: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/4b4d659b317e9018852cc43ba2b15930910e5d80/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/AbstractExportToCloudStorage.java#L78

Since Job extends from `GenericJson`, this response is just the JSON payload for this response object.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/output/BigQuerySQLEngineOutputFormat.java
fernst: The only hint in the source code is that, since not all BQ Input formats formats support the same compression codecs, they'd rather not to support compression at all: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/eaaebef7d5efaf51d97d37dfcf9c93a28e1d71ae/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/output/ForwardingBigQueryFileOutputFormat.java#L75

fernst:
@albertshau Snappy was integrated into Hadoop Common in 2011 https://code.google.com/archive/p/hadoop-snappy/

We can add a config setting for the BQ Pushdown plugin that will allow users to enable/disable compression (enabled by default).

Since BQ Pushdown can be enabled using the Pipeline execution UI, this will work effectively as a runtime argument.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/input/BigQuerySQLEngineInputFormatProvider.java
```diff
@@ -19,6 +19,8 @@
 import io.cdap.plugin.gcp.bigquery.source.BigQueryInputFormatProvider;
 import org.apache.hadoop.conf.Configuration;
 
+import java.util.Map;
```
albertshau, line 22: unintended?

albertshau (approved):

PR #702, opened 2021-07-17
CDAP-18109 validate connections in source and update connector doc

seanzhougoogle:


yaojiefeng (approved):
LGTM assuming dependent's comments are addressed

PR #706, opened 2021-07-18
PLUGIN-719 : create temp table and load in batches of 10000 and then …

sahusanket:
…importing in actual table

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
@@ -29,6 +29,7 @@ import com.google.api.services.bigquery.model.ErrorProto;
 import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.JobConfiguration;
 import com.google.api.services.bigquery.model.JobConfigurationLoad;
+import com.google.api.services.bigquery.model.JobConfigurationTableCopy;
 import com.google.api.services.bigquery.model.JobReference;
 import com.google.api.services.bigquery.model.RangePartitioning;
 import com.google.api.services.bigquery.model.RangePartitioning.Range;
@@ -61,6 +62,7 @@ import com.google.cloud.hadoop.util.ConfigurationUtil;
 import com.google.cloud.hadoop.util.ResilientOperation;
 import com.google.cloud.hadoop.util.RetryDeterminer;
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.common.GCPUtils;
@@ -88,7 +90,6 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
-
 /**
  * An Output Format which extends {@link ForwardingBigQueryFileOutputFormat}.
  * This is added to override BigQueryUtils.waitForJobCompletion error message with more useful error message.
@@ -161,6 +162,8 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
     private boolean allowSchemaRelaxation;
     private boolean allowSchemaRelaxationOnEmptyOutput;
 
+    private static final int BQ_IMPORT_MAX_BATCH_SIZE = 10000;
+
     BigQueryOutputCommitter(TaskAttemptContext context, OutputCommitter delegate) throws IOException {
       super(context, delegate);
       try {
@@ -219,10 +222,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
       try {
         importFromGcs(destProjectId, destTable, destSchema.orElse(null), kmsKeyName, outputFileFormat,
                       writeDisposition, sourceUris, partitionType, range, partitionByField,
-                      requirePartitionFilter, clusteringOrderList, tableExists, getJobIdForImportGCS(conf));
-        if (temporaryTableReference != null) {
-          operationAction(destTable, kmsKeyName, getJobIdForUpdateUpsert(conf), conf);
-        }
+                      requirePartitionFilter, clusteringOrderList, tableExists, conf);
       } catch (Exception e) {
         throw new IOException("Failed to import GCS into BigQuery. ", e);
       }
@@ -267,12 +267,13 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
                                List<String> gcsPaths, PartitionType partitionType, @Nullable Range range,
                                @Nullable String partitionByField, boolean requirePartitionFilter,
-                               List<String> clusteringOrderList, boolean tableExists, String jobId)
+                               List<String> clusteringOrderList, boolean tableExists, Configuration conf)
       throws IOException, InterruptedException {
       LOG.info("Importing into table '{}' from {} paths; path[0] is '{}'; awaitCompletion: {}",
                BigQueryStrings.toString(tableRef), gcsPaths.size(), gcsPaths.isEmpty() ? "(empty)" : gcsPaths.get(0),
                true);
 
+      String jobId = getJobIdForImportGCS(conf);
       if (gcsPaths.isEmpty()) {
         if (!bigQueryHelper.tableExists(tableRef)) {
           // If gcsPaths empty and destination table not exist - creating empty destination table.
@@ -294,6 +295,9 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
       }
       // Create load conf with minimal requirements.
       JobConfigurationLoad loadConfig = new JobConfigurationLoad();
+      loadConfig.setSourceFormat(sourceFormat.getFormatIdentifier());
+      loadConfig.setUseAvroLogicalTypes(true);
+
       // If schema change is not allowed and if the destination table already exists, use the destination table schema
       // See PLUGIN-395
       if (!allowSchemaRelaxation && tableExists) {
@@ -301,10 +305,6 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
       } else {
         loadConfig.setSchema(schema);
       }
-      loadConfig.setSourceFormat(sourceFormat.getFormatIdentifier());
-      loadConfig.setSourceUris(gcsPaths);
-      loadConfig.setWriteDisposition(writeDisposition);
-      loadConfig.setUseAvroLogicalTypes(true);
 
       Map<String, String> fieldDescriptions = new HashMap<>();
       if (JobInfo.WriteDisposition.WRITE_TRUNCATE
@@ -346,29 +346,7 @@ public class BigQueryOutputFormat extends ForwardingBigQueryFileOutputFormat<Str
         }
       }
 
-      temporaryTableReference = null;
-      if (!Operation.INSERT.equals(operation)) {
-        String temporaryTableName = tableRef.getTableId() + "_"
-          + UUID.randomUUID().toString().replaceAll("-", "_");
-        temporaryTableReference = new TableReference()
-          .setDatasetId(tableRef.getDatasetId())
-          .setProjectId(tableRef.getProjectId())
-          .setTableId(temporaryTableName);
-        loadConfig.setDestinationTable(temporaryTableReference);
-
-        if (!tableExists) {
-          if (Operation.UPSERT.equals(operation)) {
-            // For upsert operation, if the destination table does not exist, create it
-            Table table = new Table();
-            table.setTableReference(tableRef);
-            table.setSchema(schema);
-            bigQueryHelper.getRawBigquery().tables().insert(tableRef.getProjectId(), tableRef.getDatasetId(), table)
-              .execute();
-          }
-        }
-      } else {
-        loadConfig.setDestinationTable(tableRef);
-
+      if (Operation.INSERT.equals(operation)) {
         // Schema update options should only be specified with WRITE_APPEND disposition,
         // or with WRITE_TRUNCATE disposition on a table partition - The logic below should change when we support
         // insertion into single partition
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -392,26 +383,33 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
+      loadConfig.setDestinationTable(temporaryTableReference);
+      loadConfig.setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND.toString()); // Always append to temp table
 
-      // Poll until job is complete.
-      waitForJobCompletion(bigQueryHelper.getRawBigquery(), projectId, jobReference);
+      Collection<List<String>> gcsPathsInBatches = splitListInBatches(gcsPaths, BQ_IMPORT_MAX_BATCH_SIZE);
+
+      int jobcount = 1;
+      for (List<String> gcsPathBatch :  gcsPathsInBatches) {
+        LOG.info(" Running for Batch {} with number of gcs paths : {}", jobcount, gcsPathBatch.size());
```
nitinmotgi: Not as info

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.hadoop.mapreduce.JobContext;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+
```
nitinmotgi, line 41: Test is not actually testing the change

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -736,4 +771,9 @@ private static BigQuery getBigQuery(Configuration config) throws IOException {
       GCPUtils.loadServiceAccountCredentials(serviceAccount, isServiceAccountFile);
     return GCPUtils.getBigQuery(projectId, credentials);
   }
+  
+  protected static <T> Collection<List<T>> splitListInBatches(List<T> inputList, int batchSize) {
```
fernst: You can use  the `com.google.common.collect.Lists.partition` method to achieve this.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
fernst: +1

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatTest.java
sahusanket: Added some mocked unit tests to actually verify invocations.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: done

fernst:
@sahusanket did you test this code with the BQ Sink? Specially when creating new tables or updating schema for an existing table.

sahusanket:
@fernst  Yes I had tested the schema scenario. Please check https://docs.google.com/spreadsheets/d/1VQxtD3i6Gy4SONLypOeNRsCso7TzB39TgGJCDNp15S4 .I have tested these scenarios locally

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: used the suggested method. Thanks

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -421,6 +417,13 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
                                                         temporaryTableReference.getTableId(), table).execute();
       }
 
+      if (operation.equals(Operation.INSERT)) {
+        handleInsertOperation(tableRef, writeDisposition, loadConfig.getDestinationEncryptionConfiguration(), projectId,
+                              jobId, dataset, tableExists);
+      } else {
+        operationAction(tableRef, kmsKeyName, getJobIdForUpdateUpsert(conf), conf);
```
fernst: nitpick: Since this now covers the UPDATE/UPSERT operation, maybe this method should be renamed to handleUpdateUpsertOperation?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -392,26 +381,33 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
-      // Insert and run job.
-      bigQueryHelper.insertJobOrFetchDuplicate(projectId, job);
+      loadConfig.setDestinationTable(temporaryTableReference);
+      loadConfig.setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND.toString()); // Always append to temp table
 
-      // Poll until job is complete.
-      waitForJobCompletion(bigQueryHelper.getRawBigquery(), projectId, jobReference);
+      List<List<String>> gcsPathsInBatches = Lists.partition(gcsPaths, BQ_IMPORT_MAX_BATCH_SIZE);
+
+      int jobcount = 1;
```
fernst, line 450: add a comment here explaining what this logic does (load in batches of up to 10k files)

sahusanket: done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: renamed it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -267,12 +268,13 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
                                @Nullable String kmsKeyName, BigQueryFileFormat sourceFormat, String writeDisposition,
                                List<String> gcsPaths, PartitionType partitionType, @Nullable Range range,
                                @Nullable String partitionByField, boolean requirePartitionFilter,
-                               List<String> clusteringOrderList, boolean tableExists, String jobId)
-      throws IOException, InterruptedException {
+                               List<String> clusteringOrderList, boolean tableExists, Configuration conf)
+      throws Exception {
```
snehalakshmisha: Why are we throwing the generic exception instead of the specific ones?

snehalakshmisha (commented):
Why are we throwing a generic exception instead of specific ones?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: Because one of the old method was throwing "Exception". On including that in this method automatically made all other Exceptions redundant. I did not touch that old method. 
Also The parent method calling this just throws IOException in try catch, irrespective of the type of error. Check line 228 .

sahusanket: Refactored a bit and reverted the exceptions.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -392,61 +379,94 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
+
+        if (operation.equals(Operation.INSERT)) { // For the case when gcs paths is more than 10000
+          handleInsertOperation(tableRef, writeDisposition, loadConfig.getDestinationEncryptionConfiguration(),
+                                projectId, jobId, dataset, tableExists);
+        } else {
+          handleUpdateUpsertOperation(tableRef, kmsKeyName, getJobIdForUpdateUpsert(conf), conf);
+        }
+      }
+
+      setTemporaryTableExpiration();
```
fernst, line 407: I'd like for the temporary table to get created and the expiration set before the load in batches happens.

Think about it. If there are 2 batches, first succeeds and second fails, the table will not have an expiration set, thus is will stay as a dangling table.

Since I do not want to block the RC5 release, let's merge this and address this in a later PR.

fernst (approved):
Looks good, just left one comment.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
fernst: This can be achieved by creating the temp table with a TTL and then load records while allowing for the table schema to be relaxed

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatTest.java
```diff
@@ -0,0 +1,140 @@
+package io.cdap.plugin.gcp.bigquery.sink;
```
CuriousVini, line 16: Please include license  with year 2021. similar to 
https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java#L2

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -161,6 +162,8 @@ public OutputCommitter createCommitter(TaskAttemptContext context) throws IOExce
     private boolean allowSchemaRelaxation;
     private boolean allowSchemaRelaxationOnEmptyOutput;
 
+    private static final int BQ_IMPORT_MAX_BATCH_SIZE = 10000;
```
CuriousVini, line 165: Should we document this limitation in bq sink documentation?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -378,6 +356,15 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
             JobInfo.SchemaUpdateOption.ALLOW_FIELD_ADDITION.name(),
             JobInfo.SchemaUpdateOption.ALLOW_FIELD_RELAXATION.name()));
         }
+      } else {
+        if (!tableExists && Operation.UPSERT.equals(operation)) {
```
CuriousVini: nit: move else if in same line similar to:
https://github.com/data-integrations/google-cloud/blob/f054621188c1301d7d47b204b2d4888f6a7a08d3/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSArgumentSetter.java#L157

CuriousVini: We dont need this check for insert?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -392,61 +379,94 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
+
       Job job = new Job();
-      job.setConfiguration(config);
+      job.setConfiguration(jobConfiguration);
       job.setJobReference(jobReference);
-
-      // Insert and run job.
       bigQueryHelper.insertJobOrFetchDuplicate(projectId, job);
-
-      // Poll until job is complete.
```
CuriousVini, line 413: why this change is needed?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -392,61 +379,94 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
...
         new JobReference().setProjectId(projectId)
           .setJobId(jobId)
           .setLocation(dataset.getLocation());
+
       Job job = new Job();
-      job.setConfiguration(config);
+      job.setConfiguration(jobConfiguration);
       job.setJobReference(jobReference);
-
-      // Insert and run job.
```
CuriousVini, line 410: why this change is needed?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: For INSERT , since it's a Big Query Job directly into destination table, it Automatically creates the Table.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: @CuriousVini  this limitation is addressed now via creating temp table and loading and User would not have to take any extra action.  We can mention this temp table thing in doc But I went through the Docs in this repo, we do not have a section for any such internal processing notes that does not require user action.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: reverted

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
sahusanket: reverted

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormatTest.java
sahusanket: added

PR #720, opened 2021-07-23
[Cherry-Pick] [0.18]PLUGIN-719 : create temp table and load in batches of 10000 and impor…

sahusanket:
…t in actual table


fernst (approved):
LGTM

PR #722, opened 2021-07-23
do not show hidden datasets

yaojiefeng:


sechegaray (commented):
Could removing this option without adding a toggle to show all Datasets, break some user workflow?

yaojiefeng:
Browsing change won't affect any user pipeline. This is to make the behavior same as bigquery console

seanzhougoogle (approved):

PR #725, opened 2021-07-23
add option to show hidden datasets

yaojiefeng:


src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorSpecificConfig.java
```diff
@@ -29,11 +29,20 @@
...
                                          @Nullable String serviceAccountType, @Nullable String serviceFilePath,
                                          @Nullable String serviceAccountJson,
                                          @Nullable String rootDataset) {
     super(project, datasetProject, serviceAccountType, serviceFilePath, serviceAccountJson);
     this.rootDataset = rootDataset;
+    this.showHiddenDatasets = false;
+  }
+
+  public boolean showHiddenDatasets() {
+    return showHiddenDatasets == null ? false : showHiddenDatasets;
```
fernst, line 46: Since the current behavior assumes this to be true, should the default be true as well?

docs/BigQuery-connector.md
```diff
@@ -26,4 +26,6 @@ it can automatically be read from the environment. In other environments, the se
   auto-detect' when running on a Dataproc cluster. When running on other clusters, the file must be present on every
   node in the cluster.
 
-* **JSON**: Contents of the service account JSON file.
\ No newline at end of file
+* **JSON**: Contents of the service account JSON file.
+
+**Show Hidden Datasets:** Option to show hidden datasets.
```
albertshau: nit: Whether to show hidden datasets.

Also, is there a similar concept for BQ tables?

fernst (approved):
LGTM, minor question

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorSpecificConfig.java
albertshau, line 46: I think we actually want to default to true since that's the more user friendly behavior, and let the upgrade back set this to false for any pre-existing connections.

albertshau: the plan will be for connection upgrade to set this to true for any pre-existing connections.

docs/BigQuery-connector.md
yaojiefeng: No there is no such concept for table, any table starts with _ will show up in bigquery console

albertshau (approved):

PR #737, opened 2021-07-31
Feature/plugin 767 - Add BigQuery Plugin to load Array and Structs (And convert them to StructureRecords)

sechegaray:
Allow the BigQuery Plugin to load Array and Structs from BigQuery and convert them to StructureRecord to pass them downstream.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -302,17 +302,18 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
 
     for (Schema.Field field: fields) {
       String fieldName = prefix + field.getName();
+      if (depth == BigQueryTypeSize.Struct.MAX_DEPTH) {
+        collector.addFailure(
+          String.format("Field '%s' exceeds BigQuery maximum allowed depth of %d.",
+                        fieldName, BigQueryTypeSize.Struct.MAX_DEPTH),
+          "Please flatten the schema to contain fewer levels.");
+        continue;
+      }
+
       Schema fieldSchema = BigQueryUtil.getNonNullableSchema(field.getSchema());
       switch(fieldSchema.getType()) {
         case RECORD:
-          if (depth == BigQueryTypeSize.Struct.MAX_DEPTH) {
-            collector.addFailure(
-              String.format("Field '%s' exceeds BigQuery maximum allowed depth of %d.",
-                            fieldName, BigQueryTypeSize.Struct.MAX_DEPTH),
-              "Please flatten the schema to contain fewer levels.");
-          } else {
-            validateRecordDepth(fieldSchema, collector, depth + 1, fieldName + ".");
-          }
+          validateRecordDepth(fieldSchema, collector, depth + 1, fieldName + ".");
           break;
         case ARRAY:
           if (fieldSchema.getComponentSchema() == null) {
@@ -320,14 +321,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
           }
           Schema componentSchema = BigQueryUtil.getNonNullableSchema(fieldSchema.getComponentSchema());
           if (componentSchema.getType() == Schema.Type.RECORD) {
-            if (depth == BigQueryTypeSize.Struct.MAX_DEPTH) {
-              collector.addFailure(
-                String.format("Field '%s' exceeds BigQuery maximum allowed depth of %d.",
-                              fieldName, BigQueryTypeSize.Struct.MAX_DEPTH),
-                "Please flatten the schema to contain fewer levels.");
-            } else {
-              validateRecordDepth(componentSchema, collector, depth + 1, fieldName + ".");
-            }
+            validateRecordDepth(componentSchema, collector, depth + 1, fieldName + ".");
           } else {
             validateRecordDepth(componentSchema, collector, depth, fieldName + ".");
           }
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -39,6 +39,7 @@ import io.cdap.cdap.etl.api.batch.BatchSinkContext;
 import io.cdap.cdap.etl.api.validation.ValidationFailure;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryTypeSize;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.CmekUtils;
 import io.cdap.plugin.gcp.common.GCPUtils;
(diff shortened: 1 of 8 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
```diff
@@ -61,49 +62,97 @@ private BigQueryDataParser() {
...
       FieldValue fieldValue = fieldValues.get(fieldName);
-      FieldValue.Attribute attribute = fieldValue.getAttribute();
+      Attribute attribute = fieldValue.getAttribute();
       if (fieldValue.isNull()) {
         recordBuilder.set(fieldName, null);
         continue;
       }
 
-      if (attribute == FieldValue.Attribute.REPEATED) {
+      Schema.Field localSchemaField = schemaFieldList.get(i);
```
tivv: Are you assuming CDAP schema and BQ schema has same fields in same order?

tivv (commented):
Please add a feature flag to this new capability. We should guard it with feature flag until after it's ready and tested. 
Also please give a good name to this PR

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
sechegaray: Refactored to avoid the assumption.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -346,6 +346,11 @@ private void importFromGcs(String projectId, TableReference tableRef, @Nullable
         }
       }
 
+      if (!tableExists && Operation.UPSERT.equals(operation)) {
+        // If the table does not exist then an UPSERT operation becomes an INSERT operation
```
fernst: Is there a difference here?

fernst: I get it now. Much cleaner.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -267,6 +268,68 @@ protected void validateInsertSchema(Table table, @Nullable Schema tableSchema, F
...
+    }
+
+    for (Schema.Field field: fields) {
+      String fieldName = prefix + field.getName();
+      if (depth == BigQueryTypeSize.Struct.MAX_DEPTH) {
+        collector.addFailure(
+          String.format("Field '%s' exceeds BigQuery maximum allowed depth of %d.",
+                        fieldName, BigQueryTypeSize.Struct.MAX_DEPTH),
+          "Please flatten the schema to contain fewer levels.");
+        continue;
```
fernst, line 310: Shouldn't this break/return here? If we discover the depth exceeds the threshold, we know this record is not processable and we should fail the pipeline.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -267,6 +268,68 @@ protected void validateInsertSchema(Table table, @Nullable Schema tableSchema, F
...
+          break;
+        case ARRAY:
+          if (fieldSchema.getComponentSchema() == null) {
+            break;
+          }
+          Schema componentSchema = BigQueryUtil.getNonNullableSchema(fieldSchema.getComponentSchema());
+          if (componentSchema.getType() == Schema.Type.RECORD) {
+            validateRecordDepth(componentSchema, collector, depth + 1, fieldName + ".");
+          } else {
+            validateRecordDepth(componentSchema, collector, depth, fieldName + ".");
```
fernst, line 326: So this means arrays are considered "flat" (same level) as a single value for a key?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -272,16 +273,20 @@ private static BigQueryTableFieldSchema generateTableFieldSchema(Schema.Field fi
     fieldSchema.setType(type.name());
     if (type == LegacySQLTypeName.RECORD) {
       List<Schema.Field> schemaFields;
-      if (Schema.Type.ARRAY == field.getSchema().getType()) {
-        schemaFields = Objects.requireNonNull(field.getSchema().getComponentSchema()).getFields();
+      Schema fieldCdapSchema = BigQueryUtil.getNonNullableSchema(field.getSchema());
```
fernst, line 276: Why not just `cdapSchema`?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
```diff
@@ -63,49 +64,100 @@ private BigQueryDataParser() {
   }
 
   public static StructuredRecord getStructuredRecord(Schema schema, FieldList fields, FieldValueList fieldValues) {
+
     StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
-    for (Field field : fields) {
+    for (int i = 0; i < fields.size(); i++) {
```
fernst: Is this change required? It just makes accessing the list of fields O(N^2) instead of O(N)

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
```diff
@@ -63,49 +64,100 @@ private BigQueryDataParser() {
...
       FieldValue fieldValue = fieldValues.get(fieldName);
-      FieldValue.Attribute attribute = fieldValue.getAttribute();
+      Attribute attribute = fieldValue.getAttribute();
       if (fieldValue.isNull()) {
         recordBuilder.set(fieldName, null);
         continue;
       }
 
-      if (attribute == FieldValue.Attribute.REPEATED) {
+      Schema.Field localSchemaField = schema.getField(fieldName);
```
fernst: why not `localField` ?

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
```diff
@@ -63,49 +64,100 @@ private BigQueryDataParser() {
...
+      Schema localSchema;
+      if (localSchemaField != null) {
+        localSchema = localSchemaField.getSchema();
+      } else {
+        continue;
+      }
+      FieldList localFields = field.getSubFields();
```
fernst: `subFields` perhaps?

src/test/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtilTest.java
```diff
@@ -95,9 +95,11 @@ public void testGetTableSchema() {
...
         Schema.nullableOf(Schema.of(Schema.Type.STRING)));
     Schema.Field cdapSubRecordField = Schema.Field.of("subStructure",
-        Schema.nullableOf(Schema.recordOf("structure.subStructure", cdapSubStringField)));
+        Schema.nullableOf(Schema.recordOf("structure.subStructure" +
+                                            Schema.recordOf(cdapSubStringField).getRecordName(), cdapSubStringField)));
     Schema.Field cdapRecordField = Schema.Field.of("structure",
-        Schema.nullableOf(Schema.recordOf("structure", cdapSubRecordField)));
+        Schema.nullableOf(Schema.recordOf("structure" +
+          Schema.recordOf(cdapSubRecordField).getRecordName(), cdapSubRecordField)));
     cdapFieldList.add(cdapRecordField);
```
fernst, line 104: For such a large change, I would like additional tests added to validate that this doesn't change existing behaviors.

fernst (commented):
Some questions/comments

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
sechegaray: The pipeline will fail. I set continue so we can show the user ALL fields that are breaking this issue and not only the first one.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
sechegaray: Yes. Arrays in BQ are not a different level, they are the record but with mode "REPEATED."

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
sechegaray: The BQ field schema above is named `fieldSchema` so I wanted to connect both variables to make clear that they are referring to the same field.  Another solution to make it clearer would be to refactor the function and use `bqSchema` for the BQ field, and `cdapSchema` for cdap. But we haven't been using that standard across the code (prefixing bq)

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
sechegaray: Good point.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryDataParser.java
sechegaray: Good Idea.

src/test/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtilTest.java
sechegaray: Added the test to check depth validation as discussed offline.

fernst (approved):
LGTM

PR #745, opened 2021-08-03
CMEK support for Spanner sink and Bigtable sink

hsiaoching:


src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
```diff
@@ -69,7 +69,6 @@
 @Name(BigtableSink.NAME)
 @Description("This sink writes data to Google Cloud Bigtable. " +
   "Cloud Bigtable is Google's NoSQL Big Data database service.")
-@Requirements(capabilities = "bypass_cmek_check")
```
albertshau, line 72: is this supposed to be removed?

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
```diff
@@ -183,16 +190,40 @@ private boolean isTablePresent(DatabaseClient dbClient) {
...
     DatabaseAdminClient dbAdminClient) throws ExecutionException, InterruptedException {
     Database database = getDatabaseIfPresent(dbAdminClient);
 
     if (database == null) {
       LOG.debug("Database not found. Creating database {} in instance {}.", config.getDatabase(), config.getInstance());
       // Create database
+      String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
+      if (cmekKey == null) {
+        database = dbAdminClient
+        .newDatabaseBuilder(DatabaseId.of(config.getProject(), config.getInstance(), config.getDatabase()))
```
albertshau: nit: indentation should be 2 more spaces

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
```diff
@@ -183,16 +190,40 @@ private boolean isTablePresent(DatabaseClient dbClient) {
...
+        .build();
+      } else {
+        database = dbAdminClient
+        .newDatabaseBuilder(DatabaseId.of(config.getProject(), config.getInstance(), config.getDatabase()))
+        .setEncryptionConfig(EncryptionConfigs.customerManagedEncryption(cmekKey))
```
albertshau: nit: to remove code duplication, it is better to create the builder outside of the if-else and only call setEncryptionConfig() if the key is not null

src/main/java/io/cdap/plugin/gcp/bigtable/sink/BigtableSink.java
hsiaoching: Yes it is intended. Since the Bigtable sink does not create non-CMEK compliant resources, it should not have the bypass_cmek_check requirement.

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
hsiaoching: Fixed.

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java
hsiaoching: Fixed.

albertshau (approved):
lgtm

PR #746, opened 2021-08-04
[Cherrypick] [release/0.18] CMEK support for Spanner sink and Bigtable sink

hsiaoching:


albertshau (approved):
lgtm.

Please be on the lookout for integration test failures. All of these GCP libraries have the same set of dependencies, so sometimes bumping the version of one can cause a problem with another.

PR #749, opened 2021-08-05
Fixed issue with BQ Pushdown temporary buckets not getting deleted after pipeline execution

fernst:
Cleaned up some duplicated code.


albertshau (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryPushDataset.java
```diff
@@ -98,7 +98,6 @@ protected static BigQueryPushDataset getInstance(SQLPushRequest pushRequest,
     String gcsPath = BigQuerySQLEngineUtils.getGCSPath(bucket, runId, table);
     List<BigQueryTableFieldSchema> fields =
       BigQuerySinkUtils.getBigQueryTableFieldsFromSchema(pushRequest.getDatasetSchema());
-    BigQuerySinkUtils.configureBucket(configuration, bucket, runId);
```
snehalakshmisha, line 101: Just for my understanding, why don't we need to configure bucket here anymore?

fernst: Because the configuration is already set in the configuration object provided by the BigQuery SQL Engine. This instance of the configuration starts as a clone from the parent config, thus is already property configured for GCS operations.

PR #757, opened 2021-08-09
Add details about Format field in Pub/Sub plugins

snehalakshmisha:
Pub/Sub Streaming Source markdown documentation was missing the Format property. 
This was also missing from Pub/Sub sink plugin, the same information has been added here as well.

Testing:

In Source:
Screen Shot 2021-08-09 at 11 05 46 AM

In Sink:
Screen Shot 2021-08-09 at 11 06 12 AM

### DIFF
--- a/docs/GooglePublisher-batchsink.md
+++ b/docs/GooglePublisher-batchsink.md
@@ -28,6 +28,9 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Topic**: Name of the Google Cloud Pub/Sub topic to publish to.
 
+**Format**: Format of the data to read. Supported formats are avro, blob, tsv, csv, delimited, json,
+parquet, and text. Default is text.
+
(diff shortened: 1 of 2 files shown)


avirmani (approved):

chtyim (approved):

CuriousVini (approved):

docs/GooglePublisher-batchsink.md
```diff
@@ -28,6 +28,9 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Topic**: Name of the Google Cloud Pub/Sub topic to publish to.
 
+**Format**: Format of the data to read. Supported formats are avro, blob, tsv, csv, delimited, json,
```
rrielley76, line 31: @snehalakshmisha shouldn't the sink say "publish" or "write" instead of "read"?  If so, the tooltip for the pubsub sink is wrong too. It says "read"

PR #763, opened 2021-08-19
Fixed issue with GCS Source temp buckets not getting deleted after pipeline execution

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -74,6 +74,7 @@
   private BigQuerySourceConfig config;
   private Schema outputSchema;
   private Configuration configuration;
+  private Boolean deleteBucket;
```
tivv: Where is this one used?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -172,17 +176,17 @@ public static void deleteBigQueryTemporaryTable(Configuration configuration, Big
    *
    * @param configuration Hadoop Configuration.
    * @param bucket the bucket name
-   * @param runId the run ID
+   * @param bucketPath the bucket path to use.
```
tivv: I don't think it's a path. It's used as as part of  path in one place, as name in other. In both cases I don't think it has part elements (separators)

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
fernst: Oops, accidentally commited something from an initial idea. Deleted.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
fernst: It does not at this time. but I want to keep this generic for the future (anything could be used here as long as it is a valid GCS name)

tivv: That's what I mean. Don't name it path as someone can pass a path and it won't work as bucket name

fernst: I've reverted this change to match RunID. It is true that we only use this for now.

tivv (approved):

TunkVineeth:
test message

PR #767, opened 2021-09-08
[PLUGIN-883] Adding CMEK KEY as Config in GCS Batch Sink Plugin

itsankit-google:
[PLUGIN-883](https://cdap.atlassian.net/browse/PLUGIN-883)

Added CMEK Key as a config field in the GCS Batch Sink Plugin. Added validations for structure of cmek key resource ID provided by user and if the location of key and resource to be created are not same.

PR had to be reverted and merged again after bumping the cdap snapshot version. [Link](https://github.com/data-integrations/google-cloud/pull/770)


albertshau:
don't link to docs for explanation. Anything that is needed to understand what the code is doing should be in the code itself as comments.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +398,29 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    //This method validated the pattern of CMEK Key resource ID.
+    private void validateCmekKey(FailureCollector failureCollector) {
+      String[] part = cmekKey.split("/");
```
albertshau: Have you looked at the KMS library to see if there's a better way to do this? Usually not a good idea to manually parse like this.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -99,6 +99,7 @@ public ValidatingOutputFormat getOutputFormatForRun(BatchSinkContext context) th
   public void prepareRun(BatchSinkContext context) throws Exception {
     super.prepareRun(context);
     String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
+    cmekKey = config.getCmekKey(cmekKey);
```
albertshau: this logic will need to be done in multiple plugins and should be pulled into its own class instead of repeated in every plugin. GCPUtils.CMEK_KEY should be private to that class.

For example, may make sense to put this in the base GCPConfig class.

albertshau (commented):
the jira mentions additional validation checks, like if the key is in the wrong location. Is that coming in a future PR?

itsankit-google:
> the jira mentions additional validation checks, like if the key is in the wrong location. Is that coming in a future PR?

I have put a check for location in line no 412.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: You are mentioning about the getCmekKey() method, right?
If we put this in the GCPConfig class then cmek key will be included as a config in all sink plugins extending it.
Currently, all sink plugins don't support cmek.

albertshau:
the location check isn't about the location missing from the resource URI. It's about the CMEK key being in a different location than the relevant resource. For example, the key is in region1 and the GCS bucket it is being used on is in region2.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -261,6 +261,7 @@ private long extractRecordCount(Map<String, String> metadata) {
     private static final String FORMAT_DELIMITED = "delimited";
     private static final String FORMAT_ORC = "orc";
     private static final String FORMAT_PARQUET = "parquet";
+    private static final String NAME_CMEK_KEY = "cmekKey";
```
albertshau: this constant should be moved to the base GCPConfig class and used in the getCmekKey() method

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -138,4 +139,12 @@ public boolean autoServiceAccountUnavailable() {
     }
     return false;
   }
+
+  public String getCmekKey(SettableArguments arguments) {
```
albertshau: this should just be Arguments and not SettableArguments, since we don't need or want to set anything in this method.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +397,22 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    //This method validated the pattern of CMEK Key resource ID.
+    private void validateCmekKey(FailureCollector failureCollector) {
+      String[] part = cmekKey.split("/");
+      if ((part.length != 8) || !part[0].equals("projects") || !part[2].equals("locations")
+        || !part[4].equals("keyRings") || !part[6].equals("cryptoKeys")) {
+        failureCollector.addFailure("Invalid cmek key resource Id format. It should be of the form " +
```
albertshau: Invalid cmek key resource Id format. It ... -> Invalid CMEK key. It ...

itsankit-google:
> the location check isn't about the location missing from the resource URI. It's about the CMEK key being in a different location than the relevant resource. For example, the key is in region1 and the GCS bucket it is being used on is in region2.

Yes, I have added the check for that only, that the location of the key and the bucket that will be created are same or not.
if (!part[3].equals(location)) {
        failureCollector.addFailure(String.format("Cloud KMS region '%s' not available for use " +
                                                    "with GCS region '%s'.", part[3], location), null)
          .withConfigProperty(NAME_CMEK_KEY);
      }

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +397,22 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    //This method validated the pattern of CMEK Key resource ID.
+    private void validateCmekKey(FailureCollector failureCollector) {
+      String[] part = cmekKey.split("/");
+      if ((part.length != 8) || !part[0].equals("projects") || !part[2].equals("locations")
```
albertshau: nit: when doing comparisons, it's better to put the constant first to avoid possible null pointer exceptions ("projects".equals(part[0])

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +397,22 @@ public void validate(FailureCollector collector) {
       }
     }
 
+    //This method validated the pattern of CMEK Key resource ID.
+    private void validateCmekKey(FailureCollector failureCollector) {
```
albertshau: this method only validates the CMEK key from the config and not the key from the arguments. It should be validating whatever key will be used (should pass in the key)

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +397,22 @@ public void validate(FailureCollector collector) {
...
+      String[] part = cmekKey.split("/");
+      if ((part.length != 8) || !part[0].equals("projects") || !part[2].equals("locations")
+        || !part[4].equals("keyRings") || !part[6].equals("cryptoKeys")) {
+        failureCollector.addFailure("Invalid cmek key resource Id format. It should be of the form " +
+                                      "projects/<gcp-project-id>/locations/<key-location>/keyRings/" +
+                                      "<key-ring-name>/cryptoKeys/<key-name>", null)
+          .withConfigProperty(NAME_CMEK_KEY);
+      } else if (!part[3].equals(location)) {
+        failureCollector.addFailure(String.format("Cloud KMS region '%s' not available for use " +
```
albertshau: The plugin behavior (at least according to the description) is that the location is ignored if the bucket already exists.  This should only be checked if the bucket does not exist.

The message should also be more specific. For example, the message can be "CMEK key '%s' is in location '%s', while the GCS bucket will be created in location '%s'."  and the corrective action can be "Modify the CMEK key or bucket location to be the same"

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +397,22 @@ public void validate(FailureCollector collector) {
...
+                                                    "with GCS region '%s'.", part[3], location), null)
+          .withConfigProperty(NAME_CMEK_KEY);
```
albertshau: the key can be read from runtime arguments and not specified in the config, in which case this would highlight an empty box. It would be better to highlight both the CMEK key and the location properties.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: validateCmekKey() method is called when we click on the validate button in the plugin so it will be concerned about the cmek Key that is passed as config and not the arguments one.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
+            if (!cmekKeyLocation.equals(location)) {
+              failureCollector.addFailure(String.format("CMEK key '%s' is in location '%s' while the GCS bucket will " +
+                                                          "be created in location '%s'.", cmekKey,
+                                                        cmekKeyLocation, location)
+                , "Modify the CMEK key or bucket location to be the same")
+                .withConfigProperty(NAME_CMEK_KEY);
+            }
+          }
+        }
+      } catch (Exception e) {
```
albertshau: this try-catch should only be for parsing the CMEK key. If there is some other exception (for example, with loading the service account credentials), we don't want to highlight the cmek key.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
-                                      "projects/<gcp-project-id>/locations/<key-location>/keyRings/" +
-                                      "<key-ring-name>/cryptoKeys/<key-name>", null)
-          .withConfigProperty(NAME_CMEK_KEY);
-      } else if (!part[3].equals(location)) {
-        failureCollector.addFailure(String.format("Cloud KMS region '%s' not available for use " +
-                                                    "with GCS region '%s'.", part[3], location), null)
-          .withConfigProperty(NAME_CMEK_KEY);
+      try {
+        CryptoKeyName cmekKeyName = CryptoKeyName.parse(cmekKey);
+        Boolean isServiceAccountFilePath = isServiceAccountFilePath();
```
albertshau: it's not good that this returns a Boolean and not a boolean. Taking a quick look at the usages, there are possible NullPointerExceptions because of this.

Can you change isServiceAccountFilePath() and isServiceAccountJson() to return booleans? If no value is given, serviceAccountFilePath should be true and serviceAccountJson should be false.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
+        if (isServiceAccountFilePath != null) {
```
albertshau: this if condition shouldn't be here, should always try to load the credentials

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
+          Credentials credentials = getServiceAccount() == null ?
+            null : GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath);
+          Storage storage = GCPUtils.getStorage(getProject(), credentials);
+          Bucket bucket = null;
+          try {
+            bucket = storage.get(getBucket());
+          } catch (StorageException e) {
```
albertshau: safer to catch any Exception here.

Also good to add a comment here on why we ignore.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
+            return;
+          }
+          if (bucket == null) {
+            String cmekKeyLocation = cmekKeyName.getLocation();
+            if (!cmekKeyLocation.equals(location)) {
```
albertshau: if location is null, this will incorrectly throw an exception. When it is null, the GCS API decides which location, I think it defaults to US. Can you check if there is some way to get the default location from the GCS libraries?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -399,17 +399,33 @@ public void validate(FailureCollector collector) {
...
-          .withConfigProperty(NAME_CMEK_KEY);
-      } else if (!part[3].equals(location)) {
-        failureCollector.addFailure(String.format("Cloud KMS region '%s' not available for use " +
-                                                    "with GCS region '%s'.", part[3], location), null)
-          .withConfigProperty(NAME_CMEK_KEY);
+      try {
+        CryptoKeyName cmekKeyName = CryptoKeyName.parse(cmekKey);
+        Boolean isServiceAccountFilePath = isServiceAccountFilePath();
+        if (isServiceAccountFilePath != null) {
+          Credentials credentials = getServiceAccount() == null ?
```
albertshau: is the first getServiceAccount() == null needed? It gets passed to loadServiceAccountCredentials() already.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: Its already a boolean and returns true if no value is given by default. I followed the way, service account was loaded in prepareRun().

albertshau: On develop (https://github.com/data-integrations/google-cloud/blob/07800a568ec0ea9a3c4a06fbfdaee22226e7451e/src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java#L110) it returns a Boolean that can return null. Is it different on another branch or something?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -402,25 +402,30 @@ private void validateCmekKey(FailureCollector failureCollector) {
...
-            return;
+        Credentials credentials = getServiceAccount() == null ?
+          null : GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath);
+        Storage storage = GCPUtils.getStorage(getProject(), credentials);
+        Bucket bucket = null;
+        try {
+          bucket = storage.get(getBucket());
+        } catch (StorageException e) {
+          /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
+            the bucket information either because the service account used during validation is the CDF service account
```
albertshau: In general, don't mention CDF or make assumptions about where the code is running. This is the open source code that can run in all sorts of environments. Change to:

... during validation can be different than the service account that will be used at runtime.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -402,25 +402,30 @@ private void validateCmekKey(FailureCollector failureCollector) {
...
-            String cmekKeyLocation = cmekKeyName.getLocation();
-            if (!cmekKeyLocation.equals(location)) {
-              failureCollector.addFailure(String.format("CMEK key '%s' is in location '%s' while the GCS bucket will " +
-                                                          "be created in location '%s'.", cmekKey,
-                                                        cmekKeyLocation, location)
-                , "Modify the CMEK key or bucket location to be the same")
-                .withConfigProperty(NAME_CMEK_KEY);
-            }
+          String cmekKeyLocation = cmekKeyName.getLocation();
+          if (!cmekKeyLocation.equals(location)) {
```
albertshau: I see a global location for keyrings. Can a global key be used in any location?

What about case sensitivity, is the 'Us' location treated the same as 'US'?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +398,50 @@ public void validate(FailureCollector collector) {
...
+        } catch (StorageException e) {
+          /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
+            the bucket information either because the service account used during validation is the CDF service account
+            which is different than the service account that will be used at runtime (the dataproc service account)
+            (assuming the user has auto-detect for the service account) */
+          return;
+        }
+        if (bucket == null) {
+          if (Strings.isNullOrEmpty(location)) {
+            location = "US";
```
albertshau: don't modify any values in the config. They should always be the value that the user provided. Treat them like they are final.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -386,6 +398,50 @@ public void validate(FailureCollector collector) {
...
+          String cmekKeyLocation = cmekKeyName.getLocation();
+          if (!cmekKeyLocation.equals(location)) {
+            failureCollector.addFailure(String.format("CMEK key '%s' is in location '%s' while the GCS bucket will " +
+                                                        "be created in location '%s'.", cmekKey,
+                                                      cmekKeyLocation, location)
+              , "Modify the CMEK key or bucket location to be the same")
+              .withConfigProperty(NAME_CMEK_KEY);
+          }
+        }
+      } catch (ValidationException e) {
```
albertshau: Nothing in the try block throws a ValidationException. You should not have a try-catch for this entire method. As mentioned before, this should only be around the part that parses the key.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+                                   StandardCharsets.UTF_8);
+
+  }
+
+  private GCSBatchSink.GCSBatchSinkConfig getConfig(@Nullable String project, @Nullable String path,
+                                                    @Nullable String serviceAccountType,
+                                                    @Nullable String serviceFilePath,
+                                                    @Nullable String serviceAccountJson) throws NoSuchFieldException {
+    GCSBatchSink.GCSBatchSinkConfig gcsBatchSinkConfig = new GCSBatchSink.GCSBatchSinkConfig();
+    FieldSetter
```
albertshau: Don't set fields like this. It is brittle (renaming a field will break this test).

Create a Builder for the config and create it normally.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+                                                       null, serviceAccountKey);
+    testValidCmekKey(config);
+    testInvalidCmekKey(config);
+  }
+
+  public void testValidCmekKey(GCSBatchSink.GCSBatchSinkConfig config) throws Exception {
+    MockFailureCollector collector = new MockFailureCollector("gcssink");
+    config.validate(collector);
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("location"), "us-east1");
```
albertshau: as mentioned above, don't set fields like this. It assumes the class is implemented in a specific way.

Mockito shouldn't be used to test code that lives entirely in this project. It should only be used if you need to mock something that you don't have control over.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+    if (project == null) {
+      project = System.getProperty("GOOGLE_CLOUD_PROJECT");
+    }
+    if (project == null) {
+      project = System.getProperty("GCLOUD_PROJECT");
+    }
+    Assume.assumeFalse(String.format(messageTemplate, "project id"), project == null);
+    System.setProperty("GCLOUD_PROJECT", project);
+
+    bucketPath = System.getProperty("bucket.path");
```
albertshau: this shouldn't be passed in as an argument, as it modifies the expectations of the test.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("cmekKey"),
+                "keyRings/my_ring/cryptoKeys/test_key");
+    config.validate(collector);
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("location"), "us");
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("cmekKey"),
+                "projects/cdf-test-322418/locations/us-east1/keyRings/my_ring/cryptoKeys/test_key");
+    config.validate(collector);
```
albertshau: tests should be as focused as possible. Make the validateCmekKey package private and call it directly.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+    Assert.assertEquals(0, collector.getValidationFailures().size());
+  }
+
+  public void testInvalidCmekKey(GCSBatchSink.GCSBatchSinkConfig config) throws Exception {
+    MockFailureCollector collector = new MockFailureCollector("gcssink");
+    config.validate(collector);
```
albertshau: there should be assertions after each validate call, not just at the end.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkCmekKeyTest.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  public void testValidCmekKey(GCSBatchSink.GCSBatchSinkConfig config) throws Exception {
+    MockFailureCollector collector = new MockFailureCollector("gcssink");
+    config.validate(collector);
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("location"), "us-east1");
+    FieldSetter
+      .setField(config, GCSBatchSink.GCSBatchSinkConfig.class.getDeclaredField("cmekKey"),
+                "projects/cdf-test-322418/locations/us-east1/keyRings/my_ring/cryptoKeys/test_key");
```
albertshau: is this an actual project or just some value you made up? Don't put references to actual resources in the code.

itsankit-google: Its just some value I made up to test the cmek key name format.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: Global and multi-region keys are not allowed. Yeah, changed the check to equalsIgnoreCase. They are treated as same verified by running the pipeline.

albertshau: 'us' is multi-region, does that fail as well?

itsankit-google: If the location of key is also 'us' then it does not fail

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -532,5 +559,53 @@ public String getContentType() {
...
+      } catch (StorageException e) {
+        /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
+          the bucket information either because the service account used during validation can be different than
+          the service account that will be used at runtime (the dataproc service account)
+          (assuming the user has auto-detect for the service account) */
+        return;
+      }
+      if (bucket == null) {
+        String cmekKeyLocation = cmekKeyName.getLocation();
+        if ((Strings.isNullOrEmpty(location) && "US".equalsIgnoreCase(cmekKeyLocation))
```
albertshau: shouldn't this be `!"US".equals...`? Please add a unit test for this scenario.

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkConfigBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.gcs.sink;
+
+public class GCSBatchSinkConfigBuilder {
```
albertshau: our normal pattern is to put the Builder within the class that it builds. GCSBatchSinkConfig.Builder

src/test/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSinkConfigBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+    this.cmekKey = cmekKey;
+    return this;
+  }
+
+  public GCSBatchSinkConfigBuilder setLocation(String location) {
+    this.location = location;
+    return this;
+  }
+
+  public static GCSBatchSinkConfigBuilder aGCSBatchSinkConfig() {
```
albertshau: our normal pattern is to put the static method in the class that it builds and name it builder(). `GCSBatchSinkConfig.builder().setX(...)`

albertshau (commented):
minor style issues, otherwise lgtm

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: Yeah, missed it will make the change

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -597,15 +597,91 @@ the service account that will be used at runtime (the dataproc service account)
...
             , "Modify the CMEK key or bucket location to be the same")
             .withConfigProperty(NAME_CMEK_KEY);
         }
       }
     }
+
+    /**
+     * GCS Batch Sink configuration builder.
+     */
+    public static class GCSBatchSinkConfigBuilder {
```
albertshau: call this Builder

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -597,15 +597,91 @@ the service account that will be used at runtime (the dataproc service account)
...
+        this.cmekKey = cmekKey;
+        return this;
+      }
+
+      public GCSBatchSinkConfigBuilder setLocation(@Nullable String location) {
+        this.location = location;
+        return this;
+      }
+
+      public static GCSBatchSinkConfigBuilder builder() {
```
albertshau: this should be in the config class, not the builder class.

albertshau (approved):
lgtm. please squash commits before merging.

PR #772, opened 2021-09-15
Add support for the Spark BigQuery connector for SQL Engine Pull operations

fernst:
Added dependencies to Spark BQ connector and Spark SQL APIs.

Added support for methods introduced in https://github.com/cdapio/cdap/pull/13664 for the SQL engine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,8 @@
     <junit.version>4.13.1</junit.version>
     <powermock.version>2.0.2</powermock.version>
     <slf4j.version>1.7.5</slf4j.version>
-    <spark.version>2.3.1</spark.version>
+    <spark.version>3.1.1</spark.version>
+    <spark-bq-connector.version>0.22.1</spark-bq-connector.version>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -170,6 +170,27 @@
...
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-api-spark2_2.11</artifactId>
+      <version>${cdap.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.spark</groupId>
+      <artifactId>spark-bigquery-with-dependencies_2.11</artifactId>
+      <version>0.22.0</version>
+    </dependency>
+    <dependency>
```
fernst, line 191: Note that the `javax.inject` dependency is used because otherwise the Spark BQ connector throws class not found exception.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -176,6 +178,13 @@ public void onRunFinish(boolean succeeded, RuntimeContext context) {
     }
   }
 
+  @Override
+  public SparkPushDataset<StructuredRecord> getSparkPushProvider(SQLPushRequest sqlPushRequest)
```
tivv, line 182: This ties SQL Engine to spark. Can be fine for POC, but we would need more structure similar to plugin approach for merging. If we look into running full ELT in future we should not tie to Spark.

PR #773, opened 2021-09-16
[PLUGIN-897] Adding CMEK as a config in GCS Bucket Create, Copy and Move plugins

itsankit-google:
[PLUGIN-897](https://cdap.atlassian.net/browse/PLUGIN-897)
CMEK key as a config in GCP action plugins - GCS Bucket Create, GCS Copy, GCS Move 
Validation should fail with a proper message if they are not in the same location

Subtask-
[PLUGIN-898](https://cdap.atlassian.net/browse/PLUGIN-898)
Support for auto-creation of destination buckets in GCSMove and GCSCopy Plugins if they do not exist.

### DIFF
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -47,9 +47,8 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 
 **Location:** The location where the gcs bucket will get created. This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
-If the bucket already exists, you need to manually apply the CMEK to it.
-The location of bucket and key should be same.
+**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+If the bucket already exists, this is ignored.
 
 **Content Type:** The Content Type entity is used to indicate the media type of the resource.
 Defaults to 'application/octet-stream'. The following table shows valid content types for each format.
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -47,7 +47,7 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 
 **Location:** The location where the gcs bucket will get created. This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+**Encryption Key Name:** Used to encrypt data written to any bucket created by the plugin.
 If the bucket already exists, this is ignored.
 
 **Content Type:** The Content Type entity is used to indicate the media type of the resource.
--- a/docs/GCSBucketCreate-action.md
+++ b/docs/GCSBucketCreate-action.md
@@ -36,6 +36,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Location:** The location where the gcs buckets will get created. This value is ignored if the bucket already exists.
 
+**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
+If the buckets already exist, you need to manually apply the CMEK to it.
+The location of buckets and key should be same.
+
 **Service Account**  - service account key used for authorization
 
 * **File Path**: Path on the local file system of the service account key used for
--- a/docs/GCSBucketCreate-action.md
+++ b/docs/GCSBucketCreate-action.md
@@ -36,9 +36,8 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Location:** The location where the gcs buckets will get created. This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
-If the buckets already exist, you need to manually apply the CMEK to it.
-The location of buckets and key should be same.
+**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+If the bucket already exists, this is ignored.
 
 **Service Account**  - service account key used for authorization
 
--- a/docs/GCSBucketCreate-action.md
+++ b/docs/GCSBucketCreate-action.md
@@ -36,7 +36,7 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Location:** The location where the gcs buckets will get created. This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+**Encryption Key Name:** Used to encrypt data written to any bucket created by the plugin.
 If the bucket already exists, this is ignored.
 
 **Service Account**  - service account key used for authorization
--- a/docs/GCSCopy-action.md
+++ b/docs/GCSCopy-action.md
@@ -46,7 +46,7 @@ When running on other clusters, the file must be present on every node in the cl
 **Location:** The location where the destination gcs bucket will get created. 
 This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+**Encryption Key Name:** Used to encrypt data written to any bucket created by the plugin.
 If the bucket already exists, this is ignored.
 
 Example
--- a/docs/GCSCopy-action.md
+++ b/docs/GCSCopy-action.md
@@ -46,6 +46,10 @@ When running on other clusters, the file must be present on every node in the cl
 **Location:** The location where the destination gcs bucket will get created. 
 This value is ignored if the bucket already exists.
 
+**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
+If the bucket already exists, you need to manually apply the CMEK to it.
+The location of bucket and key should be same.
+
 Example
 -------
 
--- a/docs/GCSCopy-action.md
+++ b/docs/GCSCopy-action.md
@@ -26,7 +26,7 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Source Path**: Path to a source object or directory.
 
-**Destination Path**: Path to the destination. The bucket must already exist.
+**Destination Path**: Path to the destination.
 
 **Copy All Subdirectories**: If the source is a directory, copy all subdirectories.
 
@@ -42,6 +42,9 @@ authorization. Can be set to 'auto-detect' when running on a Dataproc cluster.
 When running on other clusters, the file must be present on every node in the cluster.
 
 * **JSON**: Contents of the service account JSON file.
+
+**Location:** The location where the destination gcs bucket will get created. 
+This value is ignored if the bucket already exists.
 
 Example
 -------
--- a/docs/GCSCopy-action.md
+++ b/docs/GCSCopy-action.md
@@ -46,9 +46,8 @@ When running on other clusters, the file must be present on every node in the cl
 **Location:** The location where the destination gcs bucket will get created. 
 This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
-If the bucket already exists, you need to manually apply the CMEK to it.
-The location of bucket and key should be same.
+**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
+If the bucket already exists, this is ignored.
 
 Example
 -------
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -70,23 +70,23 @@ public final class BigQuerySinkUtils {
    * @param datasetName the name of the dataset
    * @param bucketName the name of the bucket
    * @param location the location of the resources, this is only applied if both the bucket and dataset do not exist
-   * @param cmekKey the name of the cmek key
+   * @param cmekKeyName the name of the cmek key
    * @throws IOException if there was an error creating or fetching any GCP resource
    */
   public static void createResources(BigQuery bigQuery, Storage storage,
                                      String datasetName, String bucketName, @Nullable String location,
-                                     @Nullable String cmekKey) throws IOException {
+                                     @Nullable CryptoKeyName cmekKeyName) throws IOException {
     Dataset dataset = bigQuery.getDataset(datasetName);
     Bucket bucket = storage.get(bucketName);
 
     if (dataset == null && bucket == null) {
-      createBucket(storage, bucketName, location, cmekKey,
+      createBucket(storage, bucketName, location, cmekKeyName,
                    () -> String.format("Unable to create Cloud Storage bucket '%s'", bucketName));
       createDataset(bigQuery, datasetName, location,
                     () -> String.format("Unable to create BigQuery dataset '%s'", datasetName));
     } else if (bucket == null) {
       createBucket(
-        storage, bucketName, dataset.getLocation(), cmekKey,
+        storage, bucketName, dataset.getLocation(), cmekKeyName,
         () -> String.format(
           "Unable to create Cloud Storage bucket '%s' in the same location ('%s') as BigQuery dataset '%s'. "
             + "Please use a bucket that is in the same location as the dataset.",
@@ -132,15 +132,15 @@ public final class BigQuerySinkUtils {
    * @param storage GCS Client.
    * @param bucket Bucket Name.
    * @param location Location for this bucket.
-   * @param cmekKey CMEK key to use for this bucket.
+   * @param cmekKeyName CMEK key to use for this bucket.
    * @param errorMessage Supplier for the error message to output if the bucket could not be created.
    * @throws IOException if the bucket could not be created.
    */
   private static void createBucket(Storage storage, String bucket, @Nullable String location,
-                                   @Nullable String cmekKey, Supplier<String> errorMessage) throws IOException {
+                                   @Nullable CryptoKeyName cmekKeyName,
+                                   Supplier<String> errorMessage) throws IOException {
     try {
-      GCPUtils.createBucket(storage, bucket, location,
-                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
+      GCPUtils.createBucket(storage, bucket, location, cmekKeyName);
     } catch (StorageException e) {
       if (e.getCode() != 409) {
         // A conflict means the bucket already exists
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -26,9 +26,11 @@ import com.google.cloud.hadoop.io.bigquery.BigQueryFileFormat;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryOutputConfiguration;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableSchema;
+import com.google.cloud.kms.v1.CryptoKeyName;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
+import com.google.common.base.Strings;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryTypeSize.Numeric;
@@ -137,7 +139,8 @@ public final class BigQuerySinkUtils {
   private static void createBucket(Storage storage, String bucket, @Nullable String location,
                                    @Nullable String cmekKey, Supplier<String> errorMessage) throws IOException {
     try {
-      GCPUtils.createBucket(storage, bucket, location, cmekKey);
+      GCPUtils.createBucket(storage, bucket, location,
+                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
     } catch (StorageException e) {
       if (e.getCode() != 409) {
         // A conflict means the bucket already exists
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
@@ -22,6 +22,8 @@ import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
+import com.google.cloud.kms.v1.CryptoKeyName;
+import com.google.common.base.Strings;
 import io.cdap.plugin.gcp.bigquery.connector.BigQueryConnectorConfig;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
@@ -87,7 +89,7 @@ public class BigQuerySourceUtils {
       GCPUtils.createBucket(GCPUtils.getStorage(config.getProject(), credentials),
                             bucket,
                             dataset.getLocation(),
-                            cmekKey);
+                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
     }
 
     return bucket;
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
@@ -66,7 +66,7 @@ public class BigQuerySourceUtils {
    * @param bigQuery BigQuery client.
    * @param credentials Google Cloud Credentials.
    * @param bucketPath bucket path to use. Will be used as a bucket name if needed..
-   * @param cmekKey CMEK key to use for the auto-created bucket.
+   * @param cmekKeyName CMEK key to use for the auto-created bucket.
    * @return Bucket name.
    */
   public static String getOrCreateBucket(Configuration configuration,
@@ -74,7 +74,7 @@ public class BigQuerySourceUtils {
                                          BigQuery bigQuery,
                                          Credentials credentials,
                                          String bucketPath,
-                                         @Nullable String cmekKey) {
+                                         @Nullable CryptoKeyName cmekKeyName) {
     String bucket = config.getBucket();
 
     // Create a new bucket if needed
@@ -88,8 +88,7 @@ public class BigQuerySourceUtils {
       Dataset dataset = bigQuery.getDataset(DatasetId.of(config.getDatasetProject(), config.getDataset()));
       GCPUtils.createBucket(GCPUtils.getStorage(config.getProject(), credentials),
                             bucket,
-                            dataset.getLocation(),
-                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
+                            dataset.getLocation(), cmekKeyName);
     }
 
     return bucket;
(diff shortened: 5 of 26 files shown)


docs/GCSBucketCreate-action.md
```diff
@@ -36,6 +36,10 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Location:** The location where the gcs buckets will get created. This value is ignored if the bucket already exists.
 
+**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
```
albertshau: Used to encrypt data written to any bucket created by the plugin. If the bucket already exists, this is ignored.

docs/GCSCopy-action.md
```diff
@@ -26,7 +26,7 @@ It can be found on the Dashboard in the Google Cloud Platform Console.
 
 **Source Path**: Path to a source object or directory.
 
-**Destination Path**: Path to the destination. The bucket must already exist.
+**Destination Path**: Path to the destination.
```
albertshau, line 29: this is a change in behavior. Technically it is a backwards incompatible change, as somebody could have been relying on this behavior in previous versions (though this seems pretty unlikely).

@bdmogal is this intended?

Also commented on the code below, but this means an extra check is need to see if the bucket already exists. If that requires broader permissions than used to be needed, it is also a backwards incompatible change.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -43,7 +43,7 @@
  */
 public class StorageClient {
   private static final Logger LOG = LoggerFactory.getLogger(StorageClient.class);
-  private final Storage storage;
+  public final Storage storage;
```
albertshau: should almost never have a public variable, it breaks encapsulation.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -201,9 +207,27 @@ public void run(ActionContext context) throws Exception {
...
     @Nullable
     @Description("The location where the gcs buckets will get created. " +
-                   "This value is ignored if the bucket already exists.")
+      "This value is ignored if the bucket already exists.")
     protected String location;
 
+    @Name(NAME_CMEK_KEY)
+    @Macro
+    @Nullable
+    @Description("The GCP customer managed encryption key (CMEK) name used by Cloud Dataproc")
```
albertshau: update this description. Dataproc has nothing to do with this.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
```diff
@@ -221,7 +245,136 @@ void validate(FailureCollector collector) {
...
+      }
+
+      if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
+        validateCmekKey(collector);
+      }
+      collector.getOrThrowException();
+    }
+
+    //This method validated the pattern of CMEK Key resource ID.
+    void validateCmekKey(FailureCollector failureCollector) {
```
albertshau, line 261: there is a lot of duplicate code. Please refactor so it is in one place.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
```diff
@@ -57,8 +62,36 @@ public void run(ActionContext context) throws IOException {
...
+    GCSPath undo = null;
+    Storage storage = storageClient.storage;
+
+    // create the destination bucket if not exist
+    Bucket bucket = null;
+    try {
+      bucket = storage.get(destPath.getBucket());
+    } catch (StorageException e) {
+      throw new RuntimeException(
+        String.format("Unable to access or create bucket %s. ", destPath.getBucket())
```
albertshau: this message is wrong, it is not trying to create the bucket.

Also, is this an additional permission that we're adding (which would be backwards incompatible)? Seems like the plugin didn't need read access before, it just needed to be able to write?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSMove.java
```diff
@@ -57,8 +62,35 @@ public void run(ActionContext context) throws IOException {
...
+    GCSPath undo = null;
+    Storage storage = storageClient.storage;
+
+    // create the destination bucket if not exist
+    Bucket bucket = null;
+    try {
+      bucket = storage.get(destPath.getBucket());
+    } catch (StorageException e) {
+      throw new RuntimeException(
+        String.format("Unable to access or create bucket %s. ", destPath.getBucket())
```
albertshau: similar comment about the message and similar question about whether this change means the plugin needs extra permissions

src/main/java/io/cdap/plugin/gcp/gcs/actions/SourceDestConfig.java
```diff
@@ -47,6 +56,31 @@
...
+  @Macro
+  @Nullable
+  @Description("The location where the gcs buckets will get created. " +
+    "This value is ignored if the bucket already exists.")
+  protected String location;
+
+  @Name(NAME_CMEK_KEY)
+  @Macro
+  @Nullable
+  @Description("The GCP customer managed encryption key (CMEK) name used by Cloud Dataproc")
```
albertshau: similar comment, fix this description

src/main/java/io/cdap/plugin/gcp/gcs/actions/SourceDestConfig.java
```diff
@@ -79,6 +113,127 @@ public void validate(FailureCollector collector) {
...
       }
     }
+    if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
+      validateCmekKey(collector);
+    }
     collector.getOrThrowException();
   }
+
+  //This method validated the pattern of CMEK Key resource ID.
+  void validateCmekKey(FailureCollector failureCollector) {
```
albertshau, line 125: similar comment, code duplication is bad

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
itsankit-google: No, we don't need any additional permission to access the bucket. If it has write permission it will be able to access the bucket. Will be changing the message to "Unable to access the bucket"

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketCreate.java
itsankit-google: Refactored the code and moved that into functions in base class

docs/GCSCopy-action.md
bdmogal: @albertshau so long as the bucket is created only if it doesn't already exist (per your comment that check should be added), it should be fine, right?


We made this change so the behavior across the GCS plugins is exactly the same.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
albertshau: Are you sure? They look like different permissions based on https://cloud.google.com/storage/docs/access-control/iam-permissions. Did you actually try it out?

docs/GCSCopy-action.md
albertshau: It's a change in behavior, so it's possible somebody was relying on this behavior and expecting their pipeline to fail if the bucket doesn't exist. However, this seems unlikely, so I'm ok with changing it.

The bigger concern was about whether this change means the plugin now needs additional permissions. @itsankit-google  I didn't see an answer on whether the existing plugin is already checking bucket existence or not.

itsankit-google: Yeah, so basically that was my point that the existing plugin is already checking bucket existence in StorageClient class. @albertshau Here is the link to it, this method is used to perform the actual action of copy and move.
https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java#L160

albertshau: ok thanks, then there is no issue.

docs/GCSBucketCreate-action.md
albertshau: still have this comment

docs/GCS-batchsink.md
```diff
@@ -47,9 +47,8 @@ The delimiter will be ignored if the format is anything other than 'delimited'.
 
 **Location:** The location where the gcs bucket will get created. This value is ignored if the bucket already exists.
 
-**Encryption Key Name:** It is used to encrypt data written to bucket that will be created by pipeline.
-If the bucket already exists, you need to manually apply the CMEK to it.
-The location of bucket and key should be same.
+**Encryption Key Name:** It is used to encrypt data written to any bucket created by the plugin.
```
albertshau: It is used -> Used

same for other places.

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -148,4 +156,65 @@ public String getCmekKey(Arguments arguments) {
...
+  }
+
+  public Credentials getCredentials() {
+    Boolean isServiceAccountFilePath = isServiceAccountFilePath();
+    Credentials credentials = null;
+    try {
+      credentials = getServiceAccount() == null ?
+        null : GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath);
+    } catch (Exception e) {
+      /* Ignoring the exception because we don't want to highlight config if an exception occurs while
```
albertshau: this is not appropriate for a utility method, correct handling of the exception depends on what you're trying to do. For example, at pipeline runtime you would never want to ignore the exception.

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -148,4 +156,65 @@ public String getCmekKey(Arguments arguments) {
...
+    try {
+      cmekKeyName = CryptoKeyName.parse(key);
+    } catch (ValidationException e) {
+      collector.addFailure(e.getMessage(), null)
+        .withConfigProperty(NAME_CMEK_KEY).withStacktrace(e.getStackTrace());
+    }
+    return cmekKeyName;
+  }
+
+  public boolean requiredFieldsContainsMacro() {
```
albertshau: none of these are required fields, let's rename to something more descriptive, like `projectOrServiceAccountContainsMacro()`

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -148,4 +156,65 @@ public String getCmekKey(Arguments arguments) {
     }
     return arguments.get("gcp.cmek.key.name");
   }
+
+  public CryptoKeyName parseCmekKey(String key, FailureCollector collector) {
```
albertshau: should have just one `CryptoKeyName getCmekKey(Arguments arguments, FailureCollector collector)` method instead of one that returns a string and one that returns CryptoKeyName

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -148,4 +156,65 @@ public String getCmekKey(Arguments arguments) {
...
+      credentials = getServiceAccount() == null ?
+        null : GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath);
+    } catch (Exception e) {
+      /* Ignoring the exception because we don't want to highlight config if an exception occurs while
+        loading credentials */
+    }
+    return credentials;
+  }
+
+  public boolean validateCmekKeyAndBucketLocation(Storage storage, String path, CryptoKeyName cmekKey,
```
albertshau: this should not be in GCPConfig, as it is specific to GCS. Let's create a CmekKeys utility class and put it as a static method there.

albertshau: this method should take more focused input `validateKeyAndBucketLocation(Storage storage, GCSPath writePath, CryptoKeyName, @Nullable String bucketLocation, FailureCollector collector)` and have a javadoc that describes what the input is and what this method does. For example, detailing what the bucket location is (location to create the bucket in if it doesn't exist) and when should somebody call this method.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -45,6 +45,10 @@
   private static final Logger LOG = LoggerFactory.getLogger(StorageClient.class);
   private final Storage storage;
 
+  public Storage getStorage() {
```
albertshau: should not have this. Instead, make the constructor public and make the caller pass in the Storage class and use it outside of this class.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
```diff
@@ -57,8 +62,36 @@ public void run(ActionContext context) throws IOException {
...
                                                        isServiceAccountFilePath);
-    //noinspection ConstantConditions
-    storageClient.copy(config.getSourcePath(), config.getDestPath(), config.recursive, config.shouldOverwrite());
+
+    GCSPath destPath = config.getDestPath();
+    GCSPath undo = null;
+    Storage storage = storageClient.getStorage();
+
+    // create the destination bucket if not exist
+    Bucket bucket = null;
```
albertshau: this code is the same in GCSCopy and GCSMove. Can it be moved to the StorageClient so there isn't duplicate code?

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
itsankit-google: Both methods have different use cases. `parseCmekKey()` is used to validate cmek key formatted string while `getCmekKey()` is used to fetch the cmek key formatted string. I will move the `parseCmekKey()` method to CmekUtils class. This will resolve the conflict.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
itsankit-google: will do it in future commit

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
itsankit-google: I will not need this because will be moving that part of code to StorageClient class itself.

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
itsankit-google: Yeah, so I looked how it is done in existing code and made the changes accordingly.

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
albertshau: Is there is no reason to get the raw string anymore? Seems like the caller should always get a valid object (from which you can get the resource uri if needed)

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+public class CmekUtils {
+
+  /**
+   * This method validates the cmek formatted string.
+   *
+   * @param key cmek key formatted string
+   * @param collector  failure collector
+   * @return parsed CryptoKeyName object.
+   */
+  public static CryptoKeyName parseCmekKey(String key, FailureCollector collector) {
```
albertshau: please see other comment. Don't think there is a good reason for any caller to get the raw unvalidated string from GCPConfig

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param storage GCS storage
+   * @param path the path of bucket to validate
+   * @param cmekKey the cmek key name
+   * @param location the location to create the bucket in if it doesn't exists
+   * @param collector failure collector
+   * @return true if bucket does not exist, otherwise false
+   */
+  public static boolean validateCmekKeyAndBucketLocation(Storage storage, String path, CryptoKeyName cmekKey,
+                                                  @Nullable String location, FailureCollector collector) {
```
albertshau: fix indentation

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+   * This method validates the location of cmek key and bucket if it doesn't exists.
+   *
+   * @param storage GCS storage
+   * @param path the path of bucket to validate
+   * @param cmekKey the cmek key name
+   * @param location the location to create the bucket in if it doesn't exists
+   * @param collector failure collector
+   * @return true if bucket does not exist, otherwise false
+   */
+  public static boolean validateCmekKeyAndBucketLocation(Storage storage, String path, CryptoKeyName cmekKey,
```
albertshau: as mentioned in the previous comment, path should be a GCSPath not a string. Don't want this method to throw an exception from an invalid path.

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+                                                  @Nullable String location, FailureCollector collector) {
+    GCSPath gcsPath = GCSPath.from(path);
+    Bucket bucket = null;
+    try {
+      bucket = storage.get(gcsPath.getBucket());
+    } catch (StorageException e) {
+          /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
```
albertshau: fix indentation

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+      cmekKeyName = CryptoKeyName.parse(key);
+    } catch (ValidationException e) {
+      collector.addFailure(e.getMessage(), null)
+        .withConfigProperty(GCPConfig.NAME_CMEK_KEY).withStacktrace(e.getStackTrace());
+    }
+    return cmekKeyName;
+  }
+
+  /**
+   * This method validates the location of cmek key and bucket if it doesn't exists.
```
albertshau: more detail here.

This method checks if the GCS bucket exists. If not, it checks that the CMEK key is in the same location that the bucket will be created in. If there is any exception when checking for bucket existence, it will be ignored.

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param storage GCS storage
+   * @param path the path of bucket to validate
+   * @param cmekKey the cmek key name
+   * @param location the location to create the bucket in if it doesn't exists
+   * @param collector failure collector
+   * @return true if bucket does not exist, otherwise false
```
albertshau: this is wrong, it also returns false if you can't tell if the bucket exists. Why does it need to return anything?

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -44,6 +44,7 @@
 public class StorageClient {
   private static final Logger LOG = LoggerFactory.getLogger(StorageClient.class);
   private final Storage storage;
+  private List<GCSPath> undoBucket = new ArrayList<>();
```
albertshau: This makes the StorageClient stateful, which is bad. It assumes somebody will call methods in a certain order and behaves incorrectly otherwise.

Just make createBucket return the Bucket it created, or null if it didn't do anything

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -109,6 +110,37 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
   }
 
+  /**
+   * Creates the given bucket if it does not exists.
+   *
+   * @param path the path of the bucket 
+   * @param location the location of bucket
+   * @param cmekKey  the name of the cmek key
+   */
+  public void createBucket(GCSPath path, @Nullable String location, @Nullable String cmekKey) {
```
albertshau: createBucketIfNotExists

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -109,6 +110,37 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
+    if (bucket == null) {
+      GCPUtils.createBucket(storage, path.getBucket(), location, cmekKey);
+      undoBucket.add(path);
+    }
+  }
+
+  /**
+   * Deletes any bucket created by the client. It is called when the operation failed.
+   */
+  public void deleteBucket() {
```
albertshau: don't need this

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSCopy.java
```diff
@@ -64,31 +64,15 @@ public void run(ActionContext context) throws IOException {
...
+    storageClient.createBucket(destPath, config.location, config.getCmekKey(context.getArguments()));
 
     try {
       storageClient.copy(config.getSourcePath(), config.getDestPath(), config.recursive, config.shouldOverwrite());
     } catch (Exception e) {
       //deleting the destination bucket that was auto-created if the operation failed.
-      if (undo != null) {
-        storage.delete(undo.getBucket());
-      }
+      storageClient.deleteBucket();
```
albertshau: if this fails, we should log a warning and add the exception as a suppressed exception.

```
try {
  storage.delete(undo.getBucket());
} catch (Exception e2) {
  // log
  e.addSuppressed(e2);
}
```

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -143,12 +145,28 @@ public boolean autoServiceAccountUnavailable() {
...
-      return configKey;
+  /**
+   * This method validates the cmek key raw string.
+   *
+   * @param cmekKey cmek key raw string
+   * @param arguments runtime arguments
+   * @param collector  failure collector
+   * @return parsed CryptoKeyName object.
+   */
+  public CryptoKeyName getCmekKey(@Nullable String cmekKey, @Nullable Arguments arguments, FailureCollector collector) {
```
albertshau: annotate the method as nullable and include in the javadocs when it will return null

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -143,12 +145,28 @@ public boolean autoServiceAccountUnavailable() {
...
+    if (Strings.isNullOrEmpty(cmekKey)) {
+      cmekKey = arguments == null ? null : arguments.get("gcp.cmek.key.name");
+    }
+    CryptoKeyName cmekKeyName = null;
+    if (!Strings.isNullOrEmpty(cmekKey)) {
```
albertshau: nit: as a general pattern, it's more readable to check for things that let you return immediately rather than put a more code in an if block.

```
if (Strings.isNullOrEmpty(cmekKey)) {
  return null;
}
...
```

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -122,7 +123,8 @@ public void prepareRun(BatchSinkContext context) throws Exception {
           + "Ensure you entered the correct bucket path and have permissions for it.", e);
     }
     if (bucket == null) {
-      GCPUtils.createBucket(storage, config.getBucket(), config.getLocation(), cmekKey);
+      GCPUtils.createBucket(storage, config.getBucket(), config.getLocation(),
+                            cmekKeyName == null ? null : cmekKeyName.toString());
```
albertshau: can you change GCPUtils.createBucket to take a CryptoKeyName instead of a string? Then you don't need this null check in multiple places.  In general, we should be passing the object around instead of a raw string where possible.

albertshau (commented):
minor comments remaining

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
itsankit-google: done

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -137,7 +139,8 @@ private static void createDataset(BigQuery bigQuery, String dataset, @Nullable S
   private static void createBucket(Storage storage, String bucket, @Nullable String location,
                                    @Nullable String cmekKey, Supplier<String> errorMessage) throws IOException {
     try {
-      GCPUtils.createBucket(storage, bucket, location, cmekKey);
+      GCPUtils.createBucket(storage, bucket, location,
+                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
```
albertshau: should not parse it here, it should be passed in.

As a general rule, any time we are validating user input, it should happen as early as possible (usually right when fetching the value from the config/arguments). After that point, the validated object should be passed around instead of the raw user input.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -87,7 +89,7 @@ public static String getOrCreateBucket(Configuration configuration,
       GCPUtils.createBucket(GCPUtils.getStorage(config.getProject(), credentials),
                             bucket,
                             dataset.getLocation(),
-                            cmekKey);
+                            Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
```
albertshau: similar comment here

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSDoneFileMarker.java
```diff
@@ -172,7 +173,8 @@ private static void createFileMarker(String project, GCSPath path, String servic
     Storage storage = GCPUtils.getStorage(project, credentials);
     if (storage.get(path.getBucket()) == null) {
       try {
-        GCPUtils.createBucket(storage, path.getBucket(), null, cmekKey);
+        GCPUtils.createBucket(storage, path.getBucket(), null,
+                              Strings.isNullOrEmpty(cmekKey) ? null : CryptoKeyName.parse(cmekKey));
```
albertshau: similar comment here

src/main/java/io/cdap/plugin/gcp/common/GCPConfig.java
```diff
@@ -145,32 +145,6 @@ public boolean autoServiceAccountUnavailable() {
...
-  /**
-   * This method validates the cmek key formatted string.
-   *
-   * @param cmekKey cmek key raw string
-   * @param arguments runtime arguments
-   * @param collector  failure collector
-   * @return parsed CryptoKeyName object if the formatted string is valid otherwise null.
-   */
-  @Nullable
-  public CryptoKeyName getCmekKey(@Nullable String cmekKey, @Nullable Arguments arguments, FailureCollector collector) {
```
albertshau: it's better to keep this here and have it call the utils method. Except the cmekKey value should come from the plugin properties like before instead of passed in (i.e. `getCmekKey(@Nullable Arguments arguments, FailureCollector collector)`). This way the raw unvalidated cmek key is not exposed.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
itsankit-google: It will be fixed in the upcoming PR. As I will be adding cmek as a config in these plugins as well and then they will directly get the CryptoKeyName object. All this happened because I changed the createBucket() parameter to accept CryptoKeyName object instead of string. Is it okay?

albertshau (approved):
lgtm, please squash commits before merging.

PR #774, opened 2021-09-16
PLUGIN-899 fix custom formats

albertshau:
Fixed a bug where custom formats would be disallowed during
plugin validation.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/Formats.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/Formats.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.gcs;
+
+import io.cdap.plugin.format.FileFormat;
+
+/**
+ * Utility class for GCS formats.
+ */
+public class Formats {
+
+  private Formats() {
+    // no-op
+  }
+
+  /**
+   * Get the plugin name for the given format. Older versions of the source and sink allowed case insensitive
+   * format names. For example, 'Avro' and 'avro' were treated the same.
+   *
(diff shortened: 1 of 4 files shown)


CuriousVini (approved):

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -493,8 +485,8 @@ public String getPath() {
     }
 
     @Override
-    public FileFormat getFormat() {
-      return FileFormat.from(format, FileFormat::canWrite);
```
tivv, line 497: you are dropping `::canWrite` here. Is it intended?

src/main/java/io/cdap/plugin/gcp/gcs/Formats.java
```diff
@@ -0,0 +1,50 @@
...
+ * Utility class for GCS formats.
+ */
+public class Formats {
+
+  private Formats() {
+    // no-op
+  }
+
+  /**
+   * Get the plugin name for the given format. Older versions of the source and sink allowed case insensitive
```
tivv, line 31: If it returns plugin name, should it return "format-avro"? It's the artifact name I believe

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -493,8 +485,8 @@ public String getPath() {
...
+    public String getFormatName() {
```
tivv, line 488: How is it going to be used?

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
albertshau: yes, this method is unused code.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
albertshau: this gets used by https://github.com/cdapio/hydrator-plugins/blob/19a6d06a538714a5fdcde15eac9084b5bfe2fbba/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java#L90, it just asks CDAP for a plugin with this name

src/main/java/io/cdap/plugin/gcp/gcs/Formats.java
albertshau: we want the plugin name and not the artifact name, which is what gets passed to CDAP's usePlugin() API call at https://github.com/cdapio/hydrator-plugins/blob/19a6d06a538714a5fdcde15eac9084b5bfe2fbba/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java#L124

tivv (approved):

PR #777, opened 2021-09-23
[PLUGIN-733] Use cursor pagination for Datastore API

nliu132:
Use cursor pagination when reading from Datastore API in order to handle when response contains over 300 records

JIRA: https://cdap.atlassian.net/browse/PLUGIN-733

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
+++ b/src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
@@ -23,6 +23,7 @@ import com.google.datastore.v1.QueryResultBatch;
 import com.google.datastore.v1.RunQueryRequest;
 import com.google.datastore.v1.client.Datastore;
 import com.google.datastore.v1.client.DatastoreException;
+import com.google.protobuf.ByteString;
 import io.cdap.plugin.gcp.datastore.source.util.DatastoreSourceConstants;
 import io.cdap.plugin.gcp.datastore.util.DatastoreUtil;
 import org.apache.hadoop.conf.Configuration;
@@ -35,7 +36,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 
 /**
  * Datastore read reader instantiates a record reader that will read the entities from Datastore,
@@ -45,8 +48,11 @@ public class DatastoreRecordReader extends RecordReader<LongWritable, Entity> {
 
   private static final Logger LOG = LoggerFactory.getLogger(DatastoreRecordReader.class);
 
+  private Datastore datastore;
   private Iterator<EntityResult> results;
   private Entity entity;
+  private List<EntityResult> entityResultList;
+  private int batchSerializedSize;
   private long index;
   private LongWritable key;
 
@@ -54,7 +60,7 @@ public class DatastoreRecordReader extends RecordReader<LongWritable, Entity> {
   public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptContext) throws IOException {
     Configuration config = taskAttemptContext.getConfiguration();
     Query query = ((QueryInputSplit) inputSplit).getQuery();
-    Datastore datastore = DatastoreUtil.getDatastoreV1(
+    datastore = DatastoreUtil.getDatastoreV1(
       config.get(DatastoreSourceConstants.CONFIG_SERVICE_ACCOUNT),
       config.getBoolean(DatastoreSourceConstants.CONFIG_SERVICE_ACCOUNT_IS_FILE, true),
       config.get(DatastoreSourceConstants.CONFIG_PROJECT));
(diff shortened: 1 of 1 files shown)


nliu132:
Tested in Sandbox env with a simple Datastore source -> File sink pipeline. Datastore DB contains 401 entities:
- when run as is, only 300 entities are processed
image
```
 ~ $ cat 2021-09-22-17-34/part-r-00000 | wc
       300     299    1496
```

- when run with this branch, all of the Datastore entities are processed
image
```
 ~ $ cat 2021-09-22-17-41/part-r-00000 | wc
     401     400    2001
```

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -67,14 +73,30 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
-      results = batch.getEntityResultsList().iterator();
+      batchSerializedSize += batch.getSerializedSize();
+      entityResultList.addAll(batch.getEntityResultsList());
+      LOG.debug("Batch moreResultsType: {}", batch.getMoreResults());
+      // Datastore API only returns up to 300 items. Need to use cursor pagination if there is more results
+      if (batch.getMoreResults() != QueryResultBatch.MoreResultsType.NO_MORE_RESULTS) {
+        ByteString cursor = batch.getEndCursor();
+        query = query.toBuilder().setStartCursor(cursor).build();
+        request = request.toBuilder().setQuery(query).build();
+        createEntityResultsList(query, request);
```
tivv: Please don't use recursion instead of loop. You can easily run out of stack

tivv: Also, does it mean we need to read the whole split before serving it? It does not seem optimal

nliu132: I think so. It seems like the API only supports setting cursors at the query level (and not the request/response level)

tivv: And that's fine. Just don't use recursion a run next batch in "nextKeyValue" call, not "initialize" call. This would ensure we don't accumulate data in memory.
Also don't accumulate the counter, but call `increment` as needed

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -59,21 +62,33 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
+      .setNamespaceId(config.get(DatastoreSourceConstants.CONFIG_NAMESPACE))
+      .setProjectId(config.get(DatastoreSourceConstants.CONFIG_PROJECT));
+    List<EntityResult> entityResultList = new ArrayList<>();
+    QueryResultBatch batch;
     try {
-      QueryResultBatch batch = datastore.runQuery(request).getBatch();
-      taskAttemptContext.getCounter(FileInputFormatCounter.BYTES_READ).increment(batch.getSerializedSize());
-      results = batch.getEntityResultsList().iterator();
+      // Datastore API only returns up to 300 items. Need to use cursor pagination if there is more results
+      do {
```
chtyim: Why this is done in the initialize method? Shouldn't reading of the actual data be done in the `nextKeyValue()` method?

nliu132: I added it here since the original `initialize()` was already loading in the first page of data, but I can see how this could be sub-optimal if there's a lot of records being loaded in. Would it be better to 1) not load any data in `initialize()` anymore (so the first page of records is pulled when `nextKeyValue()` is first called) or 2) load the first page of data when `initialize()` is called and load subsequent pages with `nextKeyValue()`?

avirmani: I think the latter will be better, 1st batch can done in initialize.

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -45,47 +47,52 @@
...
-      QueryResultBatch batch = datastore.runQuery(request).getBatch();
-      taskAttemptContext.getCounter(FileInputFormatCounter.BYTES_READ).increment(batch.getSerializedSize());
-      results = batch.getEntityResultsList().iterator();
-    } catch (DatastoreException e) {
-      throw new IOException("Failed to run query", e);
-    }
+    partitionIdBuilder = PartitionId.newBuilder()
+      .setNamespaceId(config.get(DatastoreSourceConstants.CONFIG_NAMESPACE))
+      .setProjectId(config.get(DatastoreSourceConstants.CONFIG_PROJECT));
+    queryBuilder = query.toBuilder();
```
tivv: I am not sure about this case, but overall builders can't be reused, so storing them as a field can be problematic. Please create new builders for each query run. You can reuse code using methods as needed.

src/main/java/io/cdap/plugin/gcp/datastore/source/DatastoreRecordReader.java
```diff
@@ -107,4 +114,24 @@ public float getProgress() {
...
+      .setQuery(queryBuilder)
+      // partition id needs to be set in the RunQueryRequest in addition to being passed to QuerySplitter.getSplits.
+      // This is a quirk of the V1 API.
+      .setPartitionId(partitionIdBuilder)
+      .build();
+    try {
+      QueryResultBatch batch = datastore.runQuery(request).getBatch();
+      batchSizeCounter.increment(batch.getSerializedSize());
+      lastBatchMoreResultsType = batch.getMoreResults();
+      LOG.debug("Batch page moreResultsType: {}", lastBatchMoreResultsType);
```
tivv: Could you please rephrase, so that looking at the logs one could tell what is this line about without looking at the code? E.g. something along the lines of `Loaded Datastore batch of size {}, more results status: {}`. Also, are we going to get a lot of those (300 seems like a small number). If so, more it to trace, we don't want to spam the logs.

tivv (approved):

PR #780, opened 2021-09-29
Plugin - 883 Adding cmek as a config in GCP Plugins 

itsankit-google:
[PLUGIN-883](https://cdap.atlassian.net/browse/PLUGIN-883)

### DIFF
--- a/docs/BigQueryExecute-action.md
+++ b/docs/BigQueryExecute-action.md
@@ -43,6 +43,9 @@ cache that will be flushed whenever tables in the query are modified.
 
 **Job Location**: Location of the job. It must match the location of the dataset specified in the query.
 
+**Encryption Key Name**: Used to encrypt data written to any dataset or table created by the plugin.
+If the dataset or table already exists, this is ignored.
+
 **Row As Arguments**: Row as arguments. For example, if the query is 'select min(id) as min_id, max(id) as max_id from my_dataset.my_table',
 an arguments for 'min_id' and 'max_id' will be set based on the query results. Plugins further down the pipeline can then
 reference these values with macros ${min_id} and ${max_id}.
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -47,6 +47,9 @@ Should only be used with the Insert operation.
 **Location:** The location where the big query datasets will get created. This value is ignored
 if the dataset or temporary bucket already exist.
 
+**Encryption Key Name**: Used to encrypt data written to any bucket, dataset or table created by the plugin.
+If the bucket, dataset or table already exists, this is ignored.
+
 **Split Field:** The name of the field that will be used to determine which table to write to. If unspecified, it
 defaults to `tablename`. This field is used when Allow Flexible Schemas in Output configuration is enabled.
 
--- a/docs/BigQueryTable-batchsource.md
+++ b/docs/BigQueryTable-batchsource.md
@@ -63,6 +63,9 @@ materialized. Defaults to the same dataset in which the view is located.
 Temporary data will be deleted after it has been read. If it is not provided, a unique bucket will be
 created and then deleted after the run finishes.
 
+**Encryption Key Name**: Used to encrypt data written to any bucket created by the plugin.
+If the bucket already exists, this is ignored.
+
 **Service Account**  - service account key used for authorization
 
 * **File Path**: Path on the local file system of the service account key used for
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -19,6 +19,8 @@ package io.cdap.plugin.gcp.bigquery.action;
 import com.google.auth.Credentials;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.EncryptionConfiguration;
 import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.FieldValueList;
@@ -30,6 +32,7 @@ import com.google.cloud.bigquery.QueryJobConfiguration;
 import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableResult;
+import com.google.cloud.kms.v1.CryptoKeyName;
 import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
@@ -38,7 +41,9 @@ import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.action.Action;
 import io.cdap.cdap.etl.api.action.ActionContext;
+import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
+import io.cdap.plugin.gcp.common.CmekUtils;
 import io.cdap.plugin.gcp.common.GCPUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -74,10 +79,11 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
     }
 
-    String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
-    if (cmekKey != null) {
+    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), context.getFailureCollector());
+    context.getFailureCollector().getOrThrowException();
+    if (cmekKeyName != null) {
       builder.setDestinationEncryptionConfiguration(
-        EncryptionConfiguration.newBuilder().setKmsKeyName(cmekKey).build());
+        EncryptionConfiguration.newBuilder().setKmsKeyName(cmekKeyName.toString()).build());
     }
 
     // Save the results of the query to a permanent table.
@@ -100,9 +106,17 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
     // API request - starts the query.
     Credentials credentials = config.getServiceAccount() == null ?
-                                null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
-                                                                              config.isServiceAccountFilePath());
+      null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
+                                                    config.isServiceAccountFilePath());
     BigQuery bigQuery = GCPUtils.getBigQuery(config.getProject(), credentials);
+    //create dataset to store the results if not exists
+    String datasetName = config.getDataset();
+    String tableName = config.getTable();
+    if (!Strings.isNullOrEmpty(datasetName) && !Strings.isNullOrEmpty(tableName)) {
+      BigQuerySinkUtils.createDataset(bigQuery, DatasetId.of(datasetName), config.getLocation(), cmekKeyName,
+                                      () -> String.format("Unable to create BigQuery dataset '%s'", datasetName));
+    }
+
     Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
 
     LOG.info("Executing SQL as job {}.", jobId.getJob());
@@ -159,11 +173,12 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     private static final String SQL = "sql";
     private static final String DATASET = "dataset";
     private static final String TABLE = "table";
+    private static final String NAME_LOCATION = "location";
 
     @Description("Dialect of the SQL command. The value must be 'legacy' or 'standard'. " +
-                   "If set to 'standard', the query will use BigQuery's standard SQL: " +
-                   "https://cloud.google.com/bigquery/sql-reference/. If set to 'legacy', BigQuery's legacy SQL " +
-                   "dialect will be used for this query.")
+      "If set to 'standard', the query will use BigQuery's standard SQL: " +
+      "https://cloud.google.com/bigquery/sql-reference/. If set to 'legacy', BigQuery's legacy SQL " +
+      "dialect will be used for this query.")
     @Macro
     private String dialect;
 
@@ -174,10 +189,10 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
     @Name(MODE)
     @Description("Mode to execute the query in. The value must be 'batch' or 'interactive'. " +
-                   "An interactive query is executed as soon as possible and counts towards the concurrent rate " +
-                   "limit and the daily rate limit. A batch query is queued and started as soon as idle resources " +
-                   "are available, usually within a few minutes. If the query hasn't started within 3 hours, " +
-                   "its priority is changed to 'interactive'")
+      "An interactive query is executed as soon as possible and counts towards the concurrent rate " +
+      "limit and the daily rate limit. A batch query is queued and started as soon as idle resources " +
+      "are available, usually within a few minutes. If the query hasn't started within 3 hours, " +
+      "its priority is changed to 'interactive'")
     @Macro
     private String mode;
 
@@ -185,6 +200,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     @Macro
     private String useCache;
 
+    @Name(NAME_LOCATION)
     @Description("Location of the job. Must match the location of the dataset specified in the query. Defaults to 'US'")
     @Macro
     private String location;
@@ -201,11 +217,18 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     @Nullable
     private String table;
 
+    @Name(NAME_CMEK_KEY)
+    @Macro
+    @Nullable
+    @Description("The GCP customer managed encryption key (CMEK) name used to encrypt data written to " +
+      "any dataset/table created by the plugin. If the dataset/table already exists, this is ignored.")
+    private String cmekKey;
+
     @Description("Row as arguments. For example, if the query is " +
-                   "'select min(id) as min_id, max(id) as max_id from my_dataset.my_table'," +
-                   "an arguments for 'min_id' and 'max_id' will be set based on the query results. " +
-                   "Plugins further down the pipeline can then" +
-                   "reference these values with macros ${min_id} and ${max_id}.")
+      "'select min(id) as min_id, max(id) as max_id from my_dataset.my_table'," +
+      "an arguments for 'min_id' and 'max_id' will be set based on the query results. " +
+      "Plugins further down the pipeline can then" +
+      "reference these values with macros ${min_id} and ${max_id}.")
     @Macro
     private String rowAsArguments;
 
@@ -213,6 +236,19 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       this.dialect = dialect;
       this.sql = sql;
       this.mode = mode;
+    }
+
+    private Config(@Nullable String project, @Nullable String serviceAccountType, @Nullable String serviceFilePath,
+                   @Nullable String serviceAccountJson, @Nullable String dataset, @Nullable String table,
+                   @Nullable String location, @Nullable String cmekKey) {
+      this.project = project;
+      this.serviceAccountType = serviceAccountType;
+      this.serviceFilePath = serviceFilePath;
+      this.serviceAccountJson = serviceAccountJson;
+      this.dataset = dataset;
+      this.table = table;
+      this.location = location;
+      this.cmekKey = cmekKey;
     }
 
     public boolean isLegacySQL() {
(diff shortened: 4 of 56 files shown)


docs/BigQueryTable-batchsource.md
```diff
@@ -63,6 +63,9 @@ materialized. Defaults to the same dataset in which the view is located.
 Temporary data will be deleted after it has been read. If it is not provided, a unique bucket will be
 created and then deleted after the run finishes.
 
+**Encryption Key Name**: Used to encrypt data written to any bucket created by the plugin.
+If the bucket already exists, this is ignored.
```
albertshau, line 67: bucket or dataset?

docs/BigQueryMultiTable-batchsink.md
```diff
@@ -47,6 +47,9 @@ Should only be used with the Insert operation.
 **Location:** The location where the big query datasets will get created. This value is ignored
 if the dataset or temporary bucket already exist.
 
+**Encryption Key Name**: Used to encrypt data written to any resource created by the plugin.
```
albertshau: resource is a bit vague, better to list the resources. Should just be a GCS bucket and BQ table?

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -215,6 +239,20 @@ public Config(String dialect, String sql, String mode) {
       this.mode = mode;
     }
 
+    @VisibleForTesting
+    public Config(@Nullable String project, @Nullable String serviceAccountType, @Nullable String serviceFilePath,
```
albertshau: this should be private, as we want the Builder to always be used.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -289,9 +327,37 @@ public void validate(FailureCollector failureCollector) {
...
+      BigQuery bigQuery = getBigQuery(failureCollector);
+      if (bigQuery == null) {
+        return;
+      }
+      Dataset dataset = CmekUtils.validateCmekKeyAndDatasetLocation(bigQuery, datasetId, cmekKeyName, location,
+                                                                    failureCollector);
+      if (dataset == null) {
+        return;
+      }
+      CmekUtils.validateCmekKeyAndTableLocation(bigQuery, tableId, cmekKeyName, dataset.getLocation(),
```
albertshau: Is this needed? Isn't the table and dataset location always the same?

docs/BigQueryExecute-action.md
```diff
@@ -43,6 +43,9 @@ cache that will be flushed whenever tables in the query are modified.
 
 **Job Location**: Location of the job. It must match the location of the dataset specified in the query.
 
+**Encryption Key Name**: Used to encrypt data written to any dataset created by the plugin.
```
albertshau: does this plugin auto create dataset or just tables or both? Aren't keys set on tables? Or can they be set for an entire dataset?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -119,6 +136,33 @@ public void validate(FailureCollector collector) {
     if (!containsMacro(NAME_DATASET)) {
       BigQueryUtil.validateDataset(dataset, NAME_DATASET, collector);
     }
+    if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
+      validateCmekKey(collector);
+    }
+  }
 
+  void validateCmekKey(FailureCollector failureCollector) {
```
albertshau, line 138: this looks like mostly duplicate code, please consolidate the common parts

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -644,6 +670,37 @@ private boolean isSupportedLogicalType(Schema.LogicalType logicalType) {
     return false;
   }
 
+  @Override
+  void validateCmekKey(FailureCollector failureCollector) {
```
albertshau, line 665: similar comment, this looks like mostly duplicate code. In general, if you are copy pasting, should think how it can be refactored.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -184,6 +194,22 @@ public BigQuerySinkConfig(String referenceName, String dataset, String table,
     this.gcsChunkSize = gcsChunkSize;
   }
 
+  @VisibleForTesting
+  public BigQuerySinkConfig(@Nullable String referenceName, @Nullable String project,
```
albertshau: similar comment, make this private since this should not be directly called (also remove VisibleForTesting)

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -219,6 +246,39 @@ public void validate(FailureCollector collector) {
     if (!containsMacro(NAME_TABLE)) {
       validateTable(collector);
     }
+
+    if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
+      validateCmekKey(collector);
+    }
+  }
+
+  void validateCmekKey(FailureCollector collector) {
```
albertshau, line 255: similar comment about duplicate code

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -74,21 +75,33 @@ public static String getOrCreateBucket(Configuration configuration,
                                          BigQuery bigQuery,
                                          Credentials credentials,
                                          String bucketPath,
-                                         @Nullable CryptoKeyName cmekKeyName) {
+                                         @Nullable CryptoKeyName cmekKeyName) throws IOException {
     String bucket = config.getBucket();
+    // the dataset existence is validated before, so this cannot be null
```
albertshau: this method doesn't know it exists. Relying on the fact results in a tight coupling between this util method and the classes that call it.

If you want to ensure it's not null, the Dataset should be passed in as an argument instead of fetched. (This is also better in general, to prevent multiple calls to BigQuery).

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -74,21 +75,33 @@ public static String getOrCreateBucket(Configuration configuration,
...
-      Dataset dataset = bigQuery.getDataset(DatasetId.of(config.getDatasetProject(), config.getDataset()));
-      GCPUtils.createBucket(GCPUtils.getStorage(config.getProject(), credentials),
-                            bucket,
-                            dataset.getLocation(), cmekKeyName);
+      GCPUtils.createBucket(storage, bucket, dataset.getLocation(), cmekKeyName);
+    } else if (storage != null && storage.get(bucket) == null) {
+      try {
+        GCPUtils.createBucket(storage, bucket, dataset.getLocation(), cmekKeyName);
+      } catch (StorageException e) {
+        if (e.getCode() != 409) {
```
albertshau: more readable to do:

```
if (e.getCode() == 409) {
  // comment
  return;
}
throw new IOException(...);
```

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
```diff
@@ -74,21 +75,33 @@ public static String getOrCreateBucket(Configuration configuration,
                                          BigQuery bigQuery,
                                          Credentials credentials,
                                          String bucketPath,
-                                         @Nullable CryptoKeyName cmekKeyName) {
+                                         @Nullable CryptoKeyName cmekKeyName) throws IOException {
     String bucket = config.getBucket();
+    // the dataset existence is validated before, so this cannot be null
+    Dataset dataset = bigQuery.getDataset(DatasetId.of(config.getDatasetProject(), config.getDataset()));
+    Storage storage = GCPUtils.getStorage(config.getProject(), credentials);
```
albertshau: it's weird that Storage is created within this method but BigQuery is passed in. Would be better to pass it in to avoid reloading credentials and whatever else is done when creating the Storage client.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -124,13 +126,14 @@ public void prepareRun(RuntimeContext context) throws Exception {
...
-    String cmekKey = context.getRuntimeArguments().get(GCPUtils.CMEK_KEY);
+    String cmekKey = !Strings.isNullOrEmpty(sqlEngineConfig.cmekKey) ? sqlEngineConfig.cmekKey :
+      context.getRuntimeArguments().get(GCPUtils.CMEK_KEY);
     CryptoKeyName cmekKeyName = null;
     if (!Strings.isNullOrEmpty(cmekKey)) {
       cmekKeyName = CryptoKeyName.parse(cmekKey);
     }
     configuration = BigQueryUtil.getBigQueryConfig(sqlEngineConfig.getServiceAccount(), sqlEngineConfig.getProject(),
-                                                   cmekKey, sqlEngineConfig.getServiceAccountType());
+                                                   cmekKeyName    , sqlEngineConfig.getServiceAccountType());
```
albertshau: fix spaces

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngineConfig.java
```diff
@@ -98,4 +117,42 @@ public void validate() {
...
+    }
+    if (!containsMacro(NAME_DATASET)) {
+      BigQueryUtil.validateDataset(dataset, NAME_DATASET, failureCollector);
+    }
+    if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
+      validateCmekKey(failureCollector);
+    }
+  }
+
+  public void validateCmekKey(FailureCollector failureCollector) {
```
albertshau, line 129: similar comment on duplicate code.

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -94,4 +100,87 @@ public static CryptoKeyName getCmekKey(@Nullable String cmekKey, FailureCollecto
...
+    try {
+      dataset = bigQuery.getDataset(datasetId);
+    } catch (BigQueryException e) {
+      /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
+      the dataset information either because the service account used during validation can be different than
+      the service account that will be used at runtime (the dataproc service account)
+      (assuming the user has auto-detect for the service account) */
+      return dataset;
+    }
+    if (dataset == null) {
```
albertshau, line 142: removes a level of nesting and is more readable to do:

```
if (dataset != null) {
  return dataset;
}
...
```

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
```diff
@@ -94,4 +100,87 @@ public static CryptoKeyName getCmekKey(@Nullable String cmekKey, FailureCollecto
...
+      table = bigQuery.getTable(tableId);
+    } catch (BigQueryException e) {
+      /* Ignoring the exception because we don't want the validation to fail if there is an exception getting
+      the table information either because the service account used during validation can be different than
+      the service account that will be used at runtime (the dataproc service account)
+      (assuming the user has auto-detect for the service account) */
+      return;
+    }
+    if (table == null) {
+      String cmekKeyLocation = cmekKey.getLocation();
```
albertshau: looks like mostly duplicate code from method above

docs/BigQueryTable-batchsource.md
itsankit-google: It only creates a temporary gcs bucket

docs/BigQueryMultiTable-batchsink.md
itsankit-google: Acknowledged

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
itsankit-google: Yes it will always be same but in some cases if dataset exists, then it only auto creates table

docs/BigQueryExecute-action.md
itsankit-google: Key is supported both on dataset as well as on table. Existing plugin only auto creates table but I have added the support for auto creating dataset as suggested by @bdmogal to standardise the behaviour.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceUtils.java
itsankit-google: It was the existing behaviour will change it.

docs/BigQueryExecute-action.md
itsankit-google: Will change the description

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
itsankit-google: Acknowledged

src/main/java/io/cdap/plugin/gcp/common/CmekUtils.java
itsankit-google: Combined both methods now. So, added an else statement

albertshau:
in the future, please keep changes since the last review in a single commit so that we can easily review just the changes.

docs/BigQueryExecute-action.md
```diff
@@ -43,6 +43,9 @@ cache that will be flushed whenever tables in the query are modified.
 
 **Job Location**: Location of the job. It must match the location of the dataset specified in the query.
 
+**Encryption Key Name**: Used to encrypt data written to any dataset/table created by the plugin.
```
albertshau: dataset/table -> dataset or table. Similar comment for other places

docs/BigQueryMultiTable-batchsink.md
```diff
@@ -47,6 +47,9 @@ Should only be used with the Insert operation.
 **Location:** The location where the big query datasets will get created. This value is ignored
 if the dataset or temporary bucket already exist.
 
+**Encryption Key Name**: Used to encrypt data written to any bucket or dataset/table created by the plugin.
```
albertshau: bucket or dataset/table -> bucket, dataset, or table. similar comment for other places

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -159,11 +173,12 @@ public AbstractBigQueryActionConfig getConfig() {
     private static final String SQL = "sql";
     private static final String DATASET = "dataset";
     private static final String TABLE = "table";
+    private static final String NAME_LOCATION = "location";
 
     @Description("Dialect of the SQL command. The value must be 'legacy' or 'standard'. " +
-                   "If set to 'standard', the query will use BigQuery's standard SQL: " +
-                   "https://cloud.google.com/bigquery/sql-reference/. If set to 'legacy', BigQuery's legacy SQL " +
-                   "dialect will be used for this query.")
+      "If set to 'standard', the query will use BigQuery's standard SQL: " +
```
albertshau, line 179: were these automatically changed by your ide? In the future try not to change irrelevant code to keep the PRs smaller and more focused.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -215,6 +238,19 @@ public Config(String dialect, String sql, String mode) {
       this.mode = mode;
     }
 
+    private Config(@Nullable String project, @Nullable String serviceAccountType, @Nullable String serviceFilePath,
+                  @Nullable String serviceAccountJson, @Nullable String dataset, @Nullable String table,
```
albertshau: nit: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
```diff
@@ -84,4 +95,27 @@ public String getBucket() {
     return bucket;
   }
 
+  public void validateCmekKeyLocation(@Nullable CryptoKeyName cmekKeyName, @Nullable String tableName,
+                                         @Nullable String location, FailureCollector failureCollector) {
```
albertshau: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
```diff
@@ -84,4 +95,27 @@ public String getBucket() {
...
+    if (cmekKeyName == null || containsMacro(NAME_DATASET) || projectOrServiceAccountContainsMacro()) {
+      return;
+    }
+    String datasetProjectId = getDatasetProject();
+    String datasetName = getDataset();
+    DatasetId datasetId = DatasetId.of(datasetProjectId, datasetName);
+    TableId tableId = tableName == null ? null : TableId.of(datasetProjectId, datasetName, tableName);
+    Credentials credentials = getCredentials(failureCollector);
+    BigQuery bigQuery = GCPUtils.getBigQuery(getProject(), credentials);
```
albertshau, line 108: generally speaking, we should create these clients once and pass them around. Creating them in all these helper methods mean the credentials get loaded a bunch of times. Does passing BigQuery and Storage into this method cause a bunch of other methods to need to be changed?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -203,6 +221,15 @@ public String getServiceAccountType() {
     return connection == null ? null : connection.getServiceAccountType();
   }
 
+  @Nullable
+  public CryptoKeyName getCmekKey(Arguments arguments, FailureCollector collector) {
+    String cmekKey = getProperties().getProperties().get(NAME_CMEK_KEY);
```
albertshau: this should just use the cmekKey variable

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -219,6 +246,37 @@ public void validate(FailureCollector collector) {
...
+    }
+  }
+
+  void validateCmekKey(FailureCollector collector) {
+    CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(cmekKey, collector);
+    //these fields are needed to check if bucket exists or not and for location validation
+    if (cmekKeyName == null || !canConnect() || containsMacro(NAME_BUCKET) || Strings.isNullOrEmpty(bucket)) {
+      return;
+    }
+    DatasetId datasetId = DatasetId.of(getDatasetProject(), getDataset());
```
albertshau, line 261: I seem to remember the source does some checking to make sure the bucket and dataset are in the same location. It would be better to take the Dataset from that logic and pass it into this method, rather than looking up all the same information again.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -100,6 +101,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
 
     // Validate configuration and throw exception if the supplied configuration is invalid.
     sqlEngineConfig.validate();
+    sqlEngineConfig.validate(pipelineConfigurer.getStageConfigurer().getFailureCollector());
```
albertshau, line 103: shouldn't this replace the line above?

albertshau (commented):
some small comments. 

Also concerned that these plugins are now looking up the same information, but in a bunch of different places and times. If it's straightforward, these new methods should take Bucket, Dataset, and Table objects that were looked up in other places.

However, I suspect the current code is too much of a mess to do it in this PR. If it's more involved, can keep things as they are, but please open a jira to go through the codebase and clean it up. We should be creating one BigQuery client, one Storage client, look up the Bucket a single time, look up the Dataset a single time, etc.

docs/BigQueryExecute-action.md
itsankit-google: Acknowledged

docs/BigQueryMultiTable-batchsink.md
itsankit-google: Acknowledged

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
itsankit-google: Yes, it was automatically changed by the ide because I never touched these lines

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
itsankit-google: Source does that checking only while creating the bucket when running the pipeline. While config validation, it only checks the dataset pattern and bucket pattern. It looks up the dataset in prepareRun() method and if the bucket does not exists it checks that if it is able to create a bucket in the location of dataset.

itsankit-google:
> Also concerned that these plugins are now looking up the same information, but in a bunch of different places and times. If it's straightforward, these new methods should take Bucket, Dataset, and Table objects that were looked up in other places.
> 
> However, I suspect the current code is too much of a mess to do it in this PR. If it's more involved, can keep things as they are, but please open a jira to go through the codebase and clean it up. We should be creating one BigQuery client, one Storage client, look up the Bucket a single time, look up the Dataset a single time, etc.

So, in the existing plugin config validations we don't look up the bucket, dataset or table, we just check the string pattern. But for cmek, we need to check if the bucket, database or table already exists or not. So, these new validation methods create one client and lookup at the bucket or database. Added a JIRA [PLUGIN-934](https://cdap.atlassian.net/browse/PLUGIN-934)

src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
itsankit-google: This helper method is used by BigQuerySink, BigQueryMultiSink and BigQuerySQLEngine plugin config classes. Earlier I was creating the clients in validation method in each config class and then sending to the CmekUtils methods but that was causing code repetition which you pointed out in earlier comments so I created this helper method in base class of these plugin configs to create the client and pass it to the CmekUtils methods. Existing Plugin config validations don't need clients so they were never creating it but for cmek location validation we need to check if the bucket, dataset or table already exists.  Other than this helper method these plugin configs are not creating the clients anywhere else.

albertshau (approved):
lgtm, please squash commits before merging

PR #781, opened 2021-10-01
CDAP-18428 make sure project is where job is run and datasetProject is where dataset is in for BQSink

seanzhougoogle:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -64,49 +65,49 @@
...
+    Dataset dataset = bigQuery.getDataset(datasetId);
     Bucket bucket = storage.get(bucketName);
 
     if (dataset == null && bucket == null) {
       createBucket(storage, bucketName, location, cmekKey,
                    () -> String.format("Unable to create Cloud Storage bucket '%s'", bucketName));
-      createDataset(bigQuery, datasetName, location,
-                    () -> String.format("Unable to create BigQuery dataset '%s'", datasetName));
+      createDataset(bigQuery, datasetId, location,
+                    () -> String.format("Unable to create BigQuery dataset '%s'", datasetId));
```
tivv: It seems to me that datasetId.toString will produce `GenericData{classInfo=" + this.classInfo.names + ", " + super.toString() + "}"`.  No need to wrap it in quotes. Also consider a nicer value, especially for user-visible messages

fernst: I think if would be useful to print both the dataset name and project. I also don't like that we're simply using that the `toString()` method for DatasetID returns.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputFormat.java
```diff
@@ -65,8 +66,7 @@ public static void configure(Configuration conf,
                                                tableNameField,
                                                bucketName,
                                                bucketPathUniqueId,
-                                               projectName,
-                                               datasetName,
+                                                 DatasetId.of(projectName, datasetName),
```
fernst: nit: alignment is off

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -262,7 +262,7 @@ public SQLDataset join(SQLJoinRequest sqlJoinRequest) throws SQLEngineException
                                                                       sqlEngineConfig,
                                                                       bigQuery,
                                                                       project,
```
tivv, line 264: Is project still needed here?

fernst (approved):
LGTM, I just have 2 minor nitpicks

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -131,7 +131,7 @@ private String generateQuery(String partitionFromDate, String partitionToDate, S
       return null;
     }
     String queryTemplate = "select * from %s where %s";
-    com.google.cloud.bigquery.Table sourceTable = BigQueryUtil.getBigQueryTable(project, datasetProject, dataset, table,
+    com.google.cloud.bigquery.Table sourceTable = BigQueryUtil.getBigQueryTable(datasetProject, dataset, table,
```
tivv, line 134: I believe the only other place this `getBigQueryTable` is still used is in  `io.cdap.plugin.gcp.bigquery.sink.BigQuerySink#configureTable`. Should it also use the one without two projects?

tivv (commented):
Are there any backwards compatibility issues expected with this change?

seanzhougoogle:
> Are there any backwards compatibility issues expected with this change?

It's possible that to make the old (wrong) code work, users had to grant permissions to wrong project.
But this change is to fix the wrong behavior instead of changing the original desired behavior. IMO, it's not impacting the backward compatibility.

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
seanzhougoogle: It's also changed in the PR.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
seanzhougoogle: Yes, it will be used to create BQ jobs.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
seanzhougoogle: done:)

seanzhougoogle: done :)

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -64,49 +65,50 @@
...
+                    () -> String.format("Unable to create BigQuery dataset '%s.%s'", datasetId.getProject(),
+                                        datasetId.getDataset()));
     } else if (bucket == null) {
       createBucket(
         storage, bucketName, dataset.getLocation(), cmekKey,
         () -> String.format(
           "Unable to create Cloud Storage bucket '%s' in the same location ('%s') as BigQuery dataset '%s'. "
             + "Please use a bucket that is in the same location as the dataset.",
-          bucketName, dataset.getLocation(), datasetName));
+          bucketName, dataset.getLocation(), datasetId));
```
fernst, line 92: Let's also add the correct Project/Dataset in this error message.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -64,49 +65,50 @@
...
     } else if (dataset == null) {
       createDataset(
-        bigQuery, datasetName, bucket.getLocation(),
+        bigQuery, datasetId, bucket.getLocation(),
         () -> String.format(
           "Unable to create BigQuery dataset '%s' in the same location ('%s') as Cloud Storage bucket '%s'. "
             + "Please use a bucket that is in a supported location.",
-          datasetName, bucket.getLocation(), bucketName));
+          datasetId, bucket.getLocation(), bucketName));
```
fernst, line 99: Let's also add the correct Project/Dataset in this error message.

PR #799, opened 2021-10-14
Profile changes and TestRunner package inclusion

ksahilCF:


ksahilCF:
@tivv @rmstar profile changes along with report creation

pom.xml
```diff
@@ -619,6 +619,9 @@
           <systemPropertyVariables>
             <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
           </systemPropertyVariables>
+          <excludes>
+            <exclude>**/TestRunner.java</exclude>
+          </excludes>
```
tivv: Any changes should go into profile. I recommend disabling SureFire in the profile altogether  by setting "skip" property of the surefire plugin in the profile. Anyway, with the profile enabled regular unit test directory (src/test) is not used

pom.xml
```diff
@@ -800,6 +803,67 @@
...
       </properties>
+      <build>
+      <plugins>
+        <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.18.1</version>
+        <configuration>
+          <includes>
+            <include>TestRunner.java</include>
```
tivv: Why do you need to include TestRunner for surefire? I think you should disable surefire at all, only failsafe should run

pom.xml
ksahilCF: removed exclusions

pom.xml
ksahilCF: skiptest property used

tivv (approved):

PR #800, opened 2021-10-19
Ignoring the cmek validation exception for 6.5.1

itsankit-google:


albertshau (commented):
this looks like it changes the behavior so that an invalid key is treated as if there is no key. I would expect an invalid uri to generate a failure once the request goes to GCP, is that not the case?

albertshau:
I think it's ok to leave it as-is. If somebody is setting a completely invalid uri for the cmek key, that's a bug rather than intended behavior.

PR #808, opened 2021-10-25
[CDAP-18556] Turned on validation and added unit tests for cmek config

itsankit-google:
[CDAP-18556](https://cdap.atlassian.net/browse/CDAP-18556)

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -51,6 +51,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
 import javax.annotation.Nullable;
 
 /**
@@ -72,7 +74,8 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
   @Override
   public void run(ActionContext context) throws Exception {
-    config.validate(context.getFailureCollector());
+    FailureCollector collector = context.getFailureCollector();
+    config.validate(collector, context.getArguments().asMap());
     QueryJobConfiguration.Builder builder = QueryJobConfiguration.newBuilder(config.getSql());
     // Run at batch priority, which won't count toward concurrent rate limit.
     if (config.getMode().equals(QueryJobConfiguration.Priority.BATCH)) {
@@ -81,8 +84,8 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
     }
 
-    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), context.getFailureCollector());
-    context.getFailureCollector().getOrThrowException();
+    CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(config.cmekKey, context.getArguments().asMap(), collector);
+    collector.getOrThrowException();
     if (cmekKeyName != null) {
       builder.setDestinationEncryptionConfiguration(
         EncryptionConfiguration.newBuilder().setKmsKeyName(cmekKeyName.toString()).build());
@@ -307,6 +310,10 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
     @Override
     public void validate(FailureCollector failureCollector) {
+      validate(failureCollector, Collections.emptyMap());
+    }
+
+    public void validate(FailureCollector failureCollector, Map<String, String> arguments) {
       // check the mode is valid
       if (!containsMacro(MODE)) {
         try {
@@ -345,17 +352,15 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
         BigQueryUtil.validateTable(table, TABLE, failureCollector);
       }
 
-      /* Commenting out this code for 6.5.1
-      if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
-        validateCmekKey(failureCollector);
+      if (!containsMacro(NAME_CMEK_KEY)) {
+        validateCmekKey(failureCollector, arguments);
       }
-      */
 
       failureCollector.getOrThrowException();
     }
 
-    void validateCmekKey(FailureCollector failureCollector) {
-      CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(cmekKey, failureCollector);
+    void validateCmekKey(FailureCollector failureCollector, Map<String, String> arguments) {
+      CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(cmekKey, arguments, failureCollector);
       //these fields are needed to check if bucket exists or not and for location validation
       if (cmekKeyName == null || containsMacro(DATASET) || containsMacro(NAME_LOCATION) || containsMacro(TABLE) ||
         projectOrServiceAccountContainsMacro() || Strings.isNullOrEmpty(dataset) || Strings.isNullOrEmpty(table) ||
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
@@ -23,7 +23,9 @@ import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.kms.v1.CryptoKeyName;
+import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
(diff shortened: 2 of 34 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -81,8 +82,12 @@ public void run(ActionContext context) throws Exception {
       builder.setPriority(QueryJobConfiguration.Priority.INTERACTIVE);
     }
 
-    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), context.getFailureCollector());
-    context.getFailureCollector().getOrThrowException();
+    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), collector);
+    collector.getOrThrowException();
+    if (Strings.isNullOrEmpty(config.cmekKey) && cmekKeyName != null) {
```
albertshau: Seems like the null or empty shouldn't be required. Unless getCmekKey() doesn't return null in those situations? It should if it doesn't.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -345,16 +350,14 @@ public void validate(FailureCollector failureCollector) {
...
-        validateCmekKey(failureCollector);
+        validateCmekKey(cmekKey, failureCollector);
       }
-      */
 
       failureCollector.getOrThrowException();
     }
 
-    void validateCmekKey(FailureCollector failureCollector) {
+    void validateCmekKey(String cmekKey, FailureCollector failureCollector) {
```
albertshau: you don't need to pass the string, it's a variable in this class already.

src/main/java/io/cdap/plugin/gcp/bigquery/common/BigQueryBaseConfig.java
```diff
@@ -106,16 +122,28 @@ public void validateCmekKeyLocation(@Nullable CryptoKeyName cmekKeyName, @Nullab
...
-    if (dataset == null || storage == null || containsMacro(NAME_BUCKET) || Strings.isNullOrEmpty(bucket)) {
+    if (bigQuery == null || storage == null) {
       return;
     }
-    CmekUtils.validateCmekKeyAndBucketLocation(storage, GCSPath.from(bucket), cmekKeyName,
-                                               dataset.getLocation(), failureCollector);
+    if (Strings.isNullOrEmpty(getBucket())) {
+      CmekUtils.validateCmekKeyAndDatasetOrTableLocation(bigQuery, datasetId, tableId, cmekKeyName,
+                                                         location, failureCollector);
+    } else {
```
albertshau: avoid multiple layers of nesting when possible to make it more readable. For example:

```
if (...) {
  ...
  return;
}
Bucket bucket = ...
if (bucket != null) {
  ...
  return;
}
Dataset dataset = ...
```

albertshau: Though the flow itself here is confusing. There are 3 calls to the same method in different code blocks. It would be better to have just one:

```
String bucketName = getBucket();
Bucket bucket = bucketName == null ? null : getBucketIfExists(storage, bucketName);
String bucketLocation = bucket == null ? location : bucket.getLocation();
Dataset dataset = CmekUtils.validate....
```

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -87,8 +87,13 @@ public final void prepareRun(BatchSinkContext context) throws Exception {
       null : GCPUtils.loadServiceAccountCredentials(serviceAccount, config.isServiceAccountFilePath());
     String project = config.getProject();
     bigQuery = GCPUtils.getBigQuery(project, credentials);
-    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), context.getFailureCollector());
-    context.getFailureCollector().getOrThrowException();
+    FailureCollector collector = context.getFailureCollector();
+    CryptoKeyName cmekKeyName = config.getCmekKey(context.getArguments(), collector);
+    collector.getOrThrowException();
+    if (Strings.isNullOrEmpty(config.cmekKey) && cmekKeyName != null) {
```
albertshau: similar comment here about extra nullOrEmpty check

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -130,14 +130,12 @@ public void validate(FailureCollector collector) {
...
-    /* Commenting out this code for 6.5.1
     if (!containsMacro(NAME_CMEK_KEY) && !Strings.isNullOrEmpty(cmekKey)) {
-      validateCmekKey(collector);
+      validateCmekKey(cmekKey, collector);
     }
-    */
   }
 
-  void validateCmekKey(FailureCollector failureCollector) {
+  void validateCmekKey(String cmekKey, FailureCollector failureCollector) {
```
albertshau: similar comment, don't need to pass this in

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -662,7 +663,7 @@ private boolean isSupportedLogicalType(Schema.LogicalType logicalType) {
   }
 
   @Override
-  void validateCmekKey(FailureCollector failureCollector) {
+  void validateCmekKey(String cmekKey, FailureCollector failureCollector) {
```
albertshau: similar comment, don't need to pass this in

albertshau: same for rest of PR

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
albertshau: same for rest of PR

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteCmekKeyTest.java
```diff
@@ -63,11 +75,13 @@ public static void setupTestClass() throws Exception {
     serviceAccountFilePath = System.getProperty("service.account.file");
     Assume.assumeFalse(String.format(messageTemplate, "service account key file"), serviceAccountFilePath == null);
 
-    dataset = UUID.randomUUID().toString();
-    table = UUID.randomUUID().toString();
+    dataset = UUID.randomUUID().toString().replace("-", "_");
```
albertshau: it would be good to prepend some descriptive string, like "bq-cmektest-". In case they don't get cleaned up for some reason, somebody looking at the datasets/tables will know what they are and that they can delete them.

albertshau: same comment for other tests

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
itsankit-google: I need to pass cmekKey when it calls the validation on the key coming from runtime arguments or system preferences.

itsankit-google: Discussed offline to change the validate method and pass the arguments in it.

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteCmekKeyTest.java
```diff
@@ -95,51 +111,82 @@ public void testServiceAccountJson() throws Exception {
     testValidCmekKey(builder);
     testInvalidCmekKeyName(builder);
     testInvalidCmekKeyLocation(builder);
+    testCmekKeyWithExistingTable(builder);
   }
 
-  private void testValidCmekKey(BigQueryExecute.Config.Builder builder) throws Exception {
+  private void testValidCmekKey(BigQueryExecute.Config.Builder builder) {
+    String cmekKey = String.format("projects/%s/locations/us-east1/keyRings/my_ring/cryptoKeys/test_key", project);
```
fernst: I assume this is a mock value? If so, let's make the location something that doesn't really match a GCP location ir order for this to be clear.

For example, something like `my-location`

fernst (approved):
LGTM, just added one question/comment.

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteCmekKeyTest.java
itsankit-google: Sure, we can do that.

PR #811, opened 2021-11-02
Added location config in GCS Done File Marker Plugin

itsankit-google:
[PLUGIN-965](https://cdap.atlassian.net/browse/PLUGIN-965)

Before:- 
Screen Shot 2021-10-29 at 10 51 03 PM

After:-
Screen Shot 2021-11-02 at 11 49 41 AM

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSDoneFileMarker.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSDoneFileMarker.java
@@ -88,7 +88,7 @@ public class GCSDoneFileMarker extends PostAction {
     CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(config.cmekKey, runtimeArgs, collector);
     collector.getOrThrowException();
     createFileMarker(config.getProject(), markerFilePath, serviceAccount, config.isServiceAccountFilePath(),
-                     cmekKeyName);
+                     cmekKeyName, config.location);
   }
 
   public Map<String, String> getArgumentsAsMap(Arguments arguments) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSDoneFileMarker.java
```diff
@@ -119,6 +120,13 @@ public void run(BatchActionContext batchActionContext) throws IOException {
     @Macro
     public String path;
 
+    @Name(NAME_LOCATION)
+    @Macro
+    @Nullable
+    @Description("The location where the GCS bucket will get created. " +
+      "This value is ignored if the bucket already exists.")
+    protected String location;
```
seanzhougoogle: why protected ?

widgets/GCSDoneFileMarker-postaction.json
```diff
@@ -81,6 +73,27 @@
           "name": "serviceAccountJSON"
         }
       ]
+    },
+    {
+      "label": "Auto Create",
```
seanzhougoogle, line 78: this group name is wired, did you confirm with UX ?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSDoneFileMarker.java
itsankit-google: Changed to private.

seanzhougoogle (approved):

widgets/GCSDoneFileMarker-postaction.json
itsankit-google: Discussed offline to keep it the same as followed by other plugins.

PR #812, opened 2021-11-02
Added GCSCmekDemo feature

itsankit-google:


tivv (commented):
So, coming back to our discussion, could you please start from adding CMEK test case to the existing feature?  We can do new features, but duplicating all features would not be productive in the long run

tivv:
Closing this according to our discussion

PR #815, opened 2021-11-03
Added connection in GCSBatchSink and GCSMultiFileBatchSink

itsankit-google:

### DIFF
--- a/docs/GCS-batchsink.md
+++ b/docs/GCS-batchsink.md
@@ -26,6 +26,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection** Name of the connection to use, should use the macro function ${conn:(connection-name)} to provide.
+Project and service account information will be provided by the connection.
+
 **Project ID**: Google Cloud Project ID, which uniquely identifies a project.
 It can be found on the Dashboard in the Google Cloud Platform Console.
 
--- a/docs/GCSMultiFiles-batchsink.md
+++ b/docs/GCSMultiFiles-batchsink.md
@@ -38,6 +38,12 @@ Properties
 **Reference Name:** This along with the table name will be used to uniquely identify this sink for lineage, 
 annotating metadata, etc.
 
+**Use Connection** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection** Name of the connection to use, should use the macro function ${conn:(connection-name)} to provide.
+Project and service account information will be provided by the connection.
+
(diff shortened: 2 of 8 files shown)


yaojiefeng (commented):
These changes LGTM, we will also need to add these sinks to the connector's spec method to let UI know which plugins the spec is corresponding to, please see https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java#L158

Since the UI need some changes on getting back to studio, you can test it directly using REST endpoint:
namespaces/system/apps/pipeline/services/studio/methods/v1/contexts/default/connections/<connecton-name>/specification
with request body like 
```
{
  "path": "<path-to-file>",
  "properties": {}
}
```
to ensure these sinks are included in the response

yaojiefeng:
Please also make sure the changes work for preview, and actual pipeline run

itsankit-google:
> These changes LGTM, we will also need to add these sinks to the connector's spec method to let UI know which plugins the spec is corresponding to, please see https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/gcs/connector/GCSConnector.java#L158
> 
> Since the UI need some changes on getting back to studio, you can test it directly using REST endpoint: namespaces/system/apps/pipeline/services/studio/methods/v1/contexts/default/connections//specification with request body like
> 
> ```
> {
>   "path": "<path-to-file>",
>   "properties": {}
> }
> ```
> 
> to ensure these sinks are included in the response

Done.

yaojiefeng (approved):

docs/GCS-batchsink.md
```diff
@@ -26,6 +26,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection** Whether to use a connection, if a connection is used,
```
rrielley76, line 29: @itsankit-google Can we rewrite as "Whether to use a connection. If a connection is used,"

docs/GCS-batchsink.md
```diff
@@ -26,6 +26,12 @@ Properties
...
+the credentials does not need to be provided.
```
rrielley76, line 30: @itsankit-google Can we rewrite as "you do not need to provide the credentials."

docs/GCS-batchsink.md
```diff
@@ -26,6 +26,12 @@ Properties
...
+
+**Connection** Name of the connection to use, should use the macro function ${conn:(connection-name)} to provide.
```
rrielley76, line 32: @itsankit-google You no longer write a macro for the name. You Browse connection and select one. The connection name displays in the Connection field. The field isn't editable.  Maybe just say "Name of the connection to use. Project and service account information will be provided by the connection." (The last sentence is in the GCS source docs). For the part about using the macro, how's "You also can use the macro function ${conn:(connection-name)}."

PR #823, opened 2021-11-09
[PLUGIN-974] GCS Multi Files Plugin validation always failing due to output format always returning null

itsankit-google:
[PLUGIN-974](https://cdap.atlassian.net/browse/PLUGIN-974)

Screen Shot 2021-11-09 at 12 02 49 PM

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -207,7 +207,8 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
     }
 
     Map<String, String> fileSystemProperties = config.getFileSystemProperties();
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -207,7 +207,7 @@ private String getFilenameBase() {
     }
 
     Map<String, String> fileSystemProperties = config.getFileSystemProperties();
-    if (fileSystemProperties.containsKey(AVRO_NAMED_OUTPUT) && config.getFormat() == FileFormat.AVRO) {
+    if (fileSystemProperties.containsKey(AVRO_NAMED_OUTPUT) && "avro".equals(config.getFormatName())) {
```
yaojiefeng: use `FileFormat.AVRO.name().toLowerCase()` instead

yaojiefeng (approved):
One comment, rest LGTM

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
itsankit-google: done.

PR #824, opened 2021-11-09
Adding connection support in Spanner Sink

itsankit-google:

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
@@ -50,11 +50,11 @@ public class SpannerOutputFormat extends OutputFormat<NullWritable, Mutation> {
    * @param schema        schema for spanner table
    */
   public static void configure(Configuration configuration, SpannerSinkConfig config, Schema schema) {
-    String projectId = config.getProject();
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
```diff
@@ -50,11 +50,11 @@
    * @param schema        schema for spanner table
    */
   public static void configure(Configuration configuration, SpannerSinkConfig config, Schema schema) {
-    String projectId = config.getProject();
+    String projectId = config.connection.getProject();
```
yaojiefeng, line 53: Where does this method get called? Since it is named as `configure`, if it is called in `configurePipeline` method, this change will cause NPE

yaojiefeng (approved):
One question, otherwise LGTM, please verify the changes work for both preview and actual pipeline run.

src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerOutputFormat.java
itsankit-google: It is called in the `prepareRun`. https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/spanner/sink/SpannerSink.java#L140

PR #826, opened 2021-11-10
Added connection support in bigquery sinks

itsankit-google:

### DIFF
--- a/docs/BigQueryMultiTable-batchsink.md
+++ b/docs/BigQueryMultiTable-batchsink.md
@@ -22,6 +22,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection:** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection:** Name of the connection to use, should use the macro function ${conn(connection-name)} to provide.
+Project and service account information will be provided by the connection.
+
 **Project ID:** Google Cloud Project ID, which uniquely identifies a project.
 It can be found on the Dashboard in the Google Cloud Platform Console. This is the project
 that the BigQuery job will run in. `BigQuery Job User` role on this project must be granted to the specified service
--- a/docs/BigQueryTable-batchsink.md
+++ b/docs/BigQueryTable-batchsink.md
@@ -23,6 +23,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection:** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection:** Name of the connection to use, should use the macro function ${conn(connection-name)} to provide.
+Project and service account information will be provided by the connection.
+
 **Project ID**: Google Cloud Project ID, which uniquely identifies a project.
 It can be found on the Dashboard in the Google Cloud Platform Console. This is the project
 that the BigQuery job will run in. `BigQuery Job User` role on this project must be granted to the specified service
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -15,20 +15,34 @@
  */
 package io.cdap.plugin.gcp.bigquery.sink;
 
+import com.google.auth.Credentials;
+import com.google.cloud.ServiceOptions;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.TableId;
 import com.google.cloud.kms.v1.CryptoKeyName;
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.data.schema.Schema;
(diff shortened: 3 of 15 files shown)


docs/BigQueryMultiTable-batchsink.md
```diff
@@ -22,6 +22,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection:** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection:** Name of the connection to use, should use the macro function ${conn:(connection-name)} to provide.
```
seanzhougoogle: I think the macro should be  ${conn(connection-name)} , there is no colon

docs/BigQueryTable-batchsink.md
```diff
@@ -23,6 +23,12 @@ Properties
 ----------
 **Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
 
+**Use Connection:** Whether to use a connection, if a connection is used,
+the credentials does not need to be provided.
+
+**Connection:** Name of the connection to use, should use the macro function ${conn:(connection-name)} to provide.
```
seanzhougoogle: same here , no colon in the connection macro

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -149,4 +193,101 @@ void validateCmekKey(FailureCollector failureCollector, Map<String, String> argu
     }
     validateCmekKeyLocation(cmekKeyName, null, location, failureCollector);
   }
+
+  public String getDataset() {
+    return dataset;
+  }
+
+  public String getDatasetProject() {
+    if (connection == null) {
```
seanzhougoogle: I think connection will always have value, we made it backward compatible if connection was not provided , we get it from parent properties

seanzhougoogle: also connection is null should not be equivalent to auto-detect dataset project

docs/BigQueryMultiTable-batchsink.md
itsankit-google: I used this from here, https://github.com/data-integrations/google-cloud/blame/develop/docs/GCSFile-batchsource.md#L31, @yaojiefeng what do you think?

yaojiefeng: Sean is correct, seems like there is a typo in the description. the correct one should be `${conn(connection-name)}`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
yaojiefeng: In configure time, the connection will be null if user chooses to use connection and that should be the only case connection will be null.
I think if connection is null, we can just return null instead of the default project id.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -149,4 +193,101 @@ void validateCmekKey(FailureCollector failureCollector, Map<String, String> argu
...
+  }
+
+  @Nullable
+  public String getServiceAccountType() {
+    return connection == null ? null : connection.getServiceAccountType();
+  }
+
+  @Nullable
+  public String getBucket() {
+    if (bucket != null) {
```
yaojiefeng, line 235: Are these methods just copied from the `BigQueryBaseConfig`? Can you open a JIRA to refactor these after we support connection in BQ pushdown?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -679,11 +678,12 @@ void validateCmekKey(FailureCollector failureCollector, Map<String, String> argu
    */
   boolean shouldConnect() {
     return !containsMacro(BigQuerySinkConfig.NAME_DATASET) &&
-      !containsMacro(BigQuerySinkConfig.NAME_TABLE) &&
-      !containsMacro(NAME_SERVICE_ACCOUNT_TYPE) &&
-      !(containsMacro(NAME_SERVICE_ACCOUNT_FILE_PATH) || containsMacro(NAME_SERVICE_ACCOUNT_JSON)) &&
-      !containsMacro(BigQuerySinkConfig.NAME_PROJECT) &&
-      !containsMacro(BigQuerySinkConfig.DATASET_PROJECT_ID) &&
+      !containsMacro(BigQuerySinkConfig.NAME_TABLE) && connection != null && 
+      !connection.containsMacro(BigQueryConnectorConfig.NAME_SERVICE_ACCOUNT_TYPE) &&
```
yaojiefeng: can these be refactored to `connection != null && connection.canConnect()`?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
itsankit-google: done, https://cdap.atlassian.net/browse/PLUGIN-977.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
itsankit-google: done

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
itsankit-google: done

docs/BigQueryTable-batchsink.md
itsankit-google: done

docs/BigQueryMultiTable-batchsink.md
itsankit-google: done

yaojiefeng (approved):

PR #827, opened 2021-11-16
BQExecute plugin validation failure fix[bug fix for CDAP-18610]

sanjanasandeep:
**Ticket:**

https://cdap.atlassian.net/browse/CDAP-18610

**Context:**

This fix is to ensure that the BQ Execute plugin validation does not fail before pipeline runs. 

Problem : This specifically is to fix the failure that happens when a non-macros sql query is entered - querying for a non existent resource. 

Fix: Added a check to filter out BQ exceptions specifically for 404 errors.

**Testing:**

Tested locally on sandbox and added unit test.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -182,6 +182,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     private static final String TABLE = "table";
     private static final String NAME_LOCATION = "location";
     public static final String DATASET_PROJECT_ID = "datasetProject";
+    private static final int ERROR_CODE_NOT_FOUND = 404;
 
     @Name(DATASET_PROJECT_ID)
     @Macro
@@ -247,15 +248,10 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     @Macro
     private String rowAsArguments;
 
-    public Config(String dialect, String sql, String mode) {
-      this.dialect = dialect;
-      this.sql = sql;
-      this.mode = mode;
-    }
-
     private Config(@Nullable String project, @Nullable String serviceAccountType, @Nullable String serviceFilePath,
                    @Nullable String serviceAccountJson, @Nullable String dataset, @Nullable String table,
-                   @Nullable String location, @Nullable String cmekKey) {
+                   @Nullable String location, @Nullable String cmekKey, @Nullable String dialect, @Nullable String sql,
+                   @Nullable String mode) {
       this.project = project;
       this.serviceAccountType = serviceAccountType;
       this.serviceFilePath = serviceFilePath;
@@ -264,6 +260,9 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       this.table = table;
       this.location = location;
       this.cmekKey = cmekKey;
+      this.dialect = dialect;
+      this.sql = sql;
+      this.mode = mode;
     }
 
     public boolean isLegacySQL() {
@@ -384,8 +383,16 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       try {
         bigQuery.create(JobInfo.of(queryJobConfiguration));
       } catch (BigQueryException e) {
-        failureCollector.addFailure(String.format("%s.", e.getMessage()), "Please specify a valid query.")
-          .withConfigProperty(SQL);
+          final String errorMessage;
+          if (e.getCode() == ERROR_CODE_NOT_FOUND)  {
+            errorMessage = String.format("Resource was not found. Please verify the resource name. If the resource " +
+              "will be created at runtime, then update to use a macro for the resource name. Error message received " +
+              "was: %s", e.getMessage());
+          } else {
+               errorMessage = e.getMessage();
+          }
+          failureCollector.addFailure(String.format("%s.", errorMessage), "Please specify a valid query.")
+                    .withConfigProperty(SQL);
       }
     }
 
@@ -417,6 +424,9 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
       private String table;
       private String cmekKey;
       private String location;
+      private String dialect;
+      private String sql;
+      private String mode;
 
       public Builder setProject(@Nullable String project) {
         this.project = project;
@@ -458,6 +468,21 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
         return this;
       }
 
+      public Builder setDialect(@Nullable String dialect) {
+        this.dialect = dialect;
+        return this;
+      }
+
+      public Builder setMode(@Nullable String mode) {
+        this.mode = mode;
+        return this;
+      }
+
+      public Builder setSql(@Nullable String sql) {
+        this.sql = sql;
+        return this;
+      }
+
       public Config build() {
         return new Config(
           project,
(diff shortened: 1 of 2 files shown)


src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteConfigTest.java
```diff
@@ -47,11 +48,29 @@ public void testBigQueryExecuteValidSQL() throws Exception {
...
+    Log.warn("size : {}", failureCollector.getValidationFailures().size());
+    Assert.assertEquals(0, failureCollector.getValidationFailures().size());
+  }
+
   @Test
   public void testBigQueryExecuteInvalidSQL() throws Exception {
     String errorMessage = "Invalid sql";
     BigQueryExecute.Config config = new BigQueryExecute.Config("standard",
-                                                               "selcet * fro$m dataset-table where id=1",
+                                                               "select * fro$m dataset-table where id=1",
```
fernst: I think this is an accidental change. We want the SQL to be broken.

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteConfigTest.java
```diff
@@ -47,11 +48,29 @@ public void testBigQueryExecuteValidSQL() throws Exception {
...
   }
 
+  @Test
+  public void testBigQueryExecuteSQLWithNonExistentResource() throws Exception {
+    String errorMessage = "Not found";
+    BigQueryExecute.Config config = new BigQueryExecute.Config("standard",
+            "select * from dataset.table where id=1",
+            "batch");
+
+    FieldSetter.setField(config, GCPConfig.class.getDeclaredField("project"), "test_project");
```
itsankit-google: Please don't set fields like this, it is brittle (like changing the field name will break the test)!
Generally, we use a builder to create the config and set it normally.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -384,8 +385,13 @@ public void validateSQLSyntax(FailureCollector failureCollector, BigQuery bigQue
       try {
         bigQuery.create(JobInfo.of(queryJobConfiguration));
       } catch (BigQueryException e) {
-        failureCollector.addFailure(String.format("%s.", e.getMessage()), "Please specify a valid query.")
-          .withConfigProperty(SQL);
+        if (e.getCode() == ERROR_CODE_NOT_FOUND)  {
+          LOG.warn("The query contains a non resource that doesn't exist. Error code {} and message generated {}",
```
itsankit-google: @sanjanasandeep is this the intended behaviour?

bdmogal: I thought about this some more. I think the problem is that this specific situation (that the bug was created for) occurs when there is a dependency between the action and the preceding source.

However, the pipeline can also contain only an action plugin (or a set of action plugins). At that time, it may be desirable to fail if a table is provided that does not exist?

Unless the behavior we want to shoot for is that in general we don't want to check for table existence at deployment time, and rely upon it being a runtime error.

Ideally, this would be one of those scenarios where we would extend the failureCollector/validation APIs to also support warnings, and not just errors like it does now.


But until then, I think we may just want to leave it as-is, because it could have unintended consequences. E.g. with this change, if I mis-typed the name of the table in my query, then it won't fail until the pipeline is actually run. This will force the user to then duplicate the pipeline, fix the error and try again. As opposed to that, it might be better to fail early, let the user know, and if they want to ignore the failure, then use a macro.

fernst: @bdmogal Are we tracking the idea of improving the failure collector to add support for warnings anywhere?

bdmogal: Thinking a little more, I think the right thing to do will be, keep the validation, but change the error message. So, if the error from BQ is that the table is not found, update the messagee in the failure collector to be something to the effect of "Table was not found. Please verify the table name. If the table will be created at runtime, update to use a macro for the table name"

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteConfigTest.java
sanjanasandeep: Addressed this in the new rev

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -384,8 +383,16 @@ public void validateSQLSyntax(FailureCollector failureCollector, BigQuery bigQue
       try {
         bigQuery.create(JobInfo.of(queryJobConfiguration));
       } catch (BigQueryException e) {
-        failureCollector.addFailure(String.format("%s.", e.getMessage()), "Please specify a valid query.")
-          .withConfigProperty(SQL);
+          final String errorMessage;
+          if (e.getCode() == ERROR_CODE_NOT_FOUND)  {
+            errorMessage = String.format("Resource was not found. Please verify the resource name. If the resource " +
+              "will be created at runtime, then update to use a macro for the resource name. Error Message returned" +
```
fernst: nitpick: "... Error message received was: <error>"

fernst (commented):
Small nitpick.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -247,15 +248,10 @@ public AbstractBigQueryActionConfig getConfig() {
     @Macro
     private String rowAsArguments;
 
-    public Config(String dialect, String sql, String mode) {
```
fernst, line 250: Why did you condense these 2 constructors into one? Any particular reason?

sanjanasandeep: This was done specifically to make use of the Config build() function on L486 (these fields weren't set earlier)

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteConfigTest.java
sanjanasandeep: Didn't realize this was intentional, because of the stubbing. Will fix this

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteConfigTest.java
```diff
@@ -33,38 +33,57 @@
...
+    Assert.assertEquals(String.format("%s.", errorMessage),
+            failureCollector.getValidationFailures().get(0).getMessage());
+    Assert.assertEquals("sql",
+            failureCollector.getValidationFailures().get(0).getCauses().get(0).getAttribute("stageConfig"));
+  }
 
+  @Test
+  public void testBigQueryExecuteInvalidSQL() throws Exception {
+    String errorMessage = "Invalid sql";
+    BigQueryExecute.Config config = getConfig();
```
itsankit-google: `getConfig()` will return a config with valid SQL query or am I missing something? Though we are throwing an error later but it looks confusing! Can we pass the `sql` param to the `getConfig(String sql)` to make it look better?

sanjanasandeep: yes I refactored this in the next rev. I initially mixed up the sql query at #L50 to be used for both valid and invalid queries.

fernst (approved):
LGTM

itsankit-google (approved):

PR #831, opened 2021-11-19
Fixed issue which causes schema updates to fail when CMEK is used on the table and the input is empty.

fernst:


DJSagarAhire (approved):
LGTM

PR #832, opened 2021-11-22
[Cherry-Pick][0.18] Fixed issue which causes schema updates to fail when CMEK is used on the table and the input is empty.

fernst:
Cherry Pick of https://github.com/data-integrations/google-cloud/pull/831 into 0.18


sanjanasandeep (approved):
LGTM

PR #835, opened 2021-11-29
enclosing datasetProject.datasetName.tableName with backticks

itsankit-google:
Fix for:-
image


fernst (approved):
LGTM, please add tests

PR #836, opened 2021-11-29
enclosing datasetProject.datasetName.tableName with backticks

itsankit-google:
Context, #835, #839.


fernst (approved):

src/test/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormatTest.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
rrielley76, line 2: @itsankit-google Should this be 2022?

PR #865, opened 2021-12-16
Changed Precision and Scale for BigQueryRecordToJson to support BigNu…

sechegaray:
…meric

Fixed condition to detect if BigNumeric is present

Fixed check for difference between scale and precision

Added unit tests to test conversion


fernst (approved):
LGTM

PR #881, opened 2022-01-21
removed checks from bigqueryjson writer to remove bug and  make it si…

sechegaray:
…milar to avrowriter.


tivv (commented):
Please keep max precision / scale check.

tivv (approved):

PR #882, opened 2022-01-21
removed checks from bigqueryjson writer to remove bug and  make it si…

sechegaray:
…milar to avrowriter.


tivv (commented):
As we talked some checks make sense, e.g. `decimal.precision() > BigQueryTypeSize.BigNumeric.PRECISION ||
      decimal.scale() > BigQueryTypeSize.BigNumeric.SCALE) `

tivv (approved):

PR #884, opened 2022-01-28
Added support for all write operations in the BigQuery sink

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -353,6 +366,56 @@ private static BigQueryTableFieldSchema generateTableFieldSchema(Schema.Field fi
...
 
+  public static com.google.cloud.bigquery.Schema convertCdapSchemaToBigQuerySchema(Schema schema) {
+    List<Schema.Field> inputFields = Objects.requireNonNull(schema.getFields(), "Schema must have fields");
+    List<com.google.cloud.bigquery.Field> fields = inputFields.stream()
+      .map(BigQuerySinkUtils::convertCdapFieldToBigQueryField)
+      .collect(Collectors.toList());
+    return com.google.cloud.bigquery.Schema.of(fields);
+  }
+
+  private static Field convertCdapFieldToBigQueryField(Schema.Field field) {
```
fernst, line 375: @sechegaray can you walk through this code and let me know if I did this correctly? I'm essentially mapping CDAP types into LegacySQLTypes, and adding scale/precision for numeric fields.

Note that this also does a recursive call to handle nested records.

widgets/BigQueryPushdownEngine-sqlengine.json
```diff
@@ -189,6 +189,22 @@
...
+            "on": {
+              "value": "true",
+              "label": "YES"
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "false"
+          }
```
sanjanasandeep, line 207: should this be added to ```BigQueryPushdownEngine-sqlengine.md``` ?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -342,115 +322,30 @@ public SQLDatasetProducer getProducer(SQLPullRequest pullRequest, PullCapability
...
-    String destinationTableName = arguments.get(SQL_OUTPUT_TABLE);
-    String destinationProject = sinkConfig.getDatasetProject();
-    List<String> fields = GSON.fromJson(arguments.get(SQL_OUTPUT_FIELDS), LIST_OF_STRINGS_TYPE);
-    boolean allowSchemaRelaxation = sinkConfig.isAllowSchemaRelaxation();
-    JobInfo.WriteDisposition writeDisposition = sinkConfig.getWriteDisposition();
-
-    if (Operation.INSERT != sinkConfig.getOperation()) {
-      LOG.warn("Direct table copy is only supported for INSERT operations.");
+    // Check if direct sink write is enabled. If not, skip.
+    if (!sqlEngineConfig.shouldUseDirectSinkWrite()) {
```
sanjanasandeep, line 328: Should we add a log statement here?

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtilsTest.java
```diff
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.gcp.bigquery.sink;
 
+import com.google.cloud.bigquery.Field;
```
snehalakshmisha, line 19: Was this added by mistake?

widgets/BigQueryPushdownEngine-sqlengine.json
fernst: Good observation. I'll work on writing these docs with Robin.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Since the user didn't enable this feature, we can simply skip this

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -353,6 +364,56 @@ private static BigQueryTableFieldSchema generateTableFieldSchema(Schema.Field fi
...
+    String name = field.getName();
+    LegacySQLTypeName type = getTableDataType(field.getSchema());
```
snehalakshmisha, line 377: Also worth extracting `field.getSchema()` into a variable?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -441,4 +502,54 @@ private static BigQueryFileFormat getFileFormat(List<BigQueryTableFieldSchema> f
...
+        return "";
+    }
+  }
+
+  private static String formatPartitionFilter(String partitionFilter) {
+    String[] queryWords = partitionFilter.split(" ");
+    int index = 0;
+    for (String word: queryWords) {
+      if (COMPARISON_OPERATORS.contains(word.toUpperCase())) {
+        queryWords[index - 1] = queryWords[index - 1].replace(queryWords[index - 1],
```
snehalakshmisha, line 549: In the first iteration, if the condition is true then this will cause ArrayIndexOutOfBoundsException?

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtilsTest.java
fernst: yes! Removed now

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -353,6 +364,56 @@ private static BigQueryTableFieldSchema generateTableFieldSchema(Schema.Field fi
...
+                                           .collect(Collectors.toList()));
+
+      fieldBuilder = Field.newBuilder(name, type, subFields);
+    } else {
+      fieldBuilder = Field.newBuilder(name, type);
+    }
+
+    fieldBuilder.setMode(mode);
+
+    // Set prevision for numeric fields
```
tivv: nit: typo `prevision`

tivv (commented):
Please add tests for the schema mapping

fernst:
> Please add tests for the schema mapping

Added @tivv

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryJoinDataset.java
```diff
@@ -116,13 +120,17 @@ public void executeJoin() {
...
+    updateTableSchema(destinationTable, joinDefinition.getOutputSchema());
+
     // Run BigQuery job with generated SQL statement, store results in a new table, and set priority to BATCH
     // TODO: Make priority configurable
     QueryJobConfiguration queryConfig =
       QueryJobConfiguration.newBuilder(query)
         .setDestinationTable(destinationTable)
         .setCreateDisposition(JobInfo.CreateDisposition.CREATE_NEVER)
-        .setWriteDisposition(JobInfo.WriteDisposition.WRITE_TRUNCATE)
+        .setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND)
```
tivv, line 132: Why are we changing this?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryWrite.java
```diff
@@ -0,0 +1,461 @@
...
+        relaxTableSchema(schema, destTable);
+      }
+    } else {
+      createTable(schema, destinationTableId, sinkConfig, newDestinationTable);
+    }
+
+    // Get query job configuration based on wether the job is an insert or update/upsert
+    QueryJobConfiguration.Builder queryConfigBuilder;
+
+    if (Operation.INSERT.equals(operation)) {
```
tivv: nit: you can use `==` for enums

tivv (approved):
few nits

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
fernst: This means the expression would be invalid. A partition filter is a boolean expression so it has to be something like:

'a < b' or 'x = y' and so on.

Since we are only looking for the comparison operators when evaluating, we are guaranteed to have a "previous" element.

tivv: I think it's more about "invalid" query, e.g. if anyone tries to start expression with "<" this would fail with ArrayIndexOutOfBoundsException instead of a nice one

PR #886, opened 2022-02-02
BigQuery and GCS e2e test scenarios

sawantpritam:
@itsankit-google @tivv @rmstar Please review.


src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -118,27 +119,50 @@ public static void deleteTempTargetBQTable() throws IOException, InterruptedExce
...
+    try {
+      BigQueryClient.getSoleQueryResult("create table `test_automation." + bqSourceTable + "` " +
+                                          "(transaction_id INT64, transaction_uid STRING, transaction_date DATE ) " +
+                                          "PARTITION BY _PARTITIONDATE");
+      BigQueryClient.getSoleQueryResult("INSERT INTO `test_automation." + bqSourceTable + "` " +
+                                          "(transaction_id, transaction_uid, transaction_date) " +
+                                          "SELECT ROW_NUMBER() OVER(ORDER BY GENERATE_UUID()), GENERATE_UUID(), date " +
+                                          "FROM UNNEST(GENERATE_DATE_ARRAY('2022-01-01', current_date())) AS date");
+    } catch (NoSuchElementException e) {
+    }
```
rmstar, line 168: When would NoSuchElementException be thrown, and why do we want to ignore it? Can you add a comment?

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQuery.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.Date;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * BigQuery related stepDesigns.
+ */
+public class BigQuery implements CdfHelper {
+
+  static {
```
rmstar: Looks like BQ source and sink step designs are in the same class. I think it would be more clear if we had separate classes for BQ source and sink because the plugin configurations are different. If there are configs that are common to source and sink (ex - project id), we could put that in a base class.
WDYT @itsankit-google @tivv ?

itsankit-google: +1. I think the structure of step definitions also can be similar to structure of features, i.e. `bigquery/stepsdesign/BigQuerySource.java`. Can add a `BigQueryBase` class for common step defintions. `BigQuerySource` and `BigQuerySink` to have specific step definitions. Similar approach can be taken for GCS as well.

sawantpritam: I don't think we need to have separate step definition files.  We are already indicating in stepdef and method name if property is sink or source specific.
for eg: 
`@Then("Enter BigQuery source property filter {string}")`
`@Then("Toggle BigQuery sink property truncateTable to true")`

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
```diff
@@ -0,0 +1,33 @@
...
+    Given Open Datafusion Project to configure pipeline
+    When Source is GCS
+    When Sink is BigQuery
+    Then Open GCS source properties
+    Then Enter the GCS source mandatory properties
+    Then Validate "GCS" plugin properties
+    Then Close the GCS properties
+    Then Open BigQuery sink properties
+    Then Enter BigQuery property reference name
+    Then Enter BigQuery properties projectId "projectId" and datasetProjectId "projectId"
```
itsankit-google: Please separate the steps

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
```diff
@@ -0,0 +1,33 @@
...
+    Then Enter BigQuery sink property dataset "dataset" and table name
```
itsankit-google: Similar comment, please separate the steps

itsankit-google: Also, any specific reason to not pass table name as a string? This would make the step more flexible.

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
```diff
@@ -0,0 +1,33 @@
...
+    Then Enter BiqQuery property encryption key name if cmek is enabled
```
itsankit-google: Similar comment pass encryption key name as argument to the step definition.

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
```diff
@@ -0,0 +1,33 @@
+@BigQuery_Sink
+Feature: BigQuery sink - Verification of GCS to BigQuery successful data transfer
+
+  @CMEK @GCS_CSV_TEST @BQ_TARGET_TEST
```
itsankit-google: Change TARGET->SINK in all the tags.

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/CommonStepDefinitions.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * GCP Plugin's common stepDesigns.
+ */
+public class CommonStepDefinitions implements CdfHelper {
```
itsankit-google: I don't find anything specific to google-cloud in these step definitions. Can we move them to framework?

sawantpritam: Raised PR on e2e framework https://github.com/cdapio/cdap-e2e-tests/pull/64 
Please review.

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
sawantpritam: BQ Insert query does not return any record. Iterator on TableResult values in getSoleQueryResult method throws NoSuchElementException.

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQuery.java
sawantpritam: Created separate classes for source and sink step definitions.

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
sawantpritam: Separate steps added.

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
sawantpritam: Separate steps added.
We are creating sample table before test scenario.

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
sawantpritam: Done

src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
sawantpritam: Done

itsankit-google (commented):
Overall looks good, add cmek step in the BQ source related tests as well, it is used to encrypt data stored in temporary GCS bucket, can skip the verification step there as the temporary bucket gets cleaned up by the plugin at the end of the pipeline.

sawantpritam:
> Overall looks good, add cmek step in the BQ source related tests as well, it is used to encrypt data stored in temporary GCS bucket, can skip the verification step there as the temporary bucket gets cleaned up by the plugin at the end of the pipeline.

Added BQ source test case with cmek step.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
```diff
@@ -0,0 +1,187 @@
...
+    CdfGcsActions.closeButton();
+  }
+
+  @Then("Enter the GCS properties with blank property {string}")
+  public void enterTheGCSPropertiesWithBlankProperty(String property) throws InterruptedException {
+    if (!property.equalsIgnoreCase("referenceName")) {
+      CdfGcsActions.enterReferenceName();
+    }
+    if (!property.equalsIgnoreCase("path")) {
+      CdfGcsActions.getGcsBucket("gs://cdf-athena-test/dummy");
```
rmstar, line 81: bucket path should be a parameter, don't hardcode it

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+    if (!property.equalsIgnoreCase("format")) {
+      CdfGcsActions.selectFormat("csv");
```
rmstar, line 84: are we also testing other formats?

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
```diff
@@ -0,0 +1,187 @@
...
+import org.junit.Assert;
+import stepsdesign.BeforeActions;
+
+import java.io.IOException;
+import java.util.Optional;
+
+/**
+ * GCS Plugin related common step design.
+ */
+public class GCSBase implements CdfHelper {
```
rmstar, line 43: where is this class used?

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
```diff
@@ -0,0 +1,169 @@
...
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * BigQuery related common stepDesigns.
+ */
+public class BigQueryBase implements CdfHelper {
```
rmstar, line 40: where is this class used?

itsankit-google: I think it is containing common step definitions for BQ plugins.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
itsankit-google: And similarly this class is containing common step definitions for GCS plugins.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
sawantpritam: This dummy path is hardcoded here just to validate mandatory property error. Otherwise we are creating buckets at runtime before test scenario and passing that path to the tests.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSBase.java
sawantpritam: Yes we have validation tests with other formats in `GCSourceSchema.feature` and  `GCSSourceToBigQuery.feature`

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
rmstar: Which class inherits from  BigQueryBase? I thought BigQuerySource, BigQuerySink would inherit from BigQueryBase, but they inherit from CdfHelper

itsankit-google: I don't think cucumber allow inheritance of step definitions.

itsankit-google: The way to solve this problem is using composition or interfaces @sawantpritam PTAL.

rmstar: Thanks Ankit

rmstar (approved):

PR #888, opened 2022-02-03
Add BigQuery implementations of relational classes for BQ Aggregation pushdown

DJSagarAhire:
This PR adds BigQuery-specific implementations of `Relation`, `Expression` and `ExpressionFactory`. The BQ implementation of `Relation` currently does not support the `groupBy()` and `deduplicate()` operations, which will be added in a future PR.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
@@ -1,0 +1,117 @@
+package io.cdap.plugin.gcp.bigquery.relational;
+
+import io.cdap.cdap.etl.api.aggregation.DeduplicateAggregationDefinition;
+import io.cdap.cdap.etl.api.aggregation.GroupByAggregationDefinition;
+import io.cdap.cdap.etl.api.relational.Expression;
+import io.cdap.cdap.etl.api.relational.Relation;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An implementation of {@link Relation} designed to operate on BigQuery.
+ */
+public class BigQueryRelation implements Relation {
+  private final boolean isValid;
+  private final String validationError;
+
+  private final Map<String, Expression> columns;
+  private final Expression filter;
+  private final Relation baseRelation;
+  private final String datasetName;
+
+  /**
+   * Initializes a new {@link BigQueryRelation} using the set of columns specified.
+   * @param columnSet set of strings indicating column names in the relation.
+   */
+  public BigQueryRelation(String datasetName, Set<String> columnSet) {
+    SQLExpressionFactory factory = new SQLExpressionFactory();
+
+    Map<String, Expression> columns = new HashMap<>();
+    for (String column: columnSet) {
+      columns.put(column, factory.compile(column));
+    }
+    this.datasetName = datasetName;
+    this.columns = columns;
+    this.isValid = true;
+    this.validationError = null;
+    this.filter = null;
+    this.baseRelation = null;
+  }
+
+  private BigQueryRelation(String datasetName,
+                           Map<String, Expression> columns,
+                           Expression filter,
+                           Relation baseRelation) {
+    this(datasetName, columns, filter, baseRelation, true, null);
+  }
+
+  private BigQueryRelation(String datasetName,
+                           Map<String, Expression> columns,
+                           Expression filter,
+                           Relation baseRelation,
+                           boolean isValid,
+                           String validationError) {
+    this.datasetName = datasetName;
+    this.columns = columns;
+    this.filter = filter;
+    this.baseRelation = baseRelation;
+    this.isValid = isValid;
+    this.validationError = validationError;
+  }
+
+  private Relation getInvalidRelation(String validationError) {
+    return new BigQueryRelation(null, null, null, null, false, validationError);
+  }
+
+  @Override
+  public boolean isValid() {
+    return isValid;
+  }
+
+  @Override
+  public String getValidationError() {
+    return isValid() ? null : validationError;
+  }
+
+  public Relation getBaseRelation() {
+    return baseRelation;
+  }
+
+  @Override
+  public Relation setColumn(String column, Expression value) {
+    Map<String, Expression> newColumns = new HashMap<>(columns);
+    newColumns.put(column, value);
+
+    return new BigQueryRelation(datasetName, newColumns, filter, this);
+  }
+
+  @Override
+  public Relation dropColumn(String column) {
+    Map<String, Expression> newColumns = new HashMap<>(columns);
+    newColumns.remove(column);
+
+    return new BigQueryRelation(datasetName, newColumns, filter, this);
+  }
+
+  @Override
+  public Relation select(Map<String, Expression> columns) {
+    return new BigQueryRelation(datasetName, columns, filter, this);
+  }
+
+  @Override
+  public Relation filter(Expression filter) {
+    return new BigQueryRelation(datasetName, columns, filter, this);
+  }
+
+  @Override
+  public Relation groupBy(GroupByAggregationDefinition aggregationDefinition) {
+    return getInvalidRelation("Group by operation not supported currently.");
+  }
+
+  @Override
+  public Relation deduplicate(DeduplicateAggregationDefinition aggregationDefinition) {
+    return getInvalidRelation("Deduplicate operation not supported currently.");
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpression.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpression.java
@@ -1,0 +1,73 @@
+package io.cdap.plugin.gcp.bigquery.relational;
+
+import io.cdap.cdap.etl.api.relational.Expression;
+
+import java.util.Objects;
+
+/**
+ * A default implementation of Expression that simply stores the expression as a string.
+ * The string is assumed to be ANSI SQL compliant.
+ * It is the responsibility of the factory creating {@link SQLExpression} objects to enforce correctness of
+ * the expression.
+ */
+public class SQLExpression implements Expression {
+  private final String expression;
+
+  /**
+   * Creates a {@link SQLExpression} from the specified SQL string.
+   * @param expression a String containing the SQL expression.
+   */
+  public SQLExpression(String expression) {
+    this.expression = expression;
+  }
+
+  /**
+   *
+   * @return the SQL string contained in the {@link SQLExpression} object.
+   */
+  public String getExpression() {
+    return expression;
+  }
+
+  /**
+   * A {@link SQLExpression} is always assumed to contain valid SQL. It is the responsibility of the creator of the
+   * object to ensure correctness of the SQL string.
+   * @return This method always returns true.
+   */
+  @Override
+  public boolean isValid() {
+    return true;
+  }
+
+  /**
+   * A {@link SQLExpression} is always assumed to contain valid SQL. It is the responsibility of the creator of the
+   * object to ensure correctness of the SQL string.
+   * @return This method always returns a null string.
+   */
+  @Override
+  public String getValidationError() {
+    return null;
+  }
+
+  /**
+   * Two {@link SQLExpression} objects are considered equal if they contain equal SQL expressions.
+   * @param o other object to be compared for equality.
+   * @return true iff both objects are {@link SQLExpression}s with equal SQL strings.
+   */
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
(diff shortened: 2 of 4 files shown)


fernst (approved):
LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpression.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Creates a {@link SQLExpression} from the specified SQL string.
+   * @param expression a String containing the SQL expression.
+   */
+  public SQLExpression(String expression) {
+    this.expression = expression;
+  }
+
+  public String getExpression() {
```
fernst, line 28: nit: add a javadoc here saying "This returns the wrapped expression" or something like this.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * An implementation of {@link Relation} designed to operate on BigQuery.
+ */
+public class BigQueryRelation implements Relation {
+  private final boolean isValid;
+  private final String validationError;
+
+  private final Set<String> columns;
+  private final Map<String, Expression> columnValues;
```
tivv: Why do you need separately columns and columnValues? In which cases columns!=columnValues.keySet?

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Relation setColumn(String column, Expression value) {
+    Set<String> newColumns = new HashSet<>(columns);
+    Map<String, Expression> newColumnValues = new HashMap<>(columnValues);
+    newColumns.add(column);
+    newColumnValues.put(column, value);
+
+    return new BigQueryRelation(newColumns, newColumnValues, filter);
```
tivv: Passing filter like this won't work as filter may depend on other columns being dropped. 
E.g. `r.setFilter("a == 5").dropColumn("a")` won't work

tivv: Actually, same applies to any operations, e.g. consider `r.setColumn("a", "a+b").setColumn("a", "a/c")`. You should remember all nesting. You may do flattening, but only if you ensure correctness

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -506,6 +512,29 @@ public void cleanup(String datasetName) throws SQLEngineException {
...
+  @Override
+  public List<ExpressionFactory<?>> getExpressionFactories() {
+    return Collections.singletonList(new SQLExpressionFactory());
+  }
+
+  @Override
+  public Engine getRelationalEngine() {
+    return this;
+  }
+
```
tivv, line 451: You also need `getRelation` and `transform`

tivv (changes requested):

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
fernst: `r.setFilter("a == 5").dropColumn("a")` is a valid statement, though.

Input: columns a,b,c
`r.setFilter("a = 5").dropColumn("a")` means:

`select b,c from <table> where a = 5`

I do agree with `r.setColumn("a", "a+b").setColumn("a", "a/c")` is invalid, but I'm inclined to say this should simply override the initial definition with the latest one.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
fernst: Agree. Essentially all columns can be expressed as `a -> a , b -> b, c -> c` unless modified.

In the end, the result will be

`select a as a, b as b, c as c from <table>` into a new table.

DJSagarAhire: If you look at the constructor, initially `columnValues` is empty, and it is later populated with mappings if required, which is why we have both `columns` and `columnValues`. However, we can get rid of `columns` and use `columnValues` to store the columns as well with `columnValues` initialised to self-mappings.

tivv: And name it "columns" or "columnExpressions" as it has both names and expressions

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
tivv: `r.setColumn("a", "a+b").setColumn("a", "a/c")` is actually valid as you have a new Relation created after each call. Each time you do an op you do on top of existing relation.
In SQL is expressed as subselect, e.g. `select <columns except a>, a/c as a from (select <columns except a>, a+b as a  from table)`.
Overall quite often this can be optimized to remove nested subselects, but I am not sure if we need to do it. We can just go with subselects now, DB optimizers can often handle it. One special case can be adding a new column - in this case subselect can be omited. This would help to optimize r.setColumn("a1", expr).setColumn("a2", expr).... use case

tivv: But point is that when you have a relation it should work irrespective of if it's a table or derived relation.
So after
`Relation x = r.setColumn("a", "a+b")` you have a relation x with column a set to certain values that is different from values in relation r. And `x.setColumn("a", "a/c")` should give correct result different from `r.setColumn("a", "a/c")`

fernst: I see, I thought the pattern was more similar to a Builder.

The only potential red flag with nested statements is that there may be limits in how many "layers" of nesting the underlying engine may support. This can cause issues.

Another potential solution is to tokenize the expression and replace each token with the latest "expression" for this token.

EDIT: After some consideration, the tokenization idea seems very complicated to maintain.

fernst: So 

`rel.setColumn("a","a+b").setColumn("a","a/c")` 

becomes something like

`Select ... , (a+b)/c as a ...`

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
DJSagarAhire: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
DJSagarAhire: So currently this means that the relation will remain unchanged as long as we accept arbitrary nesting of sub-selects, correct?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
DJSagarAhire: Added `getRelation`. `transform` will be added in a subsequent PR.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpression.java
DJSagarAhire: Done.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -0,0 +1,104 @@
...
+import io.cdap.cdap.etl.api.relational.Relation;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An implementation of {@link Relation} designed to operate on BigQuery.
+ */
+public class BigQueryRelation implements Relation {
```
fernst, line 15: How does each relation keep track of the "previous" transformations applied before?

`BigQueryRelation` may also need to have a property to represent the "previous" relation. All relations start with a SQL table being the original relation.

 As more transformations are applied, the output from the previous relation gets generated (SQL statement) and this is used in the `FROM ...` for each successive call. This way we can build the SQL in a way that references the schema changes for other columns.

fernst: So every transformation essentially uses the output SQL for this Relation as the "source" table for the next transformation.

fernst: I see the source table can be obtained when executing the `transform` method. 

Still, each `BigQueryRelation` should have a reference to the "parent" relation so these nested statements can be generated.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
sanjanasandeep: Should it have ```aggregate``` and ```canAggregate``` (sorry I am following the design doc and it is  specified under "changes to BigQuerySQL" https://docs.google.com/document/d/1hYnP_jAday-RKI0TPkJX_r2iJeroZXZAZB8XWx1nTDI/edit?resourcekey=0-lS0CoG1Ji02whjnCwRZJ_w# ) or wondering if  it needs ```transform``` and ```canTransform```? (since it implements engine)

DJSagarAhire: `aggregate` and `canAggregate` will be added as helper methods which will be called from `transform` and `canTransform` respectively. In the future we can have `transform` and `canTransform` call methods such as `filter` and `canFilter`, `join` and `canJoin`, etc.

sanjanasandeep: got it. Thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
tivv: FYI: This is my original POC draft PR https://github.com/data-integrations/google-cloud/pull/779

tivv: In it I was tracking "parent" dataset: https://github.com/data-integrations/google-cloud/blob/268d4670b5e8351d780cc08e71899fd207b86197/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/dl/BigQueryDLDataSet.java#L16

tivv (changes requested):
Overall I don't see how relation relates to it's parent, being it another relation or the source table.

tivv:
Generally when you have a relation you should be able to generate a select statement out of it. So it should have "select", "from" and optionally "where" to be generated.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
tivv: Yep, relation is immutable. You can use a single relation and produce multiple derivative relations (e.g. to join). Generally it's very similar to spark DataSet  concept. It's not a builder as there is no "build" to mark "I am done with this" and introducing build would be unnecessary API complications

tivv: I tried to highlight it in the method javadocs (e.g. https://github.com/cdapio/cdap/blob/6c60f2d8ebeb871690382a10caf40aa532a384f7/cdap-app-templates/cdap-etl/cdap-etl-api/src/main/java/io/cdap/cdap/etl/api/relational/Relation.java#L72), but may be it should be made more clear in the class javadoc

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
DJSagarAhire: Added fields for `baseRelation` and `datasetName`.

fernst:
> Generally when you have a relation you should be able to generate a select statement out of it. So it should have "select", "from" and optionally "where" to be generated.

@tivv can we get this PR merged for now? This will unblock me to start building the SQL statements for Deduplicate/GroupBy and I'll implement similar logic to what you had to build each successive SQL statement and wrap the previous one.

tivv (approved):
Approving to unblock development given we fix the comments in scope of https://cdap.atlassian.net/browse/CDAP-18839 in 6.7

PR #889, opened 2022-02-07
Added the options for sampling properties in GCS connector

sechegaray:


yaojiefeng (commented):
You will need to add changes to the widget file so UI can show them

fernst (approved):
LGTM

fernst (approved):
LGTM

PR #894, opened 2022-02-10
Added support for Group By and Deduplication Transform operations in the BigQuerySQLEngine

fernst:
Implemented support for Relational Transform operations in the SQL engine.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
@@ -7,11 +7,11 @@ import io.cdap.cdap.etl.api.relational.InvalidRelation;
 import io.cdap.cdap.etl.api.relational.Relation;
 import io.cdap.plugin.gcp.bigquery.sqlengine.BigQuerySQLDataset;
 import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQueryDeduplicateSQLBuilder;
+import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQueryGroupBySQLBuilder;
 import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQueryNestedSelectSQLBuilder;
 import io.cdap.plugin.gcp.bigquery.sqlengine.builder.BigQuerySelectSQLBuilder;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
@@ -44,15 +44,14 @@ public class BigQueryRelation implements Relation {
                                              String bqDataset,
                                              BigQuerySQLDataset sourceDataset,
                                              Set<String> columnNames) {
-
-    Map<String, String> selectedColumns = getSelectedColumns(columnNames);
+    Map<String, Expression> selectedColumns = getSelectedColumns(columnNames);
     String sourceTableName = String.format("`%s.%s.%s`", bqProject, bqDataset, sourceDataset.getBigQueryTableName());
     String transformExpression = buildBaseSelect(selectedColumns, sourceTableName, sourceDataset.getDatasetName());
 
     return new BigQueryRelation(sourceDataset, columnNames, null, transformExpression, true, null);
   }
 
-  public BigQueryRelation(BigQuerySQLDataset sourceDataset,
+  protected BigQueryRelation(BigQuerySQLDataset sourceDataset,
                           Set<String> columns,
                           BigQueryRelation parent,
                           String transformExpression,
@@ -95,8 +94,8 @@ public class BigQueryRelation implements Relation {
       return getInvalidRelation("Unsupported or invalid expression type.");
     }
 
-    Map<String, String> selectedColumns = getSelectedColumns(columns);
-    selectedColumns.put(column, ((SQLExpression) value).getExpression());
+    Map<String, Expression> selectedColumns = getSelectedColumns(columns);
+    selectedColumns.put(column, value);
 
     // Build new transform expression and return new instance.
     String expression = buildNestedSelect(selectedColumns, transformExpression, sourceDataset.getDatasetName(), null);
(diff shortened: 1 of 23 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,351 @@
...
+  private final BigQuerySQLDataset sourceDataset;
+  private final BigQueryRelation parent;
+  private final String transformExpression;
 
   /**
-   * Initializes a new {@link BigQueryRelation} using the set of columns specified.
-   * @param columnSet set of strings indicating column names in the relation.
+   * Gets a new BigQueryRelation instance
+   *
+   * @param bqProject     Project where the source SQL table is stored
```
tivv: Won't sourceDataset has it? Another question is that, are you sure that sourceDataset would be available when you make the relation? It's pretty early in the game, there is no job yet. At this point we did not even decide if we will do it in BQ

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -432,23 +433,80 @@ public void cleanup(String datasetName) throws SQLEngineException {
...
+    Set<String> columnSet = new LinkedHashSet<>();
     List<Schema.Field> fields = relationDefinition.getSchema().getFields();
     if (fields != null) {
       for (Schema.Field field: fields) {
         columnSet.add(field.getName());
       }
     }
 
-    return new BigQueryRelation(relationDefinition.getDatasetName(), columnSet);
+    BigQuerySQLDataset sqlDataset = datasets.get(relationDefinition.getDatasetName());
```
tivv: I don't think you have dataset already

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
fernst: I've refactored the code to supply the source datasets once the engine has all of them available to execute.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
fernst: Refactored to handle this scenario.

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,409 @@
...
-  private final String validationError;
+  private static final SQLExpressionFactory factory = new SQLExpressionFactory();
 
-  private final Map<String, Expression> columns;
-  private final Expression filter;
-  private final Relation baseRelation;
   private final String datasetName;
+  private final Set<String> columns;
+  private final BigQueryRelation parent;
+  private Map<String, BigQuerySQLDataset> sourceDatasets;
```
tivv, line 32: nit: put non-final fields below final ones

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,409 @@
...
+  private final Supplier<String> expressionSupplier;
```
tivv, line 33: nit: name it `sqlExpressionSupplier` or even `sqlStatementSupplier` to highlight what's that. Otherwise the name is very generic and does not tell what's inside

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,409 @@
...
-  }
-
-  private BigQueryRelation(String datasetName,
-                           Map<String, Expression> columns,
-                           Expression filter,
-                           Relation baseRelation) {
-    this(datasetName, columns, filter, baseRelation, true, null);
+    this.parent = null;
+    this.expressionSupplier = () -> {
+      BigQuerySQLDataset sourceDataset = sourceDatasets.get(datasetName);
```
tivv, line 54: Please handle the case when it's null. If for whatever reason it get out of sync it would be hard to diagnose with just NPE

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,409 @@
...
+    }
+
+    // Remove column
+    Map<String, Expression> selectedColumns = getSelectedColumns(columns);
+    selectedColumns.remove(column);
 
-    return new BigQueryRelation(datasetName, newColumns, filter, this);
+    // Build new transform expression and return new instance.
+    Supplier<String> supplier =
+      () -> buildNestedSelect(selectedColumns, getTransformExpression(), datasetName, null);
```
tivv, line 176: nit: you probably don't need to build a nested select here. You can simply drop a column from the list and use the original expression

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -1,117 +1,409 @@
...
+      () -> buildNestedSelect(columns, getTransformExpression(), datasetName, null);
+    return new BigQueryRelation(datasetName, columns.keySet(), this, supplier);
   }
 
   @Override
   public Relation filter(Expression filter) {
-    return new BigQueryRelation(datasetName, columns, filter, this);
+    // check if expression is supported and valid
+    if (!supportsExpression(filter)) {
+      return getInvalidRelation("Unsupported or invalid expression type.");
```
tivv, line 197: It would be great to propagate expression validity message here (and in other calls) if expression is not valid. Otherwise it would be hard to debug.

tivv: Can be done in follow up PR though to get this one out

tivv (approved):
Left a number of comments, but those can be addressed separately

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
fernst: I'll address your comments in a follow up (I'll open it in a couple of hours probably)

PR #903, opened 2022-02-22
e2e BQ GCS tests with macro

sawantpritam:

### DIFF
--- a/src/e2e-test/features/bigquery/source/BigQueryToGCS_WithMacro.feature
+++ b/src/e2e-test/features/bigquery/source/BigQueryToGCS_WithMacro.feature
@@ -1,0 +1,64 @@
+@BigQuery_Source
+Feature: BigQuery source - Verification of BigQuery to GCS successful data transfer with macro arguments
+
+  @CMEK @BQ_SOURCE_TEST @GCS_SINK_TEST
+  Scenario:Validate successful records transfer from BigQuery to GCS with macro arguments
+    Given Open Datafusion Project to configure pipeline
+    When Source is BigQuery
+    When Sink is GCS
+    Then Open BigQuery source properties
+    Then Enter BigQuery property reference name
+    Then Enter BigQuery property "projectId" as macro argument "bqProjectId"
+    Then Enter BigQuery property "datasetProjectId" as macro argument "bqDatasetProjectId"
+    Then Enter BigQuery property "serviceAccountFilePath" as macro argument "bqServiceAccount"
+    Then Enter BigQuery property "dataset" as macro argument "bqDataset"
+    Then Enter BigQuery property "table" as macro argument "bqSourceTable"
+    Then Validate "BigQuery" plugin properties
+    Then Close the BigQuery properties
+    Then Open GCS sink properties
+    Then Enter GCS property reference name
+    Then Enter GCS property "projectId" as macro argument "gcsProjectId"
+    Then Enter GCS property "serviceAccountFilePath" as macro argument "gcsServiceAccount"
+    Then Enter GCS property "path" as macro argument "gcsSinkPath"
+    Then Enter GCS sink property "pathSuffix" as macro argument "gcsPathSuffix"
+    Then Enter GCS property "format" as macro argument "gcsFormat"
+    Then Enter GCS sink cmek property "encryptionKeyName" as macro argument "cmekGCS" if cmek is enabled
+    Then Validate "GCS" plugin properties
+    Then Close the GCS properties
+    Then Connect source as "BigQuery" and sink as "GCS" to establish connection
+    Then Save the pipeline
+    Then Preview and run the pipeline
+    Then Enter runtime argument value "projectId" for key "bqProjectId"
+    Then Enter runtime argument value "projectId" for key "bqDatasetProjectId"
+    Then Enter runtime argument value "serviceAccountAutoDetect" for key "bqServiceAccount"
+    Then Enter runtime argument value "dataset" for key "bqDataset"
+    Then Enter runtime argument value for BigQuery source table name key "bqSourceTable"
+    Then Enter runtime argument value "projectId" for key "gcsProjectId"
+    Then Enter runtime argument value "serviceAccountAutoDetect" for key "gcsServiceAccount"
+    Then Enter runtime argument value for GCS sink property path key "gcsSinkPath"
+    Then Enter runtime argument value "gcsPathDateSuffix" for key "gcsPathSuffix"
+    Then Enter runtime argument value "csvFormat" for key "gcsFormat"
+    Then Enter runtime argument value "cmekGCS" for GCS cmek property key "cmekGCS" if GCS cmek is enabled
+    Then Run the preview of pipeline with runtime arguments
+    Then Verify the preview of pipeline is "success"
+    Then Click on preview data for GCS sink
+    Then Close the preview data
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Enter runtime argument value "projectId" for key "bqProjectId"
+    Then Enter runtime argument value "projectId" for key "bqDatasetProjectId"
+    Then Enter runtime argument value "serviceAccountAutoDetect" for key "bqServiceAccount"
+    Then Enter runtime argument value "dataset" for key "bqDataset"
+    Then Enter runtime argument value for BigQuery source table name key "bqSourceTable"
+    Then Enter runtime argument value "projectId" for key "gcsProjectId"
+    Then Enter runtime argument value "serviceAccountAutoDetect" for key "gcsServiceAccount"
+    Then Enter runtime argument value for GCS sink property path key "gcsSinkPath"
+    Then Enter runtime argument value "gcsPathDateSuffix" for key "gcsPathSuffix"
+    Then Enter runtime argument value "csvFormat" for key "gcsFormat"
+    Then Enter runtime argument value "cmekGCS" for GCS cmek property key "cmekGCS" if GCS cmek is enabled
+    Then Run the Pipeline in Runtime with runtime arguments
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Verify data is transferred to target GCS bucket
+    Then Validate the cmek key "cmekGCS" of target GCS bucket if cmek is enabled
--- a/src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
+++ b/src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
@@ -19,12 +19,11 @@ import io.cdap.e2e.pages.actions.CdfBigQueryPropertiesActions;
 import io.cdap.e2e.pages.actions.CdfStudioActions;
 import io.cdap.e2e.pages.locators.CdfStudioLocators;
 import io.cdap.e2e.utils.BigQueryClient;
+import io.cdap.e2e.utils.ElementHelper;
 import io.cdap.e2e.utils.PluginPropertyUtils;
-import io.cdap.e2e.utils.SeleniumDriver;
 import io.cdap.e2e.utils.SeleniumHelper;
 import io.cdap.plugin.common.stepsdesign.TestSetupHooks;
 import io.cdap.plugin.utils.E2EHelper;
-import io.cucumber.java.en.And;
 import io.cucumber.java.en.Then;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Assert;
@@ -189,7 +188,7 @@ public class BigQueryBase implements E2EHelper {
                                                                                    String runtimeArgumentKey) {
     String cmekBQ = PluginPropertyUtils.pluginProp(value);
     if (cmekBQ != null) {
-      CdfStudioLocators.runtimeArgsValue(runtimeArgumentKey).sendKeys(cmekBQ);
+      ElementHelper.sendKeys(CdfStudioLocators.runtimeArgsValue(runtimeArgumentKey), cmekBQ);
       BeforeActions.scenario.write("BigQuery encryption key name - " + cmekBQ);
       return;
     }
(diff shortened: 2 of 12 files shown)


rmstar (commented):
Please add a test for the case where input schema is a macro

rmstar (commented):
Let's also test service account config set as macro.

sawantpritam:
> Please add a test for the case where input schema is a macro

Schema is passed as macro argument in 
`Scenario:Validate successful records transfer from GCS to BigQuery with macro arguments`

sawantpritam:
> Let's also test service account config set as macro.

Added step to pass service account path as macro

src/e2e-test/java/io/cdap/plugin/utils/CdfPluginProperty.java
```diff
@@ -0,0 +1,33 @@
+package io.cdap.plugin.utils;
+
+/**
+ * Plugin property data-cy attributes of elements which will be used to locate element by xpath.
+ */
+public enum CdfPluginProperty {
```
itsankit-google: You can create a mapping b/w `projectId` and `CdfPluginProperty.PROJECT_ID`. Similarly for all properties. This way you need to maintain only one step definition for all.
For example,
```
private static final Map<String, CdfPluginProperty> CDF_PLUGIN_PROPERTY_MAPPING;
  static {
    CDF_PLUGIN_PROPERTY_MAPPING = new ImmutableMap.Builder<String, CdfPluginProperty>()
    .put("projectId", CdfPluginProperty.PROJECT_ID)
    .build();
    }
  @Nullable
  public static CdfPluginProperty fromPropertyString(String property) {
    return CDF_PLUGIN_PROPERTY_MAPPING.get(property);
  }
```

src/e2e-test/features/bigquery/source/BigQueryToGCS_WithMacro.feature
```diff
@@ -0,0 +1,40 @@
...
+Feature: BigQuery source - Verification of BigQuery to GCS successful data transfer with macro arguments
+
+  @MACRO_TEST @CMEK @BQ_SOURCE_TEST @GCS_SINK_TEST
+  Scenario:Validate successful records transfer from BigQuery to GCS with macro arguments
+    Given Open Datafusion Project to configure pipeline
+    When Source is BigQuery
+    When Sink is GCS
+    Then Open BigQuery source properties
+    Then Enter BigQuery property reference name
+    Then Enter BigQuery property projectId as macro argument "bqProjectId" value "projectId"
```
itsankit-google: Separate out the steps of passing key and value.
What I mean is, we should first setup all configs with macro. 
Then after deploying the pipeline or before preview run set the runtime arguments.

itsankit-google: This comment applies to all the steps.

src/e2e-test/java/io/cdap/plugin/utils/CdfPluginProperty.java
sawantpritam: Added mapping

src/e2e-test/features/bigquery/source/BigQueryToGCS_WithMacro.feature
sawantpritam: Steps separated.

rmstar:
> Added step to pass service account path as macro

Can we also test the following:
1. Set `Service Account Type` as macro param. 
2. Test with `Service Account Type` = JSON.

sawantpritam:
> > Added step to pass service account path as macro
> 
> Can we also test the following:
> 
> 1. Set `Service Account Type` as macro param.
> 2. Test with `Service Account Type` = JSON.

We will add that in separate PR.

rmstar:
> We will add that in separate PR.

Sure, can you create a Jira to track this.

sawantpritam:
> * Set `Service Account Type` as macro param.
> * Test with `Service Account Type` = JSON.

JIRA created https://cdap.atlassian.net/browse/CDAP-18889

itsankit-google (commented):
I think we should also capture logs of preview run to debug if it fails at preview step. This is something that needs to be done in all scenarios. So, it can be done in a separate PR and a separate JIRA to track it. @rmstar WDYT?

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSource.java
```diff
@@ -17,19 +17,26 @@
...
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 import java.io.IOException;
 
 /**
  * GCS Source Plugin related step design.
  */
-public class GCSSource implements CdfHelper {
+public class GCSSource implements CdfHelper, E2EHelper {
```
itsankit-google: IIUC, E2EHelper already extends CdfHelper so why we need to implement both the interfaces? Implementing E2EHelper only should be enough.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
```diff
@@ -31,7 +33,7 @@
 /**
  * GCS Sink Plugin related step design.
  */
-public class GCSSink implements CdfHelper {
+public class GCSSink implements CdfHelper, E2EHelper {
```
itsankit-google: similar comment in all the places

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
```diff
@@ -177,4 +178,26 @@ public void validateTheCmekKeyOfTargetBigQueryTableIfCmekIsEnabled(String cmek)
...
+  @Then("Enter BiqQuery property {string} as macro argument {string} if cmek is enabled")
+  public void enterBiqQueryPropertyAsMacroArgumentIfCmekIsEnabled(String pluginProperty, String macroArgument) {
+    String cmekBQ =  PluginPropertyUtils.pluginProp("cmekBQ");
+    if (cmekBQ != null) {
+      enterPropertyAsMacroArgument(pluginProperty, macroArgument);
+    }
+  }
+
+  @Then("Enter runtime argument value {string} for key {string} if BQ cmek is enabled")
+  public void enterRuntimeArgumentValueForKeyIfBQCmekIsEnabled(String value, String runtimeArgumentKey) {
```
itsankit-google: This method name is misleading, it is not generic method, only used for entering cmek property value, right?

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
```diff
@@ -177,4 +178,26 @@ public void validateTheCmekKeyOfTargetBigQueryTableIfCmekIsEnabled(String cmek)
...
     SeleniumDriver.waitForPageToLoad();
   }
+
+  @Then("Enter BigQuery property {string} as macro argument {string}")
+  public void enterBigQueryPropertyAsMacroArgument(String pluginProperty, String macroArgument) {
+    enterPropertyAsMacroArgument(pluginProperty, macroArgument);
+  }
+
+  @Then("Enter BiqQuery property {string} as macro argument {string} if cmek is enabled")
+  public void enterBiqQueryPropertyAsMacroArgumentIfCmekIsEnabled(String pluginProperty, String macroArgument) {
```
itsankit-google: Also, don't need this method to be generic. It can be something like `enterBigQueryCmekPropertyAsMacroAgrumentIfCmekIsEnabled`.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
```diff
@@ -81,4 +83,32 @@ public void validateTheCmekKeyOfTargetGCSBucketIfCmekIsEnabled(String cmek) thro
...
     }
   }
+
+  @Then("Enter GCS sink property {string} as macro argument {string}")
+  public void enterGCSSinkPropertyAsMacroArgument(String pluginProperty, String macroArgument) {
+    enterPropertyAsMacroArgument(pluginProperty, macroArgument);
+  }
+
+  @Then("Enter GCS sink property {string} as macro argument {string} if cmek is enabled")
+  public void enterGCSSinkPropertyAsMacroArgumentIfCmekIsEnabled(String pluginProperty, String macroArgument) {
```
itsankit-google: similar comment for this method.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
```diff
@@ -81,4 +83,32 @@ public void validateTheCmekKeyOfTargetGCSBucketIfCmekIsEnabled(String cmek) thro
...
+  }
+
+  @Then("Enter runtime argument value for GCS sink property path key {string}")
+  public void enterRuntimeArgumentValueForGCSSinkPropertyPathKey(String runtimeArgumentKey) {
+    CdfStudioLocators.runtimeArgsValue(runtimeArgumentKey)
+      .sendKeys("gs://" + TestSetupHooks.gcsTargetBucketName);
+  }
+
+  @Then("Enter runtime argument value {string} for key {string} if GCS cmek is enabled")
+  public void enterRuntimeArgumentValueForKeyIfGCSCmekIsEnabled(String value, String runtimeArgumentKey) {
```
itsankit-google: Similar comment for this method.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSource.java
sawantpritam: Updated.

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
sawantpritam: Changed in all the places

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
sawantpritam: renamed

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
sawantpritam: renamed

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
sawantpritam: renamed

src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSink.java
sawantpritam: renamed

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
```diff
@@ -177,4 +177,27 @@ public void validateTheCmekKeyOfTargetBigQueryTableIfCmekIsEnabled(String cmek)
...
+    if (cmekBQ != null) {
+      enterPropertyAsMacroArgument(pluginProperty, macroArgument);
+    }
+  }
+
+  @Then("Enter runtime argument value {string} for BigQuery cmek property key {string} if BQ cmek is enabled")
+  public void enterRuntimeArgumentValueForBigQueryCmekPropertyKeyIfBQCmekIsEnabled(String value,
+                                                                                   String runtimeArgumentKey) {
+    String cmekBQ = PluginPropertyUtils.pluginProp(value);
+    if (cmekBQ != null) {
```
itsankit-google, line 190: Also add a log when it is null indicating that `CMEK is disabled`

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
```diff
@@ -177,4 +177,27 @@ public void validateTheCmekKeyOfTargetBigQueryTableIfCmekIsEnabled(String cmek)
...
+
+  @Then("Enter BigQuery property {string} as macro argument {string}")
+  public void enterBigQueryPropertyAsMacroArgument(String pluginProperty, String macroArgument) {
+    enterPropertyAsMacroArgument(pluginProperty, macroArgument);
+  }
+
+  @Then("Enter BigQuery cmek property {string} as macro argument {string} if cmek is enabled")
+  public void enterBigQueryCmekPropertyAsMacroArgumentIfCmekIsEnabled(String pluginProperty, String macroArgument) {
+    String cmekBQ =  PluginPropertyUtils.pluginProp("cmekBQ");
+    if (cmekBQ != null) {
```
itsankit-google, line 179: similar comment here and for GCS as well

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
sawantpritam: Added log

src/e2e-test/java/io/cdap/plugin/bigquery/stepsdesign/BigQueryBase.java
sawantpritam: Added log

sawantpritam:
> I think we should also capture logs of preview run to debug if it fails at preview step. This is something that needs to be done in all scenarios. So, it can be done in a separate PR and a separate JIRA to track it. @rmstar WDYT?

JIRA created https://cdap.atlassian.net/browse/CDAP-18892

itsankit-google (approved):
LGTM

PR #905, opened 2022-02-23
[CDAP-18778] add drive scope for bigquery to read from external tables

yaojiefeng:
According to https://cloud.google.com/bigquery/external-data-drive#java, one need to include the drive scope in the service account, by default it will not read from the drive scope

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -34,9 +34,13 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -163,6 +169,10 @@ public static String extractPrivateKey(ServiceAccountCredentials credentials) {
   public static BigQuery getBigQuery(String project, @Nullable Credentials credentials) {
     BigQueryOptions.Builder bigqueryBuilder = BigQueryOptions.newBuilder().setProjectId(project);
     if (credentials != null) {
+      if (credentials instanceof ServiceAccountCredentials
+            && ((ServiceAccountCredentials) credentials).getScopes().isEmpty()) {
```
seanzhougoogle: is that possible that the scope is not empty but doesn't contain the drive scope ?

yaojiefeng: Yea, I modified the logic to always add these scopes

seanzhougoogle (approved):

PR #907, opened 2022-02-23
[PLUGIN-670] Fixed partition boundaries query. Allow different projects to be set ...

sechegaray:
…for temporary tables

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
@@ -115,12 +115,12 @@ public class PartitionedBigQueryInputFormat extends AbstractBigQueryInputFormat<
       TableReference exportTableReference = createExportTableReference(type, datasetProjectId, datasetId,
                                                                        temporaryTableName, configuration);
       runQuery(configuration, bigQueryHelper, projectId, exportTableReference, query, location);
-      if (type == Type.VIEW || type == Type.MATERIALIZED_VIEW) {
-        configuration.set(BigQueryConfiguration.INPUT_PROJECT_ID_KEY,
-                          configuration.get(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_PROJECT));
-        configuration.set(BigQueryConfiguration.INPUT_DATASET_ID_KEY,
-                          configuration.get(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_DATASET));
-      }
+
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -115,12 +115,10 @@ private void processQuery(JobContext context) throws IOException, InterruptedExc
       TableReference exportTableReference = createExportTableReference(type, datasetProjectId, datasetId,
                                                                        temporaryTableName, configuration);
       runQuery(configuration, bigQueryHelper, projectId, exportTableReference, query, location);
-      if (type == Type.VIEW || type == Type.MATERIALIZED_VIEW) {
-        configuration.set(BigQueryConfiguration.INPUT_PROJECT_ID_KEY,
-                          configuration.get(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_PROJECT));
-        configuration.set(BigQueryConfiguration.INPUT_DATASET_ID_KEY,
-                          configuration.get(BigQueryConstants.CONFIG_VIEW_MATERIALIZATION_DATASET));
-      }
+      configuration.set(BigQueryConfiguration.INPUT_PROJECT_ID_KEY,
```
fernst, line 120: nit: Add a comment saying that these values come from the default sink config and can be overridden via config

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -291,20 +284,22 @@ private String generateTimePartitionCondition(StandardTableDefinition tableDefin
     }
 
     if (partitionFromDate != null) {
+      String columnFromTimestampName = columnName;
       if (LegacySQLTypeName.DATE.equals(columnType)) {
-        columnName = "TIMESTAMP(\"" + columnName + "\")";
+        columnFromTimestampName = "TIMESTAMP(`" + columnFromTimestampName + "`)";
```
fernst: `columnFromTimestampName` is a confusing name. Can you think of a better/simpler to understand column name?

fernst (commented):
Asked a question. Code looks good but I wonder if we can add some comments and a better variable name

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
sechegaray: done

fernst (approved):
LGTM

PR #916, opened 2022-03-03
Added logic to check if the BQ Dataset exists before trying to re-create in the BQ Execute plugin

fernst:


itsankit-google (dismissed):
LGTM. Thanks for the quick fix :)

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -119,9 +119,15 @@ public void run(ActionContext context) throws Exception {
     String tableName = config.getTable();
     String datasetProjectId = config.getDatasetProject();
     if (!Strings.isNullOrEmpty(datasetName) && !Strings.isNullOrEmpty(tableName)) {
-      BigQuerySinkUtils.createDataset(bigQuery, DatasetId.of(datasetProjectId, datasetName), config.getLocation(),
-                                      cmekKeyName, () -> String.format("Unable to create BigQuery dataset '%s.%s'",
-                                                                       datasetProjectId, datasetName));
+      DatasetId datasetId = DatasetId.of(datasetProjectId, datasetName);
```
seanzhougoogle: would it be better to put those logic to `BigQuerySinkUtils.createDataset` to make the operation idempotent ?

fernst: We always call this method with the assumption that this check has been made

seanzhougoogle: does that mean we need to check everywhere before calling this method ?

seanzhougoogle: would it be better to do check in one single place ?

seanzhougoogle: also checking in single place will avoid future bugs like this one.

itsankit-google: +1 to Sean, a second thought is it the correct behaviour that the util class method depends on the caller?

itsankit-google: Discussed offline with @fernst to make that method back to private and add another util method `createDatasetIfNotExists`.

seanzhougoogle (approved):
Thank you Ankit & Fernando

PR #918, opened 2022-03-04
Code Added for Dataplex source plugin

vikasrathee-cs:
Added code for dataplex source plugin.


src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+    return connection == null ? null : connection.isServiceAccountFilePath();
+  }
+
+  /**
+   * Check if service account is a valid one.
+   *
+   * @param failureCollector
+   */
+  public void validateServiceAccount(FailureCollector failureCollector) {
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
```
snehalakshmisha: If the connection is a service account and it does not have a service account file path, is it still valid?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+      case FLOAT:
+      case DOUBLE:
+        // float is a float64, so corresponding type becomes double
+        return Schema.of(Schema.Type.DOUBLE);
+      case BOOLEAN:
+        return Schema.of(Schema.Type.BOOLEAN);
+      case BYTE:
+      case INT16:
+      case INT32:
+      case INT64:
```
fernst, line 211: Does it make sense to treat `int32` and smaller as 'integer' in CDAP?

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+    return credentials;
+  }
+
+  protected void configureDataplexException(String dataplexConfigProperty, String dataplexConfigPropType,
+                                            ApiException e,
+                                            FailureCollector failureCollector) {
+    if (StatusCode.Code.NOT_FOUND.equals(e.getStatusCode().getCode()) ||
+      StatusCode.Code.INVALID_ARGUMENT.equals(e.getStatusCode().getCode())) {
+      failureCollector
+        .addFailure("'" + dataplexConfigProperty + "' could not be found. Please ensure that it exists in " +
```
snehalakshmisha: Use String.format() inside addFailure()

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
```
snehalakshmisha: nit: 2022
here and in other newly added files

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+  @Name(NAME_ZONE)
+  @Macro
+  @Description("Resource id for the Dataplex zone. You can find it on the zone detail page in Dataplex.")
+  protected String zone;
+
+  @Name(NAME_CONNECTION)
+  @Nullable
+  @Macro
+  @Description("The existing connection to use.")
+  protected GCPConnectorConfig connection;
```
rmstar, line 74: where is this field used?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+  private static final String SERVICE_ACCOUNT_KEYFILE = "google.cloud.auth.service.account.json.keyfile";
+
+  /**
+   * Converts Entity Schema into a CDAP Schema object.
+   *
+   * @param dataplexSchema Dataplex Schema to be converted.
+   * @param collector      Failure collector to collect failure messages for the client.
+   * @return CDAP schema object
+   */
+  public static Schema getTableSchema(com.google.cloud.dataplex.v1.Schema dataplexSchema,
```
snehalakshmisha: It is a bit confusing that the method has 'table' in it but there is no mention of a table inside the method

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+        if (collector != null) {
+          collector.addFailure(error, action);
+        } else {
+          throw new RuntimeException(error + action);
+        }
+        return null;
+    }
+  }
+
+  /**
```
snehalakshmisha, line 279: missing the description here?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+    String lake = conf.get(DataplexConstants.DATAPLEX_LAKE);
+    String taskId = conf.get(DataplexConstants.DATAPLEX_TASK_ID);
+    try (DataplexServiceClient dataplexServiceClient = DataplexUtil.getDataplexServiceClient(googleCredentials)) {
+      DataplexServiceClient.ListJobsPagedResponse
+        jobList = dataplexServiceClient.listJobs(TaskName.newBuilder().setProject(projectID).setLake(lake).
+        setLocation(location).setTask(taskId).build());
+      Job dataplexJob = jobList.iterateAll().iterator().next();
+      try {
+        Awaitility.await()
+          .atMost(30, TimeUnit.MINUTES)
```
snehalakshmisha, line 299: 30 minutes seems like quite a long time. How did we arrive at this number?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+    }
+  }
+
+  /**
+   * Return DataplexServiceClient to call dataplex methods.
+   *
+   * @param credentials GoogleCredentials
+   * @return DataplexServiceClient
+   */
+  public static DataplexServiceClient getDataplexServiceClient(GoogleCredentials credentials) {
```
snehalakshmisha: If this can return null it should be marked as `@Nullable`. Same with any other methods

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+    FileSystem pathFileSystem = FileSystem.get(path.toUri(), configuration);
+    FileStatus[] fileStatus = pathFileSystem.globStatus(path);
+    if (fileStatus == null) {
+      throw new IOException(String.format("Input path %s does not exist", path));
+    } else {
+      FileInputFormat.addInputPath(job, path);
+      Map<String, String> inputFormatConfiguration = validatingInputFormat.getInputFormatConfiguration();
+      Iterator inputFormatIterator = inputFormatConfiguration.entrySet().iterator();
+      while (inputFormatIterator.hasNext()) {
+        Map.Entry<String, String> propertyEntry = (Map.Entry) inputFormatIterator.next();
```
snehalakshmisha: You should be able to remove the casting here and declare the Iterator as `Iterator<> inoutFormatIterator = ...`

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+ * {@code StructuredRecord} is the third parameter because that is the output given by
+ * Hadoop's {@code TextInputFormat}. All the plugins included with Hydrator operate on
+ * StructuredRecord.
+ */
+@Plugin(type = BatchSource.PLUGIN_TYPE)
+@Name(DataplexBatchSource.NAME)
+@Description("Dataplex Source")
+public class DataplexBatchSource extends BatchSource<Object, Object, StructuredRecord> {
+  public static final String NAME = "Dataplex";
+  private static final String BQ_TEMP_BUCKET_NAME_PREFIX = "bq-source-bucket-";
```
fernst, line 99: this should by `dataplex-source-bucket-`

fernst: nit: this can be confuses with the BigQuery source. Can we make this `dataplex-bq-source-bucket`?

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+        Map.Entry<String, String> propertyEntry = (Map.Entry) inputFormatIterator.next();
+        configuration.set(propertyEntry.getKey(), propertyEntry.getValue());
+      }
+    }
+    configuration.set(DataplexConstants.DATAPLEX_ENTITY_TYPE, entity.getSystem().toString());
+    context.setInput(Input.of(config.getReferenceName(), new DataplexInputFormatProvider(configuration)));
+  }
+
+  /**
+   * set dataplex properties in configuration
```
fernst: nit: case

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexInputFormatProvider.java
```diff
@@ -0,0 +1,117 @@
...
+ * InputFormatProvider for DataplexSource
+ */
+public class DataplexInputFormatProvider implements InputFormatProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DataplexBatchSource.class);
+  private static Configuration configuration;
+  protected static Map<String, String> inputFormatConfiguration;
+
+  /**
+   * @param conf configuration
```
fernst, line 54: nit: add small description explaining how the configuration is used to initialize this provider

fernst (commented):
Code looks good, just asked some small questions.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+                              Credentials credentials,
+                              String bucketName) {
+    String bucket = String.format(BQ_TEMP_BUCKET_NAME_TEMPLATE, bucketName);
+    // By default, this option is false, meaning the job can not delete the bucket.
+    configuration.setBoolean("fs.gs.bucket.delete.enable", true);
+
+    // the dataset existence is validated before, so this cannot be null
+    Dataset bigQueryDataset = bigQuery.getDataset(DatasetId.of(datasetProject, dataset));
+    GCPUtils.createBucket(GCPUtils.getStorage(project, credentials),
+      bucket, bigQueryDataset.getLocation(), null);
```
itsankit-google: Do we need to add CMEK as a config/runtime args `Encryption Key Name` as we do in BigQuery Source instead of passing it as null?

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+        GCPUtils.createBucket(storage, bucket, location, null);
+
+      } catch (StorageException e) {
+        if (e.getCode() == 409) {
+          // A conflict means the bucket already exists
+          // This most likely means multiple stages in the same pipeline are trying to create the same bucket.
+          // Ignore this and move on, since all that matters is that the bucket exists.
+          return bucket;
+        }
+        throw new IOException(String.format("Unable to create Cloud Storage bucket '%s' in the same ",
```
itsankit-google: nit: unclear error message `in the same`

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
itsankit-google: also can we use `createBucket() in line#491` method here as well instead of directly using `GCPUtils#createBucket`? Since it has concurrency protection.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,514 @@
...
+   * @return Bucket name.
+   */
+  private String createBucket(Configuration configuration,
+                              Storage storage, String location,
+                              @Nullable String bucket) throws IOException {
+    // Create a new bucket if needed
+    if (storage != null && storage.get(bucket) == null) {
+      try {
+        configuration.setBoolean("fs.gs.bucket.delete.enable", true);
+        GCPUtils.createBucket(storage, bucket, location, null);
```
itsankit-google, line 508: similar comment about CMEK here as well.

src/test/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSourceTest.java
```diff
@@ -0,0 +1,119 @@
...
+    dataplexBatchSourceConfig.validateBigQueryDataset(mockFailureCollector, "projectId", "dataset", "table");
+    assertEquals(0, mockFailureCollector.getValidationFailures().size());
+  }
+
+  @Test
+  public void validateServiceAccount() throws Exception {
+    MockFailureCollector mockFailureCollector = new MockFailureCollector();
+    DataplexBatchSourceConfig config = spy(new DataplexBatchSourceConfig());
+    GCPConnectorConfig gcpConnectorConfig = spy(new GCPConnectorConfig("", "", "", ""));
+    FieldSetter.setField(config, DataplexBaseConfig.class.getDeclaredField("connection"),
```
itsankit-google: instead of `FieldSetter` can we use builder with private constructor in all the tests? This makes the tests brittle if we modified the field name in future.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+    MetadataServiceClient metadataServiceClient = null;
+    try {
+      metadataServiceSettings =
+        MetadataServiceSettings.newBuilder()
+          .setCredentialsProvider(FixedCredentialsProvider.create(credentials))
+          .build();
+      metadataServiceClient =
+        MetadataServiceClient.create(metadataServiceSettings);
+    } catch (RuntimeException | IOException e) {
+      LOG.error("Failed to get MetadataServiceClient. Check the credentials");
```
rmstar, line 389: should we throw an exception if we're unable to create a MetadataServiceClient? Otherwise we'll return null and caller will get a NullPointerException

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
yaojiefeng: Do we have a connector for this source/sink? This field is normally used when a connector is available and connection can be created from it. Using it like this will still force user use a flat structure instead of being able to specify the connection macro

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+
+  public String getReferenceName() {
+    return referenceName;
+  }
+
+  public String getLake() {
+    return lake;
+  }
+
+  public void setLake(String lake) {
```
yaojiefeng: Do we need the set method for these fields, normally these fields should be immutable

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+  }
+
+  private GoogleCredentials getCredentialsFromServiceAccount() throws IOException {
+    GoogleCredentials credentials = null;
+    // validate service account
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath())
+          .createScoped(Lists.newArrayList(DataplexConstants.DATAPLEX_SCOPE));
```
yaojiefeng: So do we have to specify this scope for the credentials, if this is required, then does the compute engine service account need to have this scope enabled in order for auto-detect to work? Have we verified if auto-detect works on ephemeral dataproc cluster? @itsankit-google you might have some better understanding on this. please check if this is correct

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+
+  protected void configureDataplexException(String dataplexConfigProperty, String dataplexConfigPropType,
+                                            ApiException e,
+                                            FailureCollector failureCollector) {
+    if (StatusCode.Code.NOT_FOUND.equals(e.getStatusCode().getCode()) ||
+      StatusCode.Code.INVALID_ARGUMENT.equals(e.getStatusCode().getCode())) {
+      failureCollector
+        .addFailure("'" + dataplexConfigProperty + "' could not be found. Please ensure that it exists in " +
+          "Dataplex.", null).withConfigProperty(dataplexConfigPropType);
+    } else if (StatusCode.Code.PERMISSION_DENIED.equals(e.getStatusCode().getCode()) ||
```
yaojiefeng: No need to have else here and below

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+   *
+   * @return service account email ID
+   * @throws IOException
+   */
+  public String getServiceAccountEmail() throws IOException {
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      ServiceAccountCredentials credentials =
+        GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath());
+      return credentials.getClientEmail();
+    } else {
```
yaojiefeng: No need to have else

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+      ComputeEngineCredentials credentials = (ComputeEngineCredentials) ServiceAccountCredentials.
```
yaojiefeng: Do we know if this always return a non-null service account, what if it is on sandbox?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.dataplex.common.util;
+
+/**
+ * Dataplex Constants
+ */
+public interface DataplexConstants {
+  String DATAPLEX_TASK_ID = "dataplex.task.id";
```
yaojiefeng: public final for constants

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
```diff
@@ -0,0 +1,34 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.dataplex.common.util;
+
+/**
+ * Dataplex Constants
+ */
+public interface DataplexConstants {
```
yaojiefeng: this should not be interface

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -0,0 +1,414 @@
...
+  /**
+   * Converts Entity Partitioned field type into a CDAP field type.
+   *
+   * @param field        Partitioned field to be converted.
+   * @param collector    Failure collector to collect failure messages for the client.
+   * @param recordPrefix String to add before a record name to ensure unique names.
+   * @return A CDAP field schema
+   */
+  @Nullable
+  private static Schema convertFieldTypeForPartitionedFields(com.google.cloud.dataplex.v1.Schema.PartitionField field,
```
yaojiefeng: there is similar method `convertFieldTypeForPartitionedFields`, where difference is it is partitioned field or not, can this be refactored to some common methods?

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
```diff
@@ -0,0 +1,412 @@
...
+
+  }
+
+  /**
+   * @param batchSourceContext
+   * @return
+   * @throws InstantiationException
+   */
+  public ValidatingInputFormat getValidatingInputFormat(BatchSourceContext batchSourceContext
+  ) throws InstantiationException {
```
yaojiefeng: move `)` to previous line

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
```diff
@@ -0,0 +1,412 @@
...
+  /**
+   * Setting validating input format as we are setting same format in task output format.
+   *
+   * @param pipelineConfigurer
+   * @param collector
+   * @param entity             Dataplex entity
+   */
+  public void setupValidatingInputFormat(PipelineConfigurer pipelineConfigurer, FailureCollector collector,
+                                         @Nullable Entity entity) {
+    ConfigUtil.validateConnection(this, false, connection, collector);
```
yaojiefeng: You don't need to do this validation if you don't have useConnection flag

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
```diff
@@ -0,0 +1,412 @@
...
+    String fileFormat = INPUT_FORMAT;
+    collector.getOrThrowException();
+    if (entity != null) {
+      if (this.getSchema(collector) == null) {
+        Schema schema = DataplexUtil.getTableSchema(entity.getSchema(), collector);
```
yaojiefeng, line 289: So this will return the avro schema if the dest is gcs?

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
```diff
@@ -0,0 +1,412 @@
...
+        pipelineConfigurer.getStageConfigurer().setOutputSchema(schema);
+      }
+    }
+    PluginProperties.Builder builder = PluginProperties.builder();
+    builder.addAll(this.getRawProperties().getProperties());
+    pipelineConfigurer.usePlugin("validatingInputFormat", fileFormat, fileFormat,
```
yaojiefeng, line 295: So we will always set the avro input format no matter the entity is bigquery or GCS right?

yaojiefeng (commented):
Took an initial pass

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
itsankit-google: @yaojiefeng Since, the scope required by dataplex api is `cloud-platform` so `auto-detect` should work on ephemeral dataproc cluster. 
Since, we set the `cloud-platform` scope, it is included in Compute Engine instance. Ref - https://github.com/cdapio/cdap/blob/c15d12f1876ef81b542ad5a4220dcae59ae3eb8a/cdap-runtime-ext-dataproc/src/main/java/io/cdap/cdap/runtime/spi/provisioner/dataproc/DataprocClient.java#L422

Here in the plugin, when we set the scope we are just requesting the access token with those scopes while creating client but if the Compute Engine VM does not have those scopes set the request using the access token will fail.

yaojiefeng: I see. Then there is no problem, then I don't think you need to add this, this will just be same as other gcp plugins

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,244 @@
...
+
+  private GoogleCredentials getCredentialsFromServiceAccount() throws IOException {
+    GoogleCredentials credentials = null;
+    // validate service account
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath())
+          .createScoped(Lists.newArrayList(DataplexConstants.DATAPLEX_SCOPE));
+    } else {
+      credentials = ServiceAccountCredentials.getApplicationDefault().createScoped(
```
itsankit-google: Is it necessary to use `ServiceAccountCredentials.getApplicationDefault()` while creating the dataplex service client? In other cloud APIs we don't set anything if credentials are null, they are expected to pick them by default.
Ref - https://github.com/data-integrations/google-cloud/blob/8f25b4807340f7a65d0190dc168d5c9b32e51639/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L186

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: yes, based on service account type, it can fetch values from properties for filepath or json.

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
vikasrathee-cs: this method will always get called when it is a GCS entity.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: dataplexSchema is also a table schema in dataplex

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: We have tested for 5 million records and it is taking around 10 min in creating dataplex task and then dataplex job and finally giving the output. For dataplex we can have millions of data, for that reason I set it to 30 min.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: CMEK will be added later, will use create bucket.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: We will add the connector in subsequent releases.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: removed.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: We are setting up the error message based on status code. If we will remove this it will show similar message for different error codes also.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: else part is for auto-detect in project and filepath field.

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
vikasrathee-cs: it will convert dataplexSchema to cdap schema. For GCS entites input will always be avro.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
vikasrathee-cs: these are public final by default.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
vikasrathee-cs: Same type of interface is also used in BigQuery.
https://github.com/data-integrations/google-cloud/blob/aedfd4e23bfb207668b3d8b1d72b2fef902fb0c5/src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryConstants.java

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: renamed dataplexSchema to dataplexTableSchema

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: Added

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: instead will throw an exception as suggested by Prashant.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: removed

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: will be added later

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: throwing exception seems to be a better option, will do that.

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfig.java
vikasrathee-cs: removed

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: It was failing while we were using REST api calls instead of client libraries but now it is working fine without this also. So removing it.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: Dataplex APIs doesn't provide any such option, so we are manually selecting this in case of auto-detect.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
vikasrathee-cs: done. Dataplex provide two type fields in schema, normal fields and partitioned fields. Renamed the method and moved the common code into single method.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: Yes, it will always be a non-null value, in case of null it will throw exception.

src/test/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSourceTest.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexInputFormatProvider.java
vikasrathee-cs: added.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
yaojiefeng: You already return in the previous if condition, so you don't need else here

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
yaojiefeng: This is not good practice, do you have use case to override this interface? If not, just mark this class as `public final`

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
yaojiefeng: I don't see this getting removed, is this still required now?

yaojiefeng (commented):
Only some small comments left unaddressed, please make sure this is tested on sandbox and dataproc, with auto-detect and provided service account, thanks

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexConstants.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: in case of service account json, it will go to if loop and return that email and in case of auto-detect it will go to else block and will fetch the email from there.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: I have removed it in latest commit.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,242 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * @return GoogleCredentials based on service account
+   */
+  public GoogleCredentials getCredentials() {
```
itsankit-google: Can we reuse `connection.getCredentials()`? This will reduce code duplication.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,242 @@
...
+      }
+    }
+  }
+
+  private GoogleCredentials getCredentialsFromServiceAccount() throws IOException {
+    GoogleCredentials credentials = null;
+    // validate service account
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      credentials =
+        GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath());
```
itsankit-google: This method expects non-null input. `getServiceAccount()` can still be null since `getServiceAccount()` can return null

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: No, it doesn't handle the auto-detect scenerio while getting credentials

itsankit-google: So for `auto-detect` we can just add an additional check if this method returned `null` and `service account file path` was set to detect instead of duplicating the method. Does that makes sense?

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: getServiceAccount will never be null as we have already checked that in if condition

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
itsankit-google: `getServiceAccount()` internally calls `connection.getServiceAccount()` which returns null if  `getServiceAccountJson()` is null or empty.

vikasrathee-cs: handled that

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,235 @@
...
+   * @param failureCollector  failureCollector to collect the exceptions and failures.
+   * @return GoogleCredentials based on service account
+   */
+  public GoogleCredentials getCredentials(FailureCollector failureCollector) {
+    GoogleCredentials credentials = null;
+    // validate service account
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      credentials =
+        (GoogleCredentials) connection.getCredentials(failureCollector);
+    } else {
```
itsankit-google: We don't need `else` here, we can directly return the output from `connection.getCredentials(failureCollector)`

vikasrathee-cs: else is for auto-detect scenerio

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,235 @@
...
+  public void validateServiceAccount(FailureCollector failureCollector) {
+    if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
+      GoogleCredentials credentials = getCredentials(failureCollector);
+      if (credentials == null) {
+        failureCollector.addFailure(String.format("Unable to load credentials from %s.",
+              connection.isServiceAccountFilePath() ? connection.getServiceAccountFilePath() : "provided JSON key"),
+            "Ensure the service account file is available on the local filesystem.")
+          .withConfigProperty("serviceFilePath")
+          .withConfigProperty("serviceAccountJSON");
+        throw failureCollector.getOrThrowException();
```
itsankit-google: Should not throw the exceptions at this time. Usually the pattern followed is - calling all the config validations from `configurePipeline` and `prepareRun` and then call `failureCollector.getOrThrowException()`

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -0,0 +1,522 @@
...
+    StageConfigurer configurer = pipelineConfigurer.getStageConfigurer();
+    FailureCollector collector = configurer.getFailureCollector();
+    if (!config.getConnection().canConnect() || config.getServiceAccountType() == null ||
+      (config.isServiceAccountFilePath() && config.autoServiceAccountUnavailable()) ||
+      (config.tryGetProject() == null)) {
+      // ValidatingInputFormat plugin setup is mandatory.Otherwise pipeline will fail at runtime in case of GCS entity.
+      config.setupValidatingInputFormat(pipelineConfigurer, collector, null);
+      return;
+    }
+    config.validateServiceAccount(collector);
```
itsankit-google: Instead of calling every config validation method individually, can we create `validate(failureCollector)` method in config class which internally calls all the config validation methods?
Same `validate(failureCollector)` method can be called in `prepareRun` as well

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
itsankit-google: What I meant is to remove nesting to increase code readability by doing something like this
```
if(condition) {
   // do something
   return credentials;
}
//else logic
```

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+    return false;
+  }
+
+  /**
+   * @param failureCollector  failureCollector to collect the exceptions and failures.
+   * @return GoogleCredentials based on service account
+   */
+  public GoogleCredentials getCredentials(FailureCollector failureCollector) {
```
itsankit-google, line 116: Add `@Nullable` annotation to the method

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: All the subsequent validations are based on this credentials. I changed it to validateAndGetServiceAccountCredentials to refrain from calling getCredentials method multiple times

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
vikasrathee-cs: merged the validation methods together for BQ entities but same can not be done for GCS entities as in configurePipeline we setValidatingInputFormat and in prepareRun we getValidatingInputFormat as set at design time.

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
itsankit-google: You can create a base validate method which calls all the subsequent validation methods. In which you can call this method first and if failureCollector is not empty can skip all the further validations.

vikasrathee-cs: done

rmstar (approved):

rmstar (approved):
please squash commits after addressing review comments

PR #919, opened 2022-03-04
Code Added for Dataplex sink plugin.

vikasrathee-cs:
Code Added for Dataplex sink plugin.


src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -0,0 +1,582 @@
...
+   * prepare Run for BigQuery Dataset Asset. It will create necessary resources, and configuration methods
+   *
+   * @param context
+   * @throws Exception
+   */
+  private void prepareRunBigQueryDataset(BatchSinkContext context) throws Exception {
+    FailureCollector collector = context.getFailureCollector();
+    Credentials credentials = config.getCredentials();
+    String project = config.getProject();
+    String cmekKey = context.getArguments().get(GCPUtils.CMEK_KEY);
```
itsankit-google: Does it make sense to also add CMEK as a config `Encryption Key Name`  like we have in other GCP plugins?

src/test/java/io/cdap/plugin/gcp/dataplex/sink/config/DataplexBatchSinkConfigTest.java
```diff
@@ -0,0 +1,651 @@
...
+public class DataplexBatchSinkConfigTest {
+
+  @Mock
+  StatusCode statusCode;
+
+  @Test
+  public void validateBigQueryDatasetTest() throws NoSuchFieldException {
+    DataplexBatchSinkConfig dataplexBatchSinkConfig = new DataplexBatchSinkConfig();
+    MockFailureCollector mockFailureCollector = new MockFailureCollector();
+    FieldSetter.setField(dataplexBatchSinkConfig, DataplexBatchSinkConfig.class.getDeclaredField("table"), "test");
```
itsankit-google: Similar comment, instead of `FieldSetter` can we use builder with Private Constructor in all the tests. This makes the tests brittle if we modified the field name in future.

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexOutputFormatProvider.java
```diff
@@ -0,0 +1,173 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * OutputFormatProvider for Dataplex
+ */
+public class DataplexOutputFormatProvider implements ValidatingOutputFormat {
+  public static final String RECORD_COUNT_FORMAT = "recordcount.%s";
+  public static final String DATAPLEX_ASSET_TYPE = "dataplexsink.assettype";
+  public static final String DATAPLEX_OUTPUT_BASE_DIR = "dataplex.output.fileoutputformat.baseoutputdir";
+  private static final String DELEGATE_OUTPUTFORMAT_CLASSNAME = "gcssink.delegate.outputformat.classname";
```
rmstar: This is also defined here - https://github.com/data-integrations/google-cloud/blob/develop/src%2Fmain%2Fjava%2Fio%2Fcdap%2Fplugin%2Fgcp%2Fgcs%2Fsink%2FGCSOutputFormatProvider.java#L23
Can we dedupe?

vikasrathee-cs: yes

src/test/java/io/cdap/plugin/gcp/dataplex/sink/config/DataplexBatchSinkConfigTest.java
vikasrathee-cs: done

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
vikasrathee-cs: As discussed, it was not part of prd and design, will incorporate this later.

rmstar (approved):

rmstar (approved):
please squash commits

PR #921, opened 2022-03-07
Added logic to check if the BQ Dataset exists before trying to re-cre…

itsankit-google:
…ate in the BQ Execute plugin

cherry-pick, #916


fernst (approved):
LGTM

PR #926, opened 2022-03-10
Made GCS Source inherit from abstractfilesourceconfig

sechegaray:


src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -250,15 +181,8 @@ protected boolean shouldGetSchema() {
     @Description("The existing connection to use.")
     private GCPConnectorConfig connection;
 
-    public GCSSourceConfig() {
```
fernst, line 253: Shouldn't we have a constructor that calls the `super()` constructor?

fernst (approved):
LGTM

PR #930, opened 2022-03-16
[PLUGIN-1133] Added support for external tables in BigQuery source

fernst:
JIRA: [PLUGIN-1133](https://cdap.atlassian.net/browse/PLUGIN-1133)

[PLUGIN-1133]: https://cdap.atlassian.net/browse/PLUGIN-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryFactoryWithScopes.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryFactoryWithScopes.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.bigquery.source;
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.cloud.hadoop.io.bigquery.BigQueryFactory;
+import com.google.cloud.hadoop.util.AccessTokenProviderClassFromConfigFactory;
+import com.google.cloud.hadoop.util.CredentialFromAccessTokenProviderClassFactory;
+import com.google.cloud.hadoop.util.HadoopCredentialConfiguration;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -100,7 +115,7 @@ private void processQuery(JobContext context) throws IOException, InterruptedExc
     Type type = Objects.requireNonNull(bigQueryTable).getDefinition().getType();
 
     String query;
-    if (type == Type.VIEW || type == Type.MATERIALIZED_VIEW) {
+    if (type == Type.VIEW || type == Type.MATERIALIZED_VIEW || type == Type.EXTERNAL) {
```
yaojiefeng, line 119: Is it easier if we just use BigQueryClient here to perform the query?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryFactoryWithScopes.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+/**
+ * Override for {@link BigQueryFactory} with additional scopes for External tables.
+ */
+public class BigQueryFactoryWithScopes extends BigQueryFactory {
+
+  @Override
```
yaojiefeng, line 42: I feel it is better for this to take the scopes in constructor in case we want to manage what scopes we add in the future.

yaojiefeng (approved):
Just one question and one minor comment, LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
fernst: The advantage of using the Hadoop client is that the table only gets materialized when Spark actually needs to read from it. For a large pipeline, several stages might run before this BQ source starts processing. This means that, if the pipeline fails in a previous stage, we now have to clean up a table that was never used.

fernst: It also allows us to reuse the pattern we have been using so far

yaojiefeng: Sounds good, let's keep this then

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQueryFactoryWithScopes.java
fernst: Added

yaojiefeng (approved):

PR #938, opened 2022-03-22
Improvements to the BQ Sink

fernst:
Ensure Nullable fields remain nullable when relaxing schemas.

Added error when the destination table doesn't exist for Update/Upsert operations.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -335,18 +335,7 @@ public final class BigQuerySinkUtils {
                                       Table destinationTable,
                                       List<Field> sourceFields,
                                       List<Field> destinationFields) {
-    // Collect all fields form the source table
-    Map<String, Field> sourceFieldMap = sourceFields.stream()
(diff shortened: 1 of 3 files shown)


itsankit-google (approved):
LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -335,26 +335,55 @@ public static void relaxTableSchema(BigQuery bigquery,
...
-    // Update table definition, relaxing field definitions.
-    com.google.cloud.bigquery.Schema newSchema = com.google.cloud.bigquery.Schema.of(resultFieldsList);
-    bigquery.update(
-      destinationTable.toBuilder().setDefinition(
-        destinationTable.getDefinition().toBuilder().setSchema(newSchema).build()
-      ).build()
-    );
+        return sourceField;
+      })
+      .forEach(resultingFields::add);
```
tivv, line 384: This would change the field order. Would it affect order of fields in the resulting table or visible CDAP schema?

fernst: This is only used to update the BigQuery schema and field order is not really something we guarantee in BigQuery (specially with existing tables) so no.

PR #941, opened 2022-03-23
[PLUGIN-1154][0.19] Fixed issue with NPE during Update/Upsert operations

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -589,9 +589,12 @@ private void handleInsertOperation(TableReference tableRef, String writeDisposit
       triggerBigqueryJob(projectId, jobId, dataset, config);
     }
 
-    private void handleUpdateUpsertOperation(TableReference tableRef, @Nullable String cmekKey, JobId jobId,
+    private void handleUpdateUpsertOperation(TableReference tableRef,
```
avirmani, line 592: Need to add the param in the calling place as well.

fernst: oof

fernst: Added now.

avirmani (approved):
lgtm

PR #942, opened 2022-03-23
[PLUGIN-1154][0.18] Fixed issue with NPE during Update/Upsert operations

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryOutputFormat.java
```diff
@@ -589,9 +589,12 @@ private void handleInsertOperation(TableReference tableRef, String writeDisposit
       triggerBigqueryJob(projectId, jobId, dataset, config);
     }
 
-    private void handleUpdateUpsertOperation(TableReference tableRef, @Nullable String cmekKey, JobId jobId,
+    private void handleUpdateUpsertOperation(TableReference tableRef,
```
avirmani, line 592: Same, need to add the new param in the calling place as well.

avirmani (approved):
lgtm

PR #967, opened 2022-04-11
Added BigQuery capability to BigQuery SQL Engine

fernst:
Depends on https://github.com/cdapio/cdap/pull/14178

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpressionFactory.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpressionFactory.java
@@ -1,5 +1,6 @@
 package io.cdap.plugin.gcp.bigquery.relational;
 
+import io.cdap.cdap.etl.api.engine.sql.StandardSQLCapabilities;
 import io.cdap.cdap.etl.api.relational.Capability;
 import io.cdap.cdap.etl.api.relational.CoreExpressionCapabilities;
 import io.cdap.cdap.etl.api.relational.Expression;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpressionFactory.java
```diff
@@ -29,7 +29,7 @@ public class SQLExpressionFactory implements ExpressionFactory<String> {
    */
   @Override
   public ExpressionFactoryType<String> getType() {
-    return StringExpressionFactoryType.SQL;
```
tivv: Please don't change expression factory type. Use capabilities

src/main/java/io/cdap/plugin/gcp/bigquery/relational/SQLExpressionFactory.java
```diff
@@ -51,6 +52,7 @@ public Expression compile(String expression) {
   @Override
   public Set<Capability> getCapabilities() {
     return new HashSet<>(Arrays.asList(StringExpressionFactoryType.SQL,
+                                       StandardSQLCapabilities.BIGQUERY,
```
tivv: nit: given this grows up, it makes sense to make it a constant

tivv: Please update javadoc, it does not reflect the implementation anymore

fernst: Updated

tivv (approved):

PR #968, opened 2022-04-12
Exclusions removed from pom for PLUGIN-1198 fix.

vikasrathee-cs:
Exclusions removed from pom for PLUGIN-1198 fix.


itsankit-google (approved):
LGTM. Please verify the spanner plugin works correctly with auto-detect after this PR is merged and raise a cherry-pick for 6.7.

PR #988, opened 2022-04-22
e2e - Create Spanner instance with smallest processing unit

sawantpritam:
Creating Spanner instance with smallest processing unit which is more than sufficient for running our tests.

@itsankit-google Please review.


itsankit-google (approved):
LGTM

PR #998, opened 2022-04-26
Enable pushdown for preview features toggle

sanjanasandeep:
** Issue**

https://cdap.atlassian.net/browse/CDAP-19072

image

### DIFF
--- a/widgets/BigQueryPushdownEngine-sqlengine.json
+++ b/widgets/BigQueryPushdownEngine-sqlengine.json
@@ -223,6 +223,22 @@
             },
             "default": "false"
           }
+        },
+        {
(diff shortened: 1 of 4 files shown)


widgets/BigQueryPushdownEngine-sqlengine.json
```diff
@@ -223,6 +223,22 @@
...
+            "on": {
+              "value": "true",
+              "label": "YES"
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "true"
+          }
```
sanjanasandeep, line 241: should we allow default true?

fernst (approved):

PR #999, opened 2022-04-26
Fixed too general DataplexBatchSourceConfigTest to check for specific…

rmstar:
Cherry-pick https://github.com/data-integrations/google-cloud/pull/947/commits/048830fce18797fef927b9e49b08237136ce59d4

### DIFF
--- a/src/test/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfigTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfigTest.java
@@ -23,9 +23,9 @@ import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
 import org.mockito.junit.MockitoJUnitRunner;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Tests for DataplexBatchSourceConfig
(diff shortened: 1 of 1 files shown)


src/test/java/io/cdap/plugin/gcp/dataplex/source/config/DataplexBatchSourceConfigTest.java
```diff
@@ -86,7 +86,14 @@ public void testValidateTable() {
       dataplexBatchSourceConfig.validateBigQueryDataset(mockFailureCollector, "project", "dataset", "table-wrong");
     } catch (Exception e) {
     }
-    assertEquals(2, mockFailureCollector.getValidationFailures().size());
+
+    assertTrue("Expected at least one error with incorrect table name",
+               mockFailureCollector.getValidationFailures().size() > 0);
+
+    mockFailureCollector.getValidationFailures().stream()
+      .filter(error -> error.getFullMessage().toLowerCase().contains("table name"))
```
avirmani, line 94: should we just check if it contains 'table-wrong' in the error message?

rmstar: yeah we could do that, but it would need to be a separate PR on develop first. This is a cherry-pick, so I don't want to make changes here that are not in develop

avirmani: got it.

avirmani (approved):

PR #1000, opened 2022-04-27
remove override

yaojiefeng:


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -156,12 +154,6 @@ protected Map<String, String> getFileSystemProperties(BatchSinkContext context)
     return properties;
   }
 
-  @Override
-  protected String getOutputDir(long logicalStartTime) {
-    this.outputPath = super.getOutputDir(logicalStartTime);
```
albertshau, line 161: this variable should be set in the prepareRun() method by calling getOutputDir(context). We don't want the correctness of this class to be tied to the implementation of the parent class and whenever it decides to set this variable.

albertshau (approved):

PR #937, opened 2022-03-21
CDAP-18889 and CDAP-18892

sawantpritam:
@itsankit-google @rmstar Please review.

PR contains changes for CDAP-18889 and CDAP-18892 :
1. Add tests with Service Account Type as macro param (JSON)
2. Capture logs of preview run

PR #1002, opened 2022-04-28
e2e tests CDAP-19055 - Override service account details

sawantpritam:
This PR depends on [https://github.com/cdapio/cdap-e2e-tests/pull/83](https://github.com/cdapio/cdap-e2e-tests/pull/83)

@rmstar @itsankit-google Please review.

### DIFF
--- a/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
+++ b/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
@@ -50,6 +50,7 @@ import java.util.stream.Collectors;
  */
 public class TestSetupHooks {
 
+  public static boolean beforeAllFlag = true;
   public static String gcsSourceBucketName = StringUtils.EMPTY;
   public static String gcsTargetBucketName = StringUtils.EMPTY;
   public static String bqTargetTable = StringUtils.EMPTY;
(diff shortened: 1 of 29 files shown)


sawantpritam:
Conflicts resolved.

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -63,6 +64,22 @@ public class TestSetupHooks {
...
 
+  @Before(order = 1)
+  public static void overrideServiceAccountFilePathIfProvided() {
+    if (beforeAllFlag) {
+      String serviceAccountType = System.getenv("SERVICE_ACCOUNT_TYPE");
+      if (serviceAccountType != null) {
+        if (serviceAccountType.equalsIgnoreCase("FilePath")) {
+          String serviceAccountFilePath = System.getenv("SERVICE_ACCOUNT_FILE_PATH");
+          if (!(serviceAccountFilePath == null) && !serviceAccountFilePath.equalsIgnoreCase("auto-detect")) {
+            PluginPropertyUtils.addPluginProp("serviceAccountFilePath", serviceAccountFilePath);
```
itsankit-google: Why is this step needed if we are reading directly from env variables?
https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/java/io/cdap/e2e/pages/actions/CdfPluginPropertiesActions.java#L591

itsankit-google (commented):
The clients used to setup input data should also be created using the given credentials. Are we taking care of this as well?

itsankit-google:
Also there is one GCS sink test failure, please ensure `Run tests` step is green in `Build e2e tests` workflow.

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
sawantpritam: This is required for macro runtime arguments. 
https://github.com/data-integrations/google-cloud/blob/2774b692b9d105595b87b2942b8e59e746d0f780/src/e2e-test/features/bigquery/sink/GCSToBigQuery_WithMacro.feature#L35

sawantpritam:
> The clients used to setup input data should also be created using the given credentials. Are we taking care of this as well?

Clients use projectId - which needs to be updated in pluginParameters.properties file as per config.

itsankit-google:
> > The clients used to setup input data should also be created using the given credentials. Are we taking care of this as well?
> 
> Clients use projectId - which needs to be updated in pluginParameters.properties file as per config.

If we don't set credentials while creating GCP API clients, they internally use the credentials from GCE VM which will be similar to auto-detect. If we want to create the clients in different projects you will need to set the service credentials which have permissions on the given project. See this for example, https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L186

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
itsankit-google: I think the macro scenario needs to be modified, what if service account type is set as "JSON" instead of "filePath". It will fail, right?

sawantpritam: We have separate macro tests with Service Account Type as macro param

https://github.com/data-integrations/google-cloud/pull/937

itsankit-google: What about the existing macro tests? They will fail if serviceAccountType is set as `JSON` in environment variable, right?

sawantpritam: Updated macro tests.

sawantpritam:
> > > The clients used to setup input data should also be created using the given credentials. Are we taking care of this as well?
> > 
> > 
> > Clients use projectId - which needs to be updated in pluginParameters.properties file as per config.
> 
> If we don't set credentials while creating GCP API clients, they internally use the credentials from GCE VM which will be similar to auto-detect. If we want to create the clients in different projects you will need to set the service credentials which have permissions on the given project. See this for example, https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java#L186

We can achieve this by setting env variable `GOOGLE_APPLICATION_CREDENTIALS` with same actual filepath or json used for overriding.

sawantpritam:
> Also there is one GCS sink test failure, please ensure `Run tests` step is green in `Build e2e tests` workflow.

All tests passed in build

itsankit-google (approved):
LGTM, please squash commits.

sawantpritam:
> LGTM, please squash commits.

Squashed commits.

PR #1024, opened 2022-05-06
Rearranging transformation pushdown widget

sanjanasandeep:
**Issue** 

https://cdap.atlassian.net/browse/PLUGIN-1246

**Description**

We must have 2 configuration groups : "basic" and "advanced" and connection properties are rearranged under "advanced" and basic has "Big query dataset" which is required

Screen Shot 2022-05-05 at 6 37 34 PM
Screen Shot 2022-05-05 at 6 37 50 PM
Screen Shot 2022-05-05 at 6 37 59 PM

### DIFF
--- a/widgets/BigQueryPushdownEngine-sqlengine.json
+++ b/widgets/BigQueryPushdownEngine-sqlengine.json
@@ -5,7 +5,29 @@
   "display-name": "BigQuery Pushdown",
   "configuration-groups": [
     {
-      "label": "Connection",
+      "label": "Basic",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "BigQuery Dataset",
+          "name": "dataset",
+          "widget-attributes": {
+            "placeholder": "Dataset to use to store temporary tables when pipeline is run."
+          }
+        },
+        {
+          "label": "browse",
+          "widget-type": "connection-browser",
+          "widget-category": "plugin",
+          "widget-attributes": {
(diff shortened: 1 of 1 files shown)


fernst (approved):
LGTM

widgets/BigQueryPushdownEngine-sqlengine.json
```diff
@@ -5,7 +5,29 @@
   "display-name": "BigQuery Pushdown",
   "configuration-groups": [
     {
-      "label": "Connection",
```
itsankit-google, line 8: Why are we removing this? The pattern followed in all other GCP plugins is that all the properties coming from `ConnectorConfig` are under `Connection` configuration group.
For eg, https://github.com/data-integrations/google-cloud/blob/develop/widgets/GCSFile-batchsource.json#L8

sanjanasandeep: so we had a discussion with @bdmogal on pushdown usability - we don't want to ask the user too much info. As part of the discussion, we want to keep the pushdown widget simple - we have only two configuration groups - `basic` which  will have the `dataset` field that is required and `advanced` has everything else including connection properties. Since this widget is for pushdown broadly and is not specific as other GCP source/sink plugins, I think it is alright to not follow the pattern.

itsankit-google: Sg, thanks for the info.

PR #1042, opened 2022-05-17
[PLUGIN-954 & PLUGIN-1256] set cmek key in BQ job in BQExecute only when storing results in a table

itsankit-google:
[PLUGIN-954](https://cdap.atlassian.net/browse/PLUGIN-954) - Plugin Config Improvement suggested by @bdmogal.
[PLUGIN-1256](https://cdap.atlassian.net/browse/PLUGIN-1256) - Do not set cmek key by default in BQ job in BQ execute plugin because it causes job with ddl statements to fail.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -42,6 +42,7 @@ import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.action.Action;
 import io.cdap.cdap.etl.api.action.ActionContext;
+import io.cdap.plugin.common.ConfigUtil;
 import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.CmekUtils;
@@ -86,14 +87,13 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
     CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(config.cmekKey, context.getArguments().asMap(), collector);
     collector.getOrThrowException();
-    if (cmekKeyName != null) {
-      builder.setDestinationEncryptionConfiguration(
-        EncryptionConfiguration.newBuilder().setKmsKeyName(cmekKeyName.toString()).build());
-    }
 
     // Save the results of the query to a permanent table.
-    if (config.getDataset() != null && config.getTable() != null) {
-      builder.setDestinationTable(TableId.of(config.getDataset(), config.getTable()));
+    String datasetName = config.getDataset();
+    String tableName = config.getTable();
+    String datasetProjectId = config.getDatasetProject();
+    if (config.getStoreResults() && datasetProjectId != null && datasetName != null && tableName != null) {
+      builder.setDestinationTable(TableId.of(datasetProjectId, datasetName, tableName));
     }
 
     // Enable or Disable the query cache to force live query evaluation.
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -279,6 +287,10 @@ public String getSql() {
       return sql;
     }
 
+    public Boolean getStoreResults() {
+      return storeResults != null && storeResults;
```
seanzhougoogle: to be backward compatible , we should take null as true.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -229,9 +231,9 @@ public static final class Config extends AbstractBigQueryActionConfig {
     @Name(NAME_CMEK_KEY)
     @Macro
     @Nullable
-    @Description("The GCP customer managed encryption key (CMEK) name used to encrypt data written to " +
-      "any dataset or table created by the plugin. If the dataset or table already exists, this is ignored. More " +
-      "information can be found at https://cloud.google.com/data-fusion/docs/how-to/customer-managed-encryption-keys")
+    @Description("The GCP customer managed encryption key (CMEK) name used to encrypt data written to the " +
```
seanzhougoogle, line 235: "The GCP customer managed encryption key (CMEK) name used to encrypt the query results. It's only applicable when users choose to store the query results in a BigQuery table.  More information can be found at " +
      "https://cloud.google.com/data-fusion/docs/how-to/customer-managed-encryption-keys"

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -115,14 +115,15 @@ public void run(ActionContext context) throws Exception {
...
     //create dataset to store the results if not exists
-    String datasetName = config.getDataset();
-    String tableName = config.getTable();
-    String datasetProjectId = config.getDatasetProject();
     if (!Strings.isNullOrEmpty(datasetName) && !Strings.isNullOrEmpty(tableName)) {
       BigQuerySinkUtils.createDatasetIfNotExists(bigQuery, DatasetId.of(datasetProjectId, datasetName),
                                                  config.getLocation(), cmekKeyName,
                                                  () -> String.format("Unable to create BigQuery dataset '%s.%s'",
                                                                      datasetProjectId, datasetName));
+      if (cmekKeyName != null) {
```
seanzhougoogle, line 124: should this cmekKey to be set before createDatasets ? or it has not impact on datasets creation ?

itsankit-google: yeah, it has no impact on dataset creation, that follows a different codepath.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -279,6 +289,10 @@ public String getSql() {
       return sql;
     }
 
+    public Boolean getStoreResults() {
```
seanzhougoogle, line 292: I don't see `getStoreResults` is used at backend, is it purely used for UI ? Do you need to make sure only when `getStoreResults` is true, you store the result in the table ?

itsankit-google: addressed it in `review comments` commit https://github.com/data-integrations/google-cloud/pull/1042/files#diff-9ccb1a50b6e5a7a6b2c016c6aaa7605605d66e7e6bfb5e4e95a16ac7c35caa89R118

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -86,14 +87,13 @@ public void run(ActionContext context) throws Exception {
...
-    }
 
     // Save the results of the query to a permanent table.
-    if (config.getDataset() != null && config.getTable() != null) {
-      builder.setDestinationTable(TableId.of(config.getDataset(), config.getTable()));
+    String datasetName = config.getDataset();
+    String tableName = config.getTable();
+    String datasetProjectId = config.getDatasetProject();
+    if (datasetProjectId != null && datasetName != null && tableName != null) {
+      builder.setDestinationTable(TableId.of(datasetProjectId, datasetName, tableName));
```
seanzhougoogle, line 96: should this be guarded by getStoreResults too ?

itsankit-google: ohh, yeah missed it

seanzhougoogle (approved):

PR #1063, opened 2022-06-09
[PLUGIN-664] Support bulk delete in GCS Delete Action

JessieYee:
Jira: https://cdap.atlassian.net/browse/PLUGIN-664
PR for integration test: https://github.com/cdapio/cdap-integration-tests/pull/1180


src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,77 @@ public void run(ActionContext context) throws Exception {
...
         } catch (IOException e) {
           LOG.warn(
             String.format("Failed to delete path '%s'", gcsPath)
           );
         }
       }
     }
+
+    for (GCSPath gcsPath : gcsPathsWild) {
+      String gcsPathStr = gcsPath.getUri().toString();
```
sumengwang: Can you group these stuff under the for loop into a function and write unit tests against the function. It's a little difficult to visualize the input/output.

JessieYee: Yeah, let me try to use the function

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,77 @@ public void run(ActionContext context) throws Exception {
       if (fs.exists(gcsPath)) {
         try {
           fs.delete(gcsPath, true);
+          deleteCount = deleteCount + 1;
```
de-lan: `deleteCount++` ?

seanzhougoogle (commented):
please update the document md as well.

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
JessieYee: done :)

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,80 @@ public void run(ActionContext context) throws Exception {
...
           LOG.warn(
             String.format("Failed to delete path '%s'", gcsPath)
           );
         }
       }
     }
+
+    for (GCSPath gcsPath : gcsPathsWild) {
+      String regex = getRegex(gcsPath);
+      try {
```
seanzhougoogle, line 142: does the logic of getting filesystem need to be in the loop ?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,80 @@ public void run(ActionContext context) throws Exception {
...
+        fs = new Path(gcsPath.getUri()).getFileSystem(configuration);
+      } catch (IOException e) {
+        LOG.info("Failed deleting file " + gcsPath.getUri().getPath() + ", " + e.getMessage());
+        // no-op.
+        continue;
+      }
+      Page<Blob> blobs = storage.list(gcsPathsWild.get(0).getBucket());
```
seanzhougoogle: `gcsPathsWild.get(0)` -> gcsPath

seanzhougoogle: test when user input `gs://bucket*` what will happen here to `getBucket()`, try to handle this error case

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,80 @@ public void run(ActionContext context) throws Exception {
...
+      for (Blob blob : blobs.iterateAll()) {
+        Path filePath = new Path("gs://" + blob.getBucket() + "/" + blob.getName());
+        if (filePath.toString().matches(regex)) {
+          if (fs.exists(filePath)) {
+            Path rootPath = getRootMatch(filePath, regex);
```
seanzhougoogle, line 155: add some comments about why to get root path

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,80 @@ public void run(ActionContext context) throws Exception {
...
+        }
+      }
+    }
+    context.getMetrics().gauge("gc.file.delete.count", deleteCount);
+  }
+
+  public Path getRootMatch(Path filePath, String regx) {
+    Path parent = filePath;
+    String child = null;
+    while (parent.getParent() != null && parent.toString().matches(regx)) {
```
seanzhougoogle: try to simplify it by making parent as `current` and test `current.getParent().toString().matches(regx)`

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,80 @@ public void run(ActionContext context) throws Exception {
...
+    if (child == null) {
+      return parent;
+    } else if (parent.getParent() == null && parent.toString().matches(regx)) {
+      return parent;
+    } else {
+      return new Path(parent.toString() + "/" + child);
+    }
+  }
+
+  public String getRegex(GCSPath gcsPath) {
```
seanzhougoogle: discussed offline, looks like this method is not necessary, when matching root path we can just append a `/` to match regex

itsankit-google (commented):
It would be good if we can use new `cdap-e2e-tests` framework for adding `e2e-tests` for these scenarios instead of cloud-itn since the plan is to move away from it. https://github.com/data-integrations/google-cloud/tree/develop/src/e2e-test/features/gcsdelete

cc - @rmstar, @tivv

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
JessieYee: Might received error when use `gs://bucket*`. Try validate and received message: 

Invalid bucket name in path 'test_664_*'. Bucket name should only contain lower case alphanumeric, '-'. '_' and '.'. Please follow GCS naming convention: https://cloud.google.com/storage/docs/naming-buckets

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
JessieYee: If we have delete objects from several bucket, we should get filesystem use different path. I am gonna to clarify if we should support delete from different buckets

src/e2e-test/resources/pluginParameters.properties
```diff
@@ -108,6 +112,10 @@ pubsubDelimiter=@
 
 ## GCSDELETE-PLUGIN-PROPERTIES-START
 gcsDeleteObjectsList=testdata/GCS_RECURSIVE_TEST/recursiveFile1.csv,testdata/GCS_RECURSIVE_TEST/recursiveFile2.csv
+gcsDeleteObjectsList2=testdata/GCS_WILDCARD_TEST/wildcardFile2.csv,testdata/GCS_WILDCARD_TEST/wildcardFile3.csv
+gcsDeleteObjectsList3=testdata/GCS_WILDCARD_TEST/test
```
sumengwang, line 116: `gcsWildcardPath3=testdata/GCS_WILDCARD_TEST/text*`
You have `text` above but `test` here. Is it a typo?

JessieYee: oh yeah!

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -119,13 +128,54 @@ public void run(ActionContext context) throws Exception {
...
             String.format("Failed to delete path '%s'", gcsPath)
           );
         }
       }
     }
+
+    for (GCSPath gcsPath : gcsPathsWild) {
+      String regex = ("\\Q" + gcsPath.getUri().toString() + "\\E").replace("*", "\\E[^/]*\\Q") + "(/.*)?";
+      try {
+        fs = new Path(gcsPathsWild.get(0).getUri()).getFileSystem(configuration);
```
seanzhougoogle, line 143: still using `gcsPathsWild.get(0)` ? then we can move it out of the iteration ?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
```diff
@@ -103,11 +108,15 @@ public void run(ActionContext context) throws Exception {
           String.format("Unable to access or create bucket %s. ", gcsPath.getBucket())
             + "Ensure you entered the correct bucket path and have permissions for it.", e);
       }
-      gcsPaths.add(new Path(gcsPath.getUri()));
+      if (gcsPath.getUri().toString().contains("*")) {
```
seanzhougoogle, line 111: I'm wondering whether we can handle it uniformly for wild and non-wild case ? non-wild case is just a special case of wild case , right ?

src/e2e-test/resources/testdata/GCS_WILDCARD_TEST/test/wildcardFile1.csv
```diff
@@ -0,0 +1,9 @@
+id,name,yearofbirth,tablename,offset
```
seanzhougoogle: do we need that much data ? for the csv file ? it's just for test deletion , right ? empty or even 1 character file will be enough , right ?

src/e2e-test/features/gcsdelete/GCSDelete.feature
```diff
@@ -55,6 +55,63 @@ Feature: GCS Delete - Verification of GCS Delete plugin
     Then Verify the pipeline status is "Succeeded"
     Then Verify multiple objects "gcsDeleteObjectsList" deleted successfully by GCS Delete action plugin
 
+  @GCS_WILDCARD_TEST
```
seanzhougoogle: nit, would it be better to call it GCS_DELETE_WILDCARD_TEST ?

src/e2e-test/resources/testdata/GCS_WILDCARD_TEST/test/wildcardFile1.csv
JessieYee: yeah, will use the empty file

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
JessieYee: For non-wild case, we do not need to iterate Blobs (storage path list) for each gcspath. If we use the same logic for wild and non-wild case, I am worries if time for searching the filed to delete will increase?

src/main/java/io/cdap/plugin/gcp/gcs/actions/GCSBucketDelete.java
JessieYee: Setting another Jira to solve the delete from multiple bucket: https://cdap.atlassian.net/browse/PLUGIN-1291.

src/e2e-test/features/gcsdelete/GCSDelete.feature
JessieYee: ok

seanzhougoogle (approved):

src/e2e-test/java/io/cdap/plugin/gcsdelete/stepsdesign/GCSDelete.java
```diff
@@ -119,6 +119,20 @@ public void verifyMultipleObjectsDeletedSuccessfullyByGCSDeleteActionPlugin(Stri
     BeforeActions.scenario.write("Multiple objects deleted successfully by GCS Delete plugin.");
   }
 
+  @Then("Verify objects {string} still exist after GCS Delete action plugin")
+  public void verifyObjectsExistByGCSDeleteActionPlugin(String commaSeparatedObjectsList)
+    throws IOException {
+    String[] objectsToDelete = PluginPropertyUtils.pluginProp(commaSeparatedObjectsList).split(",");
```
itsankit-google: nit->Is it better to call them `objectsToKeep` rather than `objectsToDelete`?

JessieYee: right :)

itsankit-google (approved):
Please squash commits before merging.

PR #1064, opened 2022-06-15
Fix BigQuerySink test cases

jockech:
**Scenario:Validate successful records transfer from GCS to BigQuery**
The preview failure was caused by format parsing exception:
And exception was related to these [two steps](https://github.com/data-integrations/google-cloud/blob/release/0.20/src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature#L10-L11):
```
Then Enter the GCS source mandatory properties 
Then Override Service account details if set in environment variables
```
Entering the GCS source mandatory properties will trigger the [get schema](https://github.com/data-integrations/google-cloud/blob/b55a3c5dd88f35d7f2d94a4b4c6635293e7031a1/src/e2e-test/java/io/cdap/plugin/gcs/stepsdesign/GCSSource.java#L90). Due to the service account missing, which was setup after this step, the GCS source failed to get the schema and caused the BigQuery table input format to be {offset: int, body: str}, which was inconsistent with the [source of truth](https://github.com/data-integrations/google-cloud/blob/release/0.20/src/e2e-test/resources/testdata/GCS_CSV_TEST.csv).

Fix: Switch the order can fix the issue

**BigQuery source - Verification of BigQuery to GCS successful data transfer with macro arguments**
The service account key is a multiple line json and can mess the runtime arguments if it's not handled correclty.
image
Fix: Parse the service account json into one line.

### DIFF
--- a/src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
+++ b/src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
@@ -7,8 +7,8 @@ Feature: BigQuery sink - Verification of GCS to BigQuery successful data transfe
     When Source is GCS
     When Sink is BigQuery
     Then Open GCS source properties
-    Then Override Service account details if set in environment variables
     Then Enter the GCS source mandatory properties
+    Then Override Service account details if set in environment variables
     Then Validate "GCS" plugin properties
     Then Close the GCS properties
     Then Open BigQuery sink properties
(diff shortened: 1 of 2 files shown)


src/e2e-test/features/bigquery/sink/GCSToBigQuery.feature
```diff
@@ -7,8 +7,8 @@ Feature: BigQuery sink - Verification of GCS to BigQuery successful data transfe
     When Source is GCS
     When Sink is BigQuery
     Then Open GCS source properties
-    Then Override Service account details if set in environment variables
     Then Enter the GCS source mandatory properties
+    Then Override Service account details if set in environment variables
```
itsankit-google, line 11: Why reordering of this step is needed??

jockech: explained in the description. The first step will trigger the `get schema` and it will fail without the service account json.

itsankit-google: ahh makes sense from the description.

itsankit-google (approved):
lgtm

PR #1065, opened 2022-06-16
e2e BigQuery Execute test scenarios

sawantpritam:
@rmstar @itsankit-google Please review.


itsankit-google (commented):
Can we add 2 configurable scenarios one with different `ddl queries` and other with `dml queries`? 
For eg, https://github.com/data-integrations/google-cloud/blob/develop/src/e2e-test/features/gcs/source/GCSSourceError.feature#L4

sawantpritam:
> Can we add 2 configurable scenarios one with different `ddl queries` and other with `dml queries`? For eg, https://github.com/data-integrations/google-cloud/blob/develop/src/e2e-test/features/gcs/source/GCSSourceError.feature#L4

Added scenarios with different ddl and dml queries.

sawantpritam:
Resolved conflicts.

src/e2e-test/features/bigqueryexecute/BQExecute.feature
```diff
@@ -0,0 +1,125 @@
...
+    Then Enter input plugin property: "referenceName" with value: "BQSinkReferenceName"
+    Then Enter input plugin property: "dataset" with value: "dataset"
+    Then Enter input plugin property: "table" with value: "bqTargetTable"
+    Then Click plugin property: "truncateTable"
+    Then Click plugin property: "updateTableSchema"
+    Then Validate "BigQuery" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Enter runtime argument value "bqeSourceOutputSchema" for key "bqeSourceOutputSchema"
```
itsankit-google, line 67: How is the value and key same here?

itsankit-google (commented):
Can we add tests for `update/upsert/select` operations??
And also test `Store results in a BQ Table` functionality??

sawantpritam:
> Can we add tests for `update/upsert/select` operations?? And also test `Store results in a BQ Table` functionality??

Added update and upsert scenario. Other scenarios already added in initial commit.

src/e2e-test/features/bigqueryexecute/BQExecute.feature
```diff
@@ -0,0 +1,175 @@
...
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Verify 1 records inserted in BigQuery table: "bqSourceTable" with query "bqExecuteCountDMLInsert"
+
+  @BQ_SOURCE_TEST @BQ_EXECUTE_UPSERT_SQL
+  Scenario: Verify BQExecute plugin functionality for DML query - Upsert data
```
itsankit-google, line 128: Can use a pipeline similar to the screenshot to simulate upsert, one plugin inserts and other plugin update/insert records, this will not need to create 2 pipelines?? @rmstar wdyt??

image

sawantpritam: We will not be able to validate first plugins result with above pipeline.

We should run pipeline twice to validate UPSERT query functionality.
In first run record should be inserted. In Second run same record should be updated.

itsankit-google (approved):

PR #1067, opened 2022-06-24
Run only required tests on push

sawantpritam:

This PR depends on https://github.com/cdapio/cdap-e2e-tests/pull/102

Added Required runner for only BigQuery. Once this PR is approved. I will add for other plugins as well.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -57,7 +57,11 @@ jobs:
           key: ${{ runner.os }}-maven-${{ github.workflow }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-${{ github.workflow }}
-      - name: Run tests
+      - name: Run required e2e tests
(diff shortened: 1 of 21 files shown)


itsankit-google (changes requested):
Can we keep one tag named as `REQUIRED` and one common runner?
Do we really need to maintain different required runners for all plugins?

sawantpritam:
> Can we keep one tag named as `REQUIRED` and one common runner? Do we really need to maintain different required runners for all plugins?

If we maintain single required runner then those required tests will get executed in sequence. Currently with different required runners 2 runners get executed in parallel.

itsankit-google (approved):

itsankit-google:
Please squash commits

sawantpritam:
> Please squash commits

Commits squashed.

itsankit-google (approved):
Please squash commits.

sawantpritam:
> Commits squashed.

Commits squashed.

PR #1072, opened 2022-06-29
CDAP-19078 update service account loading to use newer libraries

albertshau:
Configure the GCS connector to use a custom AccessTokenProvider
that uses the newer Credentials library instead of the default
behavior, which uses the Credential library. This allows the
connector to work with newer service account types, such as the
external_account type.

As part of this, refactoring the util class to clean up the code
and also remove bad assumptions, like the service account will
always be of type 'service_account' and instead let the
GoogleCredentials class look at the type and determine the correct
underlying implementation.

Also removing a bunch of useless constants and property prefixing.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -20,18 +20,19 @@ import com.google.auth.Credentials;
 import com.google.auth.oauth2.ExternalAccountCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.bigtable.repackaged.com.google.gson.Gson;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryOptions;
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS;
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem;
 import com.google.cloud.hadoop.util.AccessTokenProviderClassFromConfigFactory;
 import com.google.cloud.hadoop.util.CredentialFactory;
-import com.google.cloud.hadoop.util.HadoopCredentialConfiguration;
 import com.google.cloud.kms.v1.CryptoKeyName;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
+import com.google.gson.reflect.TypeToken;
 import io.cdap.plugin.gcp.gcs.GCSPath;
 import io.cdap.plugin.gcp.gcs.ServiceAccountAccessTokenProvider;
 import org.apache.hadoop.conf.Configuration;
@@ -41,6 +42,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Type;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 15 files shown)


src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -48,100 +57,133 @@
...
-      return ServiceAccountCredentials.fromStream(serviceAccountStream);
-    }
+  /**
+   * Load a service account from the local file system.
+   *
+   * @param path path to the service account file
+   * @return credentials generated from the service account file
+   * @throws IOException if there was an error reading the file
+   */
+  public static GoogleCredentials loadServiceAccountCredentials(String path) throws IOException {
```
dli357: nit: Can we rename this to loadFilesystemServiceAccountCredentials or something similar? This overload makes it a bit unclear that it is a subset of the other method.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -48,100 +57,133 @@
...
-    Map<String, String> properties = new HashMap<>();
-    if (serviceAccountType == null) {
-      return properties;
-    }
-    String privateKeyData = null;
-    properties.put(SERVICE_ACCOUNT_TYPE, serviceAccountType);
+  public static Map<String, String> generateGCSAuthProperties(@Nullable String serviceAccount,
+                                                              String serviceAccountType) {
+    // fs.gs prefix is for GoogleHadoopFileSystemBase.getCredential(), used by the GCS connector.
+    return generateAuthProperties(serviceAccount, serviceAccountType, "fs.gs");
```
dli357: nit: Maybe we can make "fs.gs" a constant value with a descriptive name? Then the comment above can be included with the constant.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -171,10 +213,19 @@ private static Map<String, String> getFileSystemProperties(String project, Strin
   public static BigQuery getBigQuery(String project, @Nullable Credentials credentials) {
     BigQueryOptions.Builder bigqueryBuilder = BigQueryOptions.newBuilder().setProjectId(project);
     if (credentials != null) {
+      Set<String> scopes = new HashSet<>(BIGQUERY_SCOPES);
+
       if (credentials instanceof ServiceAccountCredentials) {
```
dli357, line 236: Trying to understand the intent behind this code here. Is it necessary to retrieve scopes from the credential, and will the credential return any scopes? Seems like we should only be creating access tokens with `BIGQUERY_SCOPES` unless we explicitly need access to additional scopes.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public void refresh() throws IOException {
+    getCredentials().refresh();
+  }
+
+  private GoogleCredentials getCredentials() throws IOException {
+    if (credentials == null) {
+      credentials = GCPUtils.loadCredentialsFromConf(conf);
```
dli357, line 59: If this is passed to Hadoop GCS connector, will the conf be null? Do we need to add some logic here where we initialize the credentials using `GoogleCredentials.getApplicationDefault()` to handle the `auto-detect` case?

albertshau: if the service account is null, `loadCredentialsFromConf()` will call `GoogleCredentials.getApplicationDefault()`

the conf will not be null because `AccessTokenProvider` extends `Configurable`, whose contract is that `setConf()` will be called before anything else is called

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
albertshau: I'm not sure why this would need anything other than bigquery scopes, but I'm maintaining existing logic

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -217,9 +216,10 @@ public String getServiceAccountFilePath() {
    */
   public String getServiceAccountEmail() throws IOException {
     if (connection.isServiceAccountJson() || connection.getServiceAccountFilePath() != null) {
-      ServiceAccountCredentials credentials =
-        GCPUtils.loadServiceAccountCredentials(getServiceAccount(), isServiceAccountFilePath());
-      return credentials.getClientEmail();
+      try (InputStream inputStream = GCPUtils.openServiceAccount(getServiceAccount(), isServiceAccountFilePath())) {
+        ServiceAccountCredentials credentials = ServiceAccountCredentials.fromStream(inputStream);
```
albertshau, line 220: this will fail with external service accounts, but wanted to limit scope of this PR as fixing this likely requires adding a new property to the dataplex config

dli357: I think this should be fine because external account is specifically used for identity federation and requires a projected token (typically kubernetes, but can be any identity provider) to be on the filesystem. The typical user would not be using an external account file in the plugin config itself; instead it'd typically be set with the environment.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -48,100 +57,133 @@
...
+
+    GoogleCredentials credentials;
+    if (serviceAccount == null) {
+      credentials = GoogleCredentials.getApplicationDefault().createScoped();
+    } else {
+      boolean isFile = SERVICE_ACCOUNT_TYPE_FILE_PATH.equals(serviceAccountType);
+      credentials = loadServiceAccountCredentials(serviceAccount, isFile);
+    }
+
+    return credentials.createScoped(CredentialFactory.GCS_SCOPES);
```
dli357: Will this only ever need GCS_SCOPES? If so, can we rename it to be `loadGCSCredentialsFromConf`? Just want to make sure this method doesn't get mistakenly called for some other client. Thanks!

dli357 (approved):

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
dli357: Gotcha, sounds good

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
albertshau: good catch, it's not always supposed to be. Will introduce another version of the method that takes the scopes. If none are given, it won't scope the credential.

PR #1075, opened 2022-07-11
add user managed schema option to dataplex sink plugin

mrahanjam:
This PR adds a new config parameter (i.e., `Update Dataplex Metadata`) to dataplex sink plugin which adds support for updating metadata in dataplex for newly generated data.

JIRA: [PLUGIN-1379](https://cdap.atlassian.net/browse/PLUGIN-1379)

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
+++ b/src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
@@ -17,22 +17,31 @@
 package io.cdap.plugin.gcp.dataplex.common.util;
 
 import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.paging.Page;
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.google.cloud.dataplex.v1.DataplexServiceClient;
 import com.google.cloud.dataplex.v1.DataplexServiceSettings;
+import com.google.cloud.dataplex.v1.Entity;
 import com.google.cloud.dataplex.v1.Job;
 import com.google.cloud.dataplex.v1.JobName;
 import com.google.cloud.dataplex.v1.MetadataServiceClient;
 import com.google.cloud.dataplex.v1.MetadataServiceSettings;
+import com.google.cloud.dataplex.v1.Partition;
+import com.google.cloud.dataplex.v1.Schema.Mode;
+import com.google.cloud.dataplex.v1.Schema.PartitionField;
+import com.google.cloud.dataplex.v1.Schema.PartitionStyle;
+import com.google.cloud.dataplex.v1.Schema.SchemaField;
+import com.google.cloud.dataplex.v1.Schema.Type;
 import com.google.cloud.dataplex.v1.TaskName;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.data.schema.Schema.LogicalType;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryTypeSize;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.GCPConnectorConfig;
 import io.cdap.plugin.gcp.common.GCPUtils;
-
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 import org.awaitility.Awaitility;
@@ -43,9 +52,9 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
-
 import javax.annotation.Nullable;
 
 /**
@@ -230,8 +239,8 @@ public final class DataplexUtil {
         String error =
           String.format("Entity column '%s' is of unsupported type '%s'.", fieldName, standardType.name());
         String action = String.format("Supported column types are: %s.",
-          BigQueryUtil.BQ_TYPE_MAP.keySet().stream().map(t -> t.getStandardType().name())
-            .collect(Collectors.joining(", ")));
+                                      BigQueryUtil.BQ_TYPE_MAP.keySet().stream().map(t -> t.getStandardType().name())
+                                        .collect(Collectors.joining(", ")));
         if (collector != null) {
           collector.addFailure(error, action);
         } else {
@@ -292,7 +301,7 @@ public final class DataplexUtil {
     try (DataplexServiceClient dataplexServiceClient = DataplexUtil.getDataplexServiceClient(googleCredentials)) {
       DataplexServiceClient.ListJobsPagedResponse
         jobList = dataplexServiceClient.listJobs(TaskName.newBuilder().setProject(projectID).setLake(lake).
-        setLocation(location).setTask(taskId).build());
+                                                   setLocation(location).setTask(taskId).build());
       Job dataplexJob = jobList.iterateAll().iterator().next();
       try {
         Awaitility.await()
@@ -302,7 +311,7 @@ public final class DataplexUtil {
           .until(() -> {
             Job currentJob =
               dataplexServiceClient.getJob(JobName.newBuilder().setProject(projectID).setLocation(location)
-                .setLake(lake).setTask(taskId).setJob(dataplexJob.getUid()).build());
+                                             .setLake(lake).setTask(taskId).setJob(dataplexJob.getUid()).build());
             LOG.debug("State of the Job is still " + currentJob.getState());
             return currentJob.getState() != null && !Job.State.RUNNING.equals(currentJob.getState()) &&
               !Job.State.STATE_UNSPECIFIED.equals(currentJob.getState());
@@ -311,7 +320,8 @@ public final class DataplexUtil {
         throw new IOException("Job timed out.", e);
       }
       Job completedJob = dataplexServiceClient.getJob(JobName.newBuilder().setProject(projectID).setLocation(location)
-        .setLake(lake).setTask(taskId).setJob(dataplexJob.getUid()).build());
+                                                        .setLake(lake).setTask(taskId).setJob(dataplexJob.getUid())
+                                                        .build());
       if (!Job.State.SUCCEEDED.equals(completedJob.getState())) {
         throw new IOException("Job failed with message: " + completedJob.getMessage());
       }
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/plugin/gcp/dataplex/sink/config/DataplexBatchSinkConfig.java
```diff
@@ -455,6 +478,95 @@ public void validateAssetConfiguration(FailureCollector collector, DataplexServi
...
   }
 
+  /**
+   * Validates metadata update in Dataplex.
+   *
+   * @param collector
+   * @param dataplexServiceClient
+   * @throws IOException
+   */
+  public void configureMetadataUpdateWithUserManagedSchema(
```
rmstar: Can we move this method to DataplexBatchSink.java? The config class is just used for Dataplex configs. Entity creation logic should be in the batchsink class

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -199,6 +201,36 @@ public void transform(StructuredRecord input, Emitter<KeyValue<Object, Object>>
   public void onRunFinish(boolean succeeded, BatchSinkContext context) {
     if (this.config.getAssetType().equalsIgnoreCase(DataplexConstants.STORAGE_BUCKET_ASSET_TYPE)) {
       emitMetricsForStorageBucket(succeeded, context);
+      if (succeeded) {
```
rmstar: Can we add a comment describing why we need this logic and why it's only applicable to GCS assets

src/main/java/io/cdap/plugin/gcp/dataplex/sink/config/DataplexBatchSinkConfig.java
```diff
@@ -455,6 +478,95 @@ public void validateAssetConfiguration(FailureCollector collector, DataplexServi
...
+    FailureCollector collector, DataplexServiceClient dataplexServiceClient, GoogleCredentials credentials,
+    String assetFullPath, StorageSystem storageSystem, Schema schema
+  ) throws IOException {
+    if (isUserManagedSchema()) {
```
rmstar: We can do something like this to make it more readable:
```
if (!isUserManangedSchema()) {
  return;
}
...
```

src/e2e-test/java/io/cdap/plugin/utils/CdfPluginPropertyLocator.java
```diff
@@ -24,6 +24,7 @@ public enum CdfPluginPropertyLocator {
   SERVICE_ACCOUNT_PATH("serviceFilePath"),
   SERVICE_ACCOUNT_JSON("serviceAccountJSON"),
   TRUNCATE_TABLE("truncateTable"),
+  USER_MANAGED_SCHEMA("userManagedSchema"),
```
rmstar: where is this used?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/config/DataplexBatchSinkConfig.java
```diff
@@ -205,6 +217,12 @@ public class DataplexBatchSinkConfig extends DataplexBaseConfig {
     "operation.")
   protected Boolean truncateTable;
 
+  @Name(NAME_USER_MANAGED_SCHEMA)
+  @Nullable
+  @Macro
+  @Description("Whether the BigQuery table schema is managed by the user.")
```
rmstar: Is this BQ schema or Dataplex schema?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -574,9 +574,17 @@ protected String getOutputDir(long logicalStartTime) {
     String suffix = config.getSuffix();
     String defaultTimestampFormat = "yyyy-MM-dd-HH-mm";
     String tableName = config.getTable();
-    String timeSuffix = suffix == null || suffix.isEmpty() ?
-      "ts=" + new SimpleDateFormat(defaultTimestampFormat).format(logicalStartTime) :
-      "ts=" + new SimpleDateFormat(suffix).format(logicalStartTime);
+    String timeSuffix = suffix == null || suffix.isEmpty()
```
rmstar: nit: code can be simplified a bit.
```
        String suffix = Strings.isNullOrEmpty(config.getSuffix()) ? defaultTimestampFormat : suffix;
        String.format("%s=%s",  
          DataplexConstants.STORAGE_BUCKET_PARTITION_KEY, 
          new SimpleDateFormat(suffix).format(logicalStartTime)
        );
```

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -392,4 +398,201 @@ public static MetadataServiceClient getMetadataServiceClient(GoogleCredentials c
...
+
+    Schema schema = field.getSchema();
+
+    if (schema.getType() == Schema.Type.UNION) {
+      // Special case for UNION: a union of ["null", "non-null type"] means this is a
+      // nullable field. In Dataplex this will be a field with Mode = NULLABLE and Type = <non-null
+      // type>. So here we have to return the type of the other, non-NULL, element.
+      // A union of 3+ elements is not supported (can't be represented as a Dataplex type).
+      if (schema.isNullable()) {
+        Schema s1 = schema.getUnionSchemas().get(0);
```
rmstar: you can get the non-null schema using schema.getNonNullable()

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +605,91 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
     }
   }
 
+  /**
+   * Validates metadata update in Dataplex.
```
rmstar: Configures metadata update in Dataplex if the schema is user managed

rmstar (commented):
please squash commits after addressing minor comments

rmstar (approved):

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +608,99 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
...
+      entityBean =
+        metadataServiceClient.getEntity(GetEntityRequest.newBuilder().setName(EntityName.of(
+            config.tryGetProject(), config.getLocation(), config.getLake(), config.getZone(), entityID).toString())
+                                          .setView(GetEntityRequest.EntityView.FULL)
+                                          .build());
+    } catch (ApiException e) {
+      // ignore
+    }
+    if (entityBean != null && entityBean.getSchema().getUserManaged() == false) {
+      throw new RuntimeException("Entity already exists, but the schema is not user-managed.");
```
rmstar: add failures to failure collector

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +608,99 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
...
+  /**
+   * Configures metadata update in Dataplex if the schema is user managed
+   *
+   * @param credentials
+   * @param assetFullPath
+   * @param storageSystem
+   * @param schema
+   * @throws IOException
+   */
+  private void configureMetadataUpdateWithUserManagedSchema (
```
rmstar: nit: remove space before (

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +608,99 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
...
+  /**
+   * Configures metadata update in Dataplex if the schema is user managed
+   *
+   * @param credentials
+   * @param assetFullPath
+   * @param storageSystem
+   * @param schema
+   * @throws IOException
+   */
+  private void prepareMetadataUpdateWithUserManagedSchema (
```
rmstar: nit: remove space before (

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +608,99 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
...
+    MetadataServiceClient metadataServiceClient =
+      DataplexUtil.getMetadataServiceClient(credentials);
+    try {
+      entityBean =
+        metadataServiceClient.getEntity(GetEntityRequest.newBuilder().setName(EntityName.of(
+            config.tryGetProject(), config.getLocation(), config.getLake(), config.getZone(), entityID).toString())
+                                          .setView(GetEntityRequest.EntityView.FULL)
+                                          .build());
+    } catch (ApiException e) {
+      // ignore
```
rmstar: why are we ignoring APIException?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -469,6 +512,13 @@ private void prepareRunStorageBucket(BatchSinkContext context) throws Exception
     if (bucket == null) {
       GCPUtils.createBucket(storage, bucketName, config.getLocation(), cmekKeyName);
     }
+    if (config.isUserManagedSchema()) {
+      GoogleCredentials googleCredentials = config.validateAndGetServiceAccountCredentials(collector);
```
rmstar: move this into prepareRun(). Fail if the asset type is BigQuery because this feature is only supported with GCS assets

rmstar: Also call this code from configurePipeline() so that validation errors can be caught at configure time

mrahanjam: Re: FIrst comment above: The `User Managed` toggle only shows up for GCS assets. It's always hidden for BQ assets.

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
mrahanjam: Because if `entityBean` is `null` (which gives an ApiException (404)), we'd create one, if it exists, we update it.

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
mrahanjam: Not sure if you meant to move it to `prepareRun()` or `configurePipeline()` (or both?!). But I ended up calling it from `prepareRun()`. It swallows the `RuntimeException` when called from `configurePipeline()`.

rmstar: I meant both. 
Why?
- `configurePipeline` is called at configure time. It's easier to find and fix configuration errors while validating the pipeline in studio vs running a deployed pipeline. Note that macros are not evaluated at configure time, so macro params won't be validated here. You'll want to wrap your validation logic in if(containMacro()) checks (you can find examples of this in google-cloud repo).
- `prepareRun` is called when the pipeline is run. We want the validation logic to be present here as well because macros are resolved when prepareRun is called. So macro params can be validated here.

rmstar: > It swallows the RuntimeException when called from configurePipeline()

Don't throw RuntimeException. Add a failure to FailureCollector.

mrahanjam: addressed in [479534a](https://github.com/data-integrations/google-cloud/pull/1075/commits/479534aae257b81d9efe58ae4b29a15a7659560b). Will squash after second approval.

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -589,4 +593,44 @@ private static Type dataplexLogicalFieldType(Schema schema) {
...
+    Storage storage = GCPUtils.getStorage(project, credentials);
+    String delimiter = "/";
+    String partitionPrefix = DataplexConstants.STORAGE_BUCKET_PARTITION_KEY + "=";
+    Page<Blob> blobs =
+      storage.list(
+        bucketName,
+        Storage.BlobListOption.prefix(tableName + delimiter + partitionPrefix),
+        Storage.BlobListOption.currentDirectory());
+    String lastPartition = null;
+    for (Blob blob : blobs.iterateAll()) {
```
mrahanjam, line 625: Not too familiar with iterables in Java. Is there a better way (i.e., in O(1)) to get to the last element of the iterator (looks like it's a linked list not an array)?

rmstar: I don't think there's a way to reverse iterate over Pages

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -438,8 +438,9 @@ public static com.google.cloud.dataplex.v1.Schema getDataplexSchema(Schema schem
     if (schema == null) {
       return dataplexSchemaBuilder.build();
     }
-    dataplexSchemaBuilder.addAllFields(Objects.requireNonNull(schema.getFields()).stream().
-      map(DataplexUtil::toDataplexSchemaField).collect(Collectors.toList()));
+    dataplexSchemaBuilder.addAllFields(Objects.requireNonNull(schema.getFields()).stream()
+      .filter(avroField -> !avroField.getName().equals(DataplexConstants.STORAGE_BUCKET_PARTITION_KEY))
```
mrahanjam, line 444: to ensure old `ts` field is not present in the schema

rmstar: please add this as a comment in the code

rmstar: also explain  in the comment why `ts` can't be present in the schema

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
rmstar: I see. should we ignore the exception only if e.getCode() == 404 (add failure to collector for any other errors)?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -556,6 +606,121 @@ private void emitMetricsForStorageBucket(boolean succeeded, BatchSinkContext con
...
+        } catch (ApiException e) {
+          throw new RuntimeException(
+            String.format("%s: %s", "There was a problem creating the entity for metadata updates.", e.getMessage()));
+        }
+      }
+      try {
+        DataplexUtil.addPartitionInfo(entityBean, credentials,
+                                    asset.getResourceSpec().getName(), config.getTable(), config.getProject());
+      } catch (ApiException e) {
+        String errMessage = e.getMessage().substring(e.getMessage().length() - 14);
```
rmstar, line 720: why are we subtracting 14?

rmstar: Can you add a comment to clarify. 
Also, is it possible to figure out if it's an "already exists" error based on error code instead of parsing the message?

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -392,4 +402,238 @@ public static MetadataServiceClient getMetadataServiceClient(GoogleCredentials c
...
+    Page<Blob> blobs =
+      storage.list(
+        bucketName,
+        Storage.BlobListOption.prefix(tableName + delimiter + partitionPrefix),
+        Storage.BlobListOption.currentDirectory());
+    String lastPartition = null;
+    for (Blob blob : blobs.iterateAll()) {
+      lastPartition = blob.getName();
+    }
+    String location = "gs://" + bucketName +
```
rmstar: nit: use constant for gs://
```
private static final String SCHEME = "gs://";
```

src/main/java/io/cdap/plugin/gcp/dataplex/common/util/DataplexUtil.java
```diff
@@ -392,4 +402,238 @@ public static MetadataServiceClient getMetadataServiceClient(GoogleCredentials c
...
+      delimiter + lastPartition.substring(0, lastPartition.length() - 1);
+    String[] lastPartitionParts = lastPartition.split(delimiter);
+    Partition.Builder dataplexPartitionBuilder =
+      Partition.newBuilder();
+    try (MetadataServiceClient metadataServiceClient =
+           getMetadataServiceClient(credentials)) {
+      Partition partition = dataplexPartitionBuilder
+        .setLocation(location)
+        .addValues(lastPartitionParts[lastPartitionParts.length - 1].substring(partitionPrefix.length()))
```
rmstar, line 639: some comments to explain what we are doing and why would help here.
- Why do we need to get the last blob name
- show example of what does the blob name look like, to make the parsing logic easier to understand

pom.xml
```diff
@@ -72,7 +72,7 @@
     <avro.version>1.8.2</avro.version>
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
-    <cdap.version>6.7.0</cdap.version>
+    <cdap.version>6.8.0-SNAPSHOT</cdap.version>
```
rmstar: why are we bumping up cdap version depedency? Are we adding a dependency on a new CDAP API?

mrahanjam: @snehalakshmisha made some changes in `cdap-api`, and `hydrator-common`. After those changes were merged with develop. My pipelines started breaking and I had to update cdap version dependency for it to work again.

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
mrahanjam: The status code comes as `INVALID_ARGUMENT (400)`. I thought that may be too generic, hence parsing the last 14 chars of the error message to make sure it's failing due to `already exists` error.

mrahanjam: I can just use http status code 400 instead.

rmstar (approved):
please squash commits

PR #1076, opened 2022-07-14
[PLUGIN-1309] Document permissions required to use Dataplex plugins

rmstar:


dli357 (approved):
nit: Might be helpful to link back to a page describing these required roles if one exists. Thanks!

PR #1078, opened 2022-07-19
Added simple BigQuery Pushdown source

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryReadDataset.java
```diff
@@ -53,10 +57,10 @@
 import io.cdap.cdap.etl.api.engine.sql.request.SQLReadResult;
 import io.cdap.plugin.gcp.bigquery.source.BigQuerySourceConfig;
 import io.cdap.plugin.gcp.bigquery.sqlengine.util.BigQuerySQLEngineUtils;
+import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.lang.reflect.Type;
```
fernst: Why was this import removed?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryReadDataset.java
```diff
@@ -75,7 +79,7 @@ public class BigQueryReadDataset implements SQLDataset, BigQuerySQLDataset {
   public static final String SQL_INPUT_CONFIG = "config";
   public static final String SQL_INPUT_FIELDS = "fields";
   public static final String SQL_INPUT_SCHEMA = "schema";
-  private static final Type LIST_OF_STRINGS_TYPE = new TypeToken<ArrayList<String>>() {
+  private static final java.lang.reflect.Type LIST_OF_STRINGS_TYPE = new TypeToken<ArrayList<String>>() {
```
fernst, line 82: I don't think  this fully qualified import is needed. There are no conflicting class names.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryReadDataset.java
```diff
@@ -241,17 +250,45 @@ private SQLReadResult readInternal(SQLReadRequest readRequest,
...
+      condition.append(
+        BigQueryUtil.generateTimePartitionCondition(tableDefinition, partitionFromDate, partitionToDate));
+    }
+
+    //If filter is present add it.
+    if (!Strings.isNullOrEmpty(filter)) {
+      if (condition.length() == 0) {
+        condition.append(filter);
+      } else {
+        condition.append(" and (").append(filter).append(")");
```
fernst, line 281: nit: `WHERE` is uppercase, `AND` should be as well.

src/test/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryReadDatasetTest.java
```diff
@@ -0,0 +1,130 @@
...
+                                                         sourceTableId,
+                                                         destTableId,
+                                                         fieldList,
+                                                         filter,
+                                                         "2000-01-01", "2000-01-01")
+      .build().getQuery();
+
+    String partitionFilter = "TIMESTAMP(`_PARTITIONTIME`) >= TIMESTAMP(\"2000-01-01\") and " +
+      "TIMESTAMP(`_PARTITIONTIME`) < TIMESTAMP(\"2000-01-01\")";
+    expectedQuery = String.format("SELECT %s FROM `%s.%s.%s` WHERE %s and (%s)", String.join(",", fieldList),
```
fernst, line 125: nit: The expected query should be a fully defined string, not something that is built using code. This ensures that a bug doesn't also propagate to the test itself.

fernst: You can use the debugger to print the `expectedQuery` value and replace the string here.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -335,8 +343,26 @@ private void validatePartitionProperties(FailureCollector collector) {
...
-  private void setInputFormat(BatchSourceContext context) {
+  private void setInputFormat(BatchSourceContext context,
+                              Schema configuredSchema) {
+    // Set input for Spark
     context.setInput(Input.of(config.referenceName, new BigQueryInputFormatProvider(configuration)));
+
+    // Add output for SQL Engine Direct read
+    ImmutableMap.Builder<String, String> arguments = new ImmutableMap.Builder<>();
+
+    List<String> fieldNames = configuredSchema.getFields().stream().map(f -> f.getName()).collect(Collectors.toList());
```
fernst, line 359: We had to make a change on the BQ sink to only configure the Pushdown sink when a schema is configured in https://github.com/data-integrations/google-cloud/pull/1120/

Can you add the same here? Essentially, if `configuredSchema` is null, don't initialize the pushdown sink. Realistically, I don't see how pushdown execution would even work if the schema is not known at runtime.

PR #285, opened 2020-08-13
PLUGIN-100 added validation to pubsub source for subscription name

bajram-adapt:
Jira PLUGIN-100

PR #1084, opened 2022-07-21
Add Sampling Options for BigQuery

ellliao:
# Add Sampling Options for BigQuery

## Description
Added `random` and `proportionate` sampling options to BigQuery connections in Wrangler, checking for `sampleType` in sample properties to determine which option to use (proportionate sampling requires specifying a `strata` column to work).

Related PRs:
- [cdap #14495](https://github.com/cdapio/cdap/pull/14495)
- [database-plugins #285](https://github.com/data-integrations/database-plugins/pull/285)
- [wrangler #573](https://github.com/data-integrations/wrangler/pull/573)
- [cdap-ui #603](https://github.com/cdapio/cdap-ui/pull/603)

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: none

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
@@ -18,6 +18,7 @@ package io.cdap.plugin.gcp.bigquery.connector;
 
 import com.google.api.gax.paging.Page;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.RetryOption;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Dataset;
@@ -49,6 +50,7 @@ import io.cdap.cdap.etl.api.connector.ConnectorSpecRequest;
 import io.cdap.cdap.etl.api.connector.DirectConnector;
 import io.cdap.cdap.etl.api.connector.PluginSpec;
 import io.cdap.cdap.etl.api.connector.SampleRequest;
+import io.cdap.cdap.etl.api.connector.SampleType;
 import io.cdap.cdap.etl.api.engine.sql.BatchSQLEngine;
 import io.cdap.cdap.etl.api.validation.ValidationException;
 import io.cdap.plugin.common.ConfigUtil;
@@ -62,6 +64,7 @@ import io.cdap.plugin.gcp.bigquery.sqlengine.BigQuerySQLEngine;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryDataParser;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import org.threeten.bp.Duration;
 
 import java.io.IOException;
 import java.util.HashMap;
@@ -95,8 +98,13 @@ public final class BigQueryConnector implements DirectConnector {
       throw new IllegalArgumentException("Path should contain both dataset and table name.");
     }
     String dataset = path.getDataset();
-    return getTableData(getBigQuery(config.getProject()), config.getDatasetProject(), dataset, table,
-      sampleRequest.getLimit());
+    String query = getTableQuery(String.format("`%s.%s.%s`", config.getDatasetProject(), dataset, table),
+                                 sampleRequest.getLimit(),
+                                 SampleType.fromString(sampleRequest.getProperties().get("sampleType")),
+                                 sampleRequest.getProperties().get("strata"),
+                                 UUID.randomUUID().toString().replace("-", "_"));
+    String id = UUID.randomUUID().toString();
+    return getQueryResult(waitForJob(getBigQuery(config.getProject()), query, sampleRequest.getTimeoutMs(), id), id);
   }
 
   @Override
@@ -117,7 +125,7 @@ public final class BigQueryConnector implements DirectConnector {
           GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
       } catch (Exception e) {
         failureCollector.addFailure(String.format("Service account key provided is not valid: %s", e.getMessage()),
-          "Please provide a valid service account key.");
+                                    "Please provide a valid service account key.");
       }
     }
     // if either project or credentials cannot be loaded , no need to continue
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -233,19 +242,68 @@ private BigQuery getBigQuery(String project) throws IOException {
...
-      String.format("SELECT * FROM `%s.%s.%s` LIMIT %d", datasetProject, dataset, table, limit);
+    if (sampleType == null) {
+      sampleType = "first";
+    }
+    String tableName = String.format("`%s.%s.%s`", datasetProject, dataset, table);
+    String query;
+    switch (SampleType.fromString(sampleType)) {
+      case RANDOM:
+        query = String.format("WITH table AS (\n" +
+                                "  SELECT *, RAND() AS r\n" +
```
fernst: R should rename column `r` ro something like `r_<some_uuid>` to minimize the chance of collisions.

```java
String randomColumnName = "r_" + UUID.randomUUID().toString().replace("-", "");
```

does the trick

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -233,19 +242,68 @@ private BigQuery getBigQuery(String project) throws IOException {
...
+                                "LIMIT %d", tableName, limit, tableName, limit);
+        break;
+      case STRATIFIED:
+        if (strata == null) {
+          throw new IllegalArgumentException("No strata column given.");
+        }
+        query = String.format("SELECT * EXCEPT (`seqnum`, `cnt`)\n" +
+                        "FROM (\n" +
+                        "SELECT *, row_number() OVER (ORDER BY %s, RAND()) AS seqnum,\n" +
+                        "COUNT(*) OVER () as cnt,\n" +
```
fernst: Same advice here as before. 

```java
String auxColumnIdentifier = UUID.randomUUID().toString().replace("-", "")
String seqNumColumn = "seqnum_" + auxColumnIdentifier;
String cntColumn = "cnt_" + auxColumnIdentifierl
```

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -233,19 +242,68 @@ private BigQuery getBigQuery(String project) throws IOException {
...
+                        "ORDER BY %s\n" +
+                        "LIMIT %d",
+                strata, tableName, tableName, limit, strata, limit);
+        break;
+      default:
+        query =
+          String.format("SELECT * FROM %s LIMIT %d", tableName, limit);
+    }
+
     QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
```
fernst, line 300: I see a `setTimeoutMs` method in the `QueryJobConfiguration.Builder` class. 

https://javadoc.io/static/com.google.cloud/google-cloud-bigquery/1.137.1/com/google/cloud/bigquery/QueryJobConfiguration.Builder.html#setJobTimeoutMs-java.lang.Long-

Take a look at this method and see if it might be useful in this case.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
```diff
@@ -233,19 +242,68 @@ private BigQuery getBigQuery(String project) throws IOException {
...
+
     // Wait for the job to finish
     try {
-      queryJob = queryJob.waitFor();
+      if (timeoutMs == null) {
+        queryJob = queryJob.waitFor();
+      } else {
+        // We use the Future object to set a time limit on the query
+        Job tempQueryJob = queryJob;
+        Future<Job> future = Executors.newCachedThreadPool().submit(() -> tempQueryJob.waitFor());
```
fernst: You may not need to set up your own executor to do this. The` waitFor()` method in the QueryJob class already has logic to wait for the job for a certain period of time.

See https://javadoc.io/static/com.google.cloud/google-cloud-bigquery/1.137.1/com/google/cloud/bigquery/Job.html#waitFor-com.google.cloud.RetryOption...- for information on how to set the timeout for this method.

fernst (commented):
Left some comments/suggestions

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
fernst: This method returns either:

1. `null` if the job has been deleted. This should be treated as an error.
2. An instance of `Job` containing the latest information available for this job. You can use this to check:
    1. if the job has has an error (and display this), 
    2. if the job is still running (in which case you have to cancel it and notify of the timeout) 
    3. if the job succeeded (in which case you handle the results)

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
adinilfeld: I gave this method a try at first, but it turns out that it doesn't guarantee a timeout--in the documentation it says "BigQuery _may_ attempt to terminate the job." And when I was playing around with it, it seemed like most of the time it ran significantly longer than the inputted timeout, which would be unexpected to the user.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
ellliao: Fixed, thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
ellliao: Fixed to have same behavior as in other DB connectors that have implemented UUID.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
fernst: I see, then disregard my comment.

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnector.java
adinilfeld: Great point -- I remember seeing this before, but I can't remember why I didn't use it. Maybe I thought it was for network timeouts? Anyway, this simplified the code a lot!

fernst (commented):
Code looks good, is it possible to add a few tests for this code?

fernst (approved):
Perfect

sechegaray (approved):

PR #1090, opened 2022-07-25
Updated BigQuery table name validation to work with all valid characters

adinilfeld:
Previously, BigQuery table names could only contain letters, numbers, and underscores. The valid character set has since been expanded ([reference](https://cloud.google.com/bigquery/docs/tables#table_naming)), so certain valid BigQuery tables currently cannot be imported into CDAP (such as if their name contains a hyphen).

This PR fixes this issue by updating the regex used to validate table names. I also added relevant unit tests.

Fixes Jira issue [PLUGIN-1325](https://cdap.atlassian.net/browse/PLUGIN-1325).

[PR #172](https://github.com/data-integrations/bigquery-delta-plugins/pull/172) in `bigquery-delta-plugins` is related.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.gcp.bigquery.connector;
 
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 /**
@@ -33,7 +34,11 @@ public class BigQueryPath {
   private String dataset;
   private String table;
   private static final int NAME_MAX_LENGTH = 1024;
-  private static final String VALID_NAME_REGEX = "[\\w]+";
+  // Valid BigQuery dataset names can contain only letters, numbers, and underscores.
+  // See here: https://cloud.google.com/bigquery/docs/datasets#dataset-naming
+  private static final Pattern VALID_DATASET_NAME_REGEX = Pattern.compile("[\\w]+");
+  // Valid BigQuery table names are defined here: https://cloud.google.com/bigquery/docs/tables#table_naming
+  private static final Pattern VALID_TABLE_NAME_REGEX = Pattern.compile("[\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}\\p{Zs}]+");
 
   public BigQueryPath(String path) {
     parsePath(path);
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -33,7 +33,11 @@ public class BigQueryPath {
   private String dataset;
   private String table;
   private static final int NAME_MAX_LENGTH = 1024;
-  private static final String VALID_NAME_REGEX = "[\\w]+";
+  // Valid BigQuery dataset names can contain only letters, numbers, and underscores.
+  // See here: https://cloud.google.com/bigquery/docs/datasets#dataset-naming
+  private static final String VALID_DATASET_NAME_REGEX = "[\\w]+";
```
chtyim: All regex should be of type `Pattern` so that we have it compiled and reused.

```
Pattern VALID_DATESET_NAME = Pattern.compile("\\w+");
```

adinilfeld: Fixed! I also made the equivalent changes in SpannerPath.java, and in my bigquery-delta-plugins PR

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -88,9 +99,14 @@ private void validateName(String property, String name) {
       throw new IllegalArgumentException(
         String.format("%s is invalid, it should contain at most %d characters.", property, NAME_MAX_LENGTH));
     }
-    if (!name.matches(VALID_NAME_REGEX)) {
+    if (property.equals("Dataset") && !VALID_DATASET_NAME_REGEX.matcher(name).matches()) {
```
chtyim: Where the `"Dataset"` comes from? Is it some hardcoded default? Same for the `"Table"` below

adinilfeld: Yeah, it's hardcoded and used whenever this `validateName` function is called.
For example, in [DBPath:79](https://github.com/cdapio/hydrator-plugins/blob/2ae29b018990e419f726adcf8df50a75759be3f5/hydrator-common/src/main/java/io/cdap/plugin/common/db/DBPath.java#L79) and [DBSpecificPath:90](https://github.com/data-integrations/database-plugins/blob/ee8eda9ee2631073eecea72899bc95d72f0b8637/database-commons/src/main/java/io/cdap/plugin/db/connector/DBSpecificPath.java#L90)

fernst (approved):
LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
fernst: nit: if "Dataset" is a constant, it should be stored in a static final variable

src/main/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryPath.java
```diff
@@ -88,9 +99,14 @@ private void validateName(String property, String name) {
...
       throw new IllegalArgumentException(
-        String.format("%s is invalid, it should contain only letters, numbers, and underscores.", property));
+              String.format("%s is invalid, it should contain only letters, numbers, and underscores.", property));
+    }
+    if (property.equals("Table") && !VALID_TABLE_NAME_REGEX.matcher(name).matches()) {
```
fernst: Same here. "Table" should be a static final constant

fernst (changes requested):
Noticed a small nit

fernst (approved):
LGTM

sechegaray (approved):

PR #1091, opened 2022-07-27
[PLUGIN-1291] Support GCS Delete Action in multiple Buckets

JessieYee:
Jira: https://cdap.atlassian.net/browse/PLUGIN-1291

Update GCS Delete Action, support delete file from different buckets.
Support wildcard delete (*) for file/directory
Not support wildcard character in Bucket name.
Add e2e test.

### DIFF
--- a/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
+++ b/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
@@ -151,11 +151,29 @@ public class TestSetupHooks {
     gcsSourceBucketName = createGCSBucketWithFile(PluginPropertyUtils.pluginProp("gcsCsvRangeFile"));
   }
 
+  @Before(order = 1, value = "@GCS_DELETE_MULTIPLE_BUCKETS_TEST")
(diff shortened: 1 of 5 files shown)


sumengwang (approved):

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -396,6 +413,29 @@ private static String createGCSBucketWithMultipleFiles(String folderPath) throws
     return bucketName;
   }
 
+  private static String createMultipleGCSBucketsWithMultipleFiles(String folderPath)
```
itsankit-google: can we make this method more generic and add no of buckets to be created as a param?

itsankit-google (commented):
Are we supporting deletion from multiple buckets with wildcard as well? 
If yes, can we add a test for that as well?

src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
JessieYee: yes. will edit the function and also add some wildcard delete test

itsankit-google (approved):

PR #1093, opened 2022-07-27
[CDAP-19362] make bigquery upsert and update operations to use trigger job method

itsankit-google:
Manually tested on cdap-sandbox.


fernst (approved):
LGTM

PR #1102, opened 2022-08-03
e2e tests separate build job per plugin

sawantpritam:

This PR contains :

1. Updated e2e tests run condition.
We should run all e2e tests on PR raise, for addition of new test scenarios or updates in existing scenario.

2. When all e2e tests are executed (on PR merge / manual trigger / any change in existing tests) - create separate parallel build jobs to execute all tests per plugin.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -25,7 +25,6 @@ on:
 jobs:
   build:
     runs-on: k8s-runner-e2e
-
     # We allow builds:
     # 1) When triggered manually
     # 2) When it's a merge into a branch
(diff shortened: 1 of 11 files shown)


sawantpritam:
Workflow runs:
1. Required tests
https://github.com/data-integrations/google-cloud/actions/runs/2789291303
2. All tests
https://github.com/data-integrations/google-cloud/actions/runs/2789558834

itsankit-google (commented):
Can we also separate the required tests for different runners on different jobs? Please check if it can reduce some execution time.

.github/workflows/e2e.yml
```diff
@@ -39,7 +38,37 @@ jobs:
...
+        id: filter
+        with:
+          working-directory: plugin
+          filters: |
+            e2e-test:
+              - '**/e2e-test/**'
+      - name: Set e2e-tests list
+        id: spl
+        run: |
+          if [ ${{ github.event_name }} != 'workflow_dispatch' ] && [ ${{ github.event_name }} != 'push' ] && [ ${{ steps.filter.outputs.e2e-test }} == 'false' ]; then
```
itsankit-google: suppose there is no change in e2e-tests in the PR, it will end up running every tests, right?

sawantpritam: No. `steps.filter.outputs.e2e-test` will be `false` when there is no change in e2e-tests so only required tests will be run.

sawantpritam:
Rebased PR.

itsankit-google (approved):
Please squash commits.

sawantpritam:
> Please squash commits.

Commits squashed.

PR #1103, opened 2022-08-04
modify the connector tests to pick up information from the env

albertshau:
Modify the connector tests to pick up credentials and project id
from the environment if they are not explicitly given.

### DIFF
--- a/src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
@@ -16,6 +16,16 @@
 
 package io.cdap.plugin.gcp.bigquery.connector;
 
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.DatasetId;
+import com.google.cloud.bigquery.DatasetInfo;
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.InsertAllRequest;
+import com.google.cloud.bigquery.StandardSQLTypeName;
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.TableDefinition;
+import com.google.cloud.bigquery.TableId;
+import com.google.cloud.bigquery.TableInfo;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.batch.BatchSink;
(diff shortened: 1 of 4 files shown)


albertshau:
please don't review, checking if github action tests run properly

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -35,85 +45,101 @@
...
   @BeforeClass
   public static void setupTestClass() throws Exception {
-    // Certain properties need to be configured otherwise the whole tests will be skipped.
-
-    String messageTemplate = "%s is not configured, please refer to javadoc of this class for details.";
+    testEnvironment = TestEnvironment.load();
+
+    project = testEnvironment.getProject();
+    datasetProject = testEnvironment.getProject();
+    dataset = "bq_connector_test_" + System.currentTimeMillis();
```
yaojiefeng: can we use UUID here to avoid potential race condition where two builds trigger at the same time?

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
```diff
@@ -35,85 +45,101 @@
...
+    table = "connector_test_" + System.currentTimeMillis();
```
yaojiefeng: same here

src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
```diff
@@ -64,35 +70,63 @@ public class SpannerConnectorTest {
...
-    Assume.assumeFalse(String.format(messageTemplate, "instance id"), instance == null);
+    testEnvironment = TestEnvironment.load();
 
+    project = testEnvironment.getProject();
+    serviceAccountFilePath = testEnvironment.getServiceAccountFilePath();
+    serviceAccountKey = testEnvironment.getServiceAccountContent();
 
-    database = System.getProperty("database.name");
-    Assume.assumeFalse(String.format(messageTemplate, "database name"), database == null);
+    instance = "gcp-plugins-test-" + System.currentTimeMillis();
```
yaojiefeng: same here - can we use UUID?

yaojiefeng (approved):
Minor comments, LGTM

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
albertshau: we can if we strip out the dashes (not an allowed character), though it becomes a bit more difficult visually to tell if it is an old dataset that is safe to delete or if it is currently in use. Let me see if I can use a uuid plus a timestamp.

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
albertshau: this one I can actually change to some hardcoded value this it will be in a unique database.

src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
albertshau: changed to use timestamp in seconds followed by uuid. So it's still sorted, but also unique.

src/test/java/io/cdap/plugin/gcp/bigquery/connector/BigQueryConnectorTest.java
albertshau: changed to use timestamp in seconds followed by uuid. So it's still sorted, but also unique.

PR #1105, opened 2022-08-11
Split GCP Plugins into multiple modules

fernst:


google-cloud-bigquery/pom.xml
```diff
@@ -0,0 +1,654 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright © 2022 Cask Data, Inc.
```
sechegaray, line 3: Are the "~" part of the license?

fernst: Yep https://github.com/data-integrations/google-cloud/blob/develop/pom.xml#L2

PR #1106, opened 2022-08-12
Split GCP Plugins into multiple modules

fernst:
Applied latest repository changes and cleanup

Moved E2E tests to separate submodule

Applied latest repository changes


google-cloud-bigquery/pom.xml
```diff
@@ -0,0 +1,547 @@
...
+    <version>0.21.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>io.cdap.plugin</groupId>
+  <artifactId>google-cloud-bigquery</artifactId>
+  <version>0.21.0-SNAPSHOT</version>
+  <name>Google Cloud BigQuery Plugins</name>
+  <description>BigQuery GCP Plugins</description>
+
+  <distributionManagement>
```
greeshmaswaminathan: Do these change for submodules? I think it would keep the submodule pom files simpler if the common parts like distributionManagement, repository, checkstyle, build and even the common dependencies like cdap-api is moved to the top level pom. wdyt?

fernst: This is a great point. I will consolidate some of the common dependencies in the parent pom (CDAP API modules that are in the `provided ` scope and `test` dependencies)

greeshmaswaminathan: nit: Move the <distributionManagement> section and <repositories> section also to parent pom?

greeshmaswaminathan (approved):

yaojiefeng:
How will we handle the import for pipeline json prior to 6.8.0? With this change, the pipeline json won't be able to find the correct artifact in the new instance.

fernst:
One of the follow up items to this change is to add logic on the pipeline upgrader to handle pre-6.8 GCP artifacts

fernst:
TODOs after this PR is merged:

1. Update E2E tests to install all required jars to run E2E tests. Tests are currently disabled.

2. Update CDAP Upgrade logic to map old artifact name `google-cloud` to new artifact names `google-cloud-*`

PR #1115, opened 2022-08-22
add fqn to bq plugin

mrahanjam:
This PR adds FQN to the BigQuery plugin code.

Jira: [CDAP-19520](https://cdap.atlassian.net/browse/CDAP-19520)

Field-level lineage API output:

```json
{
  "direction": "BOTH",
  "start-ts": 0,
  "end-ts": 1661275067000,
  "entityId": {
    "dataset": "bigquery:mehrdadcdfsandbox:demo:incident",
    "namespace": "default",
    "entity": "DATASET"
  },
  "fields": [ ... ]
  ],
  "incoming": [],
  "outgoing": [
    {
      "entityId": {
        "dataset": "bigquery:mehrdadcdfsandbox:demo:incident_6",
        "namespace": "default",
        "entity": "DATASET"
      },
      "fieldCount": 88,
      "relations": [ ...]
    },
    {
      "entityId": {
        "dataset": "bigquery:mehrdadcdfsandbox:demo:incident_5",
        "namespace": "default",
        "entity": "DATASET"
      },
      "fieldCount": 88,
      "relations": [ ... ]
    }
  ]
}
```

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -31,6 +31,8 @@ import io.cdap.cdap.etl.api.Emitter;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
+import io.cdap.plugin.common.Asset;
+import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryTypeSize;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
@@ -127,8 +129,8 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
    * @param tableSchema table schema
    * @param bucket bucket name
    */
-  protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, String outputName, String tableName,
-                                  @Nullable Schema tableSchema, String bucket,
+  protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, String outputName, String fqn,
+                                  String tableName, @Nullable Schema tableSchema, String bucket,
                                   FailureCollector collector) throws IOException {
     LOG.debug("Init output for table '{}' with schema: {}", tableName, tableSchema);
 
@@ -150,7 +152,8 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    BigQuerySinkUtils.recordLineage(context, outputName, tableSchema, fieldNames);
+    Asset asset = Asset.builder(outputName).setFqn(fqn).setLocation(getConfig().getLocation()).build();
+    BigQuerySinkUtils.recordLineage(context, asset, tableSchema, fieldNames);
     context.addOutput(Output.of(outputName, getOutputFormatProvider(configuration, tableName, tableSchema)));
   }
 
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
@@ -18,6 +18,7 @@ package io.cdap.plugin.gcp.bigquery.sink;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.Table;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Metadata;
 import io.cdap.cdap.api.annotation.MetadataProperty;
@@ -32,6 +33,7 @@ import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
 import io.cdap.cdap.etl.api.connector.Connector;
+import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.gcp.bigquery.connector.BigQueryConnector;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
@@ -131,10 +133,11 @@ public class BigQueryMultiSink extends AbstractBigQuerySink {
             config.isTruncateTableSet(), config.getDataset(), collector);
 
         }
-
         String outputName = String.format("%s-%s", config.getReferenceName(), tableName);
         outputName = sanitizeOutputName(outputName);
-        initOutput(context, bigQuery, outputName, tableName, tableSchema, bucket, context.getFailureCollector());
+        initOutput(context, bigQuery, outputName,
+                   BigQueryUtil.getFQN(config.getDatasetProject(), config.getDataset(), tableName),
+                   tableName, tableSchema, bucket, context.getFailureCollector());
       } catch (IOException e) {
         collector.addFailure("Invalid schema: " + e.getMessage(), null);
       }
@@ -147,9 +150,10 @@ public class BigQueryMultiSink extends AbstractBigQuerySink {
     String splitField = config.getSplitField();
     String projectName = config.getDatasetProject();
     String datasetName = config.getDataset();
-    context.addOutput(Output.of(config.getReferenceName(),
-                                new DelegatingMultiSinkOutputFormatProvider(conf, splitField, bucket,
-                                                                            projectName, datasetName)));
+    context.addOutput(Output.of(
+      config.getReferenceName(),
+      new DelegatingMultiSinkOutputFormatProvider(conf, splitField, bucket, projectName, datasetName))
+    );
   }
 
   /**
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
@@ -50,7 +50,9 @@ import io.cdap.cdap.etl.api.batch.BatchSourceContext;
 import io.cdap.cdap.etl.api.connector.Connector;
 import io.cdap.cdap.etl.api.engine.sql.SQLEngineInput;
 import io.cdap.cdap.etl.api.validation.ValidationFailure;
+import io.cdap.plugin.common.Asset;
 import io.cdap.plugin.common.LineageRecorder;
+import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.gcp.bigquery.connector.BigQueryConnector;
 import io.cdap.plugin.gcp.bigquery.sqlengine.BigQueryReadDataset;
 import io.cdap.plugin.gcp.bigquery.sqlengine.BigQuerySQLEngine;
@@ -172,7 +174,11 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
     // Both emitLineage and setOutputFormat internally try to create an external dataset if it does not already exists.
     // We call emitLineage before since it creates the dataset with schema.
     Type sourceTableType = config.getSourceTableType();
-    emitLineage(context, configuredSchema, sourceTableType, config.getTable());
+    Asset asset = Asset.builder(config.getReferenceName())
+      .setFqn(BigQueryUtil.getFQN(config.getDatasetProject(), config.getDataset(), config.getTable()))
+      .setLocation(dataset.getLocation())
+      .build();
+    emitLineage(context, configuredSchema, sourceTableType, config.getTable(), asset);
     setInputFormat(context, configuredSchema);
   }
 
@@ -346,7 +352,7 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
   private void setInputFormat(BatchSourceContext context,
                               Schema configuredSchema) {
     // Set input for Spark
-    context.setInput(Input.of(config.referenceName, new BigQueryInputFormatProvider(configuration)));
+    context.setInput(Input.of(config.getReferenceName(), new BigQueryInputFormatProvider(configuration)));
 
     // Add output for SQL Engine Direct read
     ImmutableMap.Builder<String, String> arguments = new ImmutableMap.Builder<>();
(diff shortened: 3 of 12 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -83,6 +84,8 @@ public final class BigQuerySource extends BatchSource<LongWritable, GenericData.
   private final BigQueryAvroToStructuredTransformer transformer = new BigQueryAvroToStructuredTransformer();
   // UUID for the run. Will be used as bucket name if bucket is not provided.
   private String bucketPath;
+  private String fqn = String.join(":", "bigquery",
```
fernst: I don't think the Config object is initialized at this point. I think you should initialize this fqn during the `prepareRun` phase of the pipeline

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -717,4 +717,17 @@ public static void deleteTemporaryDirectory(Configuration configuration, String
...
+
+  /**
+   * Get fqn for a BQ table.
+   *
+   * @param datasetProject Name of the BQ project
+   * @param datasetName    Name of the BQ dataset
+   * @param tableName      Name of the BQ table
+   * @return String fqn
+   */
+  public static String getFqn(String datasetProject, String datasetName, String tableName) {
```
snehalakshmisha: since fqn is an abbreviation it should be `getFQN()`

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
```diff
@@ -77,12 +77,12 @@ public void testBigQuerySinkInvalidConfig() {
...
                                            Schema.Field.of("id", Schema.of(Schema.Type.LONG)));
 
-    BigQuerySinkConfig config = new BigQuerySinkConfig("reference!!", "ds", "tb", "buck3t$$", invalidSchema.toString(),
+    BigQuerySinkConfig config = new BigQuerySinkConfig("ds", "tb", "buck3t$$", invalidSchema.toString(),
                                                        "INTEGER", 0L, 100L, 10L, "200000");
     MockFailureCollector collector = new MockFailureCollector("bqsink");
     config.validate(collector);
     List<ValidationFailure> failures = collector.getValidationFailures();
-    Assert.assertEquals(3, failures.size());
+    Assert.assertEquals(2, failures.size());
```
snehalakshmisha: Why are there only 2 failures now?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -357,18 +357,17 @@ protected void initOutput(BatchSinkContext context, BigQuery bigQuery, String ou
...
+    BigQuerySinkUtils.configureOutput(configuration, datasetId, tableName, temporaryGcsPath, fields);
     // Both emitLineage and setOutputFormat internally try to create an external dataset if it does not already exist.
     // We call emitLineage before since it creates the dataset with schema which is used.
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    BigQuerySinkUtils.recordLineage(context, outputName, tableSchema, fieldNames);
+    String fqn = BigQueryUtil.getFqn(datasetProject, dataset, config.getTable());
+    String location = bigQuery.getDataset(datasetId).getLocation();
+    BigQuerySinkUtils.recordLineage(context, new io.cdap.plugin.common.Asset(fqn, location), tableSchema, fieldNames);
```
snehalakshmisha: import the package for Asset

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -717,4 +717,17 @@ public static void deleteTemporaryDirectory(Configuration configuration, String
       LOG.debug("Deleted temporary directory '{}'", path);
     }
   }
+
+  /**
+   * Get fqn for a BQ table.
```
snehalakshmisha: nit: expand fqn to fully-qualified name
and mention the format as well

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
mrahanjam: Asset is already imported from dataplex api:
https://github.com/data-integrations/google-cloud/blob/7abc79d048d2e6ff22a7aa6b52d81188d4c5d823/src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java#L29

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
mrahanjam: one was in regards to reference name

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -150,7 +152,12 @@ protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, Str
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    BigQuerySinkUtils.recordLineage(context, outputName, tableSchema, fieldNames);
+    String fqn = BigQueryUtil.getFQN(getConfig().getDatasetProject(),
+                                               getConfig().getDataset(), tableName);
+    Asset.Builder assetBuilder = Asset.builder(outputName);
```
snehalakshmisha: nit: let's build all fields of the asset in one line instead of splitting it between here and when we pass it as an arg to recordLineage()?
Here and in other places

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -147,9 +153,14 @@ protected void configureSchemalessOutput(BatchSinkContext context,
     String splitField = config.getSplitField();
     String projectName = config.getDatasetProject();
     String datasetName = config.getDataset();
-    context.addOutput(Output.of(config.getReferenceName(),
-                                new DelegatingMultiSinkOutputFormatProvider(conf, splitField, bucket,
-                                                                            projectName, datasetName)));
+    String fqn = BigQueryUtil.getFQN(config.getDatasetProject(),
+                                     config.getDataset(), config.getTable());
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+      ? ReferenceNames.normalizeFqn(fqn)
```
mrahanjam: There's a possibility of having clashes using this normalizing method defined [here](https://github.com/cdapio/hydrator-plugins/blob/62cee359718523171d522a0327aa7713d0a257d7/hydrator-common/src/main/java/io/cdap/plugin/common/ReferenceNames.java#L65).

We could have a case of `a://b` and `a:b` that both will be normalized to `a.b`. Wdyt @albertshau @yaojiefeng ?

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -405,18 +406,22 @@ protected void initOutput(BatchSinkContext context, BigQuery bigQuery, String ou
...
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    BigQuerySinkUtils.recordLineage(context, outputName, tableSchema, fieldNames);
+    String fqn = BigQueryUtil.getFQN(datasetProject, dataset, config.getTable());
+    String location = bigQuery.getDataset(datasetId).getLocation();
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+      ? ReferenceNames.normalizeFqn(fqn)
+      : config.getReferenceName();
+    io.cdap.plugin.common.Asset lineageAsset = io.cdap.plugin.common.Asset.builder(referenceName)
```
snehalakshmisha, line 422: Let's import the package at the beginning of the file

mrahanjam: There's already an `Asset` package imported [here](https://github.com/data-integrations/google-cloud/blob/f5f29fe038f9acefebc924ddd4493d593df83788/src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java#L30)

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
albertshau: that's fine. It is expected that if no reference name is given, the auto-generated name can have conflicts, there's no way to avoid it.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryMultiSink.java
```diff
@@ -131,10 +133,14 @@ protected void configureOutputSchemas(BatchSinkContext context,
             config.isTruncateTableSet(), config.getDataset(), collector);
 
         }
-
-        String outputName = String.format("%s-%s", config.getReferenceName(), tableName);
+        String fqn = BigQueryUtil.getFQN(config.getDatasetProject(),
+                                         config.getDataset(), config.getTable());
+        String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+          ? ReferenceNames.normalizeFqn(fqn)
+          : config.getReferenceName();
```
albertshau: looks like this logic is duplicated in a bunch of places. It would be better to have it in just a single place. I think it's better to have getFQN() and getReferenceName() in the config class rather than in the source/sink.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -759,4 +759,17 @@ public static String generateTimePartitionCondition(StandardTableDefinition tabl
...
+
+  /**
+   * Get fully-qualified name (FQN) for a BQ table (FQN format: bigquery:{projectId}.{datasetId}.{tableId}).
+   *
+   * @param datasetProject Name of the BQ project
+   * @param datasetName    Name of the BQ dataset
+   * @param tableName      Name of the BQ table
+   * @return String fqn
+   */
+  public static String getFQN(String datasetProject, String datasetName, String tableName) {
```
albertshau, line 771: We should try to move this logic into the config class(es) if possible.

In general, broad util classes like this one are a pattern we would like to avoid. They tend to gather a bunch of unrelated code and encourage poor encapsulation.

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
```diff
@@ -103,7 +103,7 @@ public void testBigQueryTimePartitionConfig() {
 
     BigQuerySinkConfig config =
       new BigQuerySinkConfig("44", "ds", "tb", "bucket", schema.toString(),
-                             "TIME", 0L, 100L, 10L, null);
+            "TIME", 0L, 100L, 10L, null);
```
albertshau: shouldn't change indentation here

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
```diff
@@ -259,7 +261,17 @@ private void configureBigQuerySource() {
 
   private void emitLineage(BatchSourceContext context, Schema schema, TableDefinition.Type sourceTableType,
                            String table) {
-    LineageRecorder lineageRecorder = new LineageRecorder(context, config.getReferenceName());
+    // entity.getDataPath() will be of format:
+    // 'projects/datasetProjectName/datasets/datasetName/tables/tableName'
+    String[] dataPathValues = entity.getDataPath().split("/");
```
albertshau: I see there is a getEntityValuesFromDataPathForBQEntities() method that sets the dataset and table variables. Can that be re-used or is there some reason that we need to parse this again? I'm also not clear why table is passed in as an argument even though it is in the entity, is it possible for the table that is passed into this method to ever differ from what is parsed from the entity?

FYI, the existing code isn't structured very well, we shouldn't be using shared state like this, as it's not clear when everything gets assigned a value. It would have been much better to parse it once into some id class that is composed of a project/dataset/table and pass that around everywhere. You don't need to do any refactoring around this in the PR, but just pointing out that this is a pattern to avoid.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
mrahanjam: The main reason I added it as a util method was to use it inside dataplex as well.

src/main/java/io/cdap/plugin/gcp/dataplex/source/DataplexBatchSource.java
mrahanjam: Ack. Must've missed `getEntityValuesFromDataPathForBQEntities()`.

src/main/java/io/cdap/plugin/gcp/dataplex/sink/DataplexBatchSink.java
```diff
@@ -405,18 +406,22 @@ protected void initOutput(BatchSinkContext context, BigQuery bigQuery, String ou
...
+    BigQuerySinkUtils.configureOutput(configuration, datasetId, tableName, temporaryGcsPath, fields);
     // Both emitLineage and setOutputFormat internally try to create an external dataset if it does not already exist.
     // We call emitLineage before since it creates the dataset with schema which is used.
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    BigQuerySinkUtils.recordLineage(context, outputName, tableSchema, fieldNames);
+    String fqn = BigQueryUtil.getFQN(datasetProject, dataset, config.getTable());
+    String location = bigQuery.getDataset(datasetId).getLocation();
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
```
mrahanjam, line 419: For dataplex (Source & Sink), I don't think I can move this logic to the config class as the config class does not readily have access to dataset value for the BQ table. It's only used once in sink and once in source

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySource.java
```diff
@@ -252,8 +258,9 @@ private void validateConfiguredSchema(Schema configuredSchema, FailureCollector
       } catch (IllegalArgumentException e) {
         // this means that the field is not present in BigQuery table.
         collector.addFailure(
-          String.format("Field '%s' is not present in table '%s:%s.%s'.", field.getName(), project, dataset, tableName),
-          String.format("Remove field '%s' from the output schema.", field.getName()))
+            String.format("Field '%s' is not present in table '%s:%s.%s'.",
```
albertshau: don't need these indentation changes

albertshau (approved):
minor style comment, otherwise lgtm

PR #1120, opened 2022-09-01
[PLUGIN-1367] Fixed null pointer exception when sink schema is not set and SQL Engine Output is initialized.

fernst:


src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkTest.java
```diff
@@ -88,18 +91,18 @@ public void testBigQuerySinkInvalidConfig() {
...
-            Schema.Field.of("price", Schema.of(Schema.Type.DOUBLE)),
-            Schema.Field.of("dt", Schema.nullableOf(Schema.of(Schema.LogicalType.DATE))),
-            Schema.Field.of("bytedata", Schema.of(Schema.Type.BYTES)),
-            Schema.Field.of("timestamp",
-                    Schema.nullableOf(Schema.of(Schema.LogicalType.TIMESTAMP_MICROS))));
+                                    Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
+                                    Schema.Field.of("name", Schema.of(Schema.Type.STRING)),
+                                    Schema.Field.of("price", Schema.of(Schema.Type.DOUBLE)),
+                                    Schema.Field.of("dt", Schema.nullableOf(Schema.of(Schema.LogicalType.DATE))),
+                                    Schema.Field.of("bytedata", Schema.of(Schema.Type.BYTES)),
```
avirmani: can you fix the tabs here, pls

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -161,6 +161,12 @@ void initSQLEngineOutput(BatchSinkContext context,
                            String tableName,
                            @Nullable Schema tableSchema,
                            FailureCollector collector) {
+    // Sink Pushdown is not supported if the sink schema is not defined.
+    if (tableSchema == null) {
+      LOG.debug("BigQuery SQL Engine Output was not initialized. Schema was empty.");
+      return;
```
avirmani, line 167: so, this just means that bqexecute will not be pushed down for schemaless and will work in the old way, correct?

fernst: BQ Execute is not a stage that can be "pushed down" in the traditional pipeline sense. BQ Execute operates on tables in BigQuery after the sink operation is completed. 

Schemaless pipelines still work in Spark as they did before version 0.20 of the plugin / CDAP 6.7

avirmani (approved):

PR #1131, opened 2022-09-07
Bump to 0.18.7 and remove Dataplex

fernst:


sechegaray (approved):

pom.xml
```diff
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.18.6</version>
+  <version>0.18.7</version>
```
rmstar, line 23: we just bumped up version to 0.18.6 yesterday - https://github.com/data-integrations/google-cloud/commit/751f43c64a80615d306f57660f003ac8a99dbbe8
Are we bumping up to 18.7 because 0.18.6 was already released?

fernst: Yes, 0.18.6 is already in mavencentral

rmstar (approved):

PR #1136, opened 2022-09-12
PLUGIN-1003:support reserved words as column names in big query

akritig187:
_**PLUGIN-1003: Support reserved words as column names in BigQuery**_
**Description:**
Currently using BigQuery reserved word  as a column name does not work and results in SQL errors.
E.g. in queries like `insert into a(DESC, field2, field3) values(1,2,3)` won't work since DESC is a reserved word.

**Solution**
This can be solved by quoting field names with back quotes. 
Fortunately unlike most DBs, BigQuery quoted identifiers are not case-sensitive, but case-retentive, so `Aaa` and `AAA` columns are considered same names, so we don't need to introduce special mode.

**Testing**
The code was thoroughly tested in all possible modes (INSERT/UPDATE/UPSERT, new vs existing table, JSON vs AVRO values, etc).

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -82,7 +82,7 @@ public final class BigQuerySinkUtils {
   private static final Type LIST_OF_FIELD_TYPE = new TypeToken<ArrayList<Field>>() { }.getType();
 
   // Fields used to build update/upsert queries
-  private static final String CRITERIA_TEMPLATE = "T.%s = S.%s";
+  private static final String CRITERIA_TEMPLATE = "T.`%s` = S.`%s`";
   private static final String SOURCE_DATA_QUERY = "(SELECT * FROM (SELECT row_number() OVER (PARTITION BY %s%s) " +
     "as rowid, * FROM %s) where rowid = 1)";
   private static final String UPDATE_QUERY = "UPDATE %s T SET %s FROM %s S WHERE %s";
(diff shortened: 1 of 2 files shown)


itsankit-google (commented):
Please add unit tests.

src/test/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtilsTest.java
```diff
@@ -522,4 +524,74 @@ public void testGetRelaxedTableFieldsMarksFieldsNullable() {
     Assert.assertEquals("c", result.get(2).getName());
     Assert.assertEquals(Field.Mode.REPEATED, result.get(2).getMode());
   }
+
+
+  @Test
+  public void testGenerateUpdateUpsertQueryUpdate() {
+    Operation operation = Operation.UPDATE;
+
+    TableId sourceTableId = TableId.of("akritigiri-gcp-project0", "test",
```
itsankit-google: nit: Is this an actual project or just some value you made up? Generally, we don't put references to actual resources in the code.

akritig187: An actual project

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -585,13 +585,19 @@ public static String generateUpdateUpsertQuery(Operation operation,
                                                        formatPartitionFilter(partitionFilter), criteria) : criteria;
     String fieldsForUpdate = tableFieldsList.stream().filter(s -> !tableKeyList.contains(s))
       .map(s -> String.format(CRITERIA_TEMPLATE, s, s)).collect(Collectors.joining(", "));
+
+    orderedByList.replaceAll(s -> "`" + Arrays.stream(s.split(" ")).map(String::trim)
```
fernst: If I understand correctly, you need to split a list containing thigs like:

["col1 ASC", "col2 DESC", "col3 ASC"] and add quotes to the column name.

If so, you can do this by using:

```
orderedByList = orderedByList.stream()
  .map(s -> s.trim()) // for safety
  .map(s -> {
     return "`" + s.split(" ")[0] + "` " + s.split(" ", 2)[1]; //the second function only splits up to the first space and then the rest of the string is left untouched
  })
```

akritig187: Done

fernst (approved):

PR #1137, opened 2022-09-12
[PLUGIN-871] Adding support for BQ source pushdown in BQ Plugin

sahusanket:


src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -717,4 +722,41 @@ public static void deleteTemporaryDirectory(Configuration configuration, String
       LOG.debug("Deleted temporary directory '{}'", path);
     }
   }
+
+  public static String generateTimePartitionCondition(StandardTableDefinition tableDefinition,
```
fernst, line 726: It would be nice to add tests for this. Can you log a JIRA for this? No need to do it right now, but it would be good to cover this piece of code with tests.

fernst (approved):

src/test/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryReadDatasetTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    generatedQuery = bigQueryReadDataset.getQueryBuilder(sourceTable,
+                                                         sourceTableId,
+                                                         destTableId,
+                                                         fieldList,
+                                                         filter,
+                                                         "2000-01-01", "2000-01-01")
+      .build().getQuery();
+
+    String partitionFilter = "TIMESTAMP(`_PARTITIONTIME`) >= TIMESTAMP(\"2000-01-01\") and " +
```
sahusanket, line 123: @fernst 
The time partition logic is somewhat Unit tested here.

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
sahusanket: This logic is covered under `generateQuery` Unit test. Have tagged you, PTAL

PR #1138, opened 2022-09-12
Set a 15 minute timeout for GCS client operations

fernst:


src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -255,6 +257,18 @@ public static Storage getStorage(String project, @Nullable Credentials credentia
...
+
+    // Add retry settings with up to 1 hour delay.
+    builder.setRetrySettings(RetrySettings.newBuilder()
+                               .setInitialRetryDelay(Duration.ofSeconds(1))
+                               .setMaxRetryDelay(Duration.ofMinutes(1))
+                               .setRetryDelayMultiplier(2.0)
+                               .setInitialRpcTimeout(Duration.ofSeconds(50))
+                               .setMaxRpcTimeout(Duration.ofSeconds(50))
+                               .setMaxAttempts(60)
+                               .setTotalTimeout(Duration.ofHours(1))
```
tivv: That's too much, especially when unconfigurable

fernst: I've updated the max delay to 15 minutes after discussing with @chtyim @albertshau and @masoud-io

albertshau (approved):

PR #1146, opened 2022-09-20
CDAP-19623: add fqn support to gcs plugins

mrahanjam:
This PR adds FQN support to the GCS plugins.

Jira: [CDAP-19623](https://cdap.atlassian.net/browse/CDAP-19623)

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
@@ -40,10 +40,12 @@ import io.cdap.cdap.etl.api.batch.BatchSink;
 import io.cdap.cdap.etl.api.batch.BatchSinkContext;
 import io.cdap.cdap.etl.api.connector.Connector;
 import io.cdap.cdap.etl.api.validation.ValidatingOutputFormat;
+import io.cdap.plugin.common.Asset;
 import io.cdap.plugin.common.ConfigUtil;
 import io.cdap.plugin.common.Constants;
 import io.cdap.plugin.common.IdUtils;
 import io.cdap.plugin.common.LineageRecorder;
+import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.format.FileFormat;
 import io.cdap.plugin.format.plugin.AbstractFileSink;
 import io.cdap.plugin.format.plugin.FileSinkProperties;
@@ -84,6 +86,7 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
 
   private final GCSBatchSinkConfig config;
   private String outputPath;
+  private Asset asset;
 
   public GCSBatchSink(GCSBatchSinkConfig config) {
     super(config);
@@ -109,7 +112,6 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    super.prepareRun(context);
     FailureCollector collector = context.getFailureCollector();
     CryptoKeyName cmekKeyName = CmekUtils.getCmekKey(config.cmekKey, context.getArguments().asMap(), collector);
     collector.getOrThrowException();
@@ -125,6 +127,7 @@ public class GCSBatchSink extends AbstractFileSink<GCSBatchSink.GCSBatchSinkConf
       null : GCPUtils.loadServiceAccountCredentials(config.connection.getServiceAccount(), isServiceAccountFilePath);
     Storage storage = GCPUtils.getStorage(config.connection.getProject(), credentials);
     Bucket bucket;
+    String location;
     try {
       bucket = storage.get(config.getBucket());
     } catch (StorageException e) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -109,7 +112,6 @@ public ValidatingOutputFormat getOutputFormatForRun(BatchSinkContext context) th
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    super.prepareRun(context);
```
yaojiefeng, line 112: Better to call super.prepareRun() first

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -136,6 +138,14 @@ public void prepareRun(BatchSinkContext context) throws Exception {
       GCPUtils.createBucket(storage, config.getBucket(), config.getLocation(), cmekKeyName);
     }
     this.outputPath = getOutputDir(context);
+
+    // create asset for lineage
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+      ? ReferenceNames.normalizeFqn(config.getPath())
+      : config.getReferenceName();
+    asset = Asset.builder(referenceName)
+      .setFqn(config.getPath()).setLocation(config.getLocation()).build();
```
yaojiefeng: Shouldn't use this location to construct the asset. This location is only used if the bucket does not exist, and will get ignored if it already exists. You should check existence of the bucket and get the correct location like you do for the GCS source

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -248,6 +277,13 @@ public String getEncryptedMetadataSuffix() {
         DEFAULT_ENCRYPTED_METADATA_SUFFIX : encryptedMetadataSuffix;
     }
 
+    /**
+     * Extract bucket name out of path with the format: gs://<bucket>/path/to/output
+     */
+    public String getBucketName() {
+      return this.getPath().replace("gs://", "").split("/")[0];
```
yaojiefeng: We have `GCSPath` class you can use to extract the bucket out

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
mrahanjam: If I call it first, i'm afraid it will use the default `LineageRecorder` instead of the overridden one.

fernst: Why? The parent class should also be calling the `getLineageRecorder` method. Because this class overrides this method, this will call the `getLineageRecorder` method in this class.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -78,6 +77,31 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+          + "Ensure you entered the correct bucket path and have permissions for it.", e);
+    }
+
+    // create asset for lineage
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+      ? ReferenceNames.normalizeFqn(config.getPath())
+      : config.getReferenceName();
+    asset = Asset.builder(referenceName)
+      .setFqn(config.getPath()).setLocation(location).build();
+    super.prepareRun(context);
```
fernst, line 105: Should the `super.prepareRun()` method be called first?

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
fernst: +1, please use the GCSPath class for this.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
mrahanjam: Don't think so. I wonder if it's the same case for S3 and FTP as well. (need to test them). if I call it first, it'll throw a `NullPointerException` at `io.cdap.plugin.common.LineageRecorder.createExternalDataset(LineageRecorder.java:68)`

fernst: Can you link to the line that causes the NPE? I want to understand that causes the NPE.

mrahanjam: @fernst I took a closer look. What happens is that if super is invoked at the top, then `asset` is null when `prepareRun` calls the `getLineageRecorder()` method, hence the NPE thrown here:

https://github.com/cdapio/hydrator-plugins/blob/fa251edb0389084a2e140ab449a79f32f5e1ef39/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java#L186

fernst: I see, thanks for the clarification.

src/main/java/io/cdap/plugin/gcp/gcs/sink/GCSBatchSink.java
```diff
@@ -125,17 +127,27 @@ public void prepareRun(BatchSinkContext context) throws Exception {
...
+      location = config.getLocation();
     }
     this.outputPath = getOutputDir(context);
+    // create asset for lineage
+    String referenceName = Strings.isNullOrEmpty(config.getReferenceName())
+      ? ReferenceNames.normalizeFqn(config.getPath())
+      : config.getReferenceName();
+    asset = Asset.builder(referenceName)
+      .setFqn(config.getPath()).setLocation(location).build();
+    super.prepareRun(context);
```
fernst, line 153: Add a comment here that states why prepare run is called at the end (to ensure this doesn't get moved to the top by accident).

fernst (approved):

PR #1152, opened 2022-09-29
PLUGIN-1413 Added argument for BQ temporary staging bucket names

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -97,7 +102,7 @@ public final class BigQueryUtil {
     .build();
 
   private static final Map<Schema.Type, Set<LegacySQLTypeName>> TYPE_MAP = ImmutableMap.<Schema.Type,
-    Set<LegacySQLTypeName>>builder()
+      Set<LegacySQLTypeName>>builder()
```
albertshau: this change isn't needed

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -678,7 +683,7 @@ public static void validateGCSChunkSize(String chunkSize, String chunkSizeProper
       try {
         if (Integer.parseInt(chunkSize) % MediaHttpUploader.MINIMUM_CHUNK_SIZE != 0) {
           collector.addFailure(
-            String.format("Value must be a multiple of %s.", MediaHttpUploader.MINIMUM_CHUNK_SIZE), null)
+              String.format("Value must be a multiple of %s.", MediaHttpUploader.MINIMUM_CHUNK_SIZE), null)
```
albertshau: indentation

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -73,6 +77,7 @@ public final class BigQueryUtil {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryUtil.class);
 
   private static final String DEFAULT_PARTITION_COLUMN_NAME = "_PARTITIONTIME";
+  private static final String BIGQUERY_BUCKET_PREFIX_PROPERTY_NAME = "io.cdap.plugin.bigquery.bucket.prefix";
```
albertshau: we do a similar thing for cmek, where the argument key is 'gcp.cmek.key.name' (see CmekUtils). Let's follow a similar pattern and name it something like 'gcp.bigquery.bucket.prefix'

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -772,4 +777,66 @@ public static String getFQN(String datasetProject, String datasetName, String ta
...
+                                           "characters or shorter.");
+    }
+  }
+
+  /**
+   * Method to generate the CRC32 checksum for a location.
+   * We use this to ensure location name length is constant (only 8 characters).
+   *
+   * @param location location to checksum
+   * @return checksum value as@ an 8 character string (hex).
```
albertshau: typo: as@

src/test/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtilTest.java
```diff
@@ -127,6 +134,129 @@ public void testValidateArraySchema() {
...
+
+    // Check there are no collisions
+    Set<String> hashValues = new HashSet<>();
+    for (String location : locations) {
+      String hash = BigQueryUtil.crc32location(location);
+      Assert.assertFalse(hashValues.contains(hash));
+      hashValues.add(hash);
+    }
+
+    System.out.println(hashValues);
```
albertshau: don't need this

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -291,4 +291,36 @@ protected Configuration getOutputConfiguration() throws IOException {
...
+    String bucketPrefix = BigQueryUtil.getBucketPrefix(context.getArguments());
+
+    // If temp bucket name is not defined in configuration, and a common bucket name prefix has been specified,
+    // we must set this prefix along with the destination location in order to create/reuse the bucket.
+    // Otherwise, if temp bucket name is defined, or a prefix is not set, the configureBucket method will prepare
+    // for a new bucket creation.
+    if (bucketName == null && bucketPrefix != null) {
+      // Check if the destination dataset exists.
+      DatasetId datasetId = DatasetId.of(config.getDatasetProject(), config.getDataset());
+      Dataset dataset = bigQuery.getDataset(datasetId);
```
albertshau: It seems like we should already be making this call as part of the existing validation/automatic bucket creation. If so, we should do some refactoring to avoid duplicate calls. It's become kind of a mess now, but ideally we do all the I/O in a single place and pass the return objects around.

fernst: I don't see an easy way to refactor this without making 2 calls to get the dataset.

The problem is the bucket name is needed even when in preview. This makes it difficult to refactor the code in a way that doesn't change the entire method signature for the abstract bigquery sink and other callers (such as the BigQuery Pushdown implementation).

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: reverted this change

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: reverted

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: Updated

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
fernst: Fixed

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
fernst: I have refactored the code a bit, hope this helps

albertshau (approved):

PR #1156, opened 2022-10-06
CDAP-19667:make reference name optional for plugins with fqns

mrahanjam:
This PR makes reference name optional for plugins with fqns


snehalakshmisha (approved):
lgtm

PR #1157, opened 2022-10-07
PLUGIN-1385 Add labels to BigQuery Execute and BigQuery Sink jobs

fernst:


src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQueryWrite.java
```diff
@@ -354,7 +354,7 @@ protected QueryJobConfiguration.Builder getInsertQueryJobBuilder(TableId sourceT
       .setCreateDisposition(JobInfo.CreateDisposition.CREATE_NEVER)
       .setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND)
       .setPriority(sqlEngineConfig.getJobPriority())
-      .setLabels(BigQuerySQLEngineUtils.getJobTags("copy"));
+      .setLabels(BigQuerySQLEngineUtils.getJobTags("write"));
```
fernst: I updated the BQ Write (BQ Sink Pushdown) labels to match the actual operation semantics (Write to the sink)

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -175,6 +179,17 @@ public AbstractBigQueryActionConfig getConfig() {
...
   }
 
+  /**
+   * Get Job Labels for the BigQuery export job.
+   * @return
+   */
+  private static Map<String, String> getJobLabels() {
+    Map<String, String> labels = new HashMap<>();
+    labels.put("job_source", "cdap");
+    labels.put("type", "bq_execute_plugin");
```
tivv: Could you please move all `getJobLabels` to one place or at least reuse constants in here and in io.cdap.plugin.gcp.bigquery.sqlengine.util.BigQuerySQLEngineUtils#getJobTags(java.lang.String)? This way anyone looking would understand what are different labels and it's possible values.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -70,6 +66,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryExecute.class);
   public static final String NAME = "BigQueryExecute";
   private static final String RECORDS_PROCESSED = "records.processed";
+  private static final String BQ_JOB_TYPE = "bq_execute_plugin";
```
tivv: nit: Consider moving it to `BigQueryUtil` as well. This would allow easily see possible values for this label.

tivv (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
fernst: Moved into the common Util class for easy reference.

PR #1161, opened 2022-10-13
Fix incorrect upsert query for BigQuery sink

zoe-L:
Fixing bug https://cdap.atlassian.net/browse/PLUGIN-676 BigQuery sink generates incorrect query when using all the input fields in relationTableKey field

Remove `WHEN MATCHED THEN` clause from upsert query when all fields are used as keys.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
@@ -47,7 +47,6 @@ import io.cdap.plugin.gcp.bigquery.util.BigQueryConstants;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryTypeSize.Numeric;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.GCPUtils;
-
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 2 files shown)


zoe-L:
Confirmed that the UPSERT query succeeds now in CDAP 6.8.0:
Screen Shot 2022-10-13 at 4 24 21 PM

Also, the e2e tests were broken upstream, not by this PR.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -624,6 +630,10 @@ public static String generateUpdateUpsertQuery(Operation operation,
         return String.format(UPDATE_QUERY, destination, fieldsForUpdate, sourceTable, criteria);
       case UPSERT:
         String insertFields = "`" + String.join("`, `", tableFieldsList) + "`";
+        if (fieldsForUpdate.isEmpty()) {
```
fernst, line 634: Add a comment explaining why we return one or the other

fernst (commented):
LGTM, but please add the comment I asked for

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
zoe-L: Added comment

fernst (approved):

PR #1162, opened 2022-10-17
fix gcs sink error when no reference name provided and some cleanup

mrahanjam:
This PR fixes GCS sink error when no reference name provided plus some cleanup
Jira: https://cdap.atlassian.net/browse/CDAP-19789


snehalakshmisha (approved):
lgtm

src/main/java/io/cdap/plugin/gcp/dataplex/common/config/DataplexBaseConfig.java
```diff
@@ -75,11 +75,7 @@ public class DataplexBaseConfig extends PluginConfig {
   @Description("The existing connection to use.")
   protected GCPConnectorConfig connection;
```
snehalakshmisha, line 77: Let's add a java doc saying it returns the reference name or the normalized FQN, since it is a public method

PR #1164, opened 2022-10-18
fix gcs sink error when no reference name provided and some cleanup

mrahanjam:
Cherry picking the fix for gcs sink and some refactoring.
Original PR: https://github.com/data-integrations/google-cloud/pull/1162


snehalakshmisha (approved):
lgtm

PR #1167, opened 2022-10-20
stop auto-populating reference name for BQ, GCS sink, and spanner sink

mrahanjam:
This PR stops auto-populating reference name for BQ, GCS sink, and spanner sink.

Jira: [CDAP-19804](https://cdap.atlassian.net/browse/CDAP-19804)

### DIFF
--- a/src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
@@ -160,10 +160,11 @@ public class SpannerConnectorTest {
     properties.put("table", table);
     properties.put("connection", null);
     properties.put("useConnection", "true");
-    properties.put("referenceName", String.format("%s.%s.%s", instance, database, table));
+    Map<String, String> sourceProperties = new HashMap<>(properties);
+    sourceProperties.put("referenceName", String.format("%s.%s.%s", instance, database, table));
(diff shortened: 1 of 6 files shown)


src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
```diff
@@ -160,10 +160,11 @@ private void testGenerateSpec(SpannerConnector connector) throws IOException {
     properties.put("table", table);
     properties.put("connection", null);
     properties.put("useConnection", "true");
```
snehalakshmisha, line 162: Can we rename this as `sinkProperties` like we did in the other file to avoid confusion?

snehalakshmisha (approved):

src/test/java/io/cdap/plugin/gcp/spanner/connector/SpannerConnectorTest.java
mrahanjam: The reason I didn't, was that `properties` is like the parent, and source diverges further down, but before that it's all shared between the two. maybe I thought about it too much.

snehalakshmisha: Right. But if I were to look at the code and I saw a `sourceProperties` and a `sinkProperties` it would be clearer to me than if it were `sourceProperties` and `properties`

PR #1171, opened 2022-10-21
send correct location for BQ sink plugin when emitting lineage info

mrahanjam:
Send correct location for BQ sink plugin when emitting lineage info

Jira: [CDAP-19520](https://cdap.atlassian.net/browse/CDAP-19520)


snehalakshmisha (changes requested):
Please add a jira to the description.

snehalakshmisha (approved):

PR #1177, opened 2022-11-03
PLUGIN-1418: add argument for BQ temp staging bucket names

itsankit-google:
Follow up PR of #1152.

It includes the support of creating common staging bucket for Dataplex and BQ pushdown plugins.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
@@ -19,6 +19,7 @@ package io.cdap.plugin.gcp.bigquery.sqlengine;
 import com.google.auth.Credentials;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.TableId;
@@ -26,7 +27,6 @@ import com.google.cloud.kms.v1.CryptoKeyName;
 import com.google.cloud.storage.Storage;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
-import io.cdap.cdap.api.RuntimeContext;
 import io.cdap.cdap.api.SQLEngineContext;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Metadata;
@@ -116,7 +116,7 @@ public class BigQuerySQLEngine
   private Configuration configuration;
   private String project;
   private String datasetProject;
-  private String dataset;
+  private String datasetName;
   private String bucket;
   private String runId;
   private Map<String, BigQuerySQLDataset> datasets;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
```diff
@@ -858,4 +859,30 @@ public static Map<String, String> getJobTags(String jobType) {
...
+    // we must set this prefix along with the destination location in order to create/reuse the bucket.
+    // Otherwise, if temp bucket name is defined, or a prefix is not set, the configureBucket method will prepare
+    // for a new bucket creation.
+    if (Strings.isNullOrEmpty(bucket) && bucketPrefix != null) {
+      // If the destination dataset exists, use the dataset location. Otherwise, use location from configuration.
+      String datasetLocation = dataset != null ? dataset.getLocation() : configLocation;
+      // Get the bucket name for the specified location.
+      bucket = BigQueryUtil.getBucketNameForLocation(bucketPrefix, datasetLocation);
+    }
+    return bucket;
```
fernst, line 887: How does this behave is the supplied Bucket is nullable and there is no prefix set? As far as I can tell, the returned `bucket` will be null.

If this is the case, please add the `@Nullable` annotation to this function.

Following question is: Should this function return a bucket name if not supplied and no prefix is set?

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -153,12 +153,14 @@ public void prepareRun(SQLEngineContext context) throws Exception {
...
-    dataset = sqlEngineConfig.getDataset();
-    bucket = sqlEngineConfig.getBucket() != null ? sqlEngineConfig.getBucket() : "bqpushdown-" + runId;
+    datasetName = sqlEngineConfig.getDataset();
 
     // Initialize BQ and GCS clients.
     bigQuery = GCPUtils.getBigQuery(project, credentials);
     storage = GCPUtils.getStorage(project, credentials);
+    Dataset dataset = bigQuery.getDataset(DatasetId.of(datasetProject, datasetName));
+    bucket = BigQueryUtil.getStagingBucketName(context.getRuntimeArguments(), sqlEngineConfig.getLocation(),
+                                               dataset, null);
```
fernst: The final argument should be `sqlEngineConfig.getBucket()`

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
```diff
@@ -169,11 +171,11 @@ public void prepareRun(SQLEngineContext context) throws Exception {
...
                                                    cmekKeyName, sqlEngineConfig.getServiceAccountType());
     // Create resources needed for this execution
-    BigQuerySinkUtils.createResources(bigQuery, storage, DatasetId.of(datasetProject, dataset), bucket,
+    BigQuerySinkUtils.createResources(bigQuery, storage, DatasetId.of(datasetProject, datasetName), bucket,
                                       sqlEngineConfig.getLocation(), cmekKeyName);
     // Configure GCS bucket that is used to stage temporary files.
-    // If the bucket is created for this run, mar it for deletion after executon is completed
-    BigQuerySinkUtils.configureBucket(configuration, bucket, runId, sqlEngineConfig.getBucket() == null);
+    // If the bucket is created for this run, mark it for deletion after executon is completed
+    BigQuerySinkUtils.configureBucket(configuration, bucket, "bqpushdown-" + runId);
```
fernst: nit: The second argument should be in a variable that outlines it's use

`String fallbackBucketName =  "bqpushdown-" + runId;`

fernst (changes requested):

src/main/java/io/cdap/plugin/gcp/bigquery/util/BigQueryUtil.java
itsankit-google: Yeah, it is expected to return `null` in that case. Added the `@Nullable` annotation.

No we don't expect this method to return a bucket name if not supplied and if no prefix is set.
`BigQuerySinkUtils#configureBucket` and `BigQuerySourceUtils#getOrCreateBucket` expect the `bucket` argument to be `nullable` and will keep the existing behaviour.

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
itsankit-google: ack

src/main/java/io/cdap/plugin/gcp/bigquery/sqlengine/BigQuerySQLEngine.java
itsankit-google: ack

fernst (approved):
lgtm

PR #1185, opened 2022-11-22
PLUGIN-1445 : BQ sink in case empty task: initialise a Null Data writer with null stream and access it in Close function

sahusanket:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -103,6 +109,7 @@ private void createFileWriter(Schema writerSchema) throws IOException {
   public void close(TaskAttemptContext context) throws IOException {
     if (mAvroFileWriter != null) {
       mAvroFileWriter.close();
+      this.outputStream.close();
```
singhravidutt: Can we also have unit test for this?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -103,6 +109,7 @@ private void createFileWriter(Schema writerSchema) throws IOException {
   public void close(TaskAttemptContext context) throws IOException {
     if (mAvroFileWriter != null) {
       mAvroFileWriter.close();
```
singhravidutt: Also, as `mAvroFileWriter` depends upon `outputStream` does it make sense to call `outputstream.close` in `mAvroFileWriter.close` instead of explicitly calling it here? I am assuming `outputstream` is used just for `mAvroFileWriter` correct me if this is wrong.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -58,6 +59,7 @@ public AvroRecordWriter(Schema writerSchema, GenericData dataModel, CodecFactory
     this.compressionCodec = compressionCodec;
     this.outputStream = outputStream;
     this.syncInterval = syncInterval;
+    createFileWriter(Schema.create(Schema.Type.NULL), ByteStreams.nullOutputStream());
```
tgmof: Hi there! If I may drop a suggestion/question:
Why not setting writerSchema here as first input and outputStream as second input and stop checking for schema change for each record in "write"? Anyway, even if it fails here because of schema issues, it's only 1 step earlier than at the BQ Load step.
At least in my specific use case, I have a "large" schema and all incoming records contain fields that are a subset of that "large" schema. Thus if we set the large schema here (coming from the output schema that I set in the UI), `mAvroFileWriter` will never need to be re-created whereas if we keep getting schemas from the records, it will have to be sync/close/recreated every X records (though unsure if that has significant performance impact, particularly if the default is 'only' 8MB/file). Correct me if I'm wrong as I'm not a CDAP expert.

But I understand that this was only intended to initialise prevSchema/mAvroFileWriter and if you prefer not to introduce additional complexity in this PR, it also makes sense to me.

PS: You never close ByteStreams.nullOutputStream() later, so even though it will never grow more than the object initialization, can it lead to some problem or is it just fine because de-referencing it will auto-close it?

tivv: Could you clarify why do we need to create a NOP writer? It makes it a bit confusing as now `mAvroFileWriter` does not always point to the `outputStream`. All we need is to always close `outputStream` in the `close`, even if writer is not there.

tgmof:
Hey Guys!
Could you have a look to this PR? I know that any good PO would answer "it's not a bug but a feature" but I am still convinced that it's a bug and it would be a bit sad that I have to create and maintain a fork just to fix a "outputStream.close()".
Thanks in advance!

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
sahusanket: I tried that 
```
if (mAvroFileWriter != null) {
      mAvroFileWriter.close();
    } else if (outputStream != null) {
      outputStream.close();
    }
```
Without any other change 
but that gives the error : 
`Error occurred while importing data to BigQuery 'Error while reading data, error message: The Apache Avro library failed to parse the header with the following error: EOF reached File : abc/xyz/part-r-00001.avro` 

I am unsure of the reason

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
sahusanket: Yes you are right `AvroFileWriter.close` internally closes `outputstream` . 
Thanks

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
sahusanket: - When the AvroRecordWriter is initialized in the constructor the value of `writerSchema` is `null`. the `writerSchema` is rather pulled from every record. Hence cannot initialize `mAvroFileWriter` with null object.

- The current is quite generic to handle different schemas . Even in your usecase, the `mAvroFileWriter.sync/close` will happen only once, since it's reloaded only if there is a new schema. 

- `ByteStreams.nullOutputStream()` is passed on to `mAvroFileWriter` , and when we call `mAvroFileWriter.close()`, it internally closes the outstream as well.

sahusanket: Also initiating like  `createFileWriter(Schema.create(Schema.Type.NULL), outputStream);` ( instead of NOP  ) fails with NPE while creating `mAvroFileWriter` when actual data is written. ( line  90 --> 103 )

tivv: Basically now we never close output stream, so the file is hanging around, but due to a side effect is probably not accessible by the importer. It still takes memory, thus the OOM. 

With the fix you propose, I don't think anything is changed as the original OutputStream is not closed anyway, so OOM will be there.

The simple `outputStream.close()` won't work as it would produce an empty file that can't be read by the AVRO reader.

We have 2 options:
 * Don't create the OutputStream if it's not needed. This would require passing a `Supplier<OutputStream>` instead of `OutputStream` to the writer. The result won't change (file won't be created if partition is empty), but the OutputStream won't be created either. In the edge case of all empty partitions, there is going to be no data files at all. I suppose this is how it is now, but it would be great to test if it works / will work after the change
 * Create both OutputStream and DataFileWriter and close in the end. This would create files with proper header, but 0 rows. Here you would also need a proper schema (I believe it's already passed to the  constructor). The downside is creating unneeded files. I would prefer to go first option unless we have issues with empty datasets.

tivv (changes requested):

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
sahusanket: Implemented the 1st Supplier approach and it worked :) 

Was able to repro original OOM issue and this code change passed without any issue. 
This also resulted in Faster execution as well ! 

Thanks

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroOutputFormat.java
```diff
@@ -65,21 +70,34 @@ public RecordWriter<AvroKey<GenericRecord>, NullWritable> getRecordWriter(TaskAt
...
-                  getAvroFileOutputStream(context), getSyncInterval(context));
+
+    //Creating a supplier for outstream , which will get triggered only if there is a record to written.
+    //This is to avoid creation of output stream for empty tasks ( no writes )
+    Supplier<OutputStream> outputStreamSupplier = () -> {
+      try {
+        return getAvroFileOutputStream(context);
+      } catch (IOException e) {
+        LOG.error("There was an IOException while creating output stream");
+        throw new RuntimeException(e);
```
tivv: Use java.io.UncheckedIOException

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroOutputFormat.java
```diff
@@ -65,21 +70,34 @@ public RecordWriter<AvroKey<GenericRecord>, NullWritable> getRecordWriter(TaskAt
...
-    return create(writerSchema, dataModel, getCompressionCodec(context),
-                  getAvroFileOutputStream(context), getSyncInterval(context));
+
+    //Creating a supplier for outstream , which will get triggered only if there is a record to written.
+    //This is to avoid creation of output stream for empty tasks ( no writes )
+    Supplier<OutputStream> outputStreamSupplier = () -> {
+      try {
+        return getAvroFileOutputStream(context);
+      } catch (IOException e) {
+        LOG.error("There was an IOException while creating output stream");
```
tivv: Do we need this? Usually if exception is rethrown you don't log it unless the caller don't log it either.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroOutputFormat.java
```diff
@@ -65,21 +70,34 @@ public RecordWriter<AvroKey<GenericRecord>, NullWritable> getRecordWriter(TaskAt
...
-   * @param outputStream The target output stream for the records.
+   * @param outputStreamSupplier A supplier to give the target output stream for the records.
    * @param syncInterval The sync interval for the writer file.
    */
   private RecordWriter<AvroKey<GenericRecord>, NullWritable> create(
     Schema writerSchema, GenericData dataModel, CodecFactory compressionCodec,
-    OutputStream outputStream, int syncInterval) throws IOException {
-    return new AvroRecordWriter(writerSchema, dataModel, compressionCodec, outputStream, syncInterval);
+    Supplier<OutputStream> outputStreamSupplier, int syncInterval)
+    throws IOException {
```
tivv: I don't think you need this `throws` in here and AvroRecordWriter constructor

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -41,35 +43,37 @@ public class AvroRecordWriter extends RecordWriter<AvroKey<GenericRecord>, NullW
   private GenericData dataModel;
   private CodecFactory compressionCodec;
   private OutputStream outputStream;
```
tivv: Do you still need the field?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -78,6 +82,11 @@ public AvroRecordWriter(Schema writerSchema, GenericData dataModel,
   public void write(AvroKey<GenericRecord> record, NullWritable ignore) throws IOException {
     // Create an Avro container file and a writer to it.
     Schema writerSchema = record.datum().getSchema();
+
+    if (this.outputStream == null) {
+      this.outputStream = outputStreamSupplier.get();
```
tivv: Since the `mAvroFileWriter` is never reset to null, you can ask supplier for a stream in the `createFileWriter` directly

tivv (commented):
Also please add a test

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriter.java
```diff
@@ -114,4 +118,20 @@ public long sync() throws IOException {
...
     return 0;
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof AvroRecordWriter)) return false;
+    AvroRecordWriter that = (AvroRecordWriter) o;
+
+    return syncInterval == that.syncInterval && Objects.equals(mAvroFileWriter, that.mAvroFileWriter)
```
sahusanket: @tivv 
Did not include supplier as I could not find a way to compare 2 `Supplier` .

Is there any way to do so ? 

This is needed for the Unit test

tivv: Why do you need it for unit test? Usually adding equals not to a data object is not a good idea as it has such challenges. You can necessary properties for testing and use org.hamcrest.beans.SamePropertyValuesAs

tivv: This also has a benefit of good description instead of "not equal, sorry", especially without good toString

sahusanket: This is for unit test for `AvroOutputFormat` , for method `getRecordWriter` which would return a `AvroRecordWriter` 

hence I wanted to have a equals method to compare Actual & expected object.

tivv: For non-DTOs you often can't compare those directly as they don't need (and often can't) implement equals. You should decide which specific properties to compare, provide getters (may be with "just for test" visibility) and compare those properties. The classes referecned above help you to do the comparison

sahusanket: Thanks. 
Implemented the same.

pom.xml
```diff
@@ -651,6 +651,12 @@
       <version>${powermock.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>java-hamcrest</artifactId>
```
tivv: I would recommend to stick to the 1.3 version we are already using (transitively from awaitility). Or fully migrate to 2.0. Having two versions in a same project often leads to issues

src/test/java/io/cdap/plugin/gcp/bigquery/sink/AvroOutputFormatTest.java
```diff
@@ -0,0 +1,55 @@
...
+    AvroOutputFormat avroOutputFormat = new AvroOutputFormat();
+
+    RecordWriter<AvroKey<GenericRecord>, NullWritable> actual =
+      avroOutputFormat.getRecordWriter(context);
+    AvroRecordWriter avroRecordWriter = (AvroRecordWriter) actual;
+
+    Supplier<OutputStream> outputStreamSupplier = avroOutputFormat.getOutputStreamSupplier(context);
+
+    Assert.assertThat(outputStreamSupplier,
+                      SamePropertyValuesAs.samePropertyValuesAs(avroRecordWriter.getOutputStreamSupplier()));
```
tivv: I don't think this does what you want. It would make sense to do `samePropertyValuesAs` for the `avroRecordWriter` as it has the property. The `Supplier` has no proeprties so that's essentially a NOP check. Also I am not sure what are you trying to test here? Why do you care that those are same or not?
Probably you care that it writes to a proper file. But in this case, just get the output stream and validate where does it write to.

sahusanket: Just wanted to have a unit test for the code change, it doesn't do anything significant. 
I will remove this unit test.

src/test/java/io/cdap/plugin/gcp/bigquery/sink/AvroRecordWriterTest.java
```diff
@@ -0,0 +1,97 @@
...
+    //Check if it matches
+    Assert.assertTrue(dataFileReader.iterator().hasNext());
+    //Record read must match the sourceGenericRecord
+    Assert.assertEquals(dataFileReader.iterator().next(), sourceGenericRecord);
+  }
+
+
+  /**
+   * If there are no records, then the output stream supplier should not be called
+   * If called then will throe exception
```
tivv: nit:  typo `throw`

tivv (approved):

PR #1196, opened 2022-12-07
Added Spark Window Frame related query changes in Bigquery

udurbha:
As per spark Window Frame(Row,Range), if a value 0 is given it becomes CURRENT ROW .If a value greater than 0 is entered it becomes value Following and if less than 0 it becomes value preceding.This PR has changes of Bigquery as per spark Window Frame


fernst (approved):
LGTM

PR #1202, opened 2023-01-06
[CDAP-20238] Remove Spark2 references and update dependencies

adityagupta94:
As part of cdapio/cdap/pull/14797, the spark2 modules and references have been removed. Similar references in this repo need to be done and subsequent dependencies need to be updated.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,6 @@
     <junit.version>4.13.1</junit.version>
     <powermock.version>2.0.2</powermock.version>
     <slf4j.version>1.7.5</slf4j.version>
-    <spark.version>2.3.1</spark.version>
     <spark3.version>3.1.1</spark3.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -711,7 +710,7 @@
     <!-- Start: dependency for Google PubSub Streaming Source -->
     <dependency>
       <groupId>org.apache.bahir</groupId>
-      <artifactId>spark-streaming-pubsub_2.11</artifactId>
+      <artifactId>spark-streaming-pubsub_2.12</artifactId>
       <version>2.4.0</version>
```
dli357, line 714: Should this be 2.4.0?

adityagupta94: The only library using scala 2.12 is 2.4.0: https://mvnrepository.com/artifact/org.apache.bahir/spark-streaming-pubsub_2.12

albertshau (approved):

PR #1222, opened 2023-03-14
Emit lineage for the plugin

snehalakshmisha:
Jira: https://cdap.atlassian.net/browse/CDAP-20441

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -141,10 +141,11 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
    * @param tableName table name
    * @param tableSchema table schema
    * @param bucket bucket name
+   * @param marker marker for the asset, if applicable
    */
   protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, String outputName, String fqn,
                                   String tableName, @Nullable Schema tableSchema, String bucket,
-                                  FailureCollector collector) throws IOException {
+                                  FailureCollector collector, @Nullable String marker) throws IOException {
     LOG.debug("Init output for table '{}' with schema: {}", tableName, tableSchema);
 
     List<BigQueryTableFieldSchema> fields = BigQuerySinkUtils.getBigQueryTableFields(bigQuery, tableName, tableSchema,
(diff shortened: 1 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -165,12 +165,8 @@ protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, Str
     List<String> fieldNames = fields.stream()
       .map(BigQueryTableFieldSchema::getName)
       .collect(Collectors.toList());
-    // Get the dataset location (if it exists)
-    DatasetId datasetId = DatasetId.of(getConfig().getDatasetProject(), getConfig().getDataset());
-    Dataset dataset = bigQuery.getDataset(datasetId);
-    String location = dataset != null ? dataset.getLocation() : getConfig().getLocation();
```
snehalakshmisha, line 171: Removed these lineage because we do not need to send location information anymore

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -169,8 +169,8 @@ protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, Str
     DatasetId datasetId = DatasetId.of(getConfig().getDatasetProject(), getConfig().getDataset());
     Dataset dataset = bigQuery.getDataset(datasetId);
     String location = dataset != null ? dataset.getLocation() : getConfig().getLocation();
-    Asset asset = Asset.builder(outputName).setFqn(fqn).setLocation(location).build();
-    BigQuerySinkUtils.recordLineage(context, asset, tableSchema, fieldNames);
+    Asset asset = Asset.builder(outputName).setFqn(fqn).setLocation(location).setTag(tableName).build();
+    BigQuerySinkUtils.recordLineage(context, asset, tableSchema, fieldNames, tableName);
```
yaojiefeng, line 175: this change should only happen in bigquery multi sink, this class is also used in bigquery sink, the change should not happen for single output sink.

snehalakshmisha: Retaining the old asset now if marker is not available for single-sink. Marker is set only for assets from multi-sink

yaojiefeng (approved):

PR #1232, opened 2023-04-19
Add a method to add schema information to a BigQueryRelation

DJSagarAhire:
This PR adds a way to add schema information to a relation. This can be called by a plugin that wishes to request validation.

The current intention is that this will be called by Wrangler in the `transform()` method to supply schema information (which Wrangler currently already has in the `oSchema` variable).

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
@@ -349,6 +349,16 @@ public class BigQueryRelation implements Relation {
     return new BigQueryRelation(datasetName, columns, featureFlagsProvider, this, supplier);
   }
 
+  /**
+   * Adds schema information to the relation. This can be used for validation purposes.
+   *
+   * @param schema The schema.
+   * @return A new relation with the schema added.
+   */
+  public Relation addSchema(@Nullable Schema schema) {
+    return new BigQueryRelation(datasetName, columns, featureFlagsProvider, this, sqlStatementSupplier, schema);
+  }
+
   private static String buildBaseSelect(Map<String, Expression> columns,
                                         String sourceTable,
                                         String datasetName) {


src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
```diff
@@ -349,6 +349,16 @@ public Relation window(WindowAggregationDefinition definition) {
     return new BigQueryRelation(datasetName, columns, featureFlagsProvider, this, supplier);
   }
 
+  /**
+   * Adds schema information to the relation. This can be used for validation purposes.
+   *
+   * @param schema The schema.
+   * @return A new relation with the schema added.
+   */
+  public Relation addSchema(@Nullable Schema schema) {
```
tivv, line 358: Please validate there is no existing schema unless you expect schema overrides

DJSagarAhire: What should we ideally be doing in case a schema already exists? Since we're returning a new `Relation` I think overriding the schema should be fine.

fernst (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/relational/BigQueryRelation.java
tivv: I don't know how do you plan to use this method. Generally it's better to introduce methods when they are needed (even with mock implementation) than visa versa.

DJSagarAhire: I've mentioned that already in the PR description above. I plan to use it in Wrangler in `transform()` just like any other `Relation` operation. I don't see any other way to add a schema to the Relation.

tivv: Whangler should have nothing to do with schema management. How would Wrangler know the schema of a SQL expression?

tivv: Also it's unneded burden on the plugin. Given known schema of the original relation and known schema of all expressions, resulting schema should be automatically contructed by the platform

PR #1238, opened 2023-04-20
PLUGIN-1537 - Refactor code to facilitate reuse and use in serializable streams.

greeshmaswaminathan:
[PLUGIN-1537](https://cdap.atlassian.net/browse/PLUGIN-1537) State management changes for PubSub part1. Follow up PR will include the changes for direct Dstream.
- Refactor code to reuse common functionality like creating subscription for receiver based Dstream and direct DStream.
- Move structured record mapping function and GoogleSubscriberConfig out of GoogleSubscriber so that they can be used in serializable Dstream.
- Introduce SerializableFunction1 for using in direct dstream.

[PLUGIN-1502]: https://cdap.atlassian.net/browse/PLUGIN-1502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PLUGIN-1537]: https://cdap.atlassian.net/browse/PLUGIN-1537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/BackoffConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/BackoffConfig.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.gcp.publisher.source;
+
+import java.io.Serializable;
+
+/**
+ * Class used to configure exponential backoff for Pub/Sub API requests.
+ */
+public class BackoffConfig implements Serializable {
+  private final int initialBackoffMs;
+  private final int maximumBackoffMs;
+  private final double backoffFactor;
+
+  static final BackoffConfig defaultInstance() {
+    return new BackoffConfig(100, 10000, 2.0);
+  }
+
+  public BackoffConfig(int initialBackoffMs, int maximumBackoffMs, double backoffFactor) {
+    this.initialBackoffMs = initialBackoffMs;
+    this.maximumBackoffMs = maximumBackoffMs;
+    this.backoffFactor = backoffFactor;
+  }
+
+  public int getInitialBackoffMs() {
+    return initialBackoffMs;
+  }
+
+  public int getMaximumBackoffMs() {
+    return maximumBackoffMs;
+  }
+
+  public double getBackoffFactor() {
+    return backoffFactor;
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriberConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriberConfig.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.gcp.publisher.source;
+
+import com.google.common.base.Strings;
+import io.cdap.cdap.api.annotation.Description;
+import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.plugin.gcp.publisher.PubSubConstants;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Extension to the PubSubSubscriberConfig class with additional fields related to record schema.
+ */
+public class GoogleSubscriberConfig extends PubSubSubscriberConfig implements Serializable {
+
+  private static final String SCHEMA = "schema";
+  private static final Schema DEFAULT_SCHEMA =
+    Schema.recordOf("event",
+                    Schema.Field.of("message", Schema.of(Schema.Type.BYTES)),
+                    Schema.Field.of("id", Schema.of(Schema.Type.STRING)),
+                    Schema.Field.of("timestamp", Schema.of(Schema.LogicalType.TIMESTAMP_MICROS)),
+                    Schema.Field.of("attributes", Schema.mapOf(Schema.of(Schema.Type.STRING),
+                                                               Schema.of(Schema.Type.STRING)))
+    );
+
+  @Macro
+  @Nullable
+  @Description("Format of the data to read. Supported formats are 'avro', 'blob', 'tsv', 'csv', 'delimited', 'json', "
+    + "'parquet' and 'text'.")
+  protected String format;
+
+  @Description("The delimiter to use if the format is 'delimited'. The delimiter will be ignored if the format "
+    + "is anything other than 'delimited'.")
+  @Macro
+  @Nullable
+  protected String delimiter;
+
+  @Name(SCHEMA)
+  @Macro
+  @Nullable
+  protected String schema;
+
+  @Override
+  public void validate(FailureCollector collector) {
+    super.validate(collector);
+
+    final Schema outputSchema = getSchema();
+    final ArrayList<String> defaultSchemaFields = getFieldsOfDefaultSchema();
+    ArrayList<String> outputSchemaFields = new ArrayList<>();
+
+    if (outputSchema != null) {
+      for (Schema.Field field : Objects.requireNonNull(outputSchema.getFields())) {
+        outputSchemaFields.add(field.getName());
+      }
+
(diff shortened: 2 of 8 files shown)


src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriberConfig.java
```diff
@@ -0,0 +1,155 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Extension to the PubSubSubscriberConfig class with additional fields related to record schema.
+ */
+public class GoogleSubscriberConfig extends PubSubSubscriberConfig implements Serializable {
```
tivv, line 34: It's serializable. Is it serialized anywhere in persistent storage?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * PubSubStructuredRecordConverter for converting a PubSubMessage to StructuredRecord.
+ */
+public class PubSubStructuredRecordConverter {
+
+  /**
+   * Converts a PubSubMessage into a StructuredRecord based on the specified schema.
+   * If no schema is specified, the default schema is used.
+   */
+  public static SerializableFunction<PubSubMessage, StructuredRecord> getFunction(GoogleSubscriberConfig config) {
```
tivv: Try not to use the language constructs in names, tell what it does. E.g. name it `getConverter` if you want a shorter name. This would give some information that's not already in the signature.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
```diff
@@ -0,0 +1,133 @@
...
+    return (pubSubMessage) -> {
+
+      Schema customMessageSchema = getCustomMessageSchema(config);
+      final Schema outputSchema = config.getSchema();
+      final String format = config.getFormat();
+
+      // Convert to a HashMap because com.google.api.client.util.ArrayMap is not serializable.
+      HashMap<String, String> hashMap = new HashMap<>();
```
tivv: Same here. Try not to use just class names as variables/fields/method names. In this case `attributes` would be a god name. Nice names make it easier to read the code later

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
tivv: BTW: Any specific reason you want to make it a lambda instead of just making this class implement the `SerializableFunction<PubSubMessage, StructuredRecord>` directly?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
```diff
@@ -34,6 +48,21 @@ public final class PubSubSubscriberUtil {
...
   protected static final Logger LOG = LoggerFactory.getLogger(PubSubSubscriberUtil.class);
 
+  // Retryable status codes. These need to be handled in case Pub/Sub throws a StatusRuntimeException.
+  private static final int RESOURCE_EXHAUSTED = StatusCode.Code.RESOURCE_EXHAUSTED.getHttpStatusCode();
+  private static final int CANCELLED = StatusCode.Code.CANCELLED.getHttpStatusCode();
+  private static final int INTERNAL = StatusCode.Code.INTERNAL.getHttpStatusCode();
+  private static final int UNAVAILABLE = StatusCode.Code.UNAVAILABLE.getHttpStatusCode();
+  private static final int DEADLINE_EXCEEDED = StatusCode.Code.DEADLINE_EXCEEDED.getHttpStatusCode();
+
+  private static final Set<Integer> RETRYABLE_STATUS_CODES = Collections.unmodifiableSet(new HashSet<Integer>() {{
```
tivv: nit: That's a weird construct with anonymous classes. I see that it was like this before, that's why it's `nit`

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
```diff
@@ -86,4 +115,91 @@ protected static JavaDStream<PubSubMessage> getInputDStream(StreamingContext str
...
+   * @param preCheck      Any checks that need to be applied before each retry.
+   * @param backoffConfig {@link BackoffConfig} for retries.
+   * @param subscription  Subscription name string.
+   * @param topic         Topic name string.
+   * @param clientSupplier Supplier for creating {@link SubscriptionAdminClient}
+   * @param isRetryableException Predicate for checking if the exception is retryable.
+   * @throws InterruptedException If the wait for retry is interrupted.
+   * @throws IOException          If {@link SubscriptionAdminClient} cannot be created.
+   */
+  public static void createSubscription(BooleanSupplier preCheck, BackoffConfig backoffConfig, String subscription,
```
tivv, line 126: This preCheck looks like a `Predicate`. signatures are similar, but `Predicate` is sematically better suited here.

src/main/java/io/cdap/plugin/gcp/publisher/source/SerializableFunction1.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Interface used to mark a {@link Function1} serializable.
+ *
+ * @param <T> the type of the input to the function
+ * @param <R> the type of the result of the function
+ */
+public interface SerializableFunction1<T, R> extends Function1<T, R>, Serializable {
```
tivv: Note: If you'll use the class instead of Lambda you won't need this, class can directly implement both interfaces

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriberConfig.java
greeshmaswaminathan: I believe this should be because Config is used in Receiver (runs on executors). Not changed from existing code.

tivv: I mean, if it's used in existing code that is persisted, the loading would fail because class names changed and old classes are not present.

greeshmaswaminathan: Ah, got it. I don't think this is persisted anywhere other than checkpoints. That should not matter since we don't support upgrade on checkpointed apps.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
greeshmaswaminathan: The original code was using it as an anonymous function. I can make this change.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
greeshmaswaminathan: Predicate takes an argument, in this case the function used in original code did not need one.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
greeshmaswaminathan: Part of existing code, but made the change.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
greeshmaswaminathan: Changed.

src/main/java/io/cdap/plugin/gcp/publisher/source/SerializableFunction1.java
greeshmaswaminathan: Removed it . Converted to class instead of lambda. Will make the additional implementation changes in the next PR.

src/main/java/io/cdap/plugin/gcp/publisher/source/BackoffConfig.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package io.cdap.plugin.gcp.publisher.source;
+
+import java.io.Serializable;
+
+/**
+ * Class used to configure exponential backoff for Pub/Sub API requests.
+ */
+public class BackoffConfig implements Serializable {
+  final int initialBackoffMs;
```
albertshau: these are better to be private, they probably were not before because it was an inner class.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
```diff
@@ -0,0 +1,133 @@
...
+      attributeMap.putAll(pubSubMessage.getAttributes());
+    }
+
+    try {
+      StructuredRecord payload = getStructuredRecord(config, customMessageSchema, format, pubSubMessage);
+
+      return StructuredRecord.builder(outputSchema)
+        .set("message", (format.equalsIgnoreCase(PubSubConstants.TEXT) ||
+          format.equalsIgnoreCase(PubSubConstants.BLOB)) ?
+          pubSubMessage.getData() : payload)
```
albertshau, line 67: this looks like existing data, but it seems wrong. text format should be a string, not a byte array. Does it work properly?

albertshau (approved):
minor comment. Also have a question about the existing logic, unrelated to this refactoring.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubStructuredRecordConverter.java
greeshmaswaminathan: Right now text is default format and the schema for message is bytes. Created https://cdap.atlassian.net/browse/PLUGIN-1593 to check further.

src/main/java/io/cdap/plugin/gcp/publisher/source/BackoffConfig.java
greeshmaswaminathan: Done.

PR #1241, opened 2023-04-25
[PLUGIN-1537] - State management changes for Pub/Sub streaming source.

greeshmaswaminathan:
[PLUGIN-1537](https://cdap.atlassian.net/browse/PLUGIN-1537)
- Direct dstream implementation is used when state store is enabled.
- A snapshot is created before each batch and deleted after the batch completes successfully.
- Seek is only done on pipeline restarts if a snapshot is present in saved state and on job restarts.
- RDD reads data using a lazy iterator that reads a fixed amount of records at a time and acknowledges them. 
- RDD finishes processing once the read data is processed and batch duration is complete.
- Retries for stages and tasks are disabled.
- RDD partition count is same as the number of readers config in the plugin.

[PLUGIN-1537]: https://cdap.atlassian.net/browse/PLUGIN-1537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <avro.version>1.8.2</avro.version>
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
-    <cdap.version>6.9.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.9.1-SNAPSHOT</cdap.version>
     <cdap.plugin.version>2.11.0-SNAPSHOT</cdap.plugin.version>
     <dropwizard.metrics-core.version>3.2.6</dropwizard.metrics-core.version>
     <flogger.system.backend.version>0.3.1</flogger.system.backend.version>
@@ -886,8 +886,8 @@
         <version>1.1.0</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.8.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[6.8.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline[6.9.1-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-streams[6.9.1-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
           </cdapArtifacts>
         </configuration>
         <executions>
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
@@ -24,8 +24,12 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.streaming.StreamingSource;
 import io.cdap.cdap.etl.api.streaming.StreamingSourceContext;
+import io.cdap.cdap.etl.api.streaming.StreamingStateHandler;
+import io.cdap.cdap.features.Feature;
 import io.cdap.plugin.common.LineageRecorder;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 /**
@@ -34,7 +38,7 @@ import java.util.stream.Collectors;
 @Plugin(type = StreamingSource.PLUGIN_TYPE)
 @Name("GoogleSubscriber")
 @Description("Streaming Source to read messages from Google PubSub.")
-public class GoogleSubscriber extends PubSubSubscriber<StructuredRecord> {
+public class GoogleSubscriber extends PubSubSubscriber<StructuredRecord> implements StreamingStateHandler {
 
   private GoogleSubscriberConfig config;
 
@@ -52,6 +56,17 @@ public class GoogleSubscriber extends PubSubSubscriber<StructuredRecord> {
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     config.validate(collector);
     pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    if (!pipelineConfigurer.isFeatureEnabled(Feature.STREAMING_PIPELINE_NATIVE_STATE_TRACKING.getFeatureFlagString())) {
+      return;
+    }
+
+    // A seek is required on the snapshot for preventing data loss when a failure retry happens.
+    // To ensure this happens and is consistent for all partitions, task and stage retries are disabled.
+    // Job has a retry which will make sure seek is done before retrying .
+    Map<String, String> additionalProps = new HashMap<>();
+    additionalProps.put("spark.task.maxFailures", "1");
+    additionalProps.put("spark.stage.maxConsecutiveAttempts", "1");
+    pipelineConfigurer.setPipelineProperties(additionalProps);
   }
 
   @Override
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
@@ -1,0 +1,291 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.gcp.publisher.source;
+
+import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.auth.Credentials;
+import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
+import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.pubsub.v1.CreateSnapshotRequest;
+import com.google.pubsub.v1.ProjectSnapshotName;
+import com.google.pubsub.v1.ProjectSubscriptionName;
+import com.google.pubsub.v1.SeekRequest;
+import com.google.pubsub.v1.Snapshot;
+import com.google.pubsub.v1.TopicName;
+import io.cdap.cdap.etl.api.streaming.StreamingEventHandler;
+import io.cdap.plugin.gcp.common.GCPUtils;
+import org.apache.spark.rdd.RDD;
+import org.apache.spark.streaming.StreamingContext;
+import org.apache.spark.streaming.Time;
+import org.apache.spark.streaming.dstream.InputDStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.Option;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * PubSubDirectDStream implementation.
+ * A snapshot is taken and saved before each batch and deleted on successful completion.
+ * On retrying a batch, snapshot is restored.
+ *
+ * @param <T> Type of object returned by RDD
+ */
+public class PubSubDirectDStream<T> extends InputDStream<T> implements StreamingEventHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(PubSubDirectDStream.class);
+  private static final String CDAP_PIPELINE = "cdap_pipeline";
+
+  private final Credentials credentials;
+  private final PubSubSubscriberConfig config;
+  private final long readDuration;
+  private final io.cdap.cdap.etl.api.streaming.StreamingContext context;
+  private final boolean autoAcknowledge;
+  private final SerializableFunction<PubSubMessage, T> mappingFn;
+  private final StreamingContext streamingContext;
+  private final String pipeline;
+
+  private SubscriptionAdminClient subscriptionAdminClient;
+  private ProjectSnapshotName currentSnapshotName;
+  private boolean takeSnapshot;
+
+  public PubSubDirectDStream(io.cdap.cdap.etl.api.streaming.StreamingContext context, PubSubSubscriberConfig config,
+                             long readDuration, boolean autoAcknowledge,
+                             SerializableFunction<PubSubMessage, T> mappingFn) {
+    super(context.getSparkStreamingContext().ssc(), scala.reflect.ClassTag$.MODULE$.apply(PubSubMessage.class));
+    this.streamingContext = context.getSparkStreamingContext().ssc();
+    this.config = config;
+    this.readDuration = readDuration;
+    this.context = context;
+    this.autoAcknowledge = autoAcknowledge;
+    this.mappingFn = mappingFn;
+    this.pipeline = context.getPipelineName();
+    this.credentials = createCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+  }
+
+  @Override
+  public Option<RDD<T>> compute(Time validTime) {
+    LOG.debug("Computing RDD for time {}.", validTime);
+    PubSubRDD pubSubRDD = new PubSubRDD(streamingContext.sparkContext(), validTime, readDuration, config,
+                                        autoAcknowledge, credentials);
+    RDD<T> mapped = pubSubRDD.map(mappingFn, scala.reflect.ClassTag$.MODULE$.apply(PubSubMessage.class));
+    return Option.apply(mapped);
+  }
+
+  @Override
+  public void start() {
+    if (config.getTopic() != null) {
+      try {
+        createSubscriptionIfNotPresent();
+      } catch (IOException | InterruptedException e) {
+        throw new RuntimeException("Subscription creation failed.", e);
+      }
+    }
+
+    try {
+      subscriptionAdminClient = buildSubscriptionAdminClient(credentials);
+    } catch (IOException e) {
+      throw new RuntimeException("SubscriptionAdminClient creation failed.", e);
+    }
+
+    // If state with snapshot is present, seek to it.
+    this.currentSnapshotName = fetchSnapShot(config.getSubscription(), context);
+    if (this.currentSnapshotName != null) {
+      seekSnapshot(currentSnapshotName, ProjectSubscriptionName.of(config.getProject(), config.getSubscription()));
+      // Can use the same snapshot till end of the batch.
+      takeSnapshot = false;
+    } else {
+      takeSnapshot = true;
+    }
+  }
+
+  @Override
+  public void stop() {
+    if (subscriptionAdminClient == null) {
+      return;
+    }
+
+    subscriptionAdminClient.close();
+  }
+
+  @Override
+  public void onBatchStarted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Starting a batch.");
+    if (!takeSnapshot) {
+      return;
+    }
+
+    String generatedSnapshotName = generateName(config.getSubscription());
+    ProjectSnapshotName projectSnapshotName = ProjectSnapshotName.of(config.getProject(), generatedSnapshotName);
+    ProjectSubscriptionName projectSubscriptionName = ProjectSubscriptionName.of(config.getProject(),
+                                                                                 config.getSubscription());
+    // Create in Pub/Sub
+    Snapshot snapshot = createSnapshot(projectSnapshotName, projectSubscriptionName);
+    // Save snapshot as state
+    try {
+      saveSnapshotAsState(snapshot, config.getSubscription(), context);
+    } catch (IOException e) {
+      // Delete the snapshot from Pub/Sub to avoid abandoned snapshots.
+      // Snapshots have a max expiry of 7 days after which it gets auto deleted.
+      deleteSnapshot(projectSnapshotName);
+
+      // Retries are already part of the state saving library, so just throw here.
+      throw new RuntimeException("Error while saving state.", e);
+    }
+    this.currentSnapshotName = projectSnapshotName;
+  }
+
+  @Override
+  public void onBatchRetry(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Batch is about to be retried. Seeking to snapshot {} for the current batch.", currentSnapshotName);
+    seekSnapshot(currentSnapshotName, ProjectSubscriptionName.of(config.getProject(), config.getSubscription()));
+  }
+
+  @Override
+  public void onBatchCompleted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Batch completed called.");
+    try {
+      streamingContext.deleteState(config.getSubscription());
+    } catch (IOException e) {
+      throw new RuntimeException("Deleting state failed. ", e);
+    }
+    // delete the snapshot from Pub/Sub .
+    deleteSnapshot(currentSnapshotName);
+
+    // For next batch
+    takeSnapshot = true;
+    LOG.debug("Batch completed successfully. Deleted snapshot {} for the current batch.", currentSnapshotName);
+  }
+
+  private String generateName(String subscription) {
+    // Rules for name
+    // Not begin with the string goog
+    // Start with a letter
+    // Contain between 3 and 255 characters
+    // Contain only the following characters:
+    // Letters [A-Za-z], numbers [0-9], dashes -, underscores _, periods ., tildes ~, plus signs +, and percent signs %
+
+    // Starting with subscription name should take care of the beginning reqs.
+    String uuidString = UUID.randomUUID().toString();
+    int maxLenForSubscriptionPrefix = 255 - uuidString.length() - 1;
+    String subscriptionPrefix = subscription.length() > maxLenForSubscriptionPrefix ?
+      subscription.substring(0, maxLenForSubscriptionPrefix) : subscription;
+    return String.format("%s-%s", subscriptionPrefix, uuidString);
+  }
+
+  private void createSubscriptionIfNotPresent() throws IOException, InterruptedException {
+    PubSubSubscriberUtil.createSubscription(() -> true, BackoffConfig.defaultInstance(),
+                                            ProjectSubscriptionName.format(config.getProject(),
+                                                                           config.getSubscription()),
+                                            TopicName.format(config.getProject(), config.getTopic()),
+                                            () -> subscriptionAdminClient,
+                                            PubSubSubscriberUtil::isApiExceptionRetryable);
+  }
+
+  private void seekSnapshot(ProjectSnapshotName projectSnapshotName,
+                            ProjectSubscriptionName projectSubscriptionName) {
+    try {
+      subscriptionAdminClient.seek(SeekRequest.newBuilder().setSnapshot(projectSnapshotName.toString())
+                                     .setSubscription(projectSubscriptionName.toString())
+                                     .build());
+    } catch (NotFoundException e) {
+      throw new RuntimeException(String.format(
+        "Saved snapshot %s not found. Please clear the application state to proceed. " +
+          "REST api for state deletion is namespaces/{namespace-id}/apps/{app-id}/state .",
+        projectSnapshotName.toString()), e);
+    }
+  }
+
+  @Nullable
+  private ProjectSnapshotName fetchSnapShot(String subscriptionId,
+                                            io.cdap.cdap.etl.api.streaming.StreamingContext context) {
+    Optional<byte[]> state = null;
+    try {
+      state = context.getState(subscriptionId);
+    } catch (IOException e) {
+      throw new RuntimeException(String.format("Error fetching saved state for subscription %s.", subscriptionId), e);
+    }
+
+    if (!state.isPresent()) {
+      LOG.debug("No saved state for {}.", subscriptionId);
+      return null;
+    }
+
+    try {
+      Snapshot snapshot = Snapshot.parseFrom(state.get());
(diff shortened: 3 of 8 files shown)


src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -52,6 +56,15 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     config.validate(collector);
     pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    if (!pipelineConfigurer.isFeatureEnabled(Feature.STREAMING_PIPELINE_NATIVE_STATE_TRACKING.getFeatureFlagString())) {
+      return;
+    }
+
+    // No retry at task and stage level. Jobs will be retried.
```
albertshau: comment should also describe why we need to do this

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+ * On retrying a batch, snapshot is restored.
+ *
+ * @param <T> Type of object returned by RDD
+ */
+public class PubSubDirectDStream<T> extends InputDStream<T> implements StreamingEventHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(PubSubDirectDStream.class);
+
+  private SubscriptionAdminClient subscriptionAdminClient;
+  private final Credentials credentials;
```
albertshau, line 59: nit: group all the final variables together

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  private PubSubSubscriberConfig config;
```
albertshau: can this (and everything else passed in the constructor) be final? Didn't see them being set anywhere.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    this.credentials = createCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+  }
+
+  @Override
+  public Option<RDD<T>> compute(Time validTime) {
+    LOG.debug("Computing RDD for time {}.", validTime);
+    PubSubRDD pubSubRDD = new PubSubRDD(streamingContext.sparkContext(), readDuration, config, autoAcknowledge,
+                                        credentials);
+    final SerializableFunction<PubSubMessage, T> localMappingFn = mappingFn;
+    // SerializableFunction1<PubSubMessage, T> pubSubMessageTFunction1 = ;
```
albertshau: can be removed

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    this.mappingFn = mappingFn;
+    this.credentials = createCredentials(config.getServiceAccount(), config.isServiceAccountFilePath());
+  }
+
+  @Override
+  public Option<RDD<T>> compute(Time validTime) {
+    LOG.debug("Computing RDD for time {}.", validTime);
+    PubSubRDD pubSubRDD = new PubSubRDD(streamingContext.sparkContext(), readDuration, config, autoAcknowledge,
+                                        credentials);
+    final SerializableFunction<PubSubMessage, T> localMappingFn = mappingFn;
```
albertshau: is there a reason we need this variable vs just using mappingFn?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  @Override
+  public void onBatchRetry(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Batch is about to be retried. Seeking to snapshot {} for the current batch.", currentSnapshotName);
+    seekSnapshot(config.getProject(), config.getSubscription(), currentSnapshotName);
+  }
+
+  @Override
+  public void onBatchCompleted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    try {
+      streamingContext.deleteSate(config.getSubscription());
```
albertshau: maybe good to wait for the rename to be published before merging

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  private void createSubscriptionIfNotPresent() {
+    try {
+      PubSubSubscriberUtil.createSubscription(() -> true, BackoffConfig.defaultInstance(),
+                                              ProjectSubscriptionName.format(config.getProject(),
+                                                                             config.getSubscription()),
+                                              TopicName.format(config.getProject(), config.getTopic()),
+                                              () -> subscriptionAdminClient,
+                                              PubSubSubscriberUtil::isApiExceptionRetryable);
+    } catch (Exception e) {
+      throw new RuntimeException("Subscription creation failed.", e);
```
albertshau: think it would be better to let the exception propagate up instead of wrapping it here.

Seems like it's only called from start(), what is the behavior if start() throws an exception?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    LOG.debug("Creating snapshot {} in Pub/Sub .", snapshotName);
+    Snapshot snapshot = createSnapshot(projectId, subscriptionId, snapshotName);
+    saveSnapshot(snapshot, subscriptionId, context);
+    return snapshot.getName();
+  }
+
+  private Snapshot createSnapshot(String projectId, String subscriptionId, String snapshotName) {
+    CreateSnapshotRequest request = CreateSnapshotRequest.newBuilder()
+      .setName(ProjectSnapshotName.of(projectId, snapshotName).toString())
+      .setSubscription(ProjectSubscriptionName.of(projectId, subscriptionId).toString()).putAllLabels(new HashMap<>())
```
albertshau: might be good to set some labels to indicate it is created by the plugin. For example, could add a pipeline label with the pipeline name as the value

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+      LOG.debug("Found existing snapshot {} .", snapshot.getName());
+      return snapshot.getName();
+    } catch (InvalidProtocolBufferException e) {
+      throw new RuntimeException(String.format("Error parsing saved state for subscription %s.", subscriptionId), e);
+    }
+  }
+
+  private String createAndSave(String projectId, String subscriptionId, String snapshotName,
+                               io.cdap.cdap.etl.api.streaming.StreamingContext context) {
+    LOG.debug("Creating snapshot {} in Pub/Sub .", snapshotName);
```
albertshau: should we mention the subscription id too?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    Snapshot snapshot = createSnapshot(projectId, subscriptionId, snapshotName);
+    saveSnapshot(snapshot, subscriptionId, context);
+    return snapshot.getName();
+  }
+
+  private Snapshot createSnapshot(String projectId, String subscriptionId, String snapshotName) {
+    CreateSnapshotRequest request = CreateSnapshotRequest.newBuilder()
+      .setName(ProjectSnapshotName.of(projectId, snapshotName).toString())
```
albertshau: does snapshot name have a character length limit or any other character restrictions that we should be aware of?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  public void onBatchStarted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Starting a batch.");
+    if (!takeSnapshot) {
+      this.currentSnapshotName = existingSnapshot;
+      return;
+    }
+
+    String subscription = config.getSubscription();
+    this.currentSnapshotName = createAndSave(config.getProject(), subscription,
+                                             subscription + "-snapshot-" + UUID.randomUUID(), context);
```
albertshau: Snapshots are already tied to a subscription, do we need the subscription name in the snapshot name? Is it because the console shows all snapshots in one page instead of grouping them per subscription? Or it expects the subscription name to be part of the snapshot name?

In any case, it would be good to add some comments to clarify

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  @Override
+  public void onBatchCompleted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    try {
+      streamingContext.deleteSate(config.getSubscription());
+    } catch (IOException e) {
+      throw new RuntimeException("Deleting state failed. ", e);
+    }
+    // delete the snapshot from Pub/Sub
```
albertshau: would be good to add comments on how long this normally takes. If it's a while, we may want to add a TODO for making this async.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+
+  private String seekSnapshot(String projectId, String subscriptionId, String snapshotName) {
+    try {
+      ProjectSnapshotName projectSnapshotName = ProjectSnapshotName.parse(snapshotName);
+      subscriptionAdminClient.seek(SeekRequest.newBuilder().setSnapshot(projectSnapshotName.toString())
+                                     .setSubscription(ProjectSubscriptionName.of(projectId, subscriptionId).toString())
+                                     .build());
+      return snapshotName;
+    } catch (NotFoundException e) {
+      //Saved snapshot is not found. Create one and save it.
```
albertshau: this comment should be updated

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException("Error saving state. ", e);
+    }
+  }
+
+  private SubscriptionAdminClient buildSubscriptionAdminClient(Credentials credentials) throws IOException {
+    SubscriptionAdminSettings.Builder builder = SubscriptionAdminSettings.newBuilder();
+
+    if (credentials != null) {
+      builder.setCredentialsProvider(FixedCredentialsProvider.create(credentials));
```
albertshau, line 272: do the Credentials get refreshed by the client library?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
albertshau: I see elsewhere we're using ProjectSnapshotName and Snapshot. I'm a bit confused about the difference, but could either of those classes be used here rather than manually concatenating?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    subscriptionAdminClient.close();
+  }
+
+  @Override
+  public void onBatchStarted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Starting a batch.");
+    if (!takeSnapshot) {
+      this.currentSnapshotName = existingSnapshot;
```
albertshau: do we need existingSnapshot? I may be missing something, but it seems like we can just have the currentSnapshot variable?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+
+    String subscription = config.getSubscription();
+    this.currentSnapshotName = createAndSave(config.getProject(), subscription,
+                                             subscription + "-snapshot-" + UUID.randomUUID(), context);
+  }
+
+  @Override
+  public void onBatchRetry(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    LOG.debug("Batch is about to be retried. Seeking to snapshot {} for the current batch.", currentSnapshotName);
+    seekSnapshot(config.getProject(), config.getSubscription(), currentSnapshotName);
```
albertshau: seems like this should also set `takeSnapshot` to false?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  @Override
+  public void onBatchCompleted(io.cdap.cdap.etl.api.streaming.StreamingContext streamingContext) {
+    try {
+      streamingContext.deleteSate(config.getSubscription());
+    } catch (IOException e) {
```
albertshau, line 169: should we change the interface to allow throwing IOException for this and onBatchStarted/onBatchRetry? Most implementations will do something with state.

albertshau: Also, would be good to add more detail to the javadoc on what happens if an exception is thrown. Is onBatchRetry() called?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,268 @@
...
+    subscriptionAdminClient.deleteSnapshot(projectSnapshotName);
+  }
+
+  private void saveSnapshot(Snapshot snapshot, String subscription,
+                            io.cdap.cdap.etl.api.streaming.StreamingContext context) {
+    try {
+      LOG.debug("Saving snapshot {} in state .", snapshot.getName());
+      context.saveState(subscription, snapshot.toByteArray());
+    } catch (IOException e) {
+      throw new RuntimeException("Error saving state. ", e);
```
albertshau: think it would be better to have this method throws IOException instead of wrapping as a RuntimeException. Will make it easier to reason about what happens if this fails.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
albertshau: let's say CDAP is down so the saveState() call keeps failing. That means we'll create a new snapshot each time this method is called, will that be a problem? Will we eventually hit some max snapshot quota or do older ones automatically get deleted?

It think it would be good to separate the calls instead of having a single createAndSave method to make it more clear that there are two calls happening, both of which can potentially fail.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -0,0 +1,169 @@
...
+      return true;
+    } catch (IOException e) {
+      throw new RuntimeException("Error reading messages from Pub/Sub. ", e);
+    }
+  }
+
+  @Override
+  public PubSubMessage next() {
+    if (receivedMessages.isEmpty()) {
+      throw new RuntimeException("Unexpected state. No messages available.");
```
albertshau: nit: IllegalStateException would be more descriptive here. Would be good to also add a comment that this should never happen.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -0,0 +1,169 @@
...
+  private List<String> ackIds;
+  private long messageCount;
+
+  public PubSubRDDIterator(PubSubSubscriberConfig config, TaskContext context, long batchDuration,
+                           boolean autoAcknowledge, Credentials credentials) {
+    this.config = config;
+    this.context = context;
+    this.batchDuration = batchDuration;
+    this.credentials = credentials;
+    this.startTime = System.currentTimeMillis();
```
albertshau: shouldn't this be the Time from InputDStream's `compute(Time validTime)` method?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -0,0 +1,169 @@
...
+    // Complete processing of acknowledged messages before finishing the batch.
+    if (!receivedMessages.isEmpty()) {
+      return true;
+    }
+
+    long currentTimeMillis = System.currentTimeMillis();
+    if (currentTimeMillis >= (startTime + batchDuration)) {
+      LOG.debug("Time exceeded for batch. Total time is {} millis. Total messages returned is {} .",
+                currentTimeMillis - startTime, messageCount);
+      return false;
```
albertshau, line 88: since we've already acked the messages, doesn't this mean anything in the receivedMessages queue will get skipped? should we always go through the list, even if we're past the duration?

greeshmaswaminathan: The check above this will make sure that if queue has pending messages hasNext() will return true. Only when queue is empty, the time is checked.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: DStream's compute time is the time when the job is generated. The actual time when the RDD is reading data from Pub/Sub may be later depending on the processing delays for each batch. Current logic reads for the whole batch duration regardless of when the batch was scheduled. If we consider batch time instead, some batches may not be able to read data.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Since batch is retried from the sink runnable, onBatchStarted will not be called again. So the snapshot creation will not be happen for retries.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
albertshau: Isn't it better not to read data in those situations though? It would basically be doing backpressure.

The concern is if it consistently takes longer to process the microbatch, the jobs will keep piling up. For example, if there are more partitions than executors available to process them in parallel, we can run into this situation.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
albertshau: this is not obvious from the API, can we be sure to update the javadocs to clarify?

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: No, from a previous implementation with lambda. Removed.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Wrapping as a runtime exception since this is an overridden method. Exception thrown from start() causes the pipeline to fail.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Added pipeline label.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Added new logic for ensuring the length is within limit.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Though snapshot is taken for subscription, it is tied only to topic. Any subscription on the topic could use this snapshot to seek. No strong reason to use subscription in name other than identifying them easily in the console as you mentioned. I have refactored a bit to avoid confusion between the use of various classes for snapshot. 

Separated the calls to saveState and snapshot creation. Also included a delete snapshot when state saving fails.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done .

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: No, this is intended to be always fixed. Since it is service account based, refresh is not necessary right?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Removed.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Will create a follow up PR with javadoc update to API in cdap.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Will create a follow up PR with javadoc update to API in cdap.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: Makes sense. Made the change to use batch time.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
albertshau: yeah I think it's not necessary, I see a refresh method on the Credentials class itself.

pom.xml
```diff
@@ -72,7 +72,7 @@
     <avro.version>1.8.2</avro.version>
     <bigquery.connector.hadoop2.version>hadoop2-1.0.0</bigquery.connector.hadoop2.version>
     <commons.codec.version>1.4</commons.codec.version>
-    <cdap.version>6.9.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.9.1-SNAPSHOT</cdap.version>
```
albertshau, line 75: need to update the parent lower bound as well

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
```diff
@@ -52,6 +56,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     config.validate(collector);
     pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    if (!pipelineConfigurer.isFeatureEnabled(Feature.STREAMING_PIPELINE_NATIVE_STATE_TRACKING.getFeatureFlagString())) {
```
albertshau, line 59: the feature flag can be on while the pipeline still decides to use Spark checkpointing right? Do we need to expose the checkpointing mode?

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,292 @@
...
+    return Option.apply(mapped);
+  }
+
+  @Override
+  public void start() {
+    if (config.getTopic() != null) {
+      try {
+        createSubscriptionIfNotPresent();
+      } catch (IOException e) {
+        throw new RuntimeException("Subscription creation failed.", e);
```
albertshau: these can be combined into one catch

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
```diff
@@ -0,0 +1,292 @@
...
+    ProjectSubscriptionName projectSubscriptionName = ProjectSubscriptionName.of(config.getProject(),
+                                                                                 config.getSubscription());
+    // Create in Pub/Sub
+    Snapshot snapshot = createSnapshot(projectSnapshotName, projectSubscriptionName);
+    // Save snapshot as state
+    try {
+      saveSnapshotAsState(snapshot, config.getSubscription(), context);
+    } catch (IOException e) {
+      // Delete the snapshot from Pub/Sub to avoid abandoned snapshots.
+      deleteSnapshot(projectSnapshotName);
```
albertshau, line 150: though unlikely, this can fail too. Is there any concept of a TTL for snapshots? If so, would be good to mention in a comment

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -0,0 +1,169 @@
...
+          subscriber.shutdown();
+          try {
+            subscriber.awaitTermination(30, TimeUnit.SECONDS);
+          } catch (InterruptedException e) {
+            LOG.warn("Exception in shutting down subscriber. ", e);
+          }
+        }
+      });
+      this.pullRequest = PullRequest.newBuilder()
+        .setMaxMessages(MAX_MESSAGES)
```
albertshau, line 140: this seems like it should be configurable, especially since we don't know the message size. Though looks like we don't have it right now, so can be done in a future PR.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
albertshau: though I guess it's not wrong to set these when using checkpointing?

albertshau (commented):
minor comments/questions

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Yes, typical expiry and deletion is after 7 days. Can be lesser depending on the oldest unacknowledged message expiry time.

src/main/java/io/cdap/plugin/gcp/publisher/source/GoogleSubscriber.java
greeshmaswaminathan: Hmm, yes, there are a few additional check for using state store like single source, no windower etc. Slightly more involved way would be to add `isRetryable` method in StreamingStateHandler and check and add these in DataStreamsSparkLauncher while launching the pipeline.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: Makes sense, Kafka has a similar option. Created https://cdap.atlassian.net/browse/PLUGIN-1598 to track this.

pom.xml
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubDirectDStream.java
greeshmaswaminathan: Done.

albertshau (approved):

PR #1245, opened 2023-05-07
PubSub state management fixes.

greeshmaswaminathan:
[PLUGIN-1603](https://cdap.atlassian.net/browse/PLUGIN-1603) : Fix NPE when subscription is not present and has to be created. 

[PLUGIN-1604](https://cdap.atlassian.net/browse/PLUGIN-1604): Read messages till end of batch duration .
Addresses the following cases:
- It is possible that PubSub synchronous pull returns 0 messages even when messages are available. (https://cloud.google.com/pubsub/docs/pull#unary_pull_code_samples ) 
- Current implementation returns when no messages are found and will not account for the messages published later in the batch duration.

[PLUGIN-1598](https://cdap.atlassian.net/browse/PLUGIN-1598): Sets max message size for each request.

[PLUGIN-1603]: https://cdap.atlassian.net/browse/PLUGIN-1603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLUGIN-1604]: https://cdap.atlassian.net/browse/PLUGIN-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLUGIN-1598]: https://cdap.atlassian.net/browse/PLUGIN-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
+++ b/src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import scala.collection.Iterator;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedDeque;
@@ -48,6 +49,8 @@ public class PubSubRDDIterator implements Iterator<PubSubMessage> {
 
   private static final Logger LOG = LoggerFactory.getLogger(PubSubRDDIterator.class);
   private static final int MAX_MESSAGES = 1000;
+  private static final int MAX_MESSAGE_SIZE = 20 * 1024 * 1024; //20 MB. Max size for "data" field of a message is 10MB.
+  private static final int RETRY_DELAY = 100;
 
   private final long startTime;
   private final PubSubSubscriberConfig config;
@@ -82,14 +85,15 @@ public class PubSubRDDIterator implements Iterator<PubSubMessage> {
     }
 
     long currentTimeMillis = System.currentTimeMillis();
-    if (currentTimeMillis >= (startTime + batchDuration)) {
+    long expiryTimeMillis = startTime + batchDuration;
+    if (currentTimeMillis >= expiryTimeMillis) {
       LOG.debug("Time exceeded for batch. Total time is {} millis. Total messages returned is {} .",
                 currentTimeMillis - startTime, messageCount);
       return false;
     }
 
     try {
-      List<ReceivedMessage> messages = fetchAndAck();
+      List<ReceivedMessage> messages = fetchAndAck(expiryTimeMillis);
       //If there are no messages to process, continue.
       if (messages.isEmpty()) {
         LOG.debug("No more messages. Total messages returned is {} .", messageCount);
@@ -119,11 +123,14 @@ public class PubSubRDDIterator implements Iterator<PubSubMessage> {
     if (credentials != null) {
       builder.setCredentialsProvider(FixedCredentialsProvider.create(credentials));
     }
+    builder.setTransportChannelProvider(
+      SubscriberStubSettings.defaultGrpcTransportProviderBuilder()
+        .setMaxInboundMessageSize(MAX_MESSAGE_SIZE).build());
     builder.getSubscriptionSettings().setRetrySettings(PubSubSubscriberUtil.getRetrySettings());
     return GrpcSubscriberStub.create(builder.build());
   }
 
-  private List<ReceivedMessage> fetchAndAck() throws IOException {
+  private List<ReceivedMessage> fetchAndAck(long expiryTimeMillis) throws IOException {
     if (this.subscriber == null) {
       this.subscriber = buildSubscriberClient();
       context.addTaskCompletionListener(context1 -> {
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubSubscriberUtil.java
```diff
@@ -148,8 +148,8 @@ public static void createSubscription(BooleanSupplier preCheck, BackoffConfig ba
 
     while (preCheck.getAsBoolean() && attempts-- > 0) {
 
-      try (SubscriptionAdminClient subscriptionAdminClient = clientSupplier.get()) {
-
+      try {
+        SubscriptionAdminClient subscriptionAdminClient = clientSupplier.get();
```
fernst, line 152: Does this client need to be manually closed after the request ends?

greeshmaswaminathan: Yes. The Receiver will close it immediately, and the PubSubDirectDStream will re-use it for other operations as well and close only in the stop() method.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -119,11 +122,14 @@ private SubscriberStub buildSubscriberClient() throws IOException {
...
     }
+    builder.setTransportChannelProvider(
+      SubscriberStubSettings.defaultGrpcTransportProviderBuilder()
+        .setMaxInboundMessageSize(MAX_MESSAGE_SIZE).build());
     builder.getSubscriptionSettings().setRetrySettings(PubSubSubscriberUtil.getRetrySettings());
     return GrpcSubscriberStub.create(builder.build());
   }
 
-  private List<ReceivedMessage> fetchAndAck() throws IOException {
+  private List<ReceivedMessage> fetchAndAck(long timeout) throws IOException {
```
albertshau: nit: the other name `expireTimeMillis` is better, timeout sounds like a duration and not a timestamp

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -142,16 +148,26 @@ private List<ReceivedMessage> fetchAndAck() throws IOException {
...
+    // request that returns messages." Since 0 messages do not indicate a lack of available messages, no delays are
+    // included  between retries.
+    // Messages will be read till batch duration completes, to account for messages published
+    // later in the batch. No documented quota on the number of synchronous subscriber requests and no issues observed
+    // in testing when repeated calls are made when no data is available in the subscription.
+    while (System.currentTimeMillis() < timeout) {
+      PullResponse pullResponse = subscriber.pullCallable().call(pullRequest);
+      List<ReceivedMessage> receivedMessagesList = pullResponse.getReceivedMessagesList();
+      if (receivedMessagesList.isEmpty()) {
+        continue;
```
albertshau, line 170: probably better to sleep a short time instead of busy-loop

albertshau: was looking at the client a little bit at https://github.com/googleapis/java-pubsub#pulling-messages, which seems to be pretty different than this API we're using. Have we looked at that one? Seems like it might be simpler to use here, as we wouldn't need to handle polling logic?

greeshmaswaminathan: Both approaches are documented here - https://cloud.google.com/pubsub/docs/pull#types_of_message_processing_modes . Preferred using the synchronous pull for controlling the number(and size) of messages that are acknowledged and pending processing.  Streaming pull is intended to decouple receiving and processing. The messages could accumulate and we need another mechanism to limit the number of messages read (may be streaming requests with smaller timeouts?).

albertshau: at a high level it seems like it would be pretty similar if the Queue were a bounded one. In any case, don't think we need to change the implementation right at this moment since it would be a larger change. Would still be good to sleep for some time though, unless that's somehow built into the pull call.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: Done.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
greeshmaswaminathan: Added sleep.

src/main/java/io/cdap/plugin/gcp/publisher/source/PubSubRDDIterator.java
```diff
@@ -150,15 +151,21 @@ private List<ReceivedMessage> fetchAndAck(long timeout) throws IOException {
...
-    while (System.currentTimeMillis() < timeout) {
+    while (System.currentTimeMillis() < expiryTimeMillis) {
       PullResponse pullResponse = subscriber.pullCallable().call(pullRequest);
       List<ReceivedMessage> receivedMessagesList = pullResponse.getReceivedMessagesList();
       if (receivedMessagesList.isEmpty()) {
+        // Delay before re-attempt
+        try {
+          TimeUnit.MILLISECONDS.sleep(RETRY_DELAY);
+        } catch (InterruptedException e) {
+          LOG.debug("Interrupted while waiting for retry. ", e);
```
albertshau, line 167: if interrupted, we should also return an empty list so that the iterator will stop.

greeshmaswaminathan: Done.

albertshau (approved):

PR #1257, opened 2023-05-30
[PLUGIN-1624] Emit bytes processed metric from BQ sink and BQ Execute

itsankit-google:


src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -218,6 +220,13 @@ void recordMetric(boolean succeeded, BatchSinkContext context) {
       context.getMetrics().count(RECORDS_UPDATED_METRIC, rowCount);
       totalRows -= rowCount;
     }
+
+    Map<String, String> tags = new ImmutableMap.Builder<String, String>()
+        .put(Constants.Metrics.Tag.APP_ENTITY_TYPE, BatchSink.PLUGIN_TYPE)
```
albertshau, line 225: this seems like something that should be done at the application level instead of forcing the plugins to know about these special tags and potentially putting the wrong values. For example, maybe just make the existing context.getMetrics() call add these tags? Can you open an improvement jira?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
```diff
@@ -257,6 +266,21 @@ private long getTotalRows(Job queryJob) {
...
+    if (type == JobConfiguration.Type.LOAD) {
+      long outputBytes = ((JobStatistics.LoadStatistics) queryJob.getStatistics()).getOutputBytes();
+      LOG.info("Job {} loaded {} bytes", queryJob.getJobId(), outputBytes);
+      return outputBytes;
+    } else if (type == JobConfiguration.Type.QUERY) {
+      long processedBytes = ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getTotalBytesProcessed();
+      LOG.info("Job {} processed {} bytes", queryJob.getJobId(), processedBytes);
+      return processedBytes;
+    }
+    LOG.warn("Unable to identify BigQuery job type. No metric will be emitted for the number of affected bytes.");
```
albertshau, line 281: is this ever expected? Would be good to add a comment in the code

albertshau (approved):

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
itsankit-google: sg, done: https://cdap.atlassian.net/browse/PLUGIN-1626

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySink.java
itsankit-google: done

PR #1283, opened 2023-08-28
gcscopy e2e tests scenarios

priyabhatnagar25:


src/e2e-test/java/io/cdap/plugin/gcscopy/actions/GCSCopyActions.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
```
GnsP: Copyright © 2023 Cask Data, Inc.

src/e2e-test/java/io/cdap/plugin/gcscopy/locators/GCSCopyLocators.java
```diff
@@ -0,0 +1,37 @@
...
+public class GCSCopyLocators {
+
+  @FindBy(how = How.XPATH, using = "//input[@data-cy='sourcePath']")
+  public static WebElement gcsCopySourcePath;
+
+  @FindBy(how = How.XPATH, using = "//input[@data-cy='destPath']")
+  public static WebElement gcsCopyDestinationPath;
+
+  @FindBy(how = How.XPATH, using = "//input[@data-cy='cmekKey']")
+  public static WebElement gcsCopyEncryptionKey;
```
GnsP: The `data-cy` attributes are for tests run with cypress. I think we should use `data-testid` attributes for locating elements with selenium (for consistency).

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+ * GCS Copy plugin related stepsdesign.
+ */
+
+public class GCSCopy {
+
+  @Then("Validate GCSCopy successfully copies object {string} to destination bucket")
+  public void validateGCSCopySuccessfullyCopiedObjectToDestinationBucket(String path) throws IOException {
+    String sourceGCSBucket = TestSetupHooks.gcsSourceBucketName;
+    String gcsObject = PluginPropertyUtils.pluginProp(path);
+    boolean isPresentatsource = false;
```
GnsP: nit: `isPresentAtSource`

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+import java.io.IOException;
+
+/**
+ * GCS Copy plugin related stepsdesign.
+ */
+
+public class GCSCopy {
+
+  @Then("Validate GCSCopy successfully copies object {string} to destination bucket")
+  public void validateGCSCopySuccessfullyCopiedObjectToDestinationBucket(String path) throws IOException {
```
GnsP, line 127: In this step only the existence of the object in both source and target buckets is validated. Shouldn't it also verify if the contents of the objects in both buckets are identical or not ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
+    Then Validate GCSCopy successfully copies object "gcsCsvFile" to destination bucket
```
GnsP, line 37: this step does not validate whether the contents of the object was copied successfully.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then Select radio button plugin property: "recursive" with value: "true"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
+    Then Validate GCSCopy successfully copies object "gcsCopyReadRecursivePath" to destination bucket
```
GnsP: does not check whether the contents (in this case the subdirectories) were copied successfully to the target bucket.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then Select radio button plugin property: "recursive" with value: "false"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
+    Then Validate GCSCopy did not copy subdirectory "gcsCopyReadRecursiveSubDirectory" to destination bucket
```
GnsP: this step does not validate the non-existence of the subdirectories in the target bucket.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,147 @@
...
+    Then Select radio button plugin property: "overwrite" with value: "true"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
+    Then Validate GCSCopy successfully copies object "gcsCsvFile" to destination bucket
```
GnsP, line 104: this step does not validate whether the contents of the objects in target bucket were overwritten or not.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Then("Enter runtime argument value for GCSCopy property destination path key {string}")
+  public void enterRuntimeArgumentValueForGCSCopyPropertyDestinationPathKey(String runtimeArgumentKey) {
+    ElementHelper.sendKeys(CdfStudioLocators.runtimeArgsValue(runtimeArgumentKey),
+                           "gs://" + TestSetupHooks.gcsTargetBucketName);
+  }
+
+  @Then("Validate GCSCopy did not copy subdirectory {string} to destination bucket")
+  public void validateGCSCopyDidNotCopySubdirectoryToDestinationBucket(String subdirectory) throws IOException {
```
GnsP: This seems to validate only the existence of the subdirectory in the source bucket. Shouldn't it also verify the non-existence of the subdirectory in the target bucket ?

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+
+public class GCSCopy {
+
+  @Then("Validate GCSCopy successfully copies object {string} to destination bucket")
+  public void validateGCSCopySuccessfullyCopiedObjectToDestinationBucket(String path) throws IOException {
+    String sourceGCSBucket = TestSetupHooks.gcsSourceBucketName;
+    String gcsObject = PluginPropertyUtils.pluginProp(path);
+    boolean isPresentatsource = false;
+    for (Blob blob : StorageClient.listObjects(sourceGCSBucket).iterateAll()) {
+      if (blob.getName().contains(gcsObject)) {
```
GnsP: why is this a `contains` check and not an `equals` check ?

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+
+    if (!isPresentatsource) {
+
+      Assert.fail("Object is not present in source bucket" + sourceGCSBucket);
+    }
+    BeforeActions.scenario.write("Object is copied from source GCS Bucket " + sourceGCSBucket + " successfully");
+    String targetGCSBucket = TestSetupHooks.gcsTargetBucketName;
+    try {
+      for (Blob blob : StorageClient.listObjects(targetGCSBucket).iterateAll()) {
+        if (blob.getName().contains(gcsObject)) {
```
GnsP: why is this a `contains` check and not an `equals` check ?

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+    ElementHelper.sendKeys(CdfStudioLocators.runtimeArgsValue(runtimeArgumentKey),
+                           "gs://" + TestSetupHooks.gcsTargetBucketName);
+  }
+
+  @Then("Validate GCSCopy did not copy subdirectory {string} to destination bucket")
+  public void validateGCSCopyDidNotCopySubdirectoryToDestinationBucket(String subdirectory) throws IOException {
+    String sourceGCSBucket = TestSetupHooks.gcsSourceBucketName;
+    String gcsObject = PluginPropertyUtils.pluginProp(subdirectory);
+    for (Blob blob : StorageClient.listObjects(sourceGCSBucket).iterateAll()) {
+      if (blob.getName().contains(gcsObject)) {
```
GnsP: why is this a contains check and not an equals check ?

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+      Assert.fail("Subdirectory is copied from source GCS Bucket " + sourceGCSBucket);
+    }
+
+  @Then("Validate GCSCopy failed to copy object {string} to destination bucket")
+  public void validateGCSCopyFailedToCopyObjectToDestinationBucket(String path) throws IOException {
+    String sourceGCSBucket = TestSetupHooks.gcsSourceBucketName;
+    String gcsObject = PluginPropertyUtils.pluginProp(path);
+    for (Blob blob : StorageClient.listObjects(sourceGCSBucket).iterateAll()) {
+      if (blob.getName().contains(gcsObject)) {
```
GnsP: why is this a contains check and not an equals check ?

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
```diff
@@ -0,0 +1,158 @@
...
+
+        BeforeActions.scenario.write("Subdirectory is not copied from source bucket" + sourceGCSBucket);
+        return;
+      }
+    }
+      Assert.fail("Subdirectory is copied from source GCS Bucket " + sourceGCSBucket);
+    }
+
+  @Then("Validate GCSCopy failed to copy object {string} to destination bucket")
+  public void validateGCSCopyFailedToCopyObjectToDestinationBucket(String path) throws IOException {
```
GnsP, line 217: this only seems to check the existence of the object in the source bucket. Shouldn't it also validate the non-existence of the object in the destination bucket ?

GnsP (changes requested):

src/e2e-test/java/io/cdap/plugin/gcscopy/actions/GCSCopyActions.java
priyabhatnagar25: Resolved , Changed the year to 2023.

src/e2e-test/java/io/cdap/plugin/gcscopy/locators/GCSCopyLocators.java
priyabhatnagar25: Resolved, Changed the attribute to data-testid

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Resolved , Changed to camel case

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Changed to equals check

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Changed to equals check.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Here it is the scenario to check that the sub-directories are not copied to the destination bucket.This is not working for equals check and the scenario is failing for it when equals check is applied. It shows the assertion error stating that subdirectory is copied from the source GCS bucket.Its working fine with contains check.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Changed to equals check.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: So here this step is used to check that the pipeline is getting failed when we are selecting overwrite existing file as false. Here we already have the file it the destination bucket and we are selecting overwrite option as false and the pipeline is failing.So here we already have the file in the destination bucket.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
rahuldash171: Hi ,
We will add the steps to verify the contents .

src/e2e-test/features/gcscopy/GCSCopy.feature
rahuldash171: Hi ,
We will add the steps to verify the contents .

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
rahuldash171: Hi , 
The default feature of gcsCopy plugin is moveAllsubDirectories : false . So here we are only validating if the file is copied to Sink . In few of the other scenarios where moveAllsubDirectories : true we are adding the steps to check the presence of subdirectories in sink folder .

src/e2e-test/features/gcscopy/GCSCopy.feature
rahuldash171: Hi , We have already added the steps for validating the presence of subdirectories in Sink Bucket . Do you want the test to validate the files present inside those copied subdirectories ?

GnsP: Yes, we should also check if the contents of the subdirectories were copied successfully. Ideally we should check the subdirectories recursively. But in this case, I guess the directory structure is 2 levels deep, so a simple check on the contents of the subdirectories should also work.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Hi, Validation step to check the contents of the objects in both buckets are identical or not is added. Please review it.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: It was not working with equals for the scenario when sub directories toggle button is set to true, so changed to contains check. As it has to check the subdirectories also.

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Added the step for validation to check the contents of the subdirectories  are also matched or not. Please review it.

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Added the step for the content verification.Please review it.

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Added the step for the content verification.Please review it.

src/e2e-test/features/gcscopy/GCSCopy.feature
rahuldash171: Hi ,
We have added the steps to verify the contents inside the files .

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Here also the target bucket is created from the hook and it doesnt have any subdirectory and the source bucket is checked that while marking subdirectories toggle button as false it must have the subdirectories in it. and at the last added the step for content verification.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: Added the step for content verification .Please review it.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    And Enter GCSCopy property source path "gcsCsvFile"
+    And Enter GCSCopy property destination path
+    Then Override Service account details if set in environment variables
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 34: Why do we need to check the logs in this scenario ? It seems to be an unnecessary check in this context.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
GnsP, line 36: May not need this step also, if checking logs is not necessary.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    And Enter GCSCopy property destination path
+    Then Override Service account details if set in environment variables
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Select radio button plugin property: "recursive" with value: "true"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 57: Why do we need to check the logs in this test ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
GnsP, line 59: May not need this step if we do not need to check the logs.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    And Enter GCSCopy property destination path
+    Then Override Service account details if set in environment variables
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Select radio button plugin property: "recursive" with value: "false"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 79: Why do we need to check the logs in this test ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
GnsP, line 81: May not need this step if we do not need to check the logs.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Enter GCSCopy property destination path "gcsCopyReadRecursivePath"
+    Then Override Service account details if set in environment variables
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Select radio button plugin property: "overwrite" with value: "true"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 101: Why do we need to check the logs in this test ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
GnsP, line 103: May not need this step if we do not need to check the logs.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Enter GCSCopy property destination path "gcsCopyReadRecursivePath"
+    Then Override Service account details if set in environment variables
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Select radio button plugin property: "overwrite" with value: "false"
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 124: Why do we need to check the logs in this test ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    And Verify the pipeline status is "Failed"
+    Then Close the pipeline logs
```
GnsP, line 126: May not need this step if we do not need to check the logs.

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    And Enter GCSCopy property destination path
+    Then Override Service account details if set in environment variables
+    Then Replace input plugin property: "location" with value: "locationEU"
+    Then Enter GCSCopy property encryption key name "cmekGCS" if cmek is enabled
+    Then Validate "GCS Copy" plugin properties
+    Then Close the Plugin Properties page
+    Then Save and Deploy Pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
```
GnsP, line 147: Why do we need to check the logs in this test ?

src/e2e-test/features/gcscopy/GCSCopy.feature
```diff
@@ -0,0 +1,153 @@
...
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
GnsP, line 149: May not need this step if we do not need to check the logs.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
GnsP: I think in case of subdirectories, we need to check if all the files in the subdirectory (in the fixtures) are present in the source bucket. 

In this context, using `String.contains` to check if the path of any of the objects in the bucket contains (as a substring) the path of the directory containing the test CSV files, seems to be fundamentally incorrect.

I think we need to restructure this check to work correctly in case of subdirectories, rather than trying to make this check pass with a substring (`String.contains`) check.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
GnsP: The comment regarding checking subdirectories in the source bucket also applies here.

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
GnsP: The comment regarding checking subdirectories in the source bucket (in the method `validateGCSCopySuccessfullyCopiedObjectToDestinationBucket`) also applies here.

GnsP (changes requested):

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Hi Ganesh , Open and Capture Pipeline logs are needed and we have used in the past for other tests across all plugins because when the report generates we can go through the logs in case of any failures during runtime . The same process we are following during Manual verification also for debugging we are dependent on raw logs

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Hi Ganesh , Open and Capture Pipeline logs are needed and we have used in the past for other tests across all plugins because when the report generates we can go through the logs in case of any failures during runtime . The same process we are following during Manual verification also for debugging we are dependent on raw logs

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Hi Ganesh , Open and Capture Pipeline logs are needed and we have used in the past for other tests across all plugins because when the report generates we can go through the logs in case of any failures during runtime . The same process we are following during Manual verification also for debugging we are dependent on raw logs

src/e2e-test/features/gcscopy/GCSCopy.feature
priyabhatnagar25: Hi Ganesh , Open and Capture Pipeline logs are needed and we have used in the past for other tests across all plugins because when the report generates we can go through the logs in case of any failures during runtime . The same process we are following during Manual verification also for debugging we are dependent on raw logs

src/e2e-test/java/io/cdap/plugin/gcscopy/stepsdesign/GCSCopy.java
priyabhatnagar25: 1. Hi, We have created two separate validation steps for subdirectories set to true and set to false. 
2. The first validation step when subdirectory sets to true is checking in the source bucket that the subdirectory present in the source bucket along with its file is copied to the destination bucket along with its file with **equals** check and validating both the source and the destination bucket.
3. The second validation step when subdirectory set to false is checking that there is one subdirectory along with its file in the source bucket but when the subdirectory toggle button is marked as false its not copying subdirectory along with its files in the destination bucket. So , the validation is checked at the destination bucket with **empty** check that if the destination bucket is empty it has not copied the subdirectory along with it files to the destination bucket.
4. We have already covered the scenario for the files (not inside any folder), when the bucket is only having the files and not subdirectory and by default overwrite and copy subdirectory toggle button is marked as **false.**

src/e2e-test/features/gcscopy/GCSCopy.feature
GnsP: got it. Thanks for clarifying.

src/e2e-test/features/gcscopy/GCSCopy.feature
GnsP: Clarified. Capturing logs is needed.

GnsP (approved):

PR #1296, opened 2023-09-04
Update Required e2e Tests

Vipinofficial11:
Updating the required e2e tests to use them for Sanity Testing on CDF. 

Note - Increasing the ratio of Run Time Scenarios to cover more use cases, Product team should suggest if any other factor should be taken into account for Sanity Builds.

### DIFF
--- a/src/e2e-test/features/bigqueryexecute/BQExecute.feature
+++ b/src/e2e-test/features/bigqueryexecute/BQExecute.feature
@@ -70,7 +70,7 @@ Feature: BigQueryExecute - Verify data transfer using BigQuery Execute plugin
     Then Open and capture logs
     Then Verify the pipeline status is "Succeeded"
 
-  @BQ_EXECUTE_DDL_CREATE_TEST
+  @BQ_EXECUTE_DDL_CREATE_TEST @BQExecute_Required
   Scenario: Verify BQExecute plugin functionality for DDL query - Create table
     Given Open Datafusion Project to configure pipeline
     When Expand Plugin group in the LHS plugins list: "Conditions and Actions"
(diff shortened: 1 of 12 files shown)


src/e2e-test/features/bigqueryexecute/BQExecute.feature
```diff
@@ -124,7 +124,7 @@ Feature: BigQueryExecute - Verify data transfer using BigQuery Execute plugin
     Then Verify the pipeline status is "Succeeded"
     Then Verify 1 records inserted in BigQuery table: "bqSourceTable" with query "bqExecuteCountDMLInsert"
 
-  @BQ_SOURCE_TEST @BQ_EXECUTE_UPSERT_SQL @BQExecute_Required
```
itsankit-google: why removing this test from required?

Vipinofficial11: No specific reason, Added this one ```Verify BQExecute plugin functionality for DDL query - Create table``` to required list so to balance the build time removed this one. (This feature file already have 2 tests marked as required)

itsankit-google: Do you have the data how much time `bq execute` required tests takes to execute in comparison to big ones like `gcs` and `bigquery`? If it is smaller then I think it should be fine to add it back because the overall execution time will remain the same.

Vipinofficial11: Ran the test on local it takes less than 2 minutes to run, In that case I believe its okay to leave it so added it back (71afa4ff800f10f49c5e0cc3361dfc310d56442a)

itsankit-google (approved):
Please squash commits before merge.

Vipinofficial11:
> Please squash commits before merge.

Squashed the commits.

PR #1304, opened 2023-09-19
[PLUGIN-1293] Gcs plugin Bug for buckets having delete multi part upload policy.

Shubhangi-cs:
Bug:
Pipeline fails with warning and error for the GCS buckets with delete multi part upload policy.

Bugfix:
Plugin should support the buckets with this policy enabled in it. Bumped up the library version to fix this.

https://cdap.atlassian.net/browse/PLUGIN-1293

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
     <google.cloud.pubsub.version>1.108.1</google.cloud.pubsub.version>
     <google.cloud.spanner.version>6.10.1</google.cloud.spanner.version>
     <google.cloud.speech.version>1.24.7</google.cloud.speech.version>
-    <google.cloud.storage.version>1.118.0</google.cloud.storage.version>
+    <google.cloud.storage.version>2.3.0</google.cloud.storage.version>
     <google.cloud.datastore.version>1.105.1</google.cloud.datastore.version>
     <google.protobuf.java.version>3.19.4</google.protobuf.java.version>
     <google.tink.version>1.3.0-rc3</google.tink.version>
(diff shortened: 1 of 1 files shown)


sahusanket (approved):

pom.xml
```diff
@@ -83,7 +83,7 @@
     <google.cloud.pubsub.version>1.108.1</google.cloud.pubsub.version>
     <google.cloud.spanner.version>6.10.1</google.cloud.spanner.version>
     <google.cloud.speech.version>1.24.7</google.cloud.speech.version>
-    <google.cloud.storage.version>1.118.0</google.cloud.storage.version>
+    <google.cloud.storage.version>2.3.0</google.cloud.storage.version>
```
itsankit-google, line 86: Can we upgrade to the latest gcloud storage version `2.27.0`?

itsankit-google (changes requested):
Please also add an `e2e test` for the scenario in the PR description.

pom.xml
Shubhangi-cs: This upgraded version supports the requirement. Is it good to upgrade it to this higher version?

itsankit-google: We should try to upgrade to the latest version available.

Shubhangi-cs: https://github.com/data-integrations/google-cloud/pull/1307. Pls have a look

itsankit-google: Why a separate PR?

Shubhangi-cs:
https://github.com/data-integrations/google-cloud/pull/1307 raised the new PR for this bug.

PR #1316, opened 2023-10-09
[PLUGIN-1692] BQ Sink Json Support

psainics:
## BQ Sink Json Support
Jira : [PLUGIN-1692](https://cdap.atlassian.net/browse/PLUGIN-1692?atlOrigin=eyJpIjoiMGFkODY1ZGUyZTU3NGI3ZGFjNGRjMWJlZTM3MGVkMDgiLCJwIjoiaiJ9)

BigQuery has json as data type and this pr adds support in the sink plugin for it.

### JSON String

As there is no json data type in cdap we are using string as json , to differentiate  string and json we let user enter all the strings that are to be treated as json via a UI field

image

### Nested Fields

Nested Fields can be targeted via the `.` as shown in the image above

image

### Code Changes

- User may skip adding the fields if table already exist. 
`getJsonStringFieldsFromBQSchema` Function can extract the json fields from the tabel schema

- `setJsonStringFields` Updates the table schema for the api request , marking user selected string fields to json

- `validateJsonStringFields` Validated if all fields given by user are present in schema

- We need fields at runtime hence they are set in config with a key.

`writeJsonObjectToWriter`
`writeJsonElementToWriter`
`writeJsonArrayToWriter`
`writeJsonPrimitiveToWriter`

All function are used to merge the user json in the main tree as json element.

### Note
Some function may be repeated , this is done to maintain a path and call them recursively for nested fields.

### Unit Test

I have added a few test cases (highlighted) to increase the code coverage for the new json writer.

image

[PLUGIN-1692]: https://cdap.atlassian.net/browse/PLUGIN-1692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -19,6 +19,7 @@ import com.google.auth.Credentials;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
+import com.google.cloud.bigquery.Table;
 import com.google.cloud.hadoop.io.bigquery.BigQueryConfiguration;
 import com.google.cloud.hadoop.io.bigquery.output.BigQueryTableFieldSchema;
 import com.google.cloud.kms.v1.CryptoKeyName;
@@ -47,7 +48,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -145,7 +150,8 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
    */
   protected final void initOutput(BatchSinkContext context, BigQuery bigQuery, String outputName, String fqn,
                                   String tableName, @Nullable Schema tableSchema, String bucket,
-                                  FailureCollector collector, @Nullable String marker) throws IOException {
+                                  FailureCollector collector, @Nullable String marker,
+                                  Table table) throws IOException {
     LOG.debug("Init output for table '{}' with schema: {}", tableName, tableSchema);
 
     List<BigQueryTableFieldSchema> fields = BigQuerySinkUtils.getBigQueryTableFields(bigQuery, tableName, tableSchema,
@@ -153,6 +159,18 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
       getConfig().getDataset(), getConfig().isTruncateTableSet(), collector);
 
     Configuration configuration = new Configuration(baseConfiguration);
+    if (table != null) {
+      com.google.cloud.bigquery.Schema bqSchema = table.getDefinition().getSchema();
+      if (bqSchema != null) {
+        String jsonStringFields = BigQuerySinkUtils.getJsonStringFieldsFromBQSchema(bqSchema);
+        configuration.set(BigQueryConstants.CONFIG_JSON_STRING_FIELDS, jsonStringFields);
+        BigQuerySinkUtils.setJsonStringFields(fields, jsonStringFields);
+      }
+    }
+
+    if (getConfig().getJsonStringFields() != null && !getConfig().getJsonStringFields().isEmpty()) {
+      BigQuerySinkUtils.setJsonStringFields(fields, getConfig().getJsonStringFields());
+    }
 
     // Build GCS storage path for this bucket output.
     String temporaryGcsPath = BigQuerySinkUtils.getTemporaryGcsPath(bucket, runUUID.toString(), tableName);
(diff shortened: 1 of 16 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -310,4 +329,82 @@ protected Configuration getOutputConfiguration() throws IOException {
     return configuration;
   }
 
+  /**
+   * Validates that the fields to be converted to JSON strings are present in the Output Schema.
+   * @param schema Output Schema.
+   * @param jsonStringFields List of fields to be converted to JSON strings comma separated.
+   * @param collector FailureCollector to collect errors.
+   */
+  public static void validateJsonStringFields(Schema schema,
```
sahusanket: The method should not be static ,the only static usage i see is in unit tests which can refactored 
Example : 
instead on `AbstractBigQuerySink.validateJsonStringFields` 
use 
`new BigQuerySink(null).validateJsonStringFields`

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -310,4 +329,82 @@ protected Configuration getOutputConfiguration() throws IOException {
...
+    jsonFields.removeAll(jsonFieldsValidated);
+    if (!jsonFields.isEmpty()) {
+      collector.addFailure(String.format("Field(s) '%s' are not present in the Output Schema.",
+                              String.join(", ", jsonFields)),
+                      "Remove the field(s) from the list of fields to be converted to JSON strings.")
+              .withConfigProperty(AbstractBigQuerySinkConfig.NAME_JSON_STRING_FIELDS);
+    }
+  }
+
+  private static void validateJsonStringFields(Schema schema, Set<String> jsonFields, ArrayList<String> path,
```
sahusanket: similar : remove static

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -279,6 +279,34 @@ public static String configureBucket(Configuration baseConfiguration,
...
+
+  private static void setJsonStringFields(List<BigQueryTableFieldSchema> fields, Set<String> jsonFields,
+                                             List<String> path) {
+    for (BigQueryTableFieldSchema field : fields) {
+      String fieldName = field.getName();
+      String fieldType = field.getType();
+      String separator = path.isEmpty() ? "" : ".";
+      String fieldPath = String.join(".", path) + separator + fieldName;
+      if (jsonFields.contains(fieldPath) && fieldType.equals(LegacySQLTypeName.STRING.name())) {
+        field.setType(LegacySQLTypeName.valueOf("JSON").name());
```
sahusanket: add a  `private static final String JSON = "JSON"`; 
and then use it use.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
```diff
@@ -919,4 +951,26 @@ public static void recordLineage(BatchSinkContext context,
...
+    getJsonStringFieldsFromBQSchema(bqSchema.getFields(), fields, new ArrayList<>());
+    return String.join(",", fields);
+  }
+  private static void getJsonStringFieldsFromBQSchema(FieldList fieldList,
+                                                      ArrayList<String> fields, ArrayList<String> path) {
+    for (Field field : fieldList) {
+      path.add(field.getName());
+      if (field.getType() == LegacySQLTypeName.RECORD) {
+        getJsonStringFieldsFromBQSchema(field.getSubFields(), fields, path);
+      } else if (field.getType().equals(LegacySQLTypeName.valueOf("JSON"))) {
```
sahusanket: Similar : use JSON from private static final String JSON

sahusanket (changes requested):
Mostly small comments . 

But from a user perspective ,  we can have a "UPDATE Schema" button after `JSON String` , which would change the data type of added fields automatically in output schema 

I am not sure if this can be taken up in this PR or not. Please consult your team and if it's not possible in this PR , please create a Jira for the same,

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
psainics: Updated !

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
psainics: Updated !

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
psainics: Added new 
```java 
private static final String JSON = "JSON"; 
```

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkUtils.java
psainics: Updated !

sahusanket (approved):

PR #1334, opened 2023-11-14
[PLUGIN-1715] Implement retry feature for BigQuery Execute Plugin

psainics:
## Implement retry feature for BigQuery Execute Plugin 

Jira : [PLUGIN-1715](https://cdap.atlassian.net/browse/PLUGIN-1715)

- In general internal errors should be retried.
- Do an exponential back-off retry on errors with reason `jobBackendError` , `jobInternalError`.
- Backoff requirements (exponential from 1 to 32 seconds with a multiplier of 2)

### Code Changes
- Add a new set that contains all reason we should retry on.
- A new custom exception that is raised when the error should be retried.
- A new function to generate the exponential back- off retry policy  
- A new dependency is added `dev.failsafe` that handles the retry logic based on the exception raised.

### Unit Tests
- Use mocks to test the behavior of the retry.

[PLUGIN-1715]: https://cdap.atlassian.net/browse/PLUGIN-1715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -33,8 +33,13 @@ import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableResult;
 import com.google.cloud.kms.v1.CryptoKeyName;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import dev.failsafe.Failsafe;
+import dev.failsafe.FailsafeException;
+import dev.failsafe.RetryPolicy;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
@@ -43,6 +48,7 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.action.Action;
 import io.cdap.cdap.etl.api.action.ActionContext;
 import io.cdap.cdap.etl.common.Constants;
+import io.cdap.plugin.gcp.bigquery.exception.BigQueryJobExecutionException;
 import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.CmekUtils;
@@ -51,8 +57,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
@@ -69,8 +77,18 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryExecute.class);
   public static final String NAME = "BigQueryExecute";
   private static final String RECORDS_PROCESSED = "records.processed";
-
   private Config config;
+  private static final String JOB_BACKEND_ERROR = "jobBackendError";
+  private static final String JOB_INTERNAL_ERROR = "jobInternalError";
+  private static final Set<String> RETRY_ON_REASON = ImmutableSet.of(JOB_BACKEND_ERROR, JOB_INTERNAL_ERROR);
+
+  BigQueryExecute() {
+   // no args constructor
+  }
+  @VisibleForTesting
+  BigQueryExecute(Config config) {
+    this.config = config;
+  }
 
   @Override
   public void run(ActionContext context) throws Exception {
@@ -103,9 +121,6 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     // Enable legacy SQL
     builder.setUseLegacySql(config.isLegacySQL());
 
-    // Location must match that of the dataset(s) referenced in the query.
-    JobId jobId = JobId.newBuilder().setRandomJob().setLocation(config.getLocation()).build();
-
     // API request - starts the query.
     Credentials credentials = config.getServiceAccount() == null ?
       null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
@@ -129,19 +144,74 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
     QueryJobConfiguration queryConfig = builder.build();
 
-    Job queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
+    // Exponential backoff
+    if (config.getRetryOnBackendError()) {
+      try {
+        executeQueryWithExponentialBackoff(bigQuery, queryConfig, context);
+      } catch (Throwable e) {
+        throw new RuntimeException(e);
+      }
+    } else {
+      executeQuery(bigQuery, queryConfig, context);
+    }
+  }
 
-    LOG.info("Executing SQL as job {}.", jobId.getJob());
-    LOG.debug("The BigQuery SQL is {}", config.getSql());
+  protected void executeQueryWithExponentialBackoff(BigQuery bigQuery,
+                                                   QueryJobConfiguration queryConfig, ActionContext context)
+          throws Throwable {
+    try {
+      Failsafe.with(getRetryPolicy()).run(() -> executeQuery(bigQuery, queryConfig, context));
+    } catch (FailsafeException e) {
+      if (e.getCause() != null) {
+        throw e.getCause();
+      }
+      throw e;
+    }
+  }
 
-    // Wait for the query to complete
-    queryJob = queryJob.waitFor();
+  private RetryPolicy<Object> getRetryPolicy() {
+    return RetryPolicy.builder()
+            .handle(BigQueryJobExecutionException.class)
+            .withBackoff(Duration.ofSeconds(config.getInitialRetryDuration()),
+                    Duration.ofSeconds(config.getMaxRetryDuration()), config.getRetryMultiplier())
+            .withMaxRetries(config.getMaxRetryCount())
+            .onRetry(event -> LOG.debug("Retrying BigQuery Execute job. Retry count: {}", event.getAttemptCount()))
+            .onSuccess(event -> LOG.debug("BigQuery Execute job executed successfully."))
+            .onRetriesExceeded(event -> LOG.error("Retry limit reached for BigQuery Execute job."))
+            .build();
+  }
+
+  private void executeQuery(BigQuery bigQuery, QueryJobConfiguration queryConfig, ActionContext context)
+          throws InterruptedException, BigQueryJobExecutionException {
+    // Location must match that of the dataset(s) referenced in the query.
+    JobId jobId = JobId.newBuilder().setRandomJob().setLocation(config.getLocation()).build();
+    Job queryJob;
+
+    try {
+      queryJob = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(jobId).build());
+
+      LOG.info("Executing SQL as job {}.", jobId.getJob());
+      LOG.debug("The BigQuery SQL is {}", config.getSql());
+
+      // Wait for the query to complete
+      queryJob = queryJob.waitFor();
+    } catch (BigQueryException e) {
+      LOG.error("The query job {} failed. Error: {}", jobId.getJob(), e.getError().getMessage());
+      if (RETRY_ON_REASON.contains(e.getError().getReason())) {
+        throw new BigQueryJobExecutionException(e.getError().getMessage(), e);
+      }
+      throw new RuntimeException(e);
+    }
 
     // Check for errors
     if (queryJob.getStatus().getError() != null) {
       // You can also look at queryJob.getStatus().getExecutionErrors() for all
       // errors, not just the latest one.
-      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
+      LOG.error("The query job {} failed. Error: {}", jobId.getJob(), queryJob.getStatus().getError());
+      if (RETRY_ON_REASON.contains(queryJob.getStatus().getError().getReason())) {
+        throw new BigQueryJobExecutionException(queryJob.getStatus().getError().getMessage());
+      }
+      throw new RuntimeException(queryJob.getStatus().getError().getMessage());
     }
 
     TableResult queryResults = queryJob.getQueryResults();
@@ -181,14 +251,14 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
 
   private void recordBytesProcessedMetric(ActionContext context, Job queryJob) {
     long processedBytes =
-        ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getTotalBytesProcessed();
+            ((JobStatistics.QueryStatistics) queryJob.getStatistics()).getTotalBytesProcessed();
     LOG.info("Job {} processed {} bytes", queryJob.getJobId(), processedBytes);
     Map<String, String> tags = new ImmutableMap.Builder<String, String>()
         .put(Constants.Metrics.Tag.APP_ENTITY_TYPE, Action.PLUGIN_TYPE)
         .put(Constants.Metrics.Tag.APP_ENTITY_TYPE_NAME, BigQueryExecute.NAME)
         .build();
     context.getMetrics().child(tags).countLong(BigQuerySinkUtils.BYTES_PROCESSED_METRIC,
-        processedBytes);
+            processedBytes);
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -43,16 +46,20 @@
 import io.cdap.cdap.etl.api.action.Action;
 import io.cdap.cdap.etl.api.action.ActionContext;
 import io.cdap.cdap.etl.common.Constants;
+import io.cdap.plugin.gcp.bigquery.exception.BigQueryJobExecutionException;
 import io.cdap.plugin.gcp.bigquery.sink.BigQuerySinkUtils;
 import io.cdap.plugin.gcp.bigquery.util.BigQueryUtil;
 import io.cdap.plugin.gcp.common.CmekUtils;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import org.jetbrains.annotations.TestOnly;
```
tivv: Don't use jetbrains annotations in production code. This may bring unnessesary dependencies

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -69,8 +76,13 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryExecute.class);
   public static final String NAME = "BigQueryExecute";
   private static final String RECORDS_PROCESSED = "records.processed";
+  private final Config config;
+  private static final Set<String> retryOnReason = ImmutableSet.of("jobBackendError", "jobInternalError");
 
-  private Config config;
+  @TestOnly
+  public BigQueryExecute(Config config) {
```
tivv: Usually package-only access is enough for testing

tivv: Adding an explicit config would effectively remove a no-argument constructor. Which constructor will be used for the production run? If this one, it should not be marked as TestOnly

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -69,8 +76,13 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryExecute.class);
   public static final String NAME = "BigQueryExecute";
   private static final String RECORDS_PROCESSED = "records.processed";
+  private final Config config;
+  private static final Set<String> retryOnReason = ImmutableSet.of("jobBackendError", "jobInternalError");
```
tivv: All constant should be uppercased

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -129,19 +138,62 @@ public void run(ActionContext context) throws Exception {
...
+  }
 
-    // Wait for the query to complete
-    queryJob = queryJob.waitFor();
+  private RetryPolicy<Object> getRetryPolicy() {
+    // Exponential backoff with initial retry of 1 second and max retry of 32 seconds.
+    Duration initialRetryDuration = Duration.ofSeconds(1);
+    Duration maxRetryDuration = Duration.ofSeconds(32);
+    int multiplier = 2;
+    int maxRetryCount = 5;
```
tivv: All of this should be configurable. Retries must be disableable

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -129,19 +138,62 @@ public void run(ActionContext context) throws Exception {
...
+    }
 
     // Check for errors
     if (queryJob.getStatus().getError() != null) {
       // You can also look at queryJob.getStatus().getExecutionErrors() for all
       // errors, not just the latest one.
-      throw new RuntimeException(queryJob.getStatus().getExecutionErrors().toString());
+      LOG.error("The query job {} failed. Error: {}", jobId.getJob(), queryJob.getStatus().getError());
+      if (retryOnReason.contains(queryJob.getStatus().getError().getReason())) {
+        throw new BigQueryJobExecutionException(queryJob.getStatus().getError().getMessage());
```
tivv, line 212: It's a checked exception. What would it be converted to when we are out of retries?

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Test(expected = java.lang.RuntimeException.class)
+  public void testExecuteQueryWithExponentialBackoffFailsRetryError() {
+    Mockito.when(bigQueryError.getReason()).thenReturn("jobBackendError");
+    bq.executeQueryWithExponentialBackoff(bigQuery, queryJobConfiguration, context);
+  }
+
+  @Test
+  public void testExecuteQueryWithExponentialBackoffSuccess() throws InterruptedException {
```
tivv: You probably want immediate retries, otherwise this test would be waiting unnesessary

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+  }
+
+  @Test(expected = java.lang.RuntimeException.class)
+  public void testExecuteQueryWithExponentialBackoffFailsWithNonRetryError() {
+    Mockito.when(bigQueryError.getReason()).thenReturn("accessDenied");
+    bq.executeQueryWithExponentialBackoff(bigQuery, queryJobConfiguration, context);
+  }
+
+  @Test(expected = java.lang.RuntimeException.class)
```
tivv: Why are you expecting just a RuntimeException here? It that it should go through retries and give some reasonable error

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
fernst: If you want to mark a method as visible only for the purposes of testing, use `@VisibleForTesting` which is already used elsewhere in the codebase.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
fernst: Also the strings here should be in their own constants.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
fernst: Agree. Please add a `Retry Configuration` section in the UI and allow the users to fine-tune this behavior.

fernst (commented):
Agree with Vitalii's suggestions here. Please address.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
vikasrathee-cs: @tivv @fernst It was mentioned previously on the bug ticket also that Bigquery SLA is for 32 seconds and that is the reason we have added these fix properties to go upto that limit. Same was confirmed in design doc also.
Can you please confirm in design doc also. Shared with you.
Also have a look at this. https://cloud.google.com/bigquery/sla#:~:text=%22-,Back%2Doff%20Requirements,-%22%C2%A0means%2C%20when

tivv: Few points here:

-  Customer may not wish to wait so long. Yep, BQ may recover, but it's also possible that it's too long for given customer
- You are changing the behaviour. We must have an option to disable this behavior
- SLAs are statistical. They may still be violated
- SLAs are point in time. They may change and we need to be able to adjust the bahvior without code change

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
psainics: Updated !

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
psainics: Removed the jetbrains annotation,  using `@ VisibleForTesting`

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
psainics: I have added new UI field to turn this behavior off (default) and let user configure the parameters.

image

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
psainics: Made this package only.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -129,19 +143,65 @@ public void run(ActionContext context) throws Exception {
...
+  }
+
+  protected void executeQueryWithExponentialBackoff(BigQuery bigQuery,
+                                                   QueryJobConfiguration queryConfig, ActionContext context)
+          throws BigQueryJobExecutionException {
+    try {
+      Failsafe.with(getRetryPolicy()).run(() -> executeQuery(bigQuery, queryConfig, context));
+    } catch (RuntimeException e) {
+      LOG.error("Failed to execute BigQuery job. Error: {}", e.getMessage());
+      throw new BigQueryJobExecutionException("Failed to execute BigQuery job. Reason: " + e.getMessage());
```
tivv: This way you are loosing all the additional information from original BigQueryException stack trace. There are only some case where you would want to hide all details of the underlying exception by using e.getMessage instead of attaching a root cause. I don't see why would we want to have it here. Same applies to the catch clause in `executeQuery`

psainics: Updated !

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
psainics: Fixed !

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
psainics: Updated !

psainics:
Changes Squashed !

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    // Mock context
+    Mockito.when(context.getMetrics()).thenReturn(stageMetrics);
+    Mockito.doNothing().when(stageMetrics).gauge(Mockito.anyString(), Mockito.anyLong());
+    Mockito.when(stageMetrics.child(Mockito.any())).thenReturn(metrics);
+    Mockito.doNothing().when(metrics).countLong(Mockito.anyString(), Mockito.anyLong());
+
+  }
+
+  @Test(expected = BigQueryJobExecutionException.class)
```
tivv: I am still a bit confused what final sequence of exceptions would it be and, what's important, what would be a message to user. Could you please use assertThrows and validate the message as described here: https://www.baeldung.com/junit-assert-exception ? I am wondering if we would need to unwrap those exceptions to prevent unneeded layers / cluttered message

psainics: The user will always see a `BigQueryJobExecutionException`

There are 3 cases for this to occur
1. Retry is not possible as the error reason is not in reason list, the user get the message `Failed to execute BigQuery job. Reason: {Reason from that exception}`

2. Retry was tried but failed , the user will get the message ~`Failed to execute BigQuery job.  io.cdap.plugin.gcp.bigquery.exception.BigQueryJobExecutionException`~ `Failed to execute BigQuery job. Reason: Retries exhausted.`
- We could provide a different message here if required, as the exception class will be  `FailsafeException` for which an additional catch could be added .

3. Someting else fails and another exception is thrown , `Failed to execute BigQuery job. Reason: {Reason from that exception}`

tivv: I am curious about the text of the message. I want to avoid something like  `Failed to execute BigQuery job. Reason: RuntimeException: BigQueryJobExecutionException: reason`

psainics: I have updated to assert throw, and made it more clear what will the end user see. in all cases.

fernst (approved):
LGTM

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  @Test
+  public void testExecuteQueryWithExponentialBackoffFailsWithNonRetryError() {
+    Mockito.when(bigQueryError.getReason()).thenReturn("accessDenied");
+    Exception exception = Assert.assertThrows(BigQueryJobExecutionException.class, () -> {
+      bq.executeQueryWithExponentialBackoff(bigQuery, queryJobConfiguration, context);
+    });
+    String actualMessage = exception.getMessage();
+    Assert.assertEquals("Failed to execute BigQuery job. Reason: " + mockErrorMessageNoRetry, actualMessage);
```
tivv: So, basically we say "Failed to execute BigQuery job. Reason: Job execution failed with error: $error" This is what I want to avoid as it's saying same thing twice

tivv: May be in top level catch we should just find the correct cause and rethrow in instead of wrapping it again

psainics: Sure, i have removed the wrapping of error message.

tivv: But this way you are loosing the underlying error. You should either: 
 - do visa versa and pop up only the underlying error message "Job execution failed with error: $error"
  - prepend it with some text that adds value, e.g. "Job retries failed after X tries. Last try failure: Job execution failed with error: $error".

psainics: Yup I am doing the 1st one , by removing wrapping I ment removing the extra messages string that was present before

tivv: But you are hiding the underlying "$error", so the only thing that we can see is that retries exhausted without much information about the underlying error.

psainics: As i already have logging when nth retry is made or the policy is out of retry should i just pop the error and throw that ?

something like this
```java
try {
      Failsafe.with(getRetryPolicy()).run(() -> executeQuery(bigQuery, queryConfig, context));
    } catch (FailsafeException e) {
      throw e.getCause();
    } 
``` 
Also i could drop the `catch (RuntimeException e)`

tivv: Sure, that's option #1 in the list above

tivv: I would check if e.getCause is not null as generally not every failure is an exception

psainics:
Thanks @tivv , i have made the changes and squashed them , hopefully you can approve this now.
I will get this QA tested before merging  👍

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
psainics: 🤔 
```java
if (e.getCause() != null) {
        throw e.getCause();  
} 
  throw e;
```

widgets/BigQueryExecute-action.json
```diff
@@ -193,6 +193,63 @@
...
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "YES"
+            },
+            "off": {
+              "value": "false",
+              "label": "NO"
+            },
+            "default": "false"
```
itsankit-google: shouldn't this be set to `true` by default, @tivv @fernst wdyt??
I thought aim is to add some resilency to the plugin to reduce intermittent failures but if the behavior is `false` by default, users will need to enable it explicitly and modify every pipeline.

psainics: Your point makes a lot of sense. Setting this behavior to true by default aligns with the goal of enhancing plugin resilience and reducing intermittent failures.

tivv: So, we usually have two things:
 * Default for UI. This is what we want default experience to be. In this case it should be "true"
 * Default for backend (when option is not set). This is usually how things worked before to ensure backward compatibility unless we have explicit OK from product to break backward compatibility

psainics: @itsankit-google we had a discussion with deepinder, the UI fields should be hidden and true by default !

psainics:
New changes include 
- Reverted changes in `docs/BigQueryExecute-action.md` as the UI fields are hidden.
- Removed filter `retryOnBackendError` with condition `retryOnBackendError == true` as this will not be required now.
- Made all the ui fields `hidden`
- Retry On Backend Error is `TRUE` by default , on widget and backed.

psainics:
Resolved conflicts caused by PR: https://github.com/data-integrations/google-cloud/pull/1344 in `bigquery/action/BigQueryExecute.java`

tivv (approved):

PR #1350, opened 2024-01-08
[PLUGIN-1692] Remove duplicate path when writing nested JSON array

psainics:
JIRA: [PLUGIN-1692](https://cdap.atlassian.net/browse/PLUGIN-1692)

Hotfix for bq sink json support , data was not formatted when writing nested json array.

This was caused due to path having duplicate entry, which is removed in this fix.

[PLUGIN-1692]: https://cdap.atlassian.net/browse/PLUGIN-1692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1354, opened 2024-01-10
[Cherrypick][HotFix] Remove duplicate path when writing nested JSON array

psainics:
[Cherrypick] Remove duplicate path when writing nested JSON array

Commit : ca5b35907d4c9f8cac80042ff24292657000b54d
PR : #1350 

---

Hotfix for bq sink json support , data was not formatted when writing nested json array.

This was caused due to path having duplicate entry, which is removed in this fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>google-cloud</artifactId>
-  <version>0.23.0</version>
+  <version>0.23.1-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


itsankit-google (commented):
The OSS release is going on, release branches are freezed at the moment.

Please bump the version to snapshot.

Please check with @samdgupi before we can merge this PR.

We might need to do a hub release now since the CDAP release has already started.

psainics:
Version bumped to `0.23.1-SNAPSHOT`

itsankit-google (approved):

PR #1359, opened 2024-01-24
[PLUGIN-1738] Fix bucket creation issue for GCS Copy/Move Plugins

dj-smart:
[JIRA](https://cdap.atlassian.net/browse/PLUGIN-1738)

### Introduction:
- The method `createBucketIfNotExists` aims to create a bucket at destination path if it doesn't exist while performing copy/move operation across GCS buckets.
- There is a bug in the current implementation due to which bucket is not created and an exception is thrown.

### Fix:
- Refactored the code to attempt Bucket creation and if error code `409` is encountered  (indicating that bucket already exists) then it is ignored and all other exceptions are thrown.

### Testing:
- Tested the changes using pipeline using `GCS Copy` plugin in CDAP sandbox.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -120,11 +120,14 @@ public class StorageClient {
       LOG.info("Bucket {} has been created successfully", path.getBucket());
     } catch (StorageException e) {
       // Don't throw error if bucket already exists
-      // https://cloud.google.com/storage/docs/xml-api/reference-status#409%E2%80%94conflict
-      if (e.getCode() != 409) {
+      // https://cloud.google.com/storage/docs/json_api/v1/status-codes#409_Conflict
+      if (e.getCode() == 409) {
+        LOG.warn("Getting 409 Conflict: {} Bucket at destination path {} may already exist.",
+                 e.getMessage(), path.getUri());
+      } else {
         throw new RuntimeException(
-        String.format("Unable to create bucket %s. ", path.getBucket())
-          + "Ensure you entered the correct bucket path and have permissions for it.", e);
+          String.format("Unable to create bucket %s. Ensure you entered the correct bucket path and " +
+                          "have permissions for it.", path.getBucket()), e);
       }
     }
   }
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -115,16 +115,17 @@ public class StorageClient {
    * @param cmekKeyName  the name of the cmek key
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
-    Bucket bucket = null;
     try {
-      bucket = storage.get(path.getBucket());
-    } catch (StorageException e) {
-      throw new RuntimeException(
-        String.format("Unable to access bucket %s. ", path.getBucket())
-          + "Ensure you entered the correct bucket path and have permissions for it.", e);
-    }
-    if (bucket == null) {
       GCPUtils.createBucket(storage, path.getBucket(), location, cmekKeyName);
+      LOG.info("Bucket {} has been created successfully", path.getBucket());
+    } catch (StorageException e) {
+      // Don't throw error if bucket already exists
+      // https://cloud.google.com/storage/docs/xml-api/reference-status#409%E2%80%94conflict
+      if (e.getCode() != 409) {
+        throw new RuntimeException(
+        String.format("Unable to create bucket %s. ", path.getBucket())
+          + "Ensure you entered the correct bucket path and have permissions for it.", e);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -115,16 +115,17 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
     try {
-      bucket = storage.get(path.getBucket());
+      GCPUtils.createBucket(storage, path.getBucket(), location, cmekKeyName);
+      LOG.info("Bucket {} has been created successfully", path.getBucket());
     } catch (StorageException e) {
-      throw new RuntimeException(
-        String.format("Unable to access bucket %s. ", path.getBucket())
+      // Don't throw error if bucket already exists
+      // https://cloud.google.com/storage/docs/xml-api/reference-status#409%E2%80%94conflict
+      if (e.getCode() != 409) {
```
sahusanket: Create a Error List/ set of https://cloud.google.com/storage/docs/xml-api/reference-status#409%E2%80%94conflict 

where throw error only on (409 AND (e get ErrorName present in Error List )

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -115,16 +115,17 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    * @param cmekKeyName  the name of the cmek key
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
```
sahusanket, line 117: Please add a Unit test

sahusanket (changes requested):

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dj-smart: Done

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dj-smart: Discussed offline. Resolving.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -114,17 +114,22 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    * @param location the location of bucket
    * @param cmekKeyName  the name of the cmek key
    */
+  @VisibleForTesting
```
itsankit-google: The method is already public we don't need this annotation.

This annotation is meant to be added when visibility of a particular method is adjusted because of unit tests.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -114,17 +114,22 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
+    } catch (StorageException e) {
+      // Don't throw error if bucket already exists
+      // https://cloud.google.com/storage/docs/json_api/v1/status-codes#409_Conflict
+      if (e.getCode() == 409) {
+        LOG.warn("Getting 409 Conflict: {} Bucket at destination path {} may already exist.",
+                 e.getMessage(), path.getUri());
+      } else {
+        throw new RuntimeException(
+          String.format("Unable to create bucket %s. ", path.getBucket())
+            + "Ensure you entered the correct bucket path and have permissions for it.", e);
```
itsankit-google:
```
String.format("Unable to create bucket %s. Ensure you entered the correct bucket path and have permissions for it.", path.getBucket())
```

itsankit-google (commented):
minor comments, otherwise LGTM.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dj-smart: Acknowledged

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dj-smart: Done

sahusanket (approved):

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -114,17 +114,22 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    * @param location the location of bucket
    * @param cmekKeyName  the name of the cmek key
    */
+
```
itsankit-google: nit: remove empty line.

itsankit-google (approved):

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
dj-smart: Done

PR #1361, opened 2024-01-25
[PLUGIN-1739] Added TIME_PARTITIONING_TYPES_UI_MAP to correctly map ui field values

psainics:
##  Added TIME_PARTITIONING_TYPES_UI_MAP to correctly map ui field values

Jira : [Plugin-1739](https://cdap.atlassian.net/browse/PLUGIN-1739)

## Description
As id and labels were different, user faced issue when using the label as macro.
This PR adds a map to MAP UI values to Backend values.

- Earlier user could only use `Day` `Hour` `Month` `Year` 
- The new UI MAP lets user use `Day` `Hour` `Month` `Year` with `Daily` `Hourly` `Monthly` `Yearly`

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
@@ -24,6 +24,7 @@ import com.google.cloud.bigquery.TimePartitioning;
 import com.google.cloud.kms.v1.CryptoKeyName;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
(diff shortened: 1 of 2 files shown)


itsankit-google (commented):
Wondering if it is okay to also keep `DAY`, `MONTH`, `HOUR` and `YEAR` also in the map as keys so that existing pipelines don't break.

@albertshau wdyt?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
```diff
@@ -282,8 +287,14 @@ public PartitionType getPartitioningType() {
...
 
   public TimePartitioning.Type getTimePartitioningType() {
-    return Strings.isNullOrEmpty(timePartitioningType) ? TimePartitioning.Type.DAY :
-            TimePartitioning.Type.valueOf(timePartitioningType.toUpperCase());
+    // Default to DAY if timePartitioningType is not set
+    if (Strings.isNullOrEmpty(timePartitioningType)) {
+      return TimePartitioning.Type.DAY;
+    }
+    String timePartitioningTypeUpperCase = timePartitioningType.toUpperCase();
+    timePartitioningTypeUpperCase  = TIME_PARTITIONING_TYPES_UI_MAP.getOrDefault(timePartitioningTypeUpperCase,
```
albertshau: nit: extra space

albertshau:
It looks like it is handled already, since the original value is used if it's not in the map. I agree that it's not the most straightforward thing to read though. In this case I think it's more clear if we just have a switch statement rather than a map.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java
psainics: 👍

itsankit-google (changes requested):
As Albert suggested, can we please use a switch-case instead of map?

psainics:
> As Albert suggested, can we please use a switch-case instead of map?

I have removed the map for a switch, also included a unit test !

itsankit-google (approved):

PR #1406, opened 2024-04-23
[PLUGIN-1780]changes done for GCS multisink issue.

vikasrathee-cs:
[PLUGIN-1780] changes done for GCS multisink issue.

GCS MultiSink plugin is failing while having Flexible schema set as true.

 io.cdap.plugin.gcp.gcs.sink.DelegatingGCSRecordWriter.close() method has a problem. It commits the job in a single partition close to the executor and not in the driver. This has the logical problem of commiting partially and too early, but also when a lot of partitions are committed at once (as it happens with datacacher), 

[PLUGIN-1780]: https://cdap.atlassian.net/browse/PLUGIN-1780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
@@ -29,8 +29,8 @@ import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter;
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
 
-import java.io.BufferedReader;
 import java.io.DataOutputStream;
+import java.io.EOFException;
 import java.io.IOException;
 import java.util.HashSet;
 import java.util.Set;
@@ -44,12 +44,11 @@ import javax.annotation.Nullable;
 public class DelegatingGCSOutputCommitter extends OutputCommitter {
 
   private TaskAttemptContext taskAttemptContext;
-  private boolean firstSplit = true;
+  private boolean firstTable = true;
   private static final String PARTITIONS_FILE_SUFFIX = "_partitions.txt";
 
-
-  public DelegatingGCSOutputCommitter(TaskAttemptContext taskContext) {
-    this.taskAttemptContext = taskContext;
+  public DelegatingGCSOutputCommitter(TaskAttemptContext taskAttemptContext) {
+    this.taskAttemptContext = taskAttemptContext;
   }
 
   /**
@@ -59,19 +58,19 @@ public class DelegatingGCSOutputCommitter extends OutputCommitter {
    */
   @SuppressWarnings("rawtypes")
   public void addGCSOutputCommitterFromOutputFormat(OutputFormat outputFormat,
-                                                    TaskAttemptContext context,
                                                     String tableName) throws IOException, InterruptedException {
     //Set output directory
-    context.getConfiguration().set(FileOutputFormat.OUTDIR,
-                                   DelegatingGCSOutputUtils.buildOutputPath(context.getConfiguration(), tableName));
+    taskAttemptContext.getConfiguration().set(FileOutputFormat.OUTDIR,
+                                              DelegatingGCSOutputUtils.buildOutputPath(
+                                                taskAttemptContext.getConfiguration(), tableName));
 
     //Wrap output committer into the GCS Output Committer.
-    GCSOutputCommitter gcsOutputCommitter = new GCSOutputCommitter(outputFormat.getOutputCommitter(context));
+    GCSOutputCommitter gcsOutputCommitter = new GCSOutputCommitter(outputFormat.getOutputCommitter(taskAttemptContext));
 
-    gcsOutputCommitter.setupJob(context);
-    gcsOutputCommitter.setupTask(context);
-    writePartitionFile(context.getConfiguration().get(FileOutputFormat.OUTDIR), context);
-    firstSplit = false;
+    gcsOutputCommitter.setupJob(taskAttemptContext);
+    gcsOutputCommitter.setupTask(taskAttemptContext);
+    writePartitionFile(taskAttemptContext.getConfiguration().get(FileOutputFormat.OUTDIR), taskAttemptContext);
+    firstTable = false;
   }
 
   @Override
@@ -79,7 +78,6 @@ public class DelegatingGCSOutputCommitter extends OutputCommitter {
     Path outputPath = new Path(jobContext.getConfiguration().get(DelegatingGCSOutputFormat.OUTPUT_PATH_BASE_DIR));
     FileSystem fs = outputPath.getFileSystem(jobContext.getConfiguration());
     Path tempPath = new Path(outputPath, getPendingDirPath(jobContext.getJobID()));
-    // creating the _temporary folder in base path
     fs.mkdirs(tempPath);
   }
 
@@ -142,21 +140,20 @@ public class DelegatingGCSOutputCommitter extends OutputCommitter {
   @Override
   public void abortJob(JobContext jobContext, JobStatus.State state) throws IOException {
     IOException ioe = null;
-
-    for (String output : getOutputPaths(jobContext)) {
-      try {
+    try {
+      for (String output : getOutputPaths(jobContext)) {
         taskAttemptContext.getConfiguration().set(FileOutputFormat.OUTDIR, output);
         FileOutputCommitter committer = new FileOutputCommitter(new Path(output), taskAttemptContext);
         committer.abortJob(jobContext, state);
-      } catch (IOException e) {
-        if (ioe == null) {
-          ioe = e;
-        } else {
-          ioe.addSuppressed(e);
-        }
-      } finally {
-        cleanupJob(jobContext);
       }
+    } catch (IOException e) {
+      if (ioe == null) {
+        ioe = e;
+      } else {
+        ioe.addSuppressed(e);
+      }
+    } finally {
+      cleanupJob(jobContext);
     }
     if (ioe != null) {
       throw ioe;
@@ -203,11 +200,13 @@ public class DelegatingGCSOutputCommitter extends OutputCommitter {
 
     for (FileStatus status : fs.listStatus(tempPath)) {
       if (status.getPath().getName().endsWith(PARTITIONS_FILE_SUFFIX)) {
-        try (FSDataInputStream dis = fs.open(status.getPath());
-             BufferedReader br = new BufferedReader(new java.io.InputStreamReader(dis))) {
-          String line;
-          while ((line = br.readLine()) != null) {
-            outputPaths.add(line);
+        try (FSDataInputStream dis = fs.open(status.getPath())) {
+          while (true) {
+            try {
+              outputPaths.add(dis.readUTF());
+            } catch (EOFException e) {
+              break;
+            }
           }
         }
       }
@@ -231,13 +230,12 @@ public class DelegatingGCSOutputCommitter extends OutputCommitter {
     Path taskPartitionFile = new Path(tempPath, String.format("%s%s", taskId, PARTITIONS_FILE_SUFFIX));
     if (!fs.exists(taskPartitionFile)) {
       fs.createNewFile(taskPartitionFile);
-    } else if (firstSplit) {
+    } else if (firstTable) {
       fs.create(taskPartitionFile, true);
     }
     try (DataOutputStream out = fs.append(taskPartitionFile)) {
-      out.writeBytes(path + "\n");
+      out.writeUTF(path);
     }
-    taskAttemptContext = context;
   }
 
   // This will create a directory with name _temporary_{jobId} to write the partition files
--- a/src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
@@ -16,32 +16,48 @@
 
 package io.cdap.plugin.gcp.gcs.sink;
 
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapreduce.JobContext;
 import org.apache.hadoop.mapreduce.JobStatus;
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -73,39 +92,45 @@ public void setupTask(TaskAttemptContext taskAttemptContext) throws IOException
...
-
-    return needsTaskCommit;
+    return true;
   }
 
   @Override
   public void commitTask(TaskAttemptContext taskAttemptContext) throws IOException {
-    for (OutputCommitter committer : committerMap.values()) {
+    for (String output : getOutputPaths(taskAttemptContext)) {
+      taskAttemptContext.getConfiguration().set(FileOutputFormat.OUTDIR, output);
```
tivv: Do you need to do this? Where is it used? You already pass the path to committer, it seems that it should be enough. If you need to do this, please  ensure you reset it after you are done

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -136,9 +163,74 @@ public void abortJob(JobContext jobContext, JobStatus.State state) throws IOExce
...
+   * @throws IOException
+   */
+  private Set<String> getOutputPathsFromTempPartitionFile(Path baseOutputPath, FileSystem fs) throws IOException {
+    Path tempPath = new Path(baseOutputPath, FileOutputCommitter.PENDING_DIR_NAME);
+    Set<String> outputPaths = new HashSet<>();
+
+    for (FileStatus status : fs.listStatus(tempPath)) {
+      if (status.getPath().getName().endsWith(PARTITIONS_FILE_SUFFIX)) {
+        try (FSDataInputStream dis = fs.open(status.getPath());
+             DataInputStream in = new DataInputStream(new BufferedInputStream(dis));
```
tivv: Why do you need a DataInputStream? I believe you can directly do `new BufferedReader(dis)`

tivv: Also consider using org.apache.commons.io.IOUtils#readLines(java.io.InputStream, java.lang.String)

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -136,9 +163,74 @@ public void abortJob(JobContext jobContext, JobStatus.State state) throws IOExce
...
+   * @param context
+   * @throws IOException
+   */
+  private void writePartitionFile(String path, TaskAttemptContext context) throws IOException {
+    Path outputPath = new Path(context.getConfiguration().get(DelegatingGCSOutputFormat.OUTPUT_PATH_BASE_DIR));
+    Path tempPath = new Path(outputPath, FileOutputCommitter.PENDING_DIR_NAME);
+    FileSystem fs = tempPath.getFileSystem(context.getConfiguration());
+    String taskId = context.getTaskAttemptID().getTaskID().toString();
+    Path taskPartitionFile = new Path(tempPath, String.format("%s%s", taskId, PARTITIONS_FILE_SUFFIX));
+    if (!fs.exists(taskPartitionFile)) {
```
tivv, line 231: So what would happen on multiple attempts? Would those be concatenated?

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSRecordWriter.java
```diff
@@ -80,11 +80,10 @@ public void close(TaskAttemptContext context) throws IOException, InterruptedExc
       delegate.close(context);
     }
 
-    // Call the Commit Task and Commit Job implementations of this plugin to copy files into their final directory.
+    // Call the Commit Task implementation of this plugin to copy files into their final directory.
     // We need to do this at this stage because the OutputCommitter needs to be aware of the different partitions
     // that have been stored so far.
     delegatingGCSOutputCommitter.commitTask(context);
```
tivv: I think this should also go away

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputFormat.java
```diff
@@ -75,6 +75,7 @@ public void checkOutputSpecs(JobContext context) throws IOException, Interrupted
 
   @Override
   public OutputCommitter getOutputCommitter(TaskAttemptContext context) throws IOException, InterruptedException {
+    outputCommitter.setTaskContext(context);
```
tivv: You should just have this as a final field in the committer and create a new committer, passing context into constructor.

tivv (commented):
As discussed, please adopt https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/test/java/org/apache/hadoop/mapred/TestFileOutputCommitter.java to test the committer

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: This needs to be done otherwise we are getting only _SUCCESS file in the final output directory after commitJob.
As this can be called from close method also, So I was thinking of leaving it as it is with committerMap.

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: taskID will be different each time, and we can clean up the partitions.txt files in case of abort task for that specific taskId

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSRecordWriter.java
vikasrathee-cs: It can go away, but reading from file and then creating outputCommitter have a lot of overheads, so will use the same committerMap to commit task and call it from here.

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
tivv: Why? My understanding is that TaskAttempId would be different, but TaskId would be same

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
tivv: Can you elaborate? I don't see OUTDIR used in the FileOutputCommiter, are there any?

itsankit-google:
Please create a JIRA and provide context in the PR description and JIRA...

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: Removed. It was not getting used.

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: Separate library needs to add for IOUtils, removed the dataInputStream for now

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: Added condition for this. It will overwrite in case file already there.

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputFormat.java
vikasrathee-cs: At the time of creating object for committer, task attempt context is not available, we have to set it up later only.

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputFormat.java
```diff
@@ -39,10 +39,10 @@ public class DelegatingGCSOutputFormat extends OutputFormat<NullWritable, Struct
   public static final String DELEGATE_CLASS = "delegating_output_format.delegate";
   public static final String OUTPUT_PATH_BASE_DIR = "delegating_output_format.output.path.base";
   public static final String OUTPUT_PATH_SUFFIX = "delegating_output_format.output.path.suffix";
-  private final DelegatingGCSOutputCommitter outputCommitter;
+  // private  DelegatingGCSOutputCommitter outputCommitter;
```
tivv: Please remove the commented code

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputFormat.java
```diff
@@ -65,7 +65,8 @@ public RecordWriter<NullWritable, StructuredRecord> getRecordWriter(TaskAttemptC
     Configuration hConf = context.getConfiguration();
     String partitionField = hConf.get(PARTITION_FIELD);
 
-    return new DelegatingGCSRecordWriter(context, partitionField, outputCommitter);
+    return new DelegatingGCSRecordWriter(context, partitionField,
+                                         (DelegatingGCSOutputCommitter) getOutputCommitter(context));
```
tivv: Just change the return type in line 78, no need to cast

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputFormat.java
```diff
@@ -74,7 +75,8 @@ public void checkOutputSpecs(JobContext context) throws IOException, Interrupted
   }
 
   @Override
-  public OutputCommitter getOutputCommitter(TaskAttemptContext context) throws IOException, InterruptedException {
+  public OutputCommitter getOutputCommitter(TaskAttemptContext context) {
+    DelegatingGCSOutputCommitter outputCommitter = new DelegatingGCSOutputCommitter(context);
```
tivv: Return right away. No need to create a variable

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -16,32 +16,45 @@
...
+
+
+  public DelegatingGCSOutputCommitter(TaskAttemptContext taskContext) {
+    this.taskAttemptContext = taskContext;
   }
 
   /**
    * Add a new GCSOutputCommitter based on a supplied Output Format and Table Name.
-   *
+   * <p>
```
tivv, line 56: No need to pass context anymore into this method. You already have it as a field

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -16,32 +16,45 @@
...
+ * <p>
  * Delegated instances are created based on a supplied Output Format and Destination Table Names.
  */
 public class DelegatingGCSOutputCommitter extends OutputCommitter {
-  private final Map<String, OutputCommitter> committerMap;
 
-  public DelegatingGCSOutputCommitter() {
-    committerMap = new HashMap<>();
+  private TaskAttemptContext taskAttemptContext;
+  private boolean firstSplit = true;
```
tivv: Can you name it firstTable? Split has a very specific meaning in Spark and we are doing tables or partition files

tivv: Also I'd consider to move file removal logic to the constructor. This way you don't need this mutable flag at all

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
```diff
@@ -125,20 +143,107 @@ public void abortTask(TaskAttemptContext taskAttemptContext) throws IOException
...
+    FileSystem fs = tempPath.getFileSystem(context.getConfiguration());
+    String taskId = context.getTaskAttemptID().getTaskID().toString();
+    Path taskPartitionFile = new Path(tempPath, String.format("%s%s", taskId, PARTITIONS_FILE_SUFFIX));
+    if (!fs.exists(taskPartitionFile)) {
+      fs.createNewFile(taskPartitionFile);
+    } else if (firstSplit) {
+      fs.create(taskPartitionFile, true);
+    }
+    try (DataOutputStream out = fs.append(taskPartitionFile)) {
+      out.writeBytes(path + "\n");
```
tivv: This won't work for any non-latin1 file names / pathes. Use writeChars / writeUTF and same on reading. This would also save you from tons of input stream wrappers on reading

tivv (commented):
Please also add tests as discussed

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSRecordWriter.java
vikasrathee-cs: removed and getting path from partitions.txt in commit task

src/main/java/io/cdap/plugin/gcp/gcs/sink/DelegatingGCSOutputCommitter.java
vikasrathee-cs: This constructor gets called by mapper job and then reducer job, creating two folders in output path, so using firstTable instead.

vikasrathee-cs:
> Please also add tests as discussed

Working on it, facing issues while using local file system to test it out, some of the methods are not supported in local file system but we are using with gs file system.

vikasrathee-cs:
created a new PR to remove merge commit.
https://github.com/data-integrations/google-cloud/pull/1414

PR #1408, opened 2024-05-13
Use securekey-textarea widget for SA json fields

GnsP:
Changes the widget type for SA json fields from `textbox` to `secure-textarea`

Please notice the shield icon in the SA json field. Clicking the shield lets the user select one of the secure keys defined in the namespace.

A corresponding [PR in cdap-ui](https://github.com/cdapio/cdap-ui/pull/1217) adds the securekeys tab in Namespace Admin and also makes the securekey-textarea widget non-editable (so that the user will only be able to add the value as a secure macro).

Screenshot 2024-05-13 at 10 21 28 AM

### DIFF
--- a/widgets/BigQuery-connector.json
+++ b/widgets/BigQuery-connector.json
@@ -76,7 +76,7 @@
           }
         },
         {
-          "widget-type": "textarea",
+          "widget-type": "securekey-textarea",
           "label": "Service Account JSON",
           "name": "serviceAccountJSON"
         }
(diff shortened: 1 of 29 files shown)


itsankit-google (commented):
Can you please add the screenshot in the PR description on how the change was tested?

GnsP:
> Can you please add the screenshot in the PR description on how the change was tested?

Added a screenshot.

itsankit-google:
> > Can you please add the screenshot in the PR description on how the change was tested?
> 
> Added a screenshot.

What will be the case for existing pipelines? How will the UI look like? Have we tested it? Also since it has been using `textarea` from sometime now, do we have a PM approval?

PR #1430, opened 2024-08-19
PLUGIN-1800: Adding read timeout to BQ Execute Config with default 120s

masivesky:
https://cdap.atlassian.net/browse/PLUGIN-1800

Retries were already configured on the BigQuery Client by default, 5xx errors are being retried 6 times before failing:

```
this.retrySettings = (RetrySettings)MoreObjects.firstNonNull(builder.retrySettings, getDefaultRetrySettings());
```

```
private static RetrySettings.Builder getDefaultRetrySettingsBuilder() {
    return RetrySettings.newBuilder().setMaxAttempts(6).setInitialRetryDelay(Duration.ofMillis(1000L)).setMaxRetryDelay(Duration.ofMillis(32000L)).setRetryDelayMultiplier(2.0).setTotalTimeout(Duration.ofMillis(50000L)).setInitialRpcTimeout(Duration.ofMillis(50000L)).setRpcTimeoutMultiplier(1.0).setMaxRpcTimeout(Duration.ofMillis(50000L));
  }
  ```

Hence, increasing the read timeout for BigQuery Execute action plugin (where we are validating SQL) to a default of 120s. Earlier timeout was the BigQuery default of 20s. 

Confirmed via local sandbox testing that the default timeout for BigQuery Execute is 120s.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -125,7 +125,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     Credentials credentials = config.getServiceAccount() == null ?
       null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
                                                     config.isServiceAccountFilePath());
-    BigQuery bigQuery = GCPUtils.getBigQuery(config.getProject(), credentials);
+    BigQuery bigQuery = GCPUtils.getBigQuery(config.getProject(), credentials, config.getReadTimeout());
     //create dataset to store the results if not exists
     if (config.getStoreResults() && !Strings.isNullOrEmpty(datasetName) &&
       !Strings.isNullOrEmpty(tableName)) {
@@ -283,11 +283,13 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     private static final String NAME_MAX_RETRY_DURATION = "maxRetryDuration";
     private static final String NAME_RETRY_MULTIPLIER = "retryMultiplier";
     private static final String NAME_MAX_RETRY_COUNT = "maxRetryCount";
+    private static final String NAME_READ_TIMEOUT = "readTimeout";
     public static final long DEFAULT_INITIAL_RETRY_DURATION_SECONDS = 1L;
     public static final double DEFAULT_RETRY_MULTIPLIER = 2.0;
     public static final int DEFAULT_MAX_RETRY_COUNT = 5;
     // Sn = a * (1 - r^n) / (r - 1)
     public static final long DEFULT_MAX_RETRY_DURATION_SECONDS = 63L;
+    public static final int DEFAULT_READ_TIMEOUT = 120;
 
     @Description("Dialect of the SQL command. The value must be 'legacy' or 'standard'. " +
       "If set to 'standard', the query will use BigQuery's standard SQL: " +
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java [hunk 1]
```diff
@@ -125,7 +125,7 @@ public void run(ActionContext context) throws Exception {
     Credentials credentials = config.getServiceAccount() == null ?
       null : GCPUtils.loadServiceAccountCredentials(config.getServiceAccount(),
                                                     config.isServiceAccountFilePath());
-    BigQuery bigQuery = GCPUtils.getBigQuery(config.getProject(), credentials);
+    BigQuery bigQuery = GCPUtils.getBigQuery(config.getProject(), credentials, config.readTimeout);
```
itsankit-google: nit: `config.readTimeout`->`config.getReadTimeout()`

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -390,13 +392,19 @@ public static final class Config extends AbstractBigQueryActionConfig {
             "are reserved keys and cannot be used as label keys.")
     protected String jobLabelKeyValue;
 
+    @Name(NAME_READ_TIMEOUT)
+    @Nullable
+    @Macro
+    @Description("Maximum time to wait before reading response. Default is 120 seconds")
```
itsankit-google: nit: `Maximum value of read timeout in seconds.`

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -604,6 +623,7 @@ public void validateSQLSyntax(FailureCollector failureCollector, BigQuery bigQue
       try {
         bigQuery.create(JobInfo.of(queryJobConfiguration));
       } catch (BigQueryException e) {
+          LOG.error(String.format("Validation for SQL command %s failed.", sql), e);
```
itsankit-google: We should not log if we are throwing, it is anti-pattern.

Instead the error code as well while adding the error message below in failure collector.

itsankit-google (commented):
Please update the markdown as well.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
masivesky: Done

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
masivesky: Modified it to be similar to the description of read timeout in GCS plugins

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
masivesky: Got it, added the error code in failure collector itself

masivesky:
> Please update the markdown as well.

As discussed offline, we will be keeping this new parameter hidden, hence not adding it to the markdown files.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -544,14 +557,15 @@ void validateJobLabelKeyValue(FailureCollector failureCollector) {
           !containsMacro(NAME_INITIAL_RETRY_DURATION) && !containsMacro(NAME_MAX_RETRY_DURATION) &&
           !containsMacro(NAME_MAX_RETRY_COUNT) && !containsMacro(NAME_RETRY_MULTIPLIER)) {
           validateRetryConfiguration(
-            failureCollector, initialRetryDuration, maxRetryDuration, maxRetryCount, retryMultiplier
+            failureCollector, initialRetryDuration, maxRetryDuration, maxRetryCount, retryMultiplier, readTimeout
```
itsankit-google, line 560: I think here `readTimeout` should be provided using the `getter`: `getReadTimeout()`, otherwise the validation will fail for existing pipelines which will have `readTimeout` as `null`.

itsankit-google: ohh I see, you already have a check later on to skip validation if it is `null`.

Nvm, resolving the comment.

itsankit-google (approved):

PR #1437, opened 2024-08-28
[Security Vulnerability] run build with unit tests without elevated permissions

itsankit-google:
context:

This change prevents any external contributor to create another workflow in the PR with same name as `Trigger Build` which allow it to run the `Build With Unit Tests` with elevated permissions.

It uses `Github Hosted Runners` to build the surefire report with elevated permissions.

Upload/Download Artifact github actions v4 support downloading artifacts from other workflow runs -  https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories


tivv (approved):

.github/workflows/build.yml
```diff
@@ -15,21 +15,29 @@
...
+      || (contains(github.event.pull_request.labels.*.name, 'build')
+         && (github.event.action != 'labeled' || github.event.label.name == 'build')
+         )
 
     steps:
-      # Pinned 1.0.0 version
-      - uses: haya14busa/action-workflow_run-status@967ed83efa565c257675ed70cfe5231f062ddd94
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.event.workflow_run.head_sha }}
```
fproulx-boostsecurity, line 43: @itsankit-google since this workflow is not trigger on `workflow_run` anymore, this `actions/checkout` `ref` will be blank, probably defaulting to `actions/checkout` without any ref specified, which with `push` and `pull_request` is doing pretty much what you want anyway.... most likely this works, but it's confusing to leave the old reference to `workflow_run` before the security fix

PR #1462, opened 2024-11-26
GCS cloudbuild e2e fixes

AnkitCLI:
This PR contains fix for the error message mismatch for GCS on cloudbuild

### DIFF
--- a/src/e2e-test/features/gcs/source/GCSToGCSAdditonalTests.feature
+++ b/src/e2e-test/features/gcs/source/GCSToGCSAdditonalTests.feature
@@ -266,7 +266,9 @@ Feature: GCS source - Verification of GCS to GCS Additional Tests successful
     Then Deploy the pipeline
     Then Run the Pipeline in Runtime
     Then Wait till pipeline is in running state
+    Then Open and capture logs
     And Verify the pipeline status is "Failed"
+    Then Close the pipeline logs
     Then Open Pipeline logs and verify Log entries having below listed Level and Message:
       | Level | Message                                                     |
       | ERROR | errorMessageMultipleFileWithFirstRowAsHeaderEnabled         |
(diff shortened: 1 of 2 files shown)


itsankit-google (changes requested):
why do these tests not have `Open and capture logs` step like other tests?

Can we ensure these tests are consistent. This will also help us in verifying the pipeline logs and assumption by e2e tests.

image

AnkitCLI:
> why do these tests not have `Open and capture logs` step like other tests?
> 
> Can we ensure these tests are consistent. This will also help us in verifying the pipeline logs and assumption by e2e tests.
> 
> image

added the step for both scenarios.

src/e2e-test/features/gcs/source/GCSToGCSAdditonalTests.feature
```diff
@@ -266,10 +266,12 @@ Feature: GCS source - Verification of GCS to GCS Additional Tests successful
     Then Deploy the pipeline
     Then Run the Pipeline in Runtime
     Then Wait till pipeline is in running state
+    Then Open and capture logs
     And Verify the pipeline status is "Failed"
+    Then Close the pipeline logs
     Then Open Pipeline logs and verify Log entries having below listed Level and Message:
       | Level | Message                                                     |
-      | ERROR | errorMessageMultipleFileWithFirstRowAsHeaderEnabled         |
+      | WARN | errorMessageMultipleFileWithFirstRowAsHeaderEnabled         |
```
itsankit-google: how is this `WARN`?

I can see the level is `ERROR`: 
image

How is this test passing right now?

AnkitCLI: Basically to make it compatible with both cdap and CDF we use warn here because in cdap it only show **NumberFormatException** error in only **WARN** level in advanced logs. You are checking it in raw logs. But the above mentioned step only checks in advanced logs section.

itsankit-google: Can we modify the step to check in raw logs? So that it is consistent in both cdap and CDF.

AnkitCLI: Changed it to the common errorMessage for both cdap and CDF in **ERROR** level.

itsankit-google (approved):

PR #1473, opened 2024-12-11
PLUGIN-1438: Added support for destination table write preference in BQ Execute Plugin

anshumanks:
### Added support for destination table write preference in BQ Execute Plugin

Jira : [PLUGIN-1438](https://cdap.atlassian.net/browse/PLUGIN-1438)

### Description

This update introduces the ability for users to select a destination table write preference when storing results in BigQuery. Users can now choose from the following options:

- Write If Empty: Write to the table only if it is empty. If the table contains data, the operation will fail.
- Append To Table: Append results to an existing table, preserving its schema.
- Overwrite Table: Overwrite the existing table, replacing both the data and schema.

### UI Field

- Modified `BigQueryExecute-action.json`

### Docs

- Modified `BigQueryExecute-action.md`

### Code change

- Modified `BigQueryExecute.java`

### Unit Tests

- Modified `BigQueryExecuteTest.java`

### Tested
image

[PLUGIN-1438]: https://cdap.atlassian.net/browse/PLUGIN-1438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
@@ -58,9 +58,11 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
@@ -111,6 +113,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     String datasetProjectId = config.getDatasetProject();
     if (config.getStoreResults() && datasetProjectId != null && datasetName != null && tableName != null) {
       builder.setDestinationTable(TableId.of(datasetProjectId, datasetName, tableName));
+      builder.setWriteDisposition(JobInfo.WriteDisposition.valueOf(config.getWritePreference()));
     }
 
     // Enable or Disable the query cache to force live query evaluation.
@@ -274,6 +277,7 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     private static final String SQL = "sql";
     private static final String DATASET = "dataset";
     private static final String TABLE = "table";
+    private static final String WRITE_PREFERENCE = "writePreference";
     private static final String NAME_LOCATION = "location";
     public static final String NAME_BQ_JOB_LABELS = "jobLabels";
     private static final int ERROR_CODE_NOT_FOUND = 404;
@@ -290,6 +294,8 @@ public final class BigQueryExecute extends AbstractBigQueryAction {
     // Sn = a * (1 - r^n) / (r - 1)
     public static final long DEFULT_MAX_RETRY_DURATION_SECONDS = 63L;
     public static final int DEFAULT_READ_TIMEOUT = 120;
+    public static final Set<String> VALID_WRITE_PREFERENCES = Arrays.stream(JobInfo.WriteDisposition.values())
+        .map(Enum::name).collect(Collectors.toSet());
 
     @Description("Dialect of the SQL command. The value must be 'legacy' or 'standard'. " +
       "If set to 'standard', the query will use BigQuery's standard SQL: " +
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -398,13 +400,23 @@ public static final class Config extends AbstractBigQueryActionConfig {
     @Description("Timeout in seconds to read data from an established HTTP connection (Default value is 120).")
     private Integer readTimeout;
 
+    @Name(WRITE_PREFERENCE)
+    @Nullable
+    @Macro
+    @Description("Specifies the write preference for the destination table. Options are: " +
+        "Write if Empty: Only write if the table is empty. " +
+        "Append to Table: Add results to existing data. " +
+        "Overwrite Table: Replace all existing data.")
```
itsankit-google: Description/Tooltip should be crisp: `Specifies whether a job should overwrite or append the existing destination table if it already exists.`

Please add these details in markdown: https://github.com/data-integrations/google-cloud/blob/develop/docs/BigQueryExecute-action.md

itsankit-google (commented):
Please add evidence/screenshots in PR description under `Tested` section.

A good PR description example: https://github.com/data-integrations/google-cloud/pull/1444#issue-2518603891

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -546,9 +560,24 @@ public void validate(FailureCollector failureCollector, Map<String, String> argu
...
 
+      if (!containsMacro(WRITE_PREFERENCE)) {
+        validateWritePreference(failureCollector, getWritePreference());
+      }
+
       failureCollector.getOrThrowException();
     }
 
+    void validateWritePreference(FailureCollector failureCollector, String writePreference) {
+      Set<String> validPreferences = ImmutableSet.of("WRITE_EMPTY", "WRITE_APPEND", "WRITE_TRUNCATE");
```
itsankit-google: use `com.google.cloud.bigquery.JobInfo.WriteDisposition.WRITE_APPEND.name()` and so on.

src/main/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecute.java
```diff
@@ -546,9 +561,33 @@ public void validate(FailureCollector failureCollector, Map<String, String> argu
...
+      if (!validPreferences.contains(writePreference)) {
+        failureCollector.addFailure(
+            String.format("Invalid write preference '%s'. Allowed values are '%s', '%s', or '%s'.",
+                writePreference,
+                JobInfo.WriteDisposition.WRITE_EMPTY.name(),
+                JobInfo.WriteDisposition.WRITE_APPEND.name(),
+                JobInfo.WriteDisposition.WRITE_TRUNCATE.name()
+            ),
+            "Please provide a valid write preference."
+        )
```
itsankit-google: you can just do `Invalid write preference '%s'. Allowed values are validPreferences.toString()`

src/test/java/io/cdap/plugin/gcp/bigquery/action/BigQueryExecuteTest.java
```diff
@@ -228,15 +230,18 @@ public void testValidateRetryConfigurationWithInvalidReadTimeout() {
 
   @Test
   public void testValidateWritePreferenceWithInvalidValue() {
+    Set<String> validPreferences = ImmutableSet.of(
+        JobInfo.WriteDisposition.WRITE_EMPTY.name(),
+        JobInfo.WriteDisposition.WRITE_APPEND.name(),
+        JobInfo.WriteDisposition.WRITE_TRUNCATE.name()
```
itsankit-google: make it a static final variable in `BQExecute.Config` class

itsankit-google: we can also do something like this so that we don't need to maintain static list.

```
Set<String> validPreferences = Arrays.stream(JobInfo.WriteDisposition.values()).map(Enum::name).collect(Collectors.toSet());
```

itsankit-google (commented):
LGTM, please squash commits before merge.

itsankit-google (approved):

PR #1541, opened 2025-04-16
[PLUGIN-1846] Fixed NPE on GCSSource when bucket does not exist

shweta-deosarkar:
Tested [PLUGIN-1846](https://cdap.atlassian.net/browse/PLUGIN-1846) in CDAP Sandbox, instead of NPE, users will see the following error message : 
Unable to access GCS bucket 'shweta_test'. Ensure you entered the correct bucket path.

[PLUGIN-1846]: https://cdap.atlassian.net/browse/PLUGIN-1846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.gcp.gcs.source;
 
 import com.google.auth.Credentials;
+import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.common.base.Strings;
@@ -55,6 +56,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
@@ -122,7 +124,16 @@ public class GCSSource extends AbstractFileSource<GCSSource.GCSSourceConfig> {
     String location = null;
     try {
       // Get location of the source for lineage
-      location = storage.get(bucketName).getLocation();
+      Bucket bucket = storage.get(bucketName);
+      if (Objects.isNull(bucket)) {
+        String errorReason = String.format("Unable to access GCS bucket '%s'.",
+            bucketName);
+        collector.addFailure(
+            String.format("%s Ensure you entered the correct bucket path.", errorReason),
+            null);
+        collector.getOrThrowException();
+      }
+      location = bucket.getLocation();
     } catch (StorageException e) {
       String errorReason = String.format("Error code: %s, Unable to access GCS bucket '%s'. ",
         e.getCode(), bucketName);


src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -119,16 +121,24 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     }
 
     Storage storage = GCPUtils.getStorage(config.connection.getProject(), credentials);
+    Bucket bucket = storage.get(bucketName);
+    if (Objects.isNull(bucket)) {
+      String errorReason = String.format("Unable to access GCS bucket '%s'.",
+              bucketName);
```
itsankit-google: indentation seems to be off here, please use `Google Configuration Checker Style`.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -119,16 +121,24 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+    }
     String location = null;
     try {
       // Get location of the source for lineage
-      location = storage.get(bucketName).getLocation();
+      location = bucket.getLocation();
     } catch (StorageException e) {
       String errorReason = String.format("Error code: %s, Unable to access GCS bucket '%s'. ",
-        e.getCode(), bucketName);
+              e.getCode(), bucketName);
```
itsankit-google: please revert unwanted changes.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -119,16 +121,24 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
       collector.addFailure(String.format("%s %s", errorReason, e.getMessage()),
-          "Ensure you entered the correct bucket path and have permissions for it.")
-        .withStacktrace(e.getStackTrace());
+                      "Ensure you entered the correct bucket path and have permissions for it.")
+              .withStacktrace(e.getStackTrace());
```
itsankit-google: similar comment here

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -119,16 +121,24 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
 
     Storage storage = GCPUtils.getStorage(config.connection.getProject(), credentials);
+    Bucket bucket = storage.get(bucketName);
+    if (Objects.isNull(bucket)) {
+      String errorReason = String.format("Unable to access GCS bucket '%s'.",
+              bucketName);
+      collector.addFailure(String.format("%s Ensure you entered the correct bucket path.", errorReason),
+              null);
+      collector.getOrThrowException();
+    }
```
itsankit-google: I think we should move this to the same try & catch because it can also throw `StorageException` when bucket is not accessible.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
shweta-deosarkar: Done

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
shweta-deosarkar: Reverted.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
shweta-deosarkar: Done

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
shweta-deosarkar: Makes sense, done.

src/main/java/io/cdap/plugin/gcp/gcs/source/GCSSource.java
```diff
@@ -122,7 +124,16 @@ public void prepareRun(BatchSourceContext context) throws Exception {
     String location = null;
     try {
       // Get location of the source for lineage
-      location = storage.get(bucketName).getLocation();
+      Bucket bucket = storage.get(bucketName);
+      if (Objects.isNull(bucket)) {
```
itsankit-google, line 128: we can use direct comparison `bucket == null` which does the same brings more clarity and is also marginally faster.

shweta-deosarkar: It is usually considered a good practice to use Objects.isNull to avoid assigning the object null by mistake due to single '=' typos or human errors & Objects Null also performs direct comparison which doesn't impact compute time much.

itsankit-google: Ack, I don't have a strong opinion on this.

itsankit-google (approved):

PR #1544, opened 2025-05-06
[PLUGIN-1808] Retry policy to service account for 5xx errors for bigquery plugin

AnkitCLI:
[PLUGIN-1808](https://cdap.atlassian.net/browse/PLUGIN-1808) Added a retry policy to handle 5xx server errors when fetching the service account access token in the BigQuery plugin

[PLUGIN-1808]: https://cdap.atlassian.net/browse/PLUGIN-1808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -91,6 +91,8 @@ public class GCPUtils {
   public static final int BQ_DEFAULT_READ_TIMEOUT_SECONDS = 120;
   public static final String DATASTORE_SUPPORTED_DOC_URL = "https://cloud.google.com/datastore/docs/concepts/errors";
   public static final String BIG_TABLE_SUPPORTED_DOC_URL = "https://cloud.google.com/bigtable/docs/status-codes";
+  public static final String GCE_METADATA_SERVER_ERROR_SUPPORTED_DOC_URL =
+    "https://cloud.google.com/compute/docs/troubleshooting/troubleshoot-metadata-server";
 
   /**
    * Load a service account from the local file system.
--- a/src/main/java/io/cdap/plugin/gcp/common/ServerErrorException.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/ServerErrorException.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2025 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.common;
+
+/**
+ * Exception indicating a server-side error (HTTP 5xx).
+ * <p>
+ * This exception is intended to be used when a server responds with an HTTP 5xx status code,
+ * which typically indicates temporary unavailability or failure on the server's part.
+ * It can be used to trigger retries in retry frameworks like Failsafe.
+ */
+public class ServerErrorException extends RuntimeException {
+  private final int statusCode;
+
+  /**
+   * Constructs a new {@code ServerErrorException} with the given status code and message.
+   *
+   * @param statusCode the HTTP status code (should be in the 5xx range)
+   * @param message    the detail message explaining the error
+   * @param cause      the original cause of the error
+   */
+  public ServerErrorException(int statusCode, String message, Throwable cause) {
+    super("Server error [" + statusCode + "]: " + message, cause);
+    this.statusCode = statusCode;
+  }
+
+  /**
+   * Returns the HTTP status code associated with this server error.
+   *
+   * @return the 5xx HTTP status code that triggered this exception
+   */
+  public int getStatusCode() {
+    return statusCode;
+  }
+}
--- a/src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
@@ -21,16 +21,24 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.bigtable.repackaged.com.google.gson.Gson;
 import com.google.cloud.hadoop.util.AccessTokenProvider;
 import com.google.cloud.hadoop.util.CredentialFactory;
-import io.cdap.cdap.api.exception.ErrorCategory;
-import io.cdap.cdap.api.exception.ErrorCategory.ErrorCategoryEnum;
+import com.google.common.annotations.VisibleForTesting;
+import dev.failsafe.Failsafe;
+import dev.failsafe.FailsafeException;
+import dev.failsafe.RetryPolicy;
 import io.cdap.cdap.api.exception.ErrorType;
-import io.cdap.cdap.api.exception.ErrorUtils;
+import io.cdap.plugin.gcp.common.GCPErrorDetailsProviderUtil;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import io.cdap.plugin.gcp.common.ServerErrorException;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -43,19 +51,65 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
   private Configuration conf;
   private GoogleCredentials credentials;
   private static final Gson GSON = new Gson();
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceAccountAccessTokenProvider.class);
+  public static final int DEFAULT_INITIAL_RETRY_DURATION_SECONDS = 5;
+  public static final int DEFAULT_MAX_RETRY_COUNT = 5;
+  public static final int DEFAULT_MAX_RETRY_DURATION_SECONDS = 80;
+  private static final RetryPolicy<Object> RETRY_POLICY = createRetryPolicy();
+  private static final Pattern SERVER_ERROR_PATTERN = Pattern.compile("Unexpected Error code 5\\d{2} trying to get " +
+  "security access token from Compute Engine metadata for the default service account.*");
 
+  @VisibleForTesting
   @Override
   public AccessToken getAccessToken() {
-    try {
-      com.google.auth.oauth2.AccessToken token = getCredentials().getAccessToken();
-      if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
-        refresh();
-        token = getCredentials().getAccessToken();
+      try {
+        return Failsafe.with(RETRY_POLICY).get(() -> {
+            com.google.auth.oauth2.AccessToken token = retrieveAccessToken();
+            if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
+              refresh();
+              token = retrieveAccessToken();
+            }
+            return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+          });
+      } catch (FailsafeException e) {
+        Throwable t = e.getCause() != null ? e.getCause() : e;
+        ErrorType errorType = (t instanceof ServerErrorException) ? ErrorType.SYSTEM : ErrorType.UNKNOWN;
+        throw GCPErrorDetailsProviderUtil.getHttpResponseExceptionDetailsFromChain(
+          e, "Unable to get service account access token after retries.", errorType, true,
+          GCPUtils.GCE_METADATA_SERVER_ERROR_SUPPORTED_DOC_URL
+        );
       }
-      return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+    }
+
+  private static RetryPolicy<Object> createRetryPolicy() {
+    return RetryPolicy.builder()
+      .handle(ServerErrorException.class)
+      .withBackoff(Duration.ofSeconds(DEFAULT_INITIAL_RETRY_DURATION_SECONDS),
+                   Duration.ofSeconds(DEFAULT_MAX_RETRY_DURATION_SECONDS))
+      .withMaxRetries(DEFAULT_MAX_RETRY_COUNT)
+      .onRetry(event -> LOG.debug("Retry attempt {} due to {}", event.getAttemptCount(), event.getLastException().
+        getMessage()))
+      .onSuccess(event -> LOG.debug("Access Token Fetched Successfully."))
+      .onRetriesExceeded(
+       event -> LOG.error("Unable to get service account access token after {} retries.", event.getAttemptCount() - 1))
+      .build();
+  }
+
+  @VisibleForTesting
+  static boolean isServerError(IOException e) {
+    String msg = e.getMessage();
+    return msg != null && SERVER_ERROR_PATTERN.matcher(msg).matches();
+  }
+
+  com.google.auth.oauth2.AccessToken retrieveAccessToken() throws IOException {
+    try {
+      return getCredentials().getAccessToken();
     } catch (IOException e) {
-      throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategoryEnum.PLUGIN),
-        "Unable to get service account access token.", e.getMessage(), ErrorType.UNKNOWN, true, e);
+      if (isServerError(e)) {
+        throw new ServerErrorException(HttpStatus.SC_SERVICE_UNAVAILABLE, "Server error while fetching access token: "
+          + e.getMessage(), e);
+      }
+      throw e;
     }
   }
 
@@ -64,9 +118,13 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
     try {
       getCredentials().refresh();
     } catch (IOException e) {
-      throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategoryEnum.PLUGIN),
-        "Unable to refresh service account access token.", e.getMessage(),
-        ErrorType.UNKNOWN, true, e);
+      if (isServerError(e)) {
+        throw new ServerErrorException(HttpStatus.SC_SERVICE_UNAVAILABLE, "Server error during refresh: " +
+          e.getMessage(), e);
+      }
+      throw GCPErrorDetailsProviderUtil.getHttpResponseExceptionDetailsFromChain(
+        e, "Unable to refresh service account access token.", ErrorType.UNKNOWN, true,
+        GCPUtils.GCE_METADATA_SERVER_ERROR_SUPPORTED_DOC_URL);
     }
   }
 
--- a/src/test/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProviderTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProviderTest.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.gcp.gcs;
 
 import com.google.api.client.auth.oauth2.Credential;
+import com.google.auth.oauth2.AccessToken;
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration;
 import com.google.cloud.hadoop.util.AccessTokenProvider;
 import com.google.cloud.hadoop.util.CredentialFactory;
@@ -24,11 +25,15 @@ import com.google.cloud.hadoop.util.CredentialFromAccessTokenProviderClassFactor
 import com.google.cloud.hadoop.util.HadoopCredentialConfiguration;
 import com.google.common.collect.ImmutableList;
 import io.cdap.plugin.gcp.common.GCPUtils;
+import io.cdap.plugin.gcp.common.ServerErrorException;
 import org.apache.hadoop.conf.Configuration;
-import org.hamcrest.CoreMatchers;
 import org.junit.Assert;
 import org.junit.Test;
+import org.mockito.Mockito;
+
 import java.io.IOException;
+import java.time.Instant;
+import java.util.Date;
 import java.util.Map;
 
 /**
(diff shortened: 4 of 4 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -334,7 +334,7 @@ public boolean canConnect() {
   }
 
     private BigQuerySourceConfig(@Nullable BigQueryConnectorConfig connection, @Nullable String dataset,
-                                @Nullable String cmekKey, @Nullable String bucket, @Nullable String table) {
+                               @Nullable String cmekKey, @Nullable String bucket, @Nullable String table) {
```
itsankit-google: nit: indentation fix

itsankit-google (commented):
on a second thought, I feel just move all the retry configs & retry policy to `ServiceAccountAccessTokenProvider` class.

For start, we can use default configs and no need to make them configurable.

This will make the PR look cleaner and changes will be limited to one class.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,70 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+        return Failsafe.with(getRetryPolicy(initialRetryDuration, maxRetryDuration, maxRetryCount))
+          .get(() -> {
+            com.google.auth.oauth2.AccessToken token = safeGetAccessToken();
+            if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
+              refresh();
+              token = safeGetAccessToken();
+            }
+            return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+          });
+      } catch (Exception e) {
```
itsankit-google: extract cause from `FailesafeException`

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,70 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+        throw ErrorUtils.getProgramFailureException(
+          new ErrorCategory(ErrorCategoryEnum.PLUGIN),
+          "Unable to get service account access token after retries.",
+          e.getMessage(),
+          ErrorType.UNKNOWN,
```
itsankit-google: for server error use `SYSTEM` error type

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,70 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
   private Configuration conf;
   private GoogleCredentials credentials;
   private static final Gson GSON = new Gson();
+  private static final Logger logger = LoggerFactory.getLogger(ServiceAccountAccessTokenProvider.class);
```
itsankit-google: usually use private static final variables in uppercase: `logger -> LOG`

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,70 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+  }
+
+  private com.google.auth.oauth2.AccessToken safeGetAccessToken() throws IOException {
+    try {
+      return getCredentials().getAccessToken();
     } catch (IOException e) {
-      throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategoryEnum.PLUGIN),
-        "Unable to get service account access token.", e.getMessage(), ErrorType.UNKNOWN, true, e);
+      if (isServerError(e)) {
+        throw new ServerErrorException(HttpStatus.SC_SERVICE_UNAVAILABLE, "Server error while fetching access token: "
```
itsankit-google, line 109: why introduce new `ServerException`, can use `ProgramFailureException` with status code `5xx`?

AnkitCLI: We generally add a custom exception when handling retry in other plugins as well.
I have added server error exception to be consistent, it is easier to maintain, in retry policy adding program failure exception may cause an incorrect retry.

what do you think?

itsankit-google: can't we choose based on `statusCode` in case of `ProgramFailureException` to retry?

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: My concern is what if some other function throws a program failure exception with status code 5xx and we trigger a retry based on status code but with custom exception we can control exactly when to retry.
As of now there is no property which can be checked to confirm if we can retry the program failure exception, they could be thrown from a dependent library.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +51,71 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
   @Override
   public AccessToken getAccessToken() {
-    try {
-      com.google.auth.oauth2.AccessToken token = getCredentials().getAccessToken();
-      if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
-        refresh();
-        token = getCredentials().getAccessToken();
+    int initialRetryDuration = DEFAULT_INITIAL_RETRY_DURATION_SECONDS;
+    int maxRetryCount = DEFAULT_MAX_RETRY_COUNT;
+    int maxRetryDuration =  DEFAULT_MAX_RETRY_DURATION_SECONDS;
```
itsankit-google: why are we creating copy of instance variables here? It eliminates the purpose of them being `final & static`

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +51,71 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+    return RetryPolicy.builder()
+      .handle(ServerErrorException.class)
+      .withBackoff(Duration.ofSeconds(initialRetryDuration), Duration.ofSeconds(maxRetryDuration))
+      .withMaxRetries(maxRetryCount)
+      .onRetry(event -> LOG.debug("Retry attempt {} due to {}", event.getAttemptCount(), event.getLastException().
+        getMessage()))
+      .onSuccess(event -> LOG.debug("Access Token Fetched Successfully."))
+      .onRetriesExceeded(event -> LOG.error("Retry limit reached for Service account."))
+      .build();
+  }
```
itsankit-google: this can be a private static method & retry policy can have a static final instance variable rather than creating retry policy for every call.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
itsankit-google: ack.

src/main/java/io/cdap/plugin/gcp/common/ServerErrorException.java
```diff
@@ -0,0 +1,48 @@
...
+package io.cdap.plugin.gcp.common;
+
+/**
+ * Exception indicating a server-side error (HTTP 5xx).
+ * <p>
+ * This exception is intended to be used when a server responds with an HTTP 5xx status code,
+ * which typically indicates temporary unavailability or failure on the server's part.
+ * It can be used to trigger retries in retry frameworks like Failsafe.
+ */
+public class ServerErrorException extends RuntimeException {
```
itsankit-google, line 26: preserve the original cause

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/common/ServerErrorException.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +51,67 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+            }
+            return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+          });
+      } catch (FailsafeException e) {
+        Throwable t = e.getCause() != null ? e.getCause() : e;
+        throw ErrorUtils.getProgramFailureException(
+          new ErrorCategory(ErrorCategoryEnum.PLUGIN),
+          "Unable to get service account access token after retries.",
+          t.getMessage(),
+          ErrorType.SYSTEM,
```
itsankit-google: `getCause()` should match first check whether it was a `ServerException` then only use `SYSTEM` otherwise `UNKNOWN`.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -64,6 +120,10 @@ public void refresh() throws IOException {
     try {
       getCredentials().refresh();
     } catch (IOException e) {
+      if (isServerError(e)) {
+        throw new ServerErrorException(HttpStatus.SC_SERVICE_UNAVAILABLE, "Server error during refresh: " +
+          e.getMessage(), e);
+      }
       throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategoryEnum.PLUGIN),
         "Unable to refresh service account access token.", e.getMessage(),
         ErrorType.UNKNOWN, true, e);
```
itsankit-google: better to use https://github.com/data-integrations/google-cloud/blob/c13becec11d9e5333bc85573ebc0ef287e9aae57/src/main/java/io/cdap/plugin/gcp/common/GCPErrorDetailsProviderUtil.java#L117 with externalDocUrl as https://cloud.google.com/compute/docs/troubleshooting/troubleshoot-metadata-server

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +51,67 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+            com.google.auth.oauth2.AccessToken token = safeGetAccessToken();
+            if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
+              refresh();
+              token = safeGetAccessToken();
+            }
+            return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+          });
+      } catch (FailsafeException e) {
+        Throwable t = e.getCause() != null ? e.getCause() : e;
+        throw ErrorUtils.getProgramFailureException(
```
itsankit-google: better to use https://github.com/data-integrations/google-cloud/blob/c13becec11d9e5333bc85573ebc0ef287e9aae57/src/main/java/io/cdap/plugin/gcp/common/GCPErrorDetailsProviderUtil.java#L117 with externalDocUrl as https://cloud.google.com/compute/docs/troubleshooting/troubleshoot-metadata-server

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -90,6 +90,8 @@ public class GCPUtils {
   public static final int BQ_DEFAULT_READ_TIMEOUT_SECONDS = 120;
   public static final String DATASTORE_SUPPORTED_DOC_URL = "https://cloud.google.com/datastore/docs/concepts/errors";
   public static final String BIG_TABLE_SUPPORTED_DOC_URL = "https://cloud.google.com/bigtable/docs/status-codes";
+  public static final String SERVER_ERROR_SUPPORTED_DOC_URL =
```
itsankit-google: `GCE_METADATA_SERVER_ERROR_SUPPORTED_DOC_URL`

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,64 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+      .onRetriesExceeded(event -> LOG.error("Retry limit reached for Service account."))
+      .build();
+  }
+
+  private boolean isServerError(IOException e) {
+    String msg = e.getMessage();
+    return msg != null && msg.matches("^5\\d{2}$"); // crude check for 5xx codes
+  }
+
+  private com.google.auth.oauth2.AccessToken safeGetAccessToken() throws IOException {
```
itsankit-google: what do we mean by `safeGetAccessToken` here?

itsankit-google: Usually method names should indicate what are we trying to do here.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,64 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
   public AccessToken getAccessToken() {
-    try {
-      com.google.auth.oauth2.AccessToken token = getCredentials().getAccessToken();
-      if (token == null || token.getExpirationTime().before(Date.from(Instant.now()))) {
-        refresh();
-        token = getCredentials().getAccessToken();
+      LOG.debug(
+        "Initializing RetryPolicy with the following configuration: MaxRetryCount: {}, InitialRetryDuration: {}s, " +
+          "MaxRetryDuration: {}s", DEFAULT_MAX_RETRY_COUNT, DEFAULT_INITIAL_RETRY_DURATION_SECONDS,
+        DEFAULT_MAX_RETRY_DURATION_SECONDS);
```
itsankit-google: no need of this log for every access token fetch.

This can be present in the constructor.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -64,9 +118,13 @@ public void refresh() throws IOException {
     try {
       getCredentials().refresh();
```
itsankit-google, line 119: We should also add retries on `getCredentials().refresh()`

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +52,64 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+      .onRetry(event -> LOG.debug("Retry attempt {} due to {}", event.getAttemptCount(), event.getLastException().
+        getMessage()))
+      .onSuccess(event -> LOG.debug("Access Token Fetched Successfully."))
+      .onRetriesExceeded(event -> LOG.error("Retry limit reached for Service account."))
+      .build();
+  }
+
+  private boolean isServerError(IOException e) {
+    String msg = e.getMessage();
+    return msg != null && msg.matches("^5\\d{2}$"); // crude check for 5xx codes
```
itsankit-google: We should broaden the matching regex to match something like : 

```
Unexpected Error code 5xx trying to get security access token from Compute Engine metadata for the default service account
```

And we could also keep the compiled pattern as prival final static variable instead of compiling it for every call.

Regex matches are cheap but pattern compilation is costly.

src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
AnkitCLI: changed

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: changed to retrieveAccessToken

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: removed

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,33 +50,81 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
   private Configuration conf;
   private GoogleCredentials credentials;
   private static final Gson GSON = new Gson();
+  private static final Logger LOG = LoggerFactory.getLogger(ServiceAccountAccessTokenProvider.class);
+  public static final int DEFAULT_INITIAL_RETRY_DURATION_SECONDS = 5;
+  public static final int DEFAULT_MAX_RETRY_COUNT = 5;
+  public static final int DEFAULT_MAX_RETRY_DURATION_SECONDS = 80;
+  private static final RetryPolicy<Object> RETRY_POLICY = createRetryPolicy();
+  private static final Pattern SERVER_ERROR_PATTERN = Pattern.compile(".*5\\d{2}.*");
```
itsankit-google: This comment is still not addressed : https://github.com/data-integrations/google-cloud/pull/1544#discussion_r2128383917

We are still looking at very broad pattern of just finding `5xx` in the error message.

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,33 +50,81 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+              refresh();
+              token = retrieveAccessToken();
+            }
+            return new AccessToken(token.getTokenValue(), token.getExpirationTime().getTime());
+          });
+      } catch (FailsafeException e) {
+        Throwable t = e.getCause() != null ? e.getCause() : e;
+        ErrorType errorType = (t instanceof ServerErrorException) ? ErrorType.SYSTEM : ErrorType.UNKNOWN;
+        throw GCPErrorDetailsProviderUtil.getHttpResponseExceptionDetailsFromChain(
+          e, "Unable to get service account access token after retries.", errorType, true,
```
itsankit-google, line 78: can we log after how many retries:

```
Unable to get service account access token after X retries.
```

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,33 +50,81 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+        getCredentials().refresh();
+      } catch (IOException e) {
+        if (isServerError(e)) {
+          throw new ServerErrorException(HttpStatus.SC_SERVICE_UNAVAILABLE,
+                                         "Server error during refresh: " + e.getMessage(), e);
+        }
+        throw GCPErrorDetailsProviderUtil.getHttpResponseExceptionDetailsFromChain(
+          e, "Unable to refresh service account access token.", ErrorType.UNKNOWN, true,
+          GCPUtils.GCE_METADATA_SERVER_ERROR_SUPPORTED_DOC_URL);
+      }
```
itsankit-google: this should be out of the retries of failsafe similarly like `getAccessToken()`

itsankit-google (changes requested):

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: The pattern is broad as we are not sure of the error message and are only matching on 5xx.
Is there a way get the error message structure to narrow this regex match.

itsankit-google: see this class : https://github.com/googleapis/google-auth-library-java/blob/d02ab85ae8d29402650d3cbf00e00363a903fe31/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L353-L354

We already know from where the error message is coming. Please look at the stacktrace of errors in the JIRA.

AnkitCLI: added done

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added in the retry policy

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: we don't need retry policy here because refresh method is already defined in the **getAccessToken** method .

itsankit-google: Sorry, I don't get this, why do we not need retries in refresh method?

AnkitCLI: **Refresh** method is getting called from **getAccessToken** method which already has retry policy. Adding a second retry here will cause a nested retry.

itsankit-google: I don't see `getAccessToken()` or `retrieveAccessToken()` method being called from `refreshToken()`. Can you please point me to the code reference?

itsankit-google (commented):
Can you please add a unit test?

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
```diff
@@ -43,19 +50,63 @@ public class ServiceAccountAccessTokenProvider implements AccessTokenProvider {
...
+      .withMaxRetries(DEFAULT_MAX_RETRY_COUNT)
+      .onRetry(event -> LOG.debug("Retry attempt {} due to {}", event.getAttemptCount(), event.getLastException().
+        getMessage()))
+      .onSuccess(event -> LOG.debug("Access Token Fetched Successfully."))
+      .onRetriesExceeded(
+       event -> LOG.error("Unable to get service account access token after {} retries.", event.getAttemptCount() - 1))
+      .build();
+  }
+
+  static boolean isServerError(IOException e) {
```
itsankit-google, line 99: please add annotation `@VisibleForTesting`.

itsankit-google (commented):
please also add unit tests to check retries, for example : https://github.com/cdapio/cdap/pull/15226/files#diff-acc8b9b649f280ee45108ce0220bc742b327acd041f1c1022489452f021ad852R53-R69

src/main/java/io/cdap/plugin/gcp/gcs/ServiceAccountAccessTokenProvider.java
AnkitCLI: added

itsankit-google (approved):

PR #1569, opened 2025-07-15
[cherry-pick][6.11]Removing htrace for security vuln

sahusanket:


pom.xml
```diff
@@ -445,86 +449,6 @@
       <artifactId>google-cloud-datastore</artifactId>
       <version>${google.cloud.datastore.version}</version>
     </dependency>
-    <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-common</artifactId>
```
itsankit-google, line 450: why are we removing this?

PR #1578, opened 2025-08-04
Removing verify tag

AnkitCLI:
This PR contains change to remove verify tag from pom.xml to get correct build details and cucumber reports.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1233,7 +1233,8 @@
           <execution>
             <goals>
               <goal>integration-test</goal>
-              <goal>verify</goal>
+              <!-- Commented out verify goal to allow Cucumber reports to generate correctly even when tests fail -->
+              <!-- <goal>verify</goal> -->
             </goals>
           </execution>
         </executions>


itsankit-google (commented):
why this change?

AnkitCLI:
> why this change?

The reason is that when the verify goal is used and the build fails due to a test failure, the Cucumber reports were still getting generated but not showing any test failures. This made it difficult to identify what exactly failed. And i checked with other repo's we have commented out this in most of the repo's as well.

itsankit-google:
> > why this change?
> 
> The reason is that when the verify goal is used and the build fails due to a test failure, the Cucumber reports were still getting generated but not showing any test failures. This made it difficult to identify what exactly failed. And i checked with other repo's we have commented out this in most of the repo's as well.

I don't see it commented out in database-plugins repo?

AnkitCLI:
> > > why this change?
> > 
> > 
> > The reason is that when the verify goal is used and the build fails due to a test failure, the Cucumber reports were still getting generated but not showing any test failures. This made it difficult to identify what exactly failed. And i checked with other repo's we have commented out this in most of the repo's as well.
> 
> I don't see it commented out in database-plugins repo?
We are also facing some issues in DB repo. This issue is when build fails no cucumber report would be generated. Currently looking into it.
But if we you see in salesforce and hydrator repo's you will see the tag is removed. In starting in google cloud also the tag is commented out.  https://github.com/data-integrations/salesforce/blob/develop/pom.xml. https://github.com/cdapio/hydrator-plugins/blob/develop/pom.xml

itsankit-google:
> > > > why this change?
> > > 
> > > 
> > > The reason is that when the verify goal is used and the build fails due to a test failure, the Cucumber reports were still getting generated but not showing any test failures. This made it difficult to identify what exactly failed. And i checked with other repo's we have commented out this in most of the repo's as well.
> > 
> > 
> > I don't see it commented out in database-plugins repo?
> > We are also facing some issues in DB repo. This issue is when build fails no cucumber report would be generated. Currently looking into it.
> > But if we you see in salesforce and hydrator repo's you will see the tag is removed. In starting in google cloud also the tag is commented out.  https://github.com/data-integrations/salesforce/blob/develop/pom.xml. https://github.com/cdapio/hydrator-plugins/blob/develop/pom.xml

Can you explain more how this helps in fixing the issue?

AnkitCLI:
In my opinion, The **failsafe:verify** goal was causing the build to terminate early on test failures, before Cucumber could generate its final report. This led to a situation where the Cucumber report incorrectly showed all tests as passed. By removing the verify goal, the build completes even if some tests fail, allowing Cucumber to generate accurate reports that include all test outcomes — both passed and failed.

pom.xml
```diff
@@ -1233,7 +1233,7 @@
           <execution>
             <goals>
               <goal>integration-test</goal>
-              <goal>verify</goal>
+<!--              <goal>verify</goal>-->
```
itsankit-google: please add a comment for disabling this check.

itsankit-google (approved):

pom.xml
AnkitCLI: done

PR #1583, opened 2025-09-23
[PLUGIN-1931] Add defaults for Time and Range Partitioning in BQ source plugin

adrikagupta:
If we pass a filter input in BQ source plugin and keep partitionFromDate and partitionToDate values as empty for a table where requirePartitionFilter is set as true, then the preview/deployment run will fail because since requirePartitionFilter is true, the query requires a mandatory “where” clause on partitioning column.
Different types of partitioning for BQ are listed in https://cloud.google.com/bigquery/docs/partitioned-tables .

To fix we are adding defaults for range, time unit and ingestion partitioning which would be used if requirePartitionFilter is set as true and no partitioning inputs are passed. The default would be "parititioning-column is NOT NULL OR parititioning-column is NULL" to get the entire data. 
A few things were noted when considering these defaults:
- Time-unit partitioning columns can have future-dated values.
- We are not using startDate and endDate for range partitioning, as it's possible to have values that do not lie within the specified range.

Also, we have added limit and order by fields in UI since filter field is only meant for WHERE clause.

Testing:
- Created a pipeline with BQ source plugin. Used a table having requirePartitioningFilter as true and **ingestion partitioning**. Provided filter as "1=1", order by and limit fields. Preview and deployment run was successful.
- Created a pipeline with BQ source plugin. Used a table having requirePartitioningFilter as true and **time-unit partitioning**. Provided filter as "1=1", order by and limit fields. Preview and deployment run was successful.
- Created a pipeline with BQ source plugin. Used a table having requirePartitioningFilter as true and **range partitioning**. Provided filter as "1=1", order by and limit fields. Preview and deployment run was successful.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
@@ -58,6 +58,7 @@ public final class BigQuerySourceConfig extends BigQueryBaseConfig {
   private static final String VALID_DATE_FORMAT = "yyyy-MM-dd";
   private static final String SCHEME = "gs://";
   private static final String WHERE = "WHERE";
+  private static final String ORDER_BY = "ORDER BY";
   private static final String NAME_READ_TIMEOUT = "readTimeout";
   public static final Set<Schema.Type> SUPPORTED_TYPES =
     ImmutableSet.of(Schema.Type.LONG, Schema.Type.STRING, Schema.Type.DOUBLE, Schema.Type.BOOLEAN, Schema.Type.BYTES,
@@ -71,6 +72,8 @@ public final class BigQuerySourceConfig extends BigQueryBaseConfig {
   public static final String NAME_ENABLE_QUERYING_VIEWS = "enableQueryingViews";
   public static final String NAME_VIEW_MATERIALIZATION_PROJECT = "viewMaterializationProject";
   public static final String NAME_VIEW_MATERIALIZATION_DATASET = "viewMaterializationDataset";
+  public static final String NAME_LIMIT = "limit";
+  public static final String NAME_ORDER_BY = "orderBy";
 
   @Name(Constants.Reference.REFERENCE_NAME)
   @Nullable
@@ -110,6 +113,19 @@ public final class BigQuerySourceConfig extends BigQueryBaseConfig {
   @Description("The WHERE clause filters out rows by evaluating each row against boolean expression, " +
     "and discards all rows that do not return TRUE (that is, rows that return FALSE or NULL).")
   private String filter;
+
+  @Name(NAME_ORDER_BY)
+  @Macro
+  @Nullable
+  @Description("The ORDER BY clause sorts the results of a query based on one or more columns. " +
+      "For example, 'name asc, age desc'.")
+  private String orderBy;
+
+  @Name(NAME_LIMIT)
+  @Macro
+  @Nullable
+  @Description("The LIMIT clause restricts the number of rows returned by the query.")
+  private Long limit;
 
   @Name(NAME_ENABLE_QUERYING_VIEWS)
   @Macro
@@ -186,6 +202,13 @@ public final class BigQuerySourceConfig extends BigQueryBaseConfig {
     }
     if (!containsMacro(NAME_CMEK_KEY)) {
       validateCmekKey(collector, arguments);
+    }
+
+    if (!containsMacro(NAME_LIMIT) && limit != null) {
+      if (limit < 0) {
+        collector.addFailure("Invalid limit value.", "Limit must be a non-negative number.")
+            .withConfigProperty(NAME_LIMIT);
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -132,19 +136,34 @@ private void processQuery(JobContext context) throws IOException, InterruptedExc
     String partitionFromDate = configuration.get(BigQueryConstants.CONFIG_PARTITION_FROM_DATE, null);
     String partitionToDate = configuration.get(BigQueryConstants.CONFIG_PARTITION_TO_DATE, null);
     String filter = configuration.get(BigQueryConstants.CONFIG_FILTER, null);
+    String limit = configuration.get(BigQueryConstants.CONFIG_LIMIT, null);
+    if (!Strings.isNullOrEmpty(limit)) {
```
vikasrathee-cs: No need to add this check here as you have already added in validate method in config class.

src/test/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfigTest.java
```diff
@@ -25,6 +25,8 @@ public void testValidatePartitionDateWithInvalidDate() throws NoSuchMethodExcept
                         collector.getValidationFailures().stream().findFirst().get().getMessage());
   }
 
+
+
```
vikasrathee-cs: remove this unintentional change

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -187,6 +203,15 @@ public void validate(FailureCollector collector, Map<String, String> arguments)
     if (!containsMacro(NAME_CMEK_KEY)) {
       validateCmekKey(collector, arguments);
     }
+
+    if (!containsMacro(NAME_LIMIT) && !Strings.isNullOrEmpty(limit)) {
```
itsankit-google: please also add a check for limit should be greater than 0

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -111,6 +114,19 @@ public final class BigQuerySourceConfig extends BigQueryBaseConfig {
...
+  @Nullable
+  @Description("The ORDER BY clause sorts the results of a query based on one or more columns. " +
+      "For example, 'name asc, age desc'.")
+  private String orderBy;
+
+  @Name(NAME_LIMIT)
+  @Macro
+  @Nullable
+  @Description("The LIMIT clause restricts the number of rows returned by the query.")
+  private String limit;
```
vikasrathee-cs: make this variable long

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -132,19 +136,34 @@ private void processQuery(JobContext context) throws IOException, InterruptedExc
     String partitionFromDate = configuration.get(BigQueryConstants.CONFIG_PARTITION_FROM_DATE, null);
     String partitionToDate = configuration.get(BigQueryConstants.CONFIG_PARTITION_TO_DATE, null);
     String filter = configuration.get(BigQueryConstants.CONFIG_FILTER, null);
+    String limit = configuration.get(BigQueryConstants.CONFIG_LIMIT, null);
```
itsankit-google, line 139: should we use `configuration.getInt`?

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
adrikagupta: Sure, done. Thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
adrikagupta: Right, done. Thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
adrikagupta: Right, done. Thanks!

src/test/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfigTest.java
adrikagupta: Done, thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
adrikagupta: Right. I kept it configuration.get() here because the limit value is only used for string concatenation to build the final SQL query. Let me know if it should be changed. Thanks!

widgets/BigQueryTable-batchsource.json
```diff
@@ -151,6 +151,18 @@
...
           }
         },
+        {
+          "widget-type": "textbox",
+          "name": "orderBy",
+          "label": "Order By",
+          "widget-attributes": {"placeholder": "Column(s) to order by, e.g., 'name asc, age desc'"}
+        },
+        {
+          "widget-type": "textbox",
```
itsankit-google, line 161: shouldn't it be a number field?

itsankit-google: I think it might be fine because we don't want a default here.

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
```diff
@@ -187,6 +203,13 @@ public void validate(FailureCollector collector, Map<String, String> arguments)
     if (!containsMacro(NAME_CMEK_KEY)) {
       validateCmekKey(collector, arguments);
     }
+
+    if (!containsMacro(NAME_LIMIT) && limit != null) {
+      if (limit < 0) {
```
itsankit-google, line 208: does `limit = 0` makes sense? when will someone want to configure it to 0? If not, please make the check  as `<=0`

itsankit-google (approved):
one small comment, otherwise LGTM

src/main/java/io/cdap/plugin/gcp/bigquery/source/BigQuerySourceConfig.java
adrikagupta: Understandable, I kept the check as < 0 because LIMIT 0 is valid BigQuery SQL. A user might set LIMIT 0 to perform a dry run of the pipeline which allows them to validate the full pipeline (eg permissions), without processing any data which is useful for testing. Let me know if I should change it. Thanks!

itsankit-google: Got it, sg.

PR #1584, opened 2025-10-01
[PLUGIN-1931] Add necessary condition to prevent query job when appropriate parameters are empty

adrikagupta:
In previous version, if a user provided no filters, no limits, and no partition dates, the code would still execute the full logic, only to determine at the end that the query should default to reading the source table directly (select * from "table") and creates a temporary table. But in such cases, there is no need to create a temporary table. Hence, added this check.

Testing:
1. Created a pipeline with BQ source plugin using a table having required partitioning filter as true. Provided no filter, partitioning inputs, limit or order by clauses. The preview/deployment run is successful.
2. Created a pipeline with BQ source plugin using a table having required partitioning filter as true. Provided filter, but no partitioning inputs, limit or order by clauses. The preview/deployment run is successful.
3. Created a pipeline with BQ source plugin using a table having required partitioning filter as false. Provided no filter, partitioning inputs, limit or order by clauses. The preview/deployment run is successful.
4. Created a pipeline with BQ source plugin using a table having required partitioning filter as false. Provided filter, limit but no partitioning inputs, order by clauses. The preview/deployment run is successful.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
@@ -182,6 +182,12 @@ public class PartitionedBigQueryInputFormat extends AbstractBigQueryInputFormat<
       String datasetProject, String dataset, String table, String limit, String orderBy,
       Boolean isPartitionFilterRequired, StandardTableDefinition tableDefinition) {
 
+    if (Strings.isNullOrEmpty(filter) && Strings.isNullOrEmpty(orderBy) && Strings.isNullOrEmpty(
+        limit)
+        && Strings.isNullOrEmpty(partitionFromDate) && Strings.isNullOrEmpty(partitionToDate)) {
+      return null;
+    }
+
     RangePartitioning rangePartitioning = tableDefinition.getRangePartitioning();
     TimePartitioning timePartitioning = tableDefinition.getTimePartitioning();
     StringBuilder condition = new StringBuilder();
--- a/src/test/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormatTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormatTest.java
@@ -34,13 +34,15 @@ import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
 public class PartitionedBigQueryInputFormatTest {
+
   private static final String TEST_PROJECT = "test-project";
   private static final String TEST_DATASET = "test-dataset";
   private static final String TEST_TABLE = "test-table";
   private static final String TEST_FILTER = "age > 10";
   private static final String TEST_LIMIT = "100";
   private static final String TEST_ORDER_BY = "name asc";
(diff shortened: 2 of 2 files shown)


src/test/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormatTest.java
```diff
@@ -260,27 +253,78 @@ public void testGenerateQuery_TimePartitionFilterNotRequiredWithDates() {
...
+    String expectedQuery = String.format("select * from %s where %s limit %s",
+        TEST_TABLE_SPEC, TEST_DEFAULT_TIME_UNIT_CONDITION, TEST_LIMIT);
+
+    String generatedQuery = format.generateQuery(null, null, null,
+        TEST_PROJECT, TEST_DATASET, TEST_TABLE,
+        TEST_LIMIT, null,
+        true, mockTableDefinition);
     Assert.assertEquals(expectedQuery, generatedQuery);
   }
+
```
itsankit-google: Nit: Remove extra empty line

itsankit-google (commented):
Please add PR description on what does this change do, how is it tested.

src/test/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormatTest.java
adrikagupta: Right, done. Thanks!

src/main/java/io/cdap/plugin/gcp/bigquery/source/PartitionedBigQueryInputFormat.java
```diff
@@ -182,6 +182,12 @@ String generateQuery(String partitionFromDate, String partitionToDate, String fi
       String datasetProject, String dataset, String table, String limit, String orderBy,
       Boolean isPartitionFilterRequired, StandardTableDefinition tableDefinition) {
 
+    if (Strings.isNullOrEmpty(filter) && Strings.isNullOrEmpty(orderBy) && Strings.isNullOrEmpty(
+        limit)
+        && Strings.isNullOrEmpty(partitionFromDate) && Strings.isNullOrEmpty(partitionToDate)) {
+      return null;
```
itsankit-google, line 188: Even though these configs are not provided but when querying from the partitioned table it still requires default partitioning condition to be added right?

adrikagupta: In the BigQuery plugin, we send a table name to a third-party library for data extraction and processing.
Before sending the table name, we determine if a subset of the data is needed. If filtering (e.g., using a WHERE clause, LIMIT, ORDER BY, or date range filters like partitionToDate and partitionFromDate) is necessary, we create a temporary BigQuery table containing the required data subset. The name of this temporary table is then sent to the 3P library.
However, if the entire table can be processed without subsetting, we send the original table name directly to the 3P library. The 3P library is responsible for extracting data from the provided table. It is also capable of handling tables where requirePartitioningFilter is set to true.
This scenario has been tested.

@vikasrathee-cs Please let me know if I am mistaken or if you have any additional information to add.

vikasrathee-cs: Yes, that is correct Adrika, without any of the above filters or limit, 3P library can extract data without providing partitioning condition even if requirePartitionFilter flag is true.

itsankit-google (approved):

PR #1590, opened 2026-01-16
[PLUGIN-1920] Add support for WRITE_TRUNCATE_DATA in BQ Sink

itsankit-google:
### Changes [PLUGIN-1920](https://cdap.atlassian.net/browse/PLUGIN-1920)

Add support for `WRITE_TRUNCATE_DATA` with `Truncate Table` config. For more details, see [here](https://docs.cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/BigQueryAuditMetadata.WriteDisposition).

image

### Tested

- Verified running a pipeline using `WRITE_TRUNCATE`.

- Verified running a pipeline using `WRITE_TRUNCATE_DATA` did not remove the existing labels.

### Limitations

- The repo is using `google-cloud-bigquery:1.37.1` which does not contain `WRITE_TRUNCATE_DATA` in `JobInfo.WriteDisposition` enum. Hence, using it as `String` everywhere.

[PLUGIN-1920]: https://cdap.atlassian.net/browse/PLUGIN-1920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
@@ -270,7 +270,7 @@ public abstract class AbstractBigQuerySink extends BatchSink<StructuredRecord, S
     baseConfiguration.setBoolean(BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION,
                                  config.isAllowSchemaRelaxation());
     baseConfiguration.setStrings(BigQueryConfiguration.OUTPUT_TABLE_WRITE_DISPOSITION.getKey(),
-                                 config.getWriteDisposition().name());
+                                 config.getWriteDisposition());
     baseConfiguration.setStrings(BigQueryConstants.CONFIG_JSON_STRING_FIELDS, config.getJsonStringFields());
     // this setting is needed because gcs has default chunk size of 64MB. This is large default chunk size which can
     // cause OOM issue if there are many tables being written. See this - CDAP-16670
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
@@ -49,6 +49,7 @@ public abstract class AbstractBigQuerySinkConfig extends BigQueryBaseConfig {
                     Schema.Type.BOOLEAN, Schema.Type.BYTES, Schema.Type.ARRAY, Schema.Type.RECORD);
 
   public static final String NAME_TRUNCATE_TABLE = "truncateTable";
+  public static final String NAME_WRITE_DISPOSITION = "writeDisposition";
   public static final String NAME_LOCATION = "location";
   private static final String NAME_GCS_CHUNK_SIZE = "gcsChunkSize";
   public static final String NAME_BQ_JOB_LABELS = "jobLabels";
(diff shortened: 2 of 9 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -82,6 +83,15 @@ public abstract class AbstractBigQuerySinkConfig extends BigQueryBaseConfig {
     + "Should only be used with the Insert operation. This could overwrite the table schema")
```
sahusanket: NIT : I think we should mention `This could overwrite the table schema based on write disposition value.` for more clarity

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
```diff
@@ -155,9 +165,16 @@ public boolean isAllowSchemaRelaxation() {
     return allowSchemaRelaxation == null ? false : allowSchemaRelaxation;
   }
 
-  public JobInfo.WriteDisposition getWriteDisposition() {
-    return isTruncateTableSet() ? JobInfo.WriteDisposition.WRITE_TRUNCATE
-      : JobInfo.WriteDisposition.WRITE_APPEND;
+  public String getTruncateTableWriteDisposition() {
```
sahusanket: since it's used in this class, can we make it private ?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
itsankit-google: good catch, thanks.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
```diff
@@ -270,7 +270,7 @@ private Configuration getBaseConfiguration(@Nullable CryptoKeyName cmekKeyName)
     baseConfiguration.setBoolean(BigQueryConstants.CONFIG_ALLOW_SCHEMA_RELAXATION,
                                  config.isAllowSchemaRelaxation());
     baseConfiguration.setStrings(BigQueryConfiguration.OUTPUT_TABLE_WRITE_DISPOSITION.getKey(),
-                                 config.getWriteDisposition().name());
+                                 config.getWriteDisposition());
```
sahusanket, line 273: nit : any reason for this change ?

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySinkConfig.java
itsankit-google: ack, updated.

src/main/java/io/cdap/plugin/gcp/bigquery/sink/AbstractBigQuerySink.java
sahusanket: my bad. got it

itsankit-google: since now `getWriteDisposition()` returns `String` instead of enum.

As mentioned in PR description under `Limitations`: 

```
The repo is using google-cloud-bigquery:1.37.1 which does not contain WRITE_TRUNCATE_DATA in JobInfo.WriteDisposition enum. Hence, using it as String everywhere.
```

sahusanket (approved):

PR #1594, opened 2026-02-12
Updated the configuration for version release

123-komal:


.github/workflows/tag-release.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2022 Cask Data, Inc.
+# Copyright © 2025 Cask Data, Inc.
```
itsankit-google: nit: `2026`

itsankit-google (commented):
This build is only used for non-snapshot releases which are tagged.

So, there is no need to maintain the step of `Publish to Maven Central`.

Also we can remove this step `Get Secrets from GCP Secret Manager`.

.github/workflows/tag-release.yml
123-komal: Done

123-komal:
> This build is only used for non-snapshot releases which are tagged.
> 
> So, there is no need to maintain the step of `Publish to Maven Central`.
> 
> Also we can remove this step `Get Secrets from GCP Secret Manager`.

Understood. Done

.github/workflows/tag-release.yml
```diff
@@ -44,25 +34,17 @@ jobs:
...
+      - name: Get Project Version
+        id: get_version
+        run: echo "VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_OUTPUT
+
+      - name: Submit Build to GCB
+        id: gcb
+        run: |
+          gcloud builds submit . \
+          --config=cloudbuild-release.yaml \
+          --project='cdapio-github-builds' \
```
itsankit-google, line 49: we can skip adding `project` in the build since the runner is running in the same project as the build.

PR #1596, opened 2026-02-18
[PLUGIN-1945] Skip bucket creation if exist (copy/move action)

psainics:
## Skip bucket creation if exist (copy/move action)

Jira : [PLUGIN-1945](https://cdap.atlassian.net/browse/PLUGIN-1945)

### Description

if the user does not have `storage.buckets.create` the flow fails with 403 error, bucket creation should be skipped if the bucket already exists.

---

### Code change

- Modified `StorageClient.java`

[PLUGIN-1945]: https://cdap.atlassian.net/browse/PLUGIN-1945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -153,15 +153,9 @@ public class StorageClient {
         return;
       }
     } catch (StorageException e) {
-      int errorCode = e.getCode();
-      if (errorCode == 403) {
-        LOG.warn(
-            "Getting 403 Forbidden: {} You may not have permission to access bucket {}. Attempting to create bucket.",
-            e.getMessage(), path.getUri());
-      } else {
-        LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
-            errorCode, e.getMessage(), path.getBucket());
-      }
+      // do not throw error if unable to access bucket for backward compatibility.
+      LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
+          e.getCode(), e.getMessage(), path.getBucket());
     }
     // Fallback to bucket creations when get returns null or throws exception.
     try {
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -147,6 +147,10 @@ public class StorageClient {
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
     try {
+      if (storage.get(path.getBucket()) != null) {
+        LOG.info("Bucket {} already exists, skipping creation.", path.getBucket());
+        return;
+      }
       GCPUtils.createBucket(storage, path.getBucket(), location, cmekKeyName);
       LOG.info("Bucket {} has been created successfully", path.getBucket());
     } catch (StorageException e) {
--- a/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
+++ b/src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
@@ -146,11 +146,25 @@ public class StorageClient {
    * @param cmekKeyName  the name of the cmek key
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
+    // Skip bucket creation if bucket already exists.
     try {
       if (storage.get(path.getBucket()) != null) {
         LOG.info("Bucket {} already exists, skipping creation.", path.getBucket());
         return;
       }
+    } catch (StorageException e) {
+      int errorCode = e.getCode();
+      if (errorCode == 403) {
+        LOG.warn(
+            "Getting 403 Forbidden: {} You may not have permission to access bucket {}. Attempting to create bucket.",
+            e.getMessage(), path.getUri());
+      } else {
+        LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
+            errorCode, e.getMessage(), path.getBucket());
+      }
+    }
+    // Fallback to bucket creations when get returns null or throws exception.
+    try {
       GCPUtils.createBucket(storage, path.getBucket(), location, cmekKeyName);
       LOG.info("Bucket {} has been created successfully", path.getBucket());
     } catch (StorageException e) {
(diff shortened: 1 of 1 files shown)


vikasrathee-cs (approved):

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
itsankit-google: we should update this error message to reflect `Unable to create or get bucket.....`

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -147,6 +147,10 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
     try {
+      if (storage.get(path.getBucket()) != null) {
```
itsankit-google, line 151: This adds an additional permission `storage.buckets.get`, do we do `storage.get` anywhere else?

I wonder if we should do `storage.list` instead which we do already later on?

itsankit-google: or may be we could just ignore `403` with `storage.get` and move forward for backward compatibility.

psainics: `storage.buckets.get` permission is required for this plugin to work, we use it to traverse the bucket in function `pairTraverse`.

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
psainics: Updated !

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
itsankit-google: `pairTraverse` function might be only getting called in `GCS Move / Copy` plugin but `StorageClient` is used in all other GCS plugins as well.

psainics: As of now `createBucketIfNotExists` is only being used in gcs copy/move action.

From gcs docs it looks like list permission is for listing all buckets in a project, much slower than get, and less secure.

Let me know what do you think.

itsankit-google: As I mentioned in the reply, we can use `storage.get` but just ignore `403` with `storage.get` and move forward for backward compatibility.

psainics: Updated

```java
// Skip bucket creation if bucket already exists.
try {
  if (storage.get(path.getBucket()) != null) {
    LOG.info("Bucket {} already exists, skipping creation.", path.getBucket());
    return;
  }
  int errorCode = e.getCode();
  if (errorCode == 403) {
    LOG.warn(
        "Getting 403 Forbidden: {} You may not have permission to access bucket {}. Attempting to create bucket.",
        e.getMessage(), path.getUri());
  } else {
    LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
        errorCode, e.getMessage(), path.getBucket());
  }
}
// Fallback to bucket creations when get returns null or throws exception.
// ... Old Flow
```

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -146,6 +146,24 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
+    } catch (StorageException e) {
+      int errorCode = e.getCode();
+      if (errorCode == 403) {
+        LOG.warn(
+            "Getting 403 Forbidden: {} You may not have permission to access bucket {}. Attempting to create bucket.",
+            e.getMessage(), path.getUri());
+      } else {
+        LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
+            errorCode, e.getMessage(), path.getBucket());
+      }
```
itsankit-google: we can simplify this : 

```
catch (StorageException e) {
  int errorCode = e.getCode();
 // do not throw error if unable to access bucket for backward compatibility.
 LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.", errorCode, e.getMessage(), path.getBucket());
}
```

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -146,6 +146,24 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
    * @param cmekKeyName  the name of the cmek key
    */
   public void createBucketIfNotExists(GCSPath path, @Nullable String location, @Nullable CryptoKeyName cmekKeyName) {
+    // Skip bucket creation if bucket already exists.
+    try {
+      if (storage.get(path.getBucket()) != null) {
+        LOG.info("Bucket {} already exists, skipping creation.", path.getBucket());
```
itsankit-google: this log can become too chatty, might be good to reduce level to trace.

PushpenderSaini0: reduced level to trace

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
PushpenderSaini0: Simplified

src/main/java/io/cdap/plugin/gcp/gcs/StorageClient.java
```diff
@@ -146,6 +146,18 @@ public void mapMetaDataForAllBlobs(String path, Consumer<Map<String, String>> fu
...
+    // Skip bucket creation if bucket already exists.
+    try {
+      if (storage.get(path.getBucket()) != null) {
+        LOG.trace("Bucket {} already exists, skipping creation.", path.getBucket());
+        return;
+      }
+    } catch (StorageException e) {
+      // do not throw error if unable to access bucket for backward compatibility.
+      LOG.warn("Getting unexpected error code {}: {} when checking if bucket {} exists. Attempting to create bucket.",
+          e.getCode(), e.getMessage(), path.getBucket());
```
itsankit-google: please also add the exception in the log.

`LOG.warn(....., e);`

psainics: Added !

itsankit-google (approved):

PR #1606, opened 2026-04-09
Enhance error handling in DelegatingMultiSinkOutputCommitter

psainics:
This pull request improves error handling and logging in the `DelegatingMultiSinkOutputCommitter` class for BigQuery multi-sink operations. The main focus is to provide more informative log messages when task or job commits fail, making it easier to diagnose issues.

**Enhanced error handling and logging:**

* Added a `Logger` instance (`LOG`) using SLF4J to the `DelegatingMultiSinkOutputCommitter` class for logging warnings. [[1]](diffhunk://#diff-9915e9b2c1184584ecab1075a66cb1d39aa98b84cd786dce71237dc1a9e9acc0R26-R27) [[2]](diffhunk://#diff-9915e9b2c1184584ecab1075a66cb1d39aa98b84cd786dce71237dc1a9e9acc0R40)
* Wrapped `commitTask` and `commitJob` calls in try-catch blocks to log warnings with detailed failure reasons when an `IOException` occurs, then re-throw the exception.
* Introduced a private `getFailureReason` helper method to extract the root cause message from an `IOException` for clearer log output.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
+++ b/src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
@@ -23,6 +23,8 @@ import org.apache.hadoop.mapreduce.JobContext;
 import org.apache.hadoop.mapreduce.JobStatus;
 import org.apache.hadoop.mapreduce.OutputCommitter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/gcp/bigquery/sink/DelegatingMultiSinkOutputCommitter.java
```diff
@@ -99,18 +102,28 @@ public boolean needsTaskCommit(TaskAttemptContext taskAttemptContext) throws IOE
...
-      configureContext(jobContext, tableName);
-
-      committerMap.get(tableName).commitJob(jobContext);
+      try {
+        configureContext(jobContext, tableName);
+        committerMap.get(tableName).commitJob(jobContext);
+      } catch (IOException e) {
+        LOG.warn("BigQuery multi-sink table '{}' failed during job commit. Reason: {}",
+                 tableName, getFailureReason(e), e);
+        throw e;
```
itsankit-google, line 125: logging and re-throwing an error is considered an anti-pattern.

Isn't this error propagated upwards and logged somewhere in pipeline run?

PR #1623, opened 2026-09-02
Fix GCP Service Agent Token Theft vulnerability in connection testing

akkaur:
Modifies GCPUtils.loadServiceAccountCredentials to use ServiceAccountCredentials.fromStream() instead of GoogleCredentials.fromStream(). Previously, GoogleCredentials.fromStream() would implicitly parse various credential types. This allowed a malicious user to supply an external_account payload with a crafted credential_source (pointing to the GCP metadata server) and an external token_url, triggering an unintended Workload Identity Federation exchange and exfiltrating the Data Fusion Service Agent's token. By explicitly restricting the parser to ServiceAccountCredentials, external and non-service-account configurations now instantly throw an IOException prior to any external token fetches and effectively blocks the Privilege Escalation attack.

Also adds GCPUtilsTest.java to explicitly test that only service_account credential types can be loaded, securing regression coverage against external account injections.

### DIFF
--- a/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
+++ b/src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
@@ -118,7 +118,7 @@ public class GCPUtils {
                                                                 boolean isServiceAccountFilePath)
     throws IOException {
     try (InputStream inputStream = openServiceAccount(serviceAccount, isServiceAccountFilePath)) {
-      return GoogleCredentials.fromStream(inputStream);
+      return ServiceAccountCredentials.fromStream(inputStream);
     }
   }
 
--- a/src/test/java/io/cdap/plugin/gcp/common/GCPUtilsTest.java
+++ b/src/test/java/io/cdap/plugin/gcp/common/GCPUtilsTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright © 2026 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.gcp.common;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import org.junit.Assert;
+import org.junit.Test;
+import java.io.IOException;
+
+public class GCPUtilsTest {
+
+  @Test
+  public void testLoadServiceAccountCredentialsNegative() {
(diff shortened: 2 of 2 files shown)


src/main/java/io/cdap/plugin/gcp/common/GCPUtils.java
```diff
@@ -118,7 +118,7 @@ public static GoogleCredentials loadServiceAccountCredentials(String serviceAcco
                                                                 boolean isServiceAccountFilePath)
     throws IOException {
     try (InputStream inputStream = openServiceAccount(serviceAccount, isServiceAccountFilePath)) {
-      return GoogleCredentials.fromStream(inputStream);
+      return ServiceAccountCredentials.fromStream(inputStream);
```
vsethi09, line 121: Changing from `GoogleCredentials.fromStream()` to `ServiceAccountCredentials.fromStream()` restricts your application to accepting only traditional, long-lived service account key JSON files.

The primary impacts across your application include:

* **Breaks Workload Identity Federation (`external_account`):** Workloads authenticating from AWS, Azure, GitHub Actions, or on-premises environments via keyless federation will crash with an `IOException` or `IllegalArgumentException` because an `external_account` JSON cannot be parsed by `ServiceAccountCredentials`.
* **Breaks Local Developer Workflows (`authorized_user`):** Developers using credential files generated by `gcloud auth application-default login` or user OAuth flows will no longer be able to supply those files to your code.
* **Blocks Impersonated Account Files (`impersonated_service_account`):** Configuration files designed to dynamically impersonate a service account without possessing its private key will fail to load.
* **Enforces Security Anti-Patterns:** It forces consumers of your application or plugin to create, download, store, and rotate static service account private keys (`"type": "service_account"`), increasing the risk of leaked credentials.
* **API Incompatibility / Code Changes:** Any calling code expecting the base `GoogleCredentials` type may need adjustments if it relies on polymorphic credential handling, though `ServiceAccountCredentials` is assignable to `GoogleCredentials`.
* **Gains Direct Access to Service Account Methods:** On the positive side, you gain compile-time access to service account-specific methods on the returned object without casting—such as `.createDelegated(userEmail)` for Google Workspace Domain-Wide Delegation, `.getClientEmail()`, and `.getPrivateKey()`.

Would it be fine?

vsethi09: To remediate you may consider implementing stricter input validation for Google Cloud connector configurations. 

Example:

*   **Implement Stricter Input Validation:** Add validation checks in the backend handlers (such as `ConnectionHandler.java` in CDAP and individual plugin connectors like `GCSConnector.java`) for the `serviceAccountJSON` payload.
*   **Restrict Credential Type Properties:** Do not allow the `credential_source` URL to point to the local metadata server (`http://metadata.google.internal/computeMetadata/v1/...`) or internal loopback IP addresses when testing connections with external accounts.
*   **Validate Token URLs:** Allowlist or validate the `token_url` domain to ensure it points to legitimate identity providers, rather than arbitrary attacker-controlled destinations.
*   **Audit All Google Connectors:** Ensure the validation applies systemically across all Google Cloud connectors (BigQuery, GCS, Spanner, Bigtable, Pub/Sub, etc.) as the vulnerability may be present in multiple plugin configurations.

vsethi09 (changes requested):

